PR #10, opened 2015-07-02
Caused fatal error in WP 4.2.2

amando96:
Caused fatal error "Call to undefined function simplexml_import_dom() in FrmXMLHelper.php on line 55"

Fixed by checking if the function exists before calling it.


stephywells:
Thanks for the pull request. I committed a patch for this a little differently here:
https://github.com/Strategy11/formidable-forms/commit/1683bbe7d0569d3689ced1bdc4082868771ac16c

Thanks again!

PR #21, opened 2015-12-12
Fixed license response on activation

knightzac19:
The response for an activated response from the license server was incorrect. I've fixed it and also #18  has additional fixes for that page as well.


zknightfsw:
This has been fixed by the devs. It can be closed.

PR #31, opened 2017-01-17
Fixing the issue 1369

gfirem:
Fixing the issue https://github.com/Strategy11/formidable/issues/1369.
Change the way to initialize the chosen jquery plugin when are load in edit in place, in this case it would load all chose inside the container

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -3734,11 +3734,11 @@ function frmFrontFormJS(){
 			if ( typeof __frmChosen !== 'undefined' ) {
 				opts = '{' + __frmChosen + '}';
 			}
-            if(typeof chosenContainer !== 'undefined' ){
-                jQuery("#"+chosenContainer).find('.frm_chzn').chosen(opts);
-            } else {
-                jQuery('.frm_chzn').chosen(opts);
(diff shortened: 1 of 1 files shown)


gfirem:
Is all are using phpStorm exist the possibility to create a file with a standard code to the environment.

image

js/formidable.js
```diff
@@ -3728,13 +3728,17 @@ function frmFrontFormJS(){
...
 
-	function loadChosen() {
+	function loadChosen(chosenContainer) {
 		if ( jQuery().chosen ) {
 			var opts = {allow_single_deselect:true,no_results_text:frm_js.no_results};
 			if ( typeof __frmChosen !== 'undefined' ) {
 				opts = '{' + __frmChosen + '}';
 			}
-			jQuery('.frm_chzn').chosen(opts);
+            if(typeof chosenContainer !== 'undefined' ){
```
stephywells: It still looks like the tabs are soft here. Can these 4 lines be adjusted to use tabs instead of spaces?

gfirem: I am using tabs

PR #32, opened 2017-01-18
Fixing the issues/1455.

gfirem:
https://github.com/Strategy11/formidable/issues/1455
This is problems with css to show the list of triggers.


gfirem:
This was the best way to do this, because if change the position it affect the others containers and need to modify almost all fathers or other place where is used the same portion of the view

PR #35, opened 2017-01-18
Spanish suggestion

gfirem:
Spanish suggestion to change Comentarios to Entradas is more intuitive for the Spanish users


gfirem:
I found the site to made the suggestion of the translation. @jwahlin Thanks

PR #36, opened 2017-01-19
Fixing the issues/1374

gfirem:
https://github.com/Strategy11/formidable/issues/1374
Fixing the validation of the field not be repeated in case of separated values in the Entries list


stephywells:
This won't quite work. If a field is set to have separate values, we want one column for the saved value, and one column for the displayed value. But a post status field should only have one column. 

This change will hide the extra column for all fields.

PR #41, opened 2017-01-21
Fixing issues/1329

gfirem:
https://github.com/Strategy11/formidable/issues/1329
Forcing the form to send with ajax and validate in the client side with js if edit in place. Now it work in a view with single entry and a list with all entries.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1048,11 +1048,11 @@ class FrmFormsController {
 
         return self::show_form(
             $shortcode_atts['id'], $shortcode_atts['key'], $shortcode_atts['title'],
-            $shortcode_atts['description'], $atts
+            $shortcode_atts['description'], $atts, $frm_vars
         );
     }
 
-    public static function show_form( $id = '', $key = '', $title = false, $description = false, $atts = array() ) {
+    public static function show_form( $id = '', $key = '', $title = false, $description = false, $atts = array(), $frm_vars = array() ) {
         if ( empty( $id ) ) {
             $id = $key;
         }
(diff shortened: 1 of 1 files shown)


classes/controllers/FrmFormsController.php
```diff
@@ -1048,11 +1048,11 @@ public static function get_form_shortcode( $atts ) {
 
         return self::show_form(
             $shortcode_atts['id'], $shortcode_atts['key'], $shortcode_atts['title'],
-            $shortcode_atts['description'], $atts
+            $shortcode_atts['description'], $atts, $frm_vars
         );
     }
 
-    public static function show_form( $id = '', $key = '', $title = false, $description = false, $atts = array() ) {
+    public static function show_form( $id = '', $key = '', $title = false, $description = false, $atts = array(), $frm_vars = array() ) {
```
stephywells: Instead of passing $frm_vars, can you use global $frm_vars instead?

classes/controllers/FrmFormsController.php
```diff
@@ -1061,6 +1061,13 @@ public static function show_form( $id = '', $key = '', $title = false, $descript
         if ( ! $form ) {
             return __( 'Please select a valid form', 'formidable' );
         }
+	
+	    if ( ! empty( $frm_vars ) && ! empty( $form ) && ! empty( $form->options ) ) {
```
stephywells: This check shouldn't be needed. $frm_vars should never be empty, and we just checked the form at line 1061.

classes/controllers/FrmFormsController.php
```diff
@@ -1061,6 +1061,13 @@ public static function show_form( $id = '', $key = '', $title = false, $descript
...
+		    if ( !empty( $frm_vars["inplace_edit"] ) ) {
+			    $form->options["js_validate"] = '1';
+			    $form->options["ajax_submit"] = '1';
```
stephywells: Since these are pro options, it would be best if we can keep it out of the lite version. Is it possible to find a cleaner way of doing this?

gfirem: The main problem come when the shortcode are process into the view, it get in the same order of the content, when the page submit and return with an error the edit shortcode show the error. but is the last part to add into the content returned and need to process the entire content at that point using DOM to process and remove if is necessary (provided by @jwahlin) or using jquery in the front. Is more secure to use the @jwahlin solution. @stephywells or @jwahlin what helper is the best to include the function i will use here because it will be usefully in others places.

PR #44, opened 2017-01-24
Fixing the issues/1455.

gfirem:
https://github.com/Strategy11/formidable/issues/1455
Providing a solution with jquery and css. Adding auto close to the trigger lit on focus out.


gfirem:
Not a valid solution

PR #49, opened 2017-01-25
Fixing issue/536

gfirem:
https://github.com/Strategy11/formidable/issues/536

Fixing the switch view to the key into the toolbox when is created a view


stephywells:
Thanks for finding the source of the bug! It looks like it would be better to use .on('click' here instead of initializing again.
This is untested, but should accomplish the same thing.
jQuery('#frm_adv_info').on('click', '.subsubsub a.frmids', function(e){toggleKeyID('frmids',e);});

stephywells:
After this change, the extra panelViewSwitchIdKey(); function can be removed. This function no longer needs to be initiated an extra time.

PR #50, opened 2017-01-25
Fixing issue 1374

gfirem:
https://github.com/Strategy11/formidable/issues/1374
Adding the post status validation


stephywells:
Instead of adding the condition back into that line, only this line needs to be changed:
$is_not_post_status = FrmField::is_option_empty( $form_col, 'post_field' );

There is also a line added in there with some white space.

PR #58, opened 2017-06-12
Blank Widget Title like WordPress Core

jeremylind:
Changed title processing to match WordPress core widgets. A blank title will output nothing, just like core widgets, instead of using the form title in the absence of a widget title. I think this is more predictable behaviour. WordPress core implementation can be found at `wp-includes/widgets/class-wp-widget-pages.php line 55`.


stephywells:
Looks great! Thanks @jeremylind!

PR #59, opened 2017-06-13
Shortcode filter added to new core text widget hook

jeremylind:
Changed the hook for widget_text_filter function to hook to widget_text_content instead of widget_text in order to match changes in WordPress 4.8. This prevents the formidable shortcode being processed before wpautop, which causes all kinds of problems.

### DIFF
--- a/classes/controllers/FrmHooksController.php
+++ b/classes/controllers/FrmHooksController.php
@@ -49,7 +49,7 @@ class FrmHooksController {
         }
 
         add_action( 'plugins_loaded', 'FrmAppController::load_lang' );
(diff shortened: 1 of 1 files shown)


jeremylind:
Hmm, this probably needs something more for backwards compatibility because the widget_text_content filter is new in WordPress 4.8.

stephywells:
Thanks for your commit! We ended up adjusting this so WP 4.9 will ignore it now since it will be reverse compatible. Thanks again!

https://github.com/Strategy11/formidable-forms/commit/90bd64d92c456e98dcfa56afd6080bbe076cd3cd#diff-56a417d312f2e5871c09d86262b0da00L52

PR #64, opened 2017-08-31
Allows text colors in frm-show-entry https://github.com/Strategy11/fo…

LauraSteiner:
…rmidable/issues/1503

@jwahlin @stephywells -- for your review.

For future proofing reasons, I'm only putting color settings through the color markup function, which only changes values that have all hex digits.

At this point, with CSS options as they are and the settings we have, it wouldn't be a problem to put all settings through this function.  But if we add settings or CSS has a new unit, this could cause problems.


jwahlin:
This is awesome! Thanks Laura! I'm merging it now.

PR #65, opened 2017-08-31
Issue/297sync

LauraSteiner:
@jwahlin Ready for next round  of reviews.  My responses to your many excellent comments are in the issue discussion:

https://github.com/Strategy11/formidable/issues/297

Thank you! 

(No changes from earlier.  I'm just turning this into a pull request for easier review.)

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -1677,6 +1677,10 @@ span.howto{display:inline;}
 	display:none;
 }
 
+.frm_add_form_row.frm_hide_add_button {
+	display: none;
+}
+
 .frm_button{
     text-decoration:none;
     border:1px solid #eee;
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -1677,10 +1677,6 @@ span.howto{display:inline;}
 	display:none;
 }
 
-.frm_add_form_row.frm_hide_add_button {
-	display: none;
-}
-
 .frm_button{
     text-decoration:none;
     border:1px solid #eee;
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -2978,6 +2978,18 @@ if(f){
 }
 }
 
+function frmCheckRepeatLimit(repeatLimit){
+
+    if ( repeatLimit.value > 1 ){
+        return;
+    }
+
+    alert("Please enter an integer greater than 1 for the Repeat Limit.");
+
+    jQuery(repeatLimit).val("");
+
+}
+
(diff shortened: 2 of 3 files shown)


LauraSteiner:
@jwahlin Ready for your next round of feedback!

js/formidable_admin.js
```diff
@@ -2978,6 +2979,17 @@ if(f){
 }
 }
 
+function checkRepeatLimit() {
```
stephywells: Would you mind moving this function inside the closure? It would be great right after the toggleRepeatButtons function.

LauraSteiner:
Thank you for the review, Steph!  I believe I've made the change you requested.  

I'm ready for further review from you and Jamie!

css/frm_admin.css
```diff
@@ -1677,6 +1677,10 @@ span.howto{display:inline;}
 	display:none;
 }
 
+.frm_add_form_row.frm_hide_add_button {
+	display: none;
+}
+
```
jwahlin: Does this have a purpose? If so, when/where does it apply?

js/formidable_admin.js
```diff
@@ -988,6 +988,17 @@ function frmAdminBuildJS(){
 		}
 	}
 
+	function checkRepeatLimit() {
+
+		if (this.value >= 2 && repeatLimit <= 200) {
```
jwahlin: The repeatLimit variable isn't defined. This should be this.value.

js/formidable_admin.js
```diff
@@ -988,6 +988,17 @@ function frmAdminBuildJS(){
...
+			return;
+		}
+
+		alert("Please enter an integer greater than 2 for the Repeat Limit.");
```
jwahlin: This should say "greater than 1", right?

css/frm_admin.css
LauraSteiner: Not sure what this is doing here.  I can't see a purpose.  I'm removing it.

js/formidable_admin.js
LauraSteiner: Thanks!  I missed that when I rewrote.

js/formidable_admin.js
LauraSteiner: Yes!   Another great catch!

 I'm concerned that some people may not know what the word "integer" means, so maybe I'll use "greater than or equal to 2."  Then they'll see the number 2 and hopefully get that that's the minimum.

PR #66, opened 2017-09-01
Catches JS errors in calculations on the front end and alerts users.

LauraSteiner:
@jwahlin @stephywells First draft of catching JS calculation errors on the front end and alert owners/users.

To be safe, I hide the submit/next button when there's an error.  I also put a red line around the form.  I'm happy to handle this another way, if you guys have a different preference.

This is phase 1.  Phase 2 will involve checking calculations in the admin for unmatched parens (and other errors, if we think of them) and alerting users.


jwahlin:
Hi Laura!

It looks like the approach for this is a bit dependent on what the team decides.

Once that is decided, I can better review this pull request. At the moment, I'm seeing a couple of minor code styling items that could be adjusted:
1. In our JavaScript, we use camel case. That is the norm for JavaScript according to the research I've done on it.
2. There is some inconsistent spacing in the PR. That could be cleaned up. We include spaces within function calls in PHP, but not JavaScript. Each language has its own code styling. A couple examples of places where I'm seeing inconsistent spacing:
- function create_calculations_error_message(err, field_key, thisFullCalc ){ (should not have space after thisFullCalc)
- if (action ==="frm_forms_preview"){ (should be space after ===)

LauraSteiner:
@jwahlin This is ready for your review so we can start a discussion.  I'd especially like your input on what to put in error messages.

classes/controllers/FrmFormsController.php
```diff
@@ -751,12 +751,15 @@ private static function get_shortcode_helpers( $settings_tab ) {
...
-    // Insert the form class setting into the form
+	// Insert the form class setting and maybe the admin-viewing class into the form
 	public static function form_classes( $form ) {
-        if ( isset($form->options['form_class']) ) {
+		if ( isset($form->options['form_class']) ) {
 			echo esc_attr( sanitize_text_field( $form->options['form_class'] ) );
-        }
-    }
+		}
+		if ( current_user_can( 'activate_plugins' ) ) {
```
stephywells: Looks like this needs to be removed?

js/formidable.js
```diff
@@ -2775,14 +2792,20 @@ function frmFrontFormJS(){
...
+			try {
+				total = parseFloat(eval(thisFullCalc));
+			}
+
+			catch(err) {
+				maybeShowCalculationsErrorAlert(err, field_key, thisFullCalc);
+			}
 
-			if ( typeof total === 'undefined' || isNaN(total) ) {
+			if (typeof total === 'undefined' || isNaN(total)) {
```
stephywells: Would you mind putting the spacing back in this line?

js/formidable.js
```diff
@@ -2775,14 +2792,20 @@ function frmFrontFormJS(){
...
 				total = 0;
 			}
 
 			// Set decimal points
-			if ( isNumeric( dec ) ) {
+			if (isNumeric(dec)) {
```
stephywells: Can we get the spacing back in this line too?

js/formidable_admin.js
```diff
@@ -505,6 +505,152 @@ function frmAdminBuildJS(){
...
+	function isTextCalculation(inputElement) {
+		return jQuery(inputElement).siblings("label[for^='calc_type']").children("input").prop("checked");
+	}
+	
+	/**
+	 * Returns a regular expression of shortcodes that can't be used in numeric calculations.
+	 * @returns {RegExp}
+	 */
+	function getNonNumericShortcodes() {
+		return /\[(date|time|email|ip\])/;
```
stephywells: Our age calculation actually uses the time shortcode. Since both date and time can use formats that make them numeric, it may be better to allow them. Moving the \] after the ) in this regex should do the trick.

js/formidable.js
```diff
@@ -2735,6 +2735,23 @@ function frmFrontFormJS(){
...
+		var alertMessage = '';
+
+		if ((!jQuery('form').hasClass('frm-admin-viewing'))) {
+			return;
+		}
+
+		alertMessage += 'Oops! There\'s an error in the calculation in field ' + field_key + ':\n\n';
+		alertMessage += thisFullCalc + '\n\n';
+		if (err.message) {
+			alertMessage += 'The error message is: ' + err.message + '\n\n';
```
stephywells: Maybe we can remove 'The error message is:' part? That will simplify the localization if there's only one string or two that we need to add.

js/formidable.js
```diff
@@ -2735,6 +2735,23 @@ function frmFrontFormJS(){
...
+		}
+		alertMessage += 'No worries!  These are generally easy to fix.';
```
stephywells: Maybe we can remove this line too?

classes/controllers/FrmFormsController.php
stephywells: Let's make sure they also have permission to edit forms with current_user_can( 'frm_edit_forms' )

js/formidable.js
```diff
@@ -2735,6 +2735,23 @@ function frmFrontFormJS(){
...
 
+	function maybeShowCalculationsErrorAlert(err, field_key, thisFullCalc) {
+
+		var alertMessage = '';
+
+		if ((!jQuery('form').hasClass('frm-admin-viewing'))) {
+			return;
+		}
+
+		alertMessage += 'Oops! There\'s an error in the calculation in field ' + field_key + ':\n\n';
```
stephywells: We'll need to move any text strings to the localization. This is included in FrmAppHelper::localize_script()

js/formidable_admin.js
```diff
@@ -505,6 +505,152 @@ function frmAdminBuildJS(){
...
+				if (top !== closing[formula_array[i]]) {
+					unmatchedClosing.push(formula_array[i])
+				}
+			}
+		}
+
+		if (stack.length > 0 || unmatchedClosing.length > 0) {
+			msg = '***This calculation has at least one unmatched ( ) { } [ ].***\n\n';
+			/*
+			//a more specific message option that has opening and closing grouped, with only the first unmatched
```
stephywells: Does this commented section need to stick around?

js/formidable_admin.js
```diff
@@ -505,6 +505,152 @@ function frmAdminBuildJS(){
...
+	}
+	
+	/**
+	 * Returns a regular expression of shortcodes that can't be used in forms (but can be used in Views, Email
+	 * Notifications, and other Formidable areas.
+	 * 
+	 * @returns {RegExp}
+	 */
+	function getNonFormShortcodes() {
+		return /\[(key|id\]|created|updated)/;
```
stephywells: This regex needs a little modification to make sure it only applies to the shortcodes we are checking for. Right now, it flags [keyoffield]. Should we use created-at, created-by, updated-at, updated-by?

LauraSteiner:
Thank you for your comments, which were very helpful, as always!  I'm glad to know about storing message strings in wp_localize_script so they can be translated.  

I've added more shortcodes to the regex checking for non-form shortcodes.  As it's written, most don't require a closing bracket to match.  I'd appreciate your review of this (and everything else, of course) to make sure the net I'm throwing isn't too wide.

js/formidable_admin.js
```diff
@@ -534,6 +534,140 @@ function frmAdminBuildJS(){
...
+
+		return msg;
+	}
+
+	function checkCalculationCreatedByUser() {
+		var calculation = this.value;
+		var warningMessage = checkMatchingParens( calculation );
+		warningMessage += checkShortcodes( calculation, this );
+
+		if ( warningMessage != '' ) {
```
stephywells: Do you have JS hint enabled in PHP storm? JS hint prefers !== instead of != whenever possible.

js/formidable.js
```diff
@@ -2817,6 +2817,23 @@ function frmFrontFormJS(){
 		return hidden;
 	}
 
+	function maybeShowCalculationsErrorAlert( err, field_key, thisFullCalc ) {
+
+		var alertMessage = '';
+
+		if ( ( !jQuery( 'form' ).hasClass( 'frm-admin-viewing' ) ) ) {
```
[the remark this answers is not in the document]
LauraSteiner: So smart of you to think of this!  Login forms didn't cross my mind.  The jQuery docs say that hasClass will return true if any of the matched elements have the specified class.  So this happens to work properly as it is, I believe.

js/formidable_admin.js
LauraSteiner: Great hint. (pun!)  Thanks.

LauraSteiner:
Thank you very much for doing this review!  I need to figure out a better system for following GitHub updates.  This somehow slipped past my radar, and I didn't see it until today. 

The JS has been split since I wrote this.  I'm sure there's a cool GitHub way to incorporate everything, but I just created a new branch and added the changes by hand.  I'll submit for a pull request shortly.

js/formidable_admin.js
LauraSteiner: Great hint. (pun!) Thanks.

js/formidable.js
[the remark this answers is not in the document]
LauraSteiner: So smart of you to think of this! Login forms didn't cross my mind. The jQuery docs say that hasClass will return true if any of the matched elements have the specified class. So this happens to work properly as it is, I believe.

PR #67, opened 2017-09-02
Show default val icons for address fields.

LauraSteiner:
@jwahlin My changes are the same as in the other branch from earlier.  To deal with merge conflicts, I downloaded the latest plugin code and moved my changes onto it.


js/formidable_admin.js
```diff
@@ -692,16 +692,6 @@ function frmAdminBuildJS(){
...
-	function showDefaults(n,fval){
-		var defaults = jQuery('#frm_clear_on_focus_'+n+',#frm_clear_on_focus_'+n+' a, #frm_clear_on_focus_'+n+'_conf,#frm_clear_on_focus_'+n+'_conf a');
-		if(fval){
-			defaults.css('visibility','visible').fadeIn('slow');
-		}else{
-			defaults.css('visibility','visible').fadeOut('slow');
-		}
-		return false;
-	}
-
```
jwahlin: I removed this to avoid duplicated code.

js/formidable_admin.js
```diff
@@ -1422,6 +1416,38 @@ function frmAdminBuildJS(){
 		});
 	}
 
+	function showOrHideDefaultValIcons(showDefaultValIcons, $innerField) {
```
jwahlin: This could use a JSdoc - would you mind adding that? I like to include docs for the arguments, version that function was added (if you added it), and a short description of the function's purpose.

js/formidable_admin.js
```diff
@@ -1422,6 +1416,38 @@ function frmAdminBuildJS(){
...
+		var $defaultValueIcons = $innerField.find('.frm_default_val_icons');
+
+        if (showDefaultValIcons) {
```
jwahlin: Spacing inconsistency on this line.

js/formidable_admin.js
```diff
@@ -1422,6 +1416,38 @@ function frmAdminBuildJS(){
...
+			$defaultValueIcons.css('visibility', 'visible').fadeIn('slow');
+		}
+		else {
```
jwahlin: I believe this should be } else {, all on the same line. If you're following WordPress codestyling recommendations on this, let me know!

js/formidable_admin.js
```diff
@@ -1422,6 +1416,38 @@ function frmAdminBuildJS(){
...
+			$defaultValueIcons.css('visibility', 'hidden').fadeOut('slow');
+		}
+	}
+
+	function maybeShowDefaultValIcons(fieldId) {
```
jwahlin: Needs JSdoc.

js/formidable_admin.js
LauraSteiner: You're correct.  Fixed.  Thanks!

js/formidable_admin.js
LauraSteiner: Happy to!

js/formidable_admin.js
LauraSteiner: Added.  Thanks!

js/formidable_admin.js
LauraSteiner: Is this still an issue?  It looks like you may have fixed this with your last commit.

LauraSteiner:
Great improvements by you, as always, Jamie!  I've adjusted the code to take confirmation fields into account.  Changes are pushed.

stephywells:
I made a few changes, and it looks good to go. The check for the length of the default value was causing trouble for me, so I had to remove it. Maybe I was typing too fast, so the first keystroke didn't trigger it?

PR #69, opened 2017-09-28
Make "Add" button properly hide and show

sojohnnysaid:


js/formidable_admin.js
```diff
@@ -1644,6 +1644,11 @@ function frmAdminBuildJS(){
 				jQuery(document.getElementById('frm_posttax_rows')).append(html);
 			}
 		});
+
+		var current_num_of_rows = jQuery(this).parent().siblings('#frm_posttax_rows').length;
+		if (current_num_of_rows > 0) {
+			jQuery(this).hide();
```
stephywells: This would be a bit more reliable if it can be included inside the success:function(html){.... 
That way, it'll for sure be run after the new row is added.

Once this is inside the success function, you only need the jQuery(this).hide(); line. The rest can go since we already know the row was added.

js/formidable_admin.js
```diff
@@ -2587,6 +2600,7 @@ function frmAdminBuildJS(){
             $formActions.on('change', '.frm_post_type', switchPostType);
 			$formActions.on('click', '.frm_add_postmeta_row', addPostmetaRow);
 			$formActions.on('click', '.frm_add_posttax_row', addPosttaxRow);
+			$formActions.on('click', '#frm_posttax_rows .frm_remove_tag', togglePostTaxAddButton);
```
stephywells: It looks like this could be removed. The frm_remove_tag class triggers the removeThisTag function, so the button show could be moved inside that function instead.

js/formidable_admin.js
```diff
@@ -1710,6 +1715,14 @@ function frmAdminBuildJS(){
 		});
 	}
 
+	function togglePostTaxAddButton(){
+		var current_num_of_rows = jQuery(this).parentsUntil('#frm_posttax_rows').siblings().length;
```
stephywells: There is a bit of duplication here with existing code. Maybe this could fit in around here?
https://github.com/Strategy11/formidable-forms/blob/7ed3a7918ff58e6fbf85a1676ff9dab6e8ab3802/js/formidable_admin.js#L75

js/formidable_admin.js
```diff
@@ -3035,5 +3049,6 @@ function frmImportCsv(formID){
 			}, 2000);
         }
     }
+
     });
-}
+}
```
stephywells: Would you mind reverting this change? We like to keep an empty line at the end of files.

sojohnnysaid:
Thanks again! Made those changes from the last review

sojohnnysaid:
Fixed the button being weird issue you were telling me about earlier

PR #70, opened 2017-10-08
Issue/361 cc bcc reply buttons

WileESpaghetti:
Fixes responsive issues for the CC/BCC/Reply To buttons for [#361](https://github.com/Strategy11/formidable/issues/361)

I moved the buttons to their own table row and made them bigger click targets on mobile


stephywells:
Thanks for working on this! It's looking a lot better on small screens. But I'm not loving the moved buttons on larger screens. It would be best if the buttons can stay as they are on bigger screens, but look better when small. Here are screenshots of the way it looked before these changes.
screen shot 2017-10-09 at 11 02 52 am
screen shot 2017-10-09 at 11 02 40 am

I think we need some CSS adjustments instead of HTML on this.

WileESpaghetti:
I removed the version with the moved buttons and pushed a new branch that keeps the preferred large screen alignment.

you might need to do a `git pull -f` to get the new changes.

PR #71, opened 2017-10-16
Issues #1614/Arrows missing in WP 4.8

endifmedia:
reference https://github.com/Strategy11/formidable/issues/1614

* added recommendations from issue tracking (above)
* changed **widget-title** h4 to h3
* updated css to fix down arrow not rotating when dropdown is open
* fixed vertical alignment on action icons left of title

### DIFF
--- a/css/font_icons.css
+++ b/css/font_icons.css
@@ -1,9 +1,39 @@
-.frm_form_action_settings .widget-top .toggle-indicator:before {
+#frm_notification_settings .widget-top .widget-action,
+#frm_form_editor_container .widget-top .widget-action {
+	border: 0;
+	margin: 0;
+	padding: 10px;
+	background: 0 0;
+	cursor: pointer;
+	outline: 0;
+}
+
+#frm_notification_settings .widget-top .widget-action .toggle-indicator:before,
+#frm_form_editor_container .widget-top .widget-action .toggle-indicator:before {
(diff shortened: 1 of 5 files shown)


stephywells:
Thank you so much! Great solution. It looks great for the field options box in the current WordPress version. I checked the field options and form actions boxes in WP 4.6 too, and it looks like it'll need a little extra love. Here are a few screenshots of what I'm seeing.

WP 4.6:
4 6-field
4 6-action

Also, the arrow isn't switching on the form actions in 4.8. And is it possible to give that left icon a bit more margin before the action title?
4 8-action

endifmedia:
Good catch. Added a bit of css for backwards compatibility and fixed an issue with the dropdown arrow not rotating correctly when a new action is added. Tested back to 4.0, looks good.

css/font_icons.css
```diff
@@ -210,6 +247,8 @@ a.frm_icon_font:hover{
 }
 .frm_email_icon:before {
 	content: "\e626";
+	vertical-align: text-top;
+	margin-right: 5px;
```
stephywells: Since this class can be used in other locations, and on the front-end, we'll need to make this addition in the frm_admin.js file.

endifmedia:
good call. css moved over to frm_admin.css

stephywells:
Thank you! This is all merged.

PR #73, opened 2017-10-19
Add class to align inline submit button with fields that don't have t…

LauraSteiner:
Add class to align inline submit button with fields that don't have top labels

I couldn't think of a good class name for this.  I'd appreciate suggestions!

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -372,23 +372,23 @@ DEFAULT_HTML;
 		return $position;
 	}
 
-
 	/**
-	 * Determine if a form has fields with top labels to align the submit button properly
+	 * Determine if a form has fields with top labels so submit button can be aligned properly
 	 *
 	 * @param $fields
 	 * @param $form
 	 *
 	 * @return bool
 	 */
-	public static function form_has_top_labels( $fields, $form ) {
+
+	private static function form_has_top_labels( $form ) {
+		$fields = $form['fields'];
 		if ( ! isset( $fields[0] ) ) {
 			return false;
 		}
 
 		return self::field_has_top_label( $fields[0], $form );
 	}
-
 
 	/**
 	 * Check if a field's label position is set to "top"
@@ -400,12 +400,24 @@ DEFAULT_HTML;
 	 */
 	private static function field_has_top_label( $field, $form ) {
 
-		$label_pos = self::label_position( $field['label'], $field, $form );
-		if ( 'top' == $label_pos ) {
-			return true;
+		$label_position = self::label_position( $field['label'], $field, $form );
+		return in_array( $label_position, array( 'top', 'inside', 'hidden' ) );
+	}
+
+	/**
+	 * Returns appropriate class if form has top labels
+	 *
+	 * @param $form
+	 *
+	 * @return string
+	 */
+	public static function maybe_return_inline_submit_top_class( $form ) {
+
+		if ( self::form_has_top_labels( $form ) ) {
+			return ' frm_inline_submit_top';
 		}
 
-		return false;
+		return '';
 	}
 
 	/**
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -375,7 +375,6 @@ DEFAULT_HTML;
 	/**
 	 * Determine if a form has fields with top labels so submit button can be aligned properly
 	 *
-	 * @param $fields
 	 * @param $form
 	 *
 	 * @return bool
@@ -383,11 +382,12 @@ DEFAULT_HTML;
 
 	private static function form_has_top_labels( $form ) {
 		$fields = $form['fields'];
-		if ( ! isset( $fields[0] ) ) {
+
+		if ( count( $fields ) <= 0 ) {
 			return false;
 		}
 
-		return self::field_has_top_label( $fields[0], $form );
+		return self::field_has_top_label( reset( $fields ), $form );
 	}
 
 	/**
@@ -399,6 +399,10 @@ DEFAULT_HTML;
 	 * @return bool
 	 */
 	private static function field_has_top_label( $field, $form ) {
+
+		if ($field === false){
+			return false;
+		}
 
 		$label_position = self::label_position( $field['label'], $field, $form );
 		return in_array( $label_position, array( 'top', 'inside', 'hidden' ) );
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -372,6 +372,42 @@ DEFAULT_HTML;
 		return $position;
 	}
 
+
+	/**
+	 * Determine if a form has fields with top labels to align the submit button properly
+	 *
+	 * @param $fields
+	 * @param $form
+	 *
+	 * @return bool
+	 */
+	public static function form_has_top_labels( $fields, $form ) {
+		if ( ! isset( $fields[0] ) ) {
+			return false;
+		}
+
+		return self::field_has_top_label( $fields[0], $form );
+	}
+
+
+	/**
+	 * Check if a field's label position is set to "top"
+	 *
+	 * @param $field
+	 * @param $form
+	 *
+	 * @return bool
+	 */
+	private static function field_has_top_label( $field, $form ) {
+
+		$label_pos = self::label_position( $field['label'], $field, $form );
+		if ( 'top' == $label_pos ) {
+			return true;
+		}
+
+		return false;
+	}
+
 	/**
 	 * Check if this field type allows placeholders
 	 * @since 2.05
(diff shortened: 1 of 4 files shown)


css/_single_theme.css.php
```diff
@@ -1115,6 +1121,12 @@
 		width:100%<?php echo esc_html( $important ) ?>;
 	}
 
+	.<?php echo esc_html( $style_class ) ?> .frm_align_top_submit input[type=submit],
+	.<?php echo esc_html( $style_class ) ?> .frm_align_top_submit.frm_submit input[type=button],
+	.<?php echo esc_html( $style_class ) ?> .frm_align_top_submit.frm_submit button {
+		margin-top: 10px;
```
stephywells: I'm thinking this should use the number from the settings rather than hard-code it. It's called submit_margin.

css/_single_theme.css.php
```diff
@@ -553,6 +553,12 @@
     vertical-align:middle;
 }
 
+.<?php echo esc_html( $style_class ) ?> .frm_align_top_submit input[type=submit],
```
stephywells: Since we'll be having people enter this value for now, maybe we could make the class more explanatory? Would frm_top_submit be any better?

LauraSteiner: I'm concerned that people will think "frm_top_submit" would move the submit button to the top of the form.

What do you think of some variation of frm_inline_submit, like frm_inline_submit_top?  It's not completely clear what it does, IMO, but I think it may get people in the right ballpark.

css/_single_theme.css.php
LauraSteiner: Perfect.  Thank you.

LauraSteiner:
Wonderful idea!  

With fields in a single row, checking the last would work perfectly.  If there are multiple rows, the submit button will align with the top row.  What do you think about checking the first field instead of the last?

LauraSteiner:
I've made the changes based on your awesome suggestions.  Making the inline submit setting useful for people is exciting!  Looking forward to your comments.

classes/helpers/FrmFormsHelper.php
```diff
@@ -267,7 +267,7 @@ public static function get_default_html( $loc ) {
 		if ( $loc == 'submit' ) {
             $draft_link = self::get_draft_link();
             $default_html = <<<SUBMIT_HTML
-<div class="frm_submit">
+<div class="frm_submit [inline_submit]">
```
stephywells: Can we revert this change and add the class with frm_inline_form in the form container instead? Since we're adding the styling, we don't really need an extra class for this so we can use ".frm_inline_form .frm_submit button" instead. It gets complicated when we make HTML changes, so we try to avoid it when possible.

classes/helpers/FrmFieldsHelper.php
```diff
@@ -372,6 +372,42 @@ private static function &label_position( $position, $field, $form ) {
...
+	 *
+	 * @param $field
+	 * @param $form
+	 *
+	 * @return bool
+	 */
+	private static function field_has_top_label( $field, $form ) {
+
+		$label_pos = self::label_position( $field['label'], $field, $form );
+		if ( 'top' == $label_pos ) {
```
stephywells: There are a few label positions that would need this. Hidden and inside also have a space above the field.

stephywells: This line can also be simplified to something more like:
return in_array( $label_pos, array( 'top', 'inside', 'hidden' ) );

css/_single_theme.css.php
```diff
@@ -553,6 +553,12 @@
     vertical-align:middle;
 }
 
+.<?php echo esc_html( $style_class ) ?> .frm_inline_submit_top input[type=submit],
+.<?php echo esc_html( $style_class ) ?> .frm_inline_submit_top.frm_submit input[type=button],
+.<?php echo esc_html( $style_class ) ?> .frm_inline_submit_top.frm_submit button {
+	margin-top: 0;
```
stephywells: This looks better for me with margin instead of margin-top

classes/helpers/FrmFormsHelper.php
LauraSteiner: Happy to set this up as you'd like!  I'm not completely sure what you're suggesting here.  Would the extra class be added to the form container?  So the two options would be something like:

.frm_inline_form.frm_form_inline_submit  .form_submit button  
.frm_inline_form.frm_form_inline_submit_top .form_submit button

Or somewhere else?

css/_single_theme.css.php
LauraSteiner: Sounds good.

classes/helpers/FrmFormsHelper.php
LauraSteiner: Moving the styling to the form container is so much cleaner and more elegant.  Thank you for this wonderful idea!

classes/helpers/FrmFormsHelper.php
```diff
@@ -291,8 +291,34 @@ public static function get_draft_link() {
...
+				$inline_class = 'frm_inline_submit ';
+			} else {
+				$inline_class = 'frm_inline_submit_top ';
+			}
+		}
+
+		return str_replace( '[inline_submit]', $inline_class, $button_html );
+	}
+
+	public static function get_custom_submit( $html, $form, $submit, $form_action, $values, $form_has_top_labels ) {
```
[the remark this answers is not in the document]
LauraSteiner: Great thinking!  With your awesome idea of moving the styling to the form container, I no longer need to interfere with the submit button HTML at all.  So much nicer this way!

classes/helpers/FrmFormsHelper.php
```diff
@@ -291,6 +291,9 @@ public static function get_draft_link() {
         return $link;
     }
 
+
+
+
```
stephywells: Looks like there is some extra space here that needs to be removed.

classes/helpers/FrmFieldsHelper.php
```diff
@@ -372,6 +372,58 @@ private static function &label_position( $position, $field, $form ) {
...
+	 *
+	 * @return string
+	 */
+	public static function maybe_return_inline_submit_top_class( $form ) {
+
+		if ( self::form_has_top_labels( $form ) ) {
+			return ' frm_inline_submit_top';
+		}
+
+		return '';
```
stephywells: Does this feel any cleaner to you?

return self::form_has_top_labels( $form ) ? ' frm_inline_submit_top' : '';

classes/helpers/FrmFormsHelper.php
```diff
@@ -481,10 +499,12 @@ public static function get_form_style_class( $form = false ) {
 			$form = $form->options;
 		}
 
+
```
stephywells: Do you see the Scrutinizer notices when you view this page?
https://github.com/Strategy11/formidable-forms/pull/73/files

Looks like there's an extra line here too.

classes/helpers/FrmFieldsHelper.php
```diff
@@ -372,6 +372,58 @@ private static function &label_position( $position, $field, $form ) {
...
+	 */
+
+	private static function form_has_top_labels( $form ) {
+		$fields = $form['fields'];
+
+		if ( count( $fields ) <= 0 ) {
+			return false;
+		}
+
+		return self::field_has_top_label( reset( $fields ), $form );
```
stephywells: This is checking the label position for the first field. Do you think it would be any better to check the last field instead? Which way would be most reliable?

css/_single_theme.css.php
```diff
@@ -553,6 +556,12 @@
     vertical-align:middle;
 }
 
+.frm_inline_form.<?php echo esc_html( $style_class ) ?> input[type=submit],
+.frm_inline_form.<?php echo esc_html( $style_class ) ?> .frm_submit input[type=button],
+.frm_inline_form.<?php echo esc_html( $style_class ) ?> .frm_submit button{
+	margin: 0 <?php echo esc_html( $important ) ?>;
+}
```
stephywells: Maybe we should combine this section with the one at line 508?
https://github.com/Strategy11/formidable-forms/pull/73/files#diff-a1eccabbb05681e6395f97699a181f58R509

css/_single_theme.css.php
```diff
@@ -504,6 +505,8 @@
 	visibility:hidden;
 }
 
+.frm_inline_form.frm_inline_submit_top.<?php echo esc_html( $style_class ) ?> .frm_submit input,
+.frm_inline_form.frm_inline_submit_top.<?php echo esc_html( $style_class ) ?> .frm_submit button,
 .<?php echo esc_html( $style_class ) ?> .frm_submit.frm_inline_submit input,
 .<?php echo esc_html( $style_class ) ?> .frm_submit.frm_inline_submit button {
 	margin-top: 0 <?php echo esc_html( $important ) ?>;
```
stephywells: This works better for me with margin:0 instead of margin-top:0

stephywells:
These changes have been added into 3.0, so I won't merge this one. But it's there. Thanks for all your work on this Laura!

PR #75, opened 2017-11-23
Fix unit tests (2.x) (Work In Progress)

jorgeatorres:
I'm going through the unit tests to see why some are not passing and fixing what needs to be fixed.

In Travis some of the builds are failing due to the tests using some reflection functions that weren't introduced until PHP 5.3.2 (this should be fixed now).
Some other builds seem to be failing due to Composer version constraints (not addressed yet).

I'm still running some of the tests, so this is *work in progress*. Will update the PR when I'm done.

### DIFF
--- a/tests/entries/test_FrmEntry.php
+++ b/tests/entries/test_FrmEntry.php
@@ -40,6 +40,10 @@ class WP_Test_FrmEntry extends FrmUnitTest {
 			$this->markTestSkipped( 'Pro is not active' );
 		}
 
+		if ( ! version_compare( PHP_VERSION, '5.3.2', '>=' ) ) {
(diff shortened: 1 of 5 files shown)


stephywells:
Thanks so much for starting on this. I actually added a function in the 3.0 tests to help with this. It would be used like this:

$this->check_php_version( '5.3.2' );

Maybe it would be better to work with this in the 3.0. It's possible we are making the same fixes.

jorgeatorres:
Got it. I'm reviewing the 3.0 branch now. Maybe after applying the fixes to 3.0 we can come port them to 2.x so all branches pass.

stephywells:
That would be wonderful!

stephywells:
I'm going to go ahead and close this since it's outdated and in the 3.0 version.

PR #77, opened 2018-01-09
Workaround WP unit testing bootstrap issue for multisite

jorgeatorres:
The WP-develop bootstrap file only hooks into `pre_option` for options added to the `wp_tests_options` array, but `pre_option` is not used by `get_site_option()`, which affects multisite testing. Even though Pro is installed and active, the `is_pro_active` var shared by all unit tests is set to `false` because of this, and producing some errors such as field counts not matching.


stephywells:
Awesome! Thanks so much!

PR #78, opened 2018-01-19
Adds field search by name, id, and key to customization panel.

LauraSteiner:


stephywells:
Awesome! This works just perfectly!

I think it would look a lot better if the search box spans 100% by moving right after the alignleft ul.

Also, perhaps we should only show the search box if there are more than 6 fields listed? I'm not sure that this one really matters or not.

Once the search box is bigger, I'm good to merge this. Great job! Simple and works great.

LauraSteiner:
Thank you for your wonderful message!  I've added a commit with the full-width search box.

stephywells:
Love it! This has been merged.

PR #80, opened 2018-02-11
Fix sort by a number/scale/star on entries page

jorgeatorres:
Fixes https://github.com/Strategy11/formidable/issues/1881.


stephywells:
Looks good. Thanks so much!

PR #82, opened 2018-02-16
Fix Travis build on PHP 7.1 + master

jorgeatorres:
Travis uses PHPUnit 7 on PHP 7.1, which results in this [compat code from WordPress-develop](https://github.com/WordPress/wordpress-develop/blob/master/tests/phpunit/includes/phpunit6-compat.php#L18) being executed:

```
class PHPUnit_Util_Test extends PHPUnit\Util\Test {
....
}
```

but `PHPUnit\Util\Test` is a **final** class on PHPUnit 7.0, which produces a fatal error and the failure of the build:

```
PHP Fatal error:  Class PHPUnit_Util_Test may not inherit from final class (PHPUnit\Util\Test) in /tmp/wordpress/tests/phpunit/includes/phpunit6-compat.php on line 18
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ install:
 - if [[ $TRAVIS_PHP_VERSION == "5.2.17" || $TRAVIS_PHP_VERSION == "5.3" ]]; then phpenv local --unset; fi
 
 before_script:
+- phpunit --version
 - export PHPCS_DIR=/tmp/phpcs
 - export SNIFFS_DIR=/tmp/sniffs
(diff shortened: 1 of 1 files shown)


jorgeatorres:
Ok, so it seems it's something related to the `precise` environment in particular, which was used to run some of the past Travis builds.
I still think it'd be useful to print the PHPUnit version during the build, just to be able to double-check when things like this happen.

@stephywells: what do you think?

stephywells:
Definitely. Thank you!

PR #83, opened 2018-02-24
Extra space in failed_msg text

jeremylind:
There is an extra space in the default `failed_msg` text string.


stephywells:
Thanks so much! And sorry for the delay on this as well.

PR #84, opened 2018-02-28
Turkish Translation 1

mugurcagdas:
Dear Sirs,
I created language file for my national people. formidable file translated from EN to TUR.
This is part1. I will continue to translate whole file. 
Could you please merge it?
Thanks in advance.


stephywells:
Thanks so much for the translation and sorry for the slow response. It would actually be best if you could add your translations here:
https://translate.wordpress.org/locale/tr/default/wp-plugins/formidable

Thanks again! This is awesome!

PR #85, opened 2018-03-20
Add new frm_include_field_in_content filter in include_field_in_content method

benjaminniess:
So we can choose which fields has to be hidden from [default-message] email content.


stephywells:
Thanks for the PR! Looks good to me.

PR #89, opened 2018-04-12
Detects possible errors in user calculations in the admin and alerts …

LauraSteiner:
…user.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -845,6 +845,10 @@ class FrmFormsController {
 		if ( isset( $form->options['js_validate'] ) && $form->options['js_validate'] ) {
 			echo ' frm_js_validate ';
 		}
+
+		if ( current_user_can( 'activate_plugins' ) && current_user_can( 'frm_edit_forms' ) ) {
+			echo ' frm-admin-viewing ';
+		}
 	}
 
 	public static function get_email_html() {
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -607,6 +607,140 @@ function frmAdminBuildJS(){
 		initiateMultiselect();
 	}
 
+	/**
+	 * Checks a string for parens, brackets, and curly braces and returns a message if any unmatched are found.
+	 * @param formula
+	 * @returns {string}
+	 */
+	function checkMatchingParens( formula ) {
+
+		var stack = [],
+			formula_array = formula.split( '' ),
+			length = formula_array.length,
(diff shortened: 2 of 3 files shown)


js/formidable_admin.js
```diff
@@ -607,6 +607,140 @@ function frmAdminBuildJS(){
...
+	 * Checks a calculation for shortcodes that shouldn't be in it and returns a message if found.
+	 * @param calculation
+	 * @param inputElement
+	 * @returns {string}
+	 */
+	function checkShortcodes( calculation, inputElement ) {
+		var nonNumericShortcodesMessage = checkNonNumericShortcodes( calculation, inputElement );
+		var nonFormShortcodesMessage = checkNonFormShortcodes( calculation );
+
+		var msg = nonNumericShortcodesMessage;
```
stephywells: This could be simplified a bit. How about 
```
var msg = checkNonNumericShortcodes( calculation, inputElement );
msg += checkNonFormShortcodes( calculation );
```

js/formidable_admin.js
```diff
@@ -607,6 +607,140 @@ function frmAdminBuildJS(){
...
+		msg += nonFormShortcodesMessage;
+
+		return msg;
+	}
+
+	function checkCalculationCreatedByUser() {
```
stephywells: Would you mind reordering the functions a bit so we can read them top to bottom like you did with your math shortcode? I find it hard to follow when I have to scroll up to find the functions.

js/formidable_admin.js
LauraSteiner: Nice!  Done.

js/formidable_admin.js
LauraSteiner: Happy to!  Done.

LauraSteiner:
Thanks, Steph!  I appreciate your insights.

classes/controllers/FrmFormsController.php
```diff
@@ -845,6 +845,10 @@ public static function form_classes( $form ) {
 		if ( isset( $form->options['js_validate'] ) && $form->options['js_validate'] ) {
 			echo ' frm_js_validate ';
 		}
+
+		if ( current_user_can( 'activate_plugins' ) && current_user_can( 'frm_edit_forms' ) ) {
```
stephywells: Sorry... One more change. It looks like this should be in Pro since it isn't used here.

LauraSteiner:
frm-admin-viewing has been removed from free and added to pro.

PR #90, opened 2018-05-19
Adds conditional logic to submit button.

LauraSteiner:


LauraSteiner:
I'm not sure where to hook this.  In the previous commit, I added $form as an argument to frm_before_gt_form.  In this latest commit, I added a new action.   I'd appreciate hearing your thoughts on this.

PR #93, opened 2018-06-02
Adds safeguards against submitting when not allowed.

LauraSteiner:


js/formidable.js
```diff
@@ -877,6 +877,12 @@ function frmFrontFormJS(){
 
 			e.preventDefault();
 
+			if ( typeof frmProForm !== 'undefined' ) {
+				if ( ! frmProForm.submitAllowed( object) ) {
```
stephywells: It's possible that frmProForm could be defined, but not submitAllowed. This could happen if the free version is updated and Pro is not. Maybe we can add an additional check for this?

LauraSteiner: Great thinking!  Added.

js/formidable.js
```diff
@@ -877,6 +877,12 @@ function frmFrontFormJS(){
 
 			e.preventDefault();
 
+			if ( typeof frmProForm !== 'undefined' && typeof frmProForm.submitAllowed === 'function' ) {
+				if (    ! frmProForm.submitAllowed( object) ) {
```
stephywells: Looks like some extra space ended up in here.

PR #96, opened 2018-09-19
Reduces cases that trigger admin calculation warning.

LauraSteiner:
Thank you for this -- https://github.com/Strategy11/formidable/issues/2049.

I like this approach much more.  It should eliminate all or most of the false positives that annoy people.  

I've allowed more formats with created_at to trigger the warning.  Is that okay?  If not, I'd be happy to do just Y-m-d.  And make any other changes you'd like.


stephywells:
Looks great. Thank you!

PR #105, opened 2019-01-11
Hides Advanced Settings in blocks using script tags.

LauraSteiner:
A non-elegant approach to hiding blocks' advanced settings that works in my testing.

The formatting fixes to radio.js were from running eslint with the --fix flag on the whole js/src directory.  I left them in in case you like the changes.


stephywells:
Looks great! Thank you!

I was still getting an error when I added a block, added a class, and then selected a form. I added className to the block parameters to prevent the error. We can decide later whether to use the class or remove it once there is a way.

LauraSteiner:
Awesome!  So glad you added this.

PR #106, opened 2019-01-22
React styler -- proof of concept

LauraSteiner:
1. The React styler is on the normal Styles page, near the top.  It takes a second or two to load.
2. The code isn't efficient, organized, consistent, etc.  
3. I like the color picker used, but it turns a 12 kb file into a 164 kb file.
4. React gives us tremendous flexibility.  I've included two different ways to choose colors, which is a taste of the possibilities.


stephywells:
This looks like a great start! I'm going to go ahead and close this since we won't be merging it yet.

PR #108, opened 2019-02-11
iOS Safari can't add default values

tiagohillebrandt:
Fixes https://github.com/Strategy11/formidable/issues/2124.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -9,7 +9,7 @@ function frmAdminBuildJS(){
 
 	// Global settings
 	var s;
-	
+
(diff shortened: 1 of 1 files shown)


stephywells:
Thanks @tiagohillebrandt! This change looks like a good approach. However, it doesn't look like it solves the issue completely. It seems this doesn't work in Chrome either, and the problem is with almost every text field in the field settings.

But... I've started working on our new 4.0 design, and this isn't an issue there. Let's not spend more time on this since it will be a non-issue in a couple months. Thank you!

PR #109, opened 2019-03-01
Resend and other single action types only fire if toggled on.

LauraSteiner:
I didn't want to change this much out of concern that something else somewhere would break.  Do you think this approach is okay?  Would you prefer a different one?

### DIFF
--- a/classes/models/FrmFormAction.php
+++ b/classes/models/FrmFormAction.php
@@ -386,7 +386,7 @@ class FrmFormAction {
 		$limit = apply_filters( 'frm_form_action_limit', $atts['limit'], compact( 'type', 'form_id' ) );
 
         if ( 'all' != $type ) {
-            return $action_controls->get_all( $form_id, $limit );
+			return $action_controls->get_all( $form_id, $limit, 'publish' );
         }
 
 		$args = self::action_args( $form_id, $limit );
@@ -458,7 +458,7 @@ class FrmFormAction {
 		return ! empty( $payment_actions );
 	}
 
-	public function get_all( $form_id = false, $limit = 99 ) {
+	public function get_all( $form_id = false, $limit = 99, $post_status = 'any' ) {
 	    if ( $form_id ) {
 	        $this->form_id = $form_id;
 	    }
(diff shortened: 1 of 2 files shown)


classes/models/FrmFormAction.php
```diff
@@ -386,7 +386,7 @@ public static function get_action_for_form( $form_id, $type = 'all', $atts = arr
 		$limit = apply_filters( 'frm_form_action_limit', $atts['limit'], compact( 'type', 'form_id' ) );
 
         if ( 'all' != $type ) {
-            return $action_controls->get_all( $form_id, $limit );
+			return $action_controls->get_all( $form_id, $limit, 'publish' );
```
stephywells: I’m thinking this may cause unexpected changes. Perhaps we should replace $limit here with $atts and use self::prepare_get_action( $atts ) in the get_all function so we don’t add too many parameters. 

The, we can step back and add the publish post status in the trigger function where we know for sure that we only want actions that are turned on. 

What do you think?

LauraSteiner: I think I may not understand your suggestion properly.

If we use self::prepare_get_action($atts) in the get_all function, that looks to me like it will set the post_status default to "publish."  Currently, "any" is the post_status used in get_all.  We definitely want to change to "publish" for triggering actions.  In duplicate, we'd want 'any,' I'd think. 

Have I misunderstood your suggestion?  I'd like to understand it.  I'd appreciate any clarification!

stephywells: You're completely right. I missed that. We would need to pass a param to set the default when set from the get_all function.

classes/models/FrmFormAction.php
```diff
@@ -458,7 +458,7 @@ public static function form_has_action_type( $form_id, $type ) {
 		return ! empty( $payment_actions );
 	}
 
-	public function get_all( $form_id = false, $limit = 99 ) {
+	public function get_all( $form_id = false, $limit = 99, $post_status = 'any' ) {
```
stephywells:
```suggestion
	public function get_all( $form_id = false, $atts = array ) {
		self::prepare_get_action( $atts, 'any' );
```

classes/models/FrmFormAction.php
stephywells:
```suggestion
			return $action_controls->get_all( $form_id, $atts );
```

classes/models/FrmFormAction.php
LauraSteiner: To make sure I understand, is your suggestion that we add an extra param to prepare_get_action which is the status we want to set? 

Currently, prepare_get_action only has one arg, $args/$atts.

How do you think it should work?  If status isn't set in $atts, then use the status passed as a separate arg?  If that doesn't have a value, the default of publish will be used?

stephywells: Yep, exactly.

stephywells:
I went ahead and finished it up so I can get this release out. Thanks for starting on this @LauraSteiner!

PR #110, opened 2019-03-06
Hides other input when Formidable styling disabled

LauraSteiner:


stephywells:
Thanks @LauraSteiner!

PR #114, opened 2019-06-06
Add better sanitization of submitted input to solve the '& entity' problem

lanresmith:


classes/models/fields/FrmFieldUrl.php
```diff
@@ -86,4 +86,11 @@ protected function prepare_display_value( $value, $atts ) {
 
 		return $value;
 	}
+
+	/**
+	 * @since 4.0
+	 */
+	public function sanitize_value( &$value ) {
+		FrmAppHelper::sanitize_value( 'esc_url_raw', $value );
```
stephywells: I'm not sure about this one... Would sanitize_text_field be better here rather than escaping?

classes/models/fields/FrmFieldTextarea.php
```diff
@@ -64,4 +64,11 @@ public function front_field_input( $args, $shortcode_atts ) {
 			FrmAppHelper::esc_textarea( $this->field['value'] ) .
 			'</textarea>';
 	}
+
+	/**
+	 * @since 4.0
+	 */
+	public function sanitize_value( &$value ) {
+		FrmAppHelper::sanitize_value( 'sanitize_textarea_field', $value );
```
stephywells: Textarea fields may have HTML, so we can't strip the HTML here.

classes/models/fields/FrmFieldHTML.php
```diff
@@ -53,4 +53,15 @@ public function get_container_class() {
 	protected function include_form_builder_file() {
 		return FrmAppHelper::plugin_path() . '/classes/views/frm-fields/back-end/field-html.php';
 	}
+
+	/**
+	 * @since 4.0
+	 */
+	public function sanitize_value( &$value ) {
```
stephywells: I don't think this is needed since HTML fields aren't saved with the entry. They are only saved from the backend while building the form.

classes/models/fields/FrmFieldType.php
```diff
@@ -1051,4 +1051,11 @@ protected function fill_values( &$value, $defaults ) {
 			$value = array_merge( $defaults, (array) $value );
 		}
 	}
+
+	/**
+	 * @since 4.0
+	 */
+	public function sanitize_value( &$value ) {
+		FrmAppHelper::sanitize_value( 'sanitize_text_field', $value );
```
stephywells: For reverse compatibility, perhaps we should use wp_kses_post here and override it in the fields we know it is safe to use sanitize_text_field. With custom plugins, we don't have a way of knowing what should be allowed and what shouldn't. I'm concerned that switching the default may cause issues.

classes/models/fields/FrmFieldUrl.php
lanresmith: Actually, there used to be sanitize_url which was like the sanitize_text_field for URL's, but it was later deprecated & esc_url_raw is the replacement. Kindly see https://developer.wordpress.org/reference/functions/sanitize_url/

classes/models/fields/FrmFieldHTML.php
lanresmith: OK, will fix that

classes/models/fields/FrmFieldTextarea.php
lanresmith: OK, wp_kses_post (with wp_specialchars_decode) should do right? like for rte

classes/models/fields/FrmFieldType.php
lanresmith: OK. So, should it be wp_kses_post (with wp_specialchars_decode)?

classes/models/fields/FrmFieldHTML.php
lanresmith: so, it should be an empty override, right?

stephywells: I don't think an override is needed at all, since it shouldn't ever be called anyway. We should be able to safely remove this function.

classes/models/fields/FrmFieldUrl.php
lanresmith: @stephywells should I leave it like that or change it to sanitize_text_field?

stephywells: Okay, looks good then. Thanks for pointing that out!

classes/models/fields/FrmFieldTextarea.php
stephywells: Yep. This one wouldn't need an override once we adjust the sanitizing in FrmFieldType.

lanresmith: yes

classes/models/fields/FrmFieldHTML.php
lanresmith: OK

classes/models/fields/FrmFieldType.php
stephywells: I'm not sure about using wp_specialchars_decode. What does this do to the < and > characters in these two cases?
1. '1 > 2'
2. <div class="here"></div>

I'm thinking unit tests would be helpful for this as well.

lanresmith: They're left unchanged in both cases.

wp_kses_post **without** wp_specialchars_decode will get us back to square one as I've put in the comment. cos wp_kses_* family turn '&' to '&amp;'. Turning '&' to '&amp;' is what's causing that Salesforce issue - Salesforce API receives '&amp;' (and that's what is actually stored in our db), so when they try to escape it on their page on output, it then results in double encoding (it's like their - I mean Salesforce - escaping mechanism works similar to PHP's htmlspecialchars; the latter too doesn't check if there's already encoding of entities, like WP's esc_html would do). So the '&' that begins '&amp;' is encoded again & then you get '&amp;amp;' (in the view source), which then shows in the browser as &amp;

lanresmith: @stephywells please review the new changes. Thanks.

lanresmith:
@stephywells please review the new changes. Thanks.

classes/models/fields/FrmFieldDefault.php
```diff
@@ -50,4 +50,11 @@ public function front_field_input( $args, $shortcode_atts ) {
 
 		return $input_html;
 	}
+
+	/**
+	 * @since 4.0
+	 */
+	public function sanitize_value( &$value ) {
+		FrmAppHelper::sanitize_value( 'sanitize_text_field', $value );
```
stephywells: Let's not override this one since we don't know what the input could be. This function can be removed.

lanresmith: OK

lanresmith:
@stephywells please review the new changes. Thanks.

lanresmith:
@stephywells I just pushed the new changes. Please review. Thanks.

PR #116, opened 2019-06-11
Adds frm-width-auto class to admin CSS

LauraSteiner:
Adds frm-width-auto class for text and select fields in the admin.  https://github.com/Strategy11/formidable-pro/issues/2257


stephywells:
Thanks @LauraSteiner! This actually already exists:
https://github.com/Strategy11/formidable-forms/blob/master/css/frm_admin.css#L2522-L2525

PR #117, opened 2019-06-16
Issue/pro/2248 - Required repeater field not redisplayed on form error if removed

lanresmith:


classes/helpers/FrmAppHelper.php
```diff
@@ -2233,6 +2233,25 @@ public static function locales( $type = 'date' ) {
 		return $locales;
 	}
 
+	/*
+	 * Check if array has at least one numeric key.
+	 */
```
stephywells: Let's include `@since` on this to make it easier to use in pro/add-ons. This is good for reference so we can later remove the is_callable check.

Is this function being used in the lite version? If not, perhaps we should move it to Pro.

classes/helpers/FrmAppHelper.php
```diff
@@ -2233,6 +2233,25 @@ public static function locales( $type = 'date' ) {
...
+	public static function has_numeric_key( $arr ) {
+		if ( ! is_array( $arr ) || empty( $arr ) ) {
+			return false;
+		}
+
+		$has_numeric_key = false;
+		foreach ( $arr as $k => $v ) {
+			if ( is_int( $k ) ) {
```
stephywells: With repeaters, on edit the key is not numeric. It's "i232" where the number is the id of the nested entry. So this isn't working when editing an entry.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -300,6 +300,19 @@ public static function set_posted_value( $field, $value, $args ) {
 		if ( empty( $args['parent_field_id'] ) ) {
 			$_POST['item_meta'][ $field->id ] = $value;
 		} else {
+			// init arrays if necessary, else we get fatal error
+			if ( isset( $_POST['item_meta'][ $args['parent_field_id'] ] ) && 
+				 is_array( $_POST['item_meta'][ $args['parent_field_id'] ] ) ) {
+
+				if ( ! isset( $_POST['item_meta'][ $args['parent_field_id'] ][ $args['key_pointer'] ] ) ||
```
stephywells: Maybe we can shorten this line a bit by setting a variable?
$row = $_POST['item_meta'][ $args['parent_field_id'] ]

classes/helpers/FrmEntriesHelper.php
```diff
@@ -300,6 +300,19 @@ public static function set_posted_value( $field, $value, $args ) {
 		if ( empty( $args['parent_field_id'] ) ) {
 			$_POST['item_meta'][ $field->id ] = $value;
 		} else {
+			// init arrays if necessary, else we get fatal error
```
stephywells: Let's simplify the code a bit by moving these new additions to their own function.

classes/helpers/FrmEntriesHelper.php
lanresmith: Hmm... The issue is that $row = $_POST['item_meta'][ $args['parent_field_id'] ] will throw a notice if the key equivalent to the value of $args['parent_field_id'] is not set on $_POST['item_meta']; meaning that line itself requires an 'if isset', or at least a ternary op. This will introduce more control flow logic & maybe increase cyclomatic complexity... Or what do you think?

stephywells: That's fine. It looks a lot better now that it's broken out. Thanks!

PR #118, opened 2019-07-01
Enable comma-separated formatting for Mailchimp tags settings field

lanresmith:


js/formidable_admin.js
```diff
@@ -3269,21 +3269,64 @@ function frmAdminBuildJS() {
 
 	function insertContent( content_box, variable ) {
 		if ( document.selection ) {
-			content_box[0].focus();
-			document.selection.createRange().text = variable;
-		} else if ( content_box[0].selectionStart ) {
+			// ie8
+			if ( 'Text' === document.selection.type && content_box[0] === document.selection.createRange().parentElement() ) {
```
stephywells: Since we no longer support IE8 in the backend, let's go ahead and remove the IE8 specific fallback.

js/formidable_admin.js
```diff
@@ -3269,21 +3269,64 @@ function frmAdminBuildJS() {
...
+				var input = content_box[0];
+				input.focus();
+				variable = maybeFormatInsertedContent( content_box, variable, input.value.length, input.value.length );
+				input.value = input.value + variable; // simply append
+			}
+		} else {
 			obj = content_box[0];
 			var e = obj.selectionEnd;
+
+			variable = maybeFormatInsertedContent( content_box, variable, content_box[0].selectionStart, content_box[0].selectionEnd );
```
stephywells: Maybe obj.selectionStart and e could be used since they are already defined?

js/formidable_admin.js
```diff
@@ -3269,21 +3269,64 @@ function frmAdminBuildJS() {
...
+		var input          = textRange.parentElement(),
+			selectionStart = input.value.indexOf( textRange.text ),
+			selectionEnd   = selectionStart + textRange.text.length;
+
+		return [ selectionStart, selectionEnd ];
+	}
+
+	function maybeFormatInsertedContent( input, textToInsert, selectionStart, selectionEnd ) {
+		var name = input.attr( 'name' );
+		if ( false === /\[FRM_TAGS\]$/.test( name ) ) {
```
stephywells: Maybe we can make this more flexible by using a data attribute like data-sep=",". If data-sep is undefined, then return.

js/formidable_admin.js
```diff
@@ -3269,21 +3269,64 @@ function frmAdminBuildJS() {
...
+			return textToInsert;
+		}
+
+		var value = input.val();
+
+		if ( ! jQuery.trim( value ).length ) {
```
stephywells: We've been working toward removing jQuery. Perhaps we can change this to value.trim.length?

js/formidable_admin.js
```diff
@@ -3268,22 +3268,45 @@ function frmAdminBuildJS() {
 	}
 
 	function insertContent( content_box, variable ) {
-		if ( document.selection ) {
```
stephywells: Looks like the original if else was lost. This behaves a bit differently depending on whether text on the box has been selected, if the cursor is in the box, or if it isn't.

lanresmith: Thought we were no longer supporting ie8. No problem, I'll add it back.

js/formidable_admin.js
```diff
@@ -3271,19 +3271,46 @@ function frmAdminBuildJS() {
 		if ( document.selection ) {
 			content_box[0].focus();
 			document.selection.createRange().text = variable;
-		} else if ( content_box[0].selectionStart ) {
+		} else {
```
stephywells: Looks like the previous else if here is missing, along with the final else

lanresmith: It was intentionally left out as it was giving some undesirable results; You can try the old structure - I mean the `else if` & 'else' with their content & see how it 'maltreats' delimiting when the cursor is at the beginning while trying to insert a shortcode . Actually, it shouldn't matter whether the cursor is at the beginning or elsewhere when trying to insert a shortcode. The current code covers all cases well.

Thanks.

stephywells: Sound good to me. Thanks for double checking that!

PR #120, opened 2019-07-08
Pro: Limit the number of boxes checked

lanresmith:


stephywells:
Thanks @lanresmith! Can you tell me a bit more about why this limit is here? Are there performance issues without it?

lanresmith:
> Thanks @lanresmith! Can you tell me a bit more about why this limit is here? Are there performance issues without it?

@stephywells You're welcome. Actually, it's just to be kind of 'consistent', as the script checking that the limit setting is within reasonable bounds for the repeater field type also exists in this js file of the lite version.

stephywells:
@lanresmith can you tell me why the limit check for the setting is needed? Would something bad happen if the limit is set to 1 or 250?

lanresmith:
> @lanresmith can you tell me why the limit check for the setting is needed? Would something bad happen if the limit is set to 1 or 250?

@stephywells nothing bad should happen with 1 or 250. Just putting the, maybe, seemingly obvious into consideration - like the Repeater field's case which starts at 2 too. E.g. if you want 1, then you probably would be using radio buttons instead. & the 200... well, I'm yet to see a form with up to 200 checkboxes for a field let alone more than :)

lanresmith:
@stephywells Please review the new changes. Thanks.

As for `FrmEntriesHelper::get_posted_value` I only modified it to better support array type for `$field`, not that it gave any problems. If you check the old code properly, you may realise that the check for array type should also have been done at the beginning - when trying to get `$field_id` - not only when trying to get `$field_obj`, but we've been getting away with it cos the `$field` param hardly arrives at the method as an array.

I also shifted the retrieval of `$field_obj` upward just to avoid double checking for the type of the `$field` param. Hope you get me.

js/formidable_admin.js
```diff
@@ -1799,6 +1799,15 @@ function frmAdminBuildJS() {
 		}
 	}
 
+	function checkCheckboxSelectionsLimit() {
+		/*jshint validthis:true */
+		var val = this.value;
+		if ( val !== '' && ( val < 2 || val > 200 ) ) {
```
stephywells: I'm thinking we should allow 1 here. I know it isn't best practice, but we do unfortunately get a lot of people who ask how to only check 1 box in a checkbox. I think allowing 1 would reduce support tickets on this.

lanresmith: OK

classes/views/frm-fields/front-end/checkbox-field.php
```diff
@@ -19,7 +19,13 @@
 		$field_val = FrmFieldsHelper::get_value_from_array( $opt, $opt_key, $field );
 		$opt = FrmFieldsHelper::get_label_from_array( $opt, $opt_key, $field );
 
-		$checked = FrmAppHelper::check_selected( $field['value'], $field_val ) ? ' checked="checked"' : '';
+		$checked = ''; // init
+		if ( ! FrmFieldsHelper::is_other_opt( $opt_key ) ) {
+			// Let the checked state of 'Other' fields be determined solely by FrmFieldsHelper::prepare_other_input as below.
+			// Without this check, one 'Other' field being checked leads to making all 'Other' fields checked on submit error
+			// since they all have the same value attr of 'Other'.
+			$checked = FrmAppHelper::check_selected( $field['value'], $field_val ) ? ' checked="checked"' : '';
```
stephywells: This prevents all other options from being checked after an error. I think the prior way is preferable. In most cases, if there are multiple other options, the labels on each on would be different.

lanresmith: This doesn't seem to be happening at my side. Can you please give an example of how to reproduce this?

stephywells: 1. Add a checkbox field and a required text field.
2. Add 4 'Other' options. Change the labels on 2 of them, and leave 2 as 'Other'.
3. Preview the form and check all 4 of the other boxes.
4. Leave the text field blank and hit submit.
5. Now there are no Other boxes checked.

lanresmith: Hmm... But it's only the boxes which were only checked while their text inputs contained nothing that won't be re-checked on submit error, which IMHO, seems OK.

Do you still feel we should leave it the old way?

stephywells: Oh! You're right! Your way looks better. We can merge it.

PR #126, opened 2019-09-02
Add Summary Field

lanresmith:


css/frm_admin.css
```diff
@@ -722,6 +722,10 @@ p.frm_has_shortcodes {
 	font-size: 15px;
 }
 
+.howto a {
+	font-size: inherit !important;
```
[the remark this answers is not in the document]
lanresmith: The size of the link text in the 'Exclude Fields' box is being affected by `.frm-right-panel .frm-inline-modal .inside a` (I met this there) - that's what's making it bigger than the surrounding text - & I thought that there's actually no reason for a link placed in a 'how to' instruction not to have the same size as its siblings in the 'how to' parent, so I added `important` pro-actively, not actually as a fast way to fix the issue.

Do you still feel that we should change it?

css/frm_admin.css
```diff
@@ -5768,6 +5789,12 @@ ul .frm_col_two {
 	float: left;
 }
 
+.field_type_list::after {
```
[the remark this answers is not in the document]
lanresmith: @stephywells It's not actually for the summary field. It's just to make `.field_type_list` have a non-zero height as it contains floated children. You may inspect with dev tools to see what I mean better.

css/frm_admin.css
```diff
@@ -5802,10 +5829,18 @@ ul .frm_col_two {
 	border-radius: var(--border-radius) 0 0 var(--border-radius);
 }
 
-#frm-insert-fields li a:hover {
+#frm-insert-fields li a:not(.disabled):hover {
 	background: var(--sidebar-hover);
 }
 
+#frm-insert-fields li a.disabled:hover {
```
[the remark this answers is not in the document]
lanresmith: @stephywells I really don't get what you mean here because I added this myself manually, not from an old commit.

css/frm_admin.css
```diff
@@ -5876,8 +5911,8 @@ input[disabled],
 	font-size: inherit;
 }
 
-#frm-insert-fields li a:hover .frmsvg,
-#frm-insert-fields li a:hover i {
+#frm-insert-fields li a:not(.disabled):hover .frmsvg,
```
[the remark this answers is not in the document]
lanresmith: @stephywells It's just to prevent the darker shade of grey that shows when you hover from showing for a disabled item, specifically when a summary field is already in the form & you cannot add more than one of it to a form. All these (including the `#frm-insert-fields li a.disabled:hover` rule above) are just to make the UI show that the item is indeed disabled & cannot accept clicks.

Do you still feel that we should change it?

js/formidable_admin.js
```diff
@@ -715,9 +737,25 @@ function frmAdminBuildJS() {
 	function addFieldClick() {
 		/*jshint validthis:true */
 		var $thisObj = jQuery( this );
-		var field_type = $thisObj.closest( 'li' ).attr( 'id' );
+		// there is no real way to disable a <a> (with a valid href attribute) in HTML - https://css-tricks.com/how-to-disable-links/
+		if ( $thisObj.hasClass( 'disabled' ) ) {
+			return false;
+		}
+
+		var $button = $thisObj.closest( 'li' );
```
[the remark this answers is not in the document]
lanresmith: I don't think it can get the wrong element, but I'll change it to `frmbutton`

css/frm_admin.css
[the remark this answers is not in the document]
stephywells: Perhaps it can be more specific so it only applies to the howtos you mentioned?

css/frm_admin.css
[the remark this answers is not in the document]
stephywells: I was thinking this applied to the inactive Pro fields that are listed in Lite. But it doesn’t look that way. I think it can stay then. Thanks for checking!

lanresmith:
@stephywells So, I've merged in the issue38-preview branch & made the changes requested. Kindly review. Thanks!

css/frm_admin.css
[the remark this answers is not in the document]
stephywells: Is the zero height causing a problem? If we add this, the ul will also need margin-bottom:0 added.

css/frm_admin.css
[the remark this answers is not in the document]
stephywells: I see. I was thinking this applied to the fields that are disabled when pro isn't installed. Nevermind.

css/frm_admin.css
```diff
@@ -2541,6 +2545,15 @@ a.frm_option_icon:hover::before {
 	left: 25%
 }
 
+.frm-summary-message {
+	top: 50%;
+	left: 50%;
+	padding: 15px 0;
+	color: var(--primary-color);
```
stephywells: Let's remove the color here so it matches the other fields. We ended up making them black instead of blue so they don't look like links.

js/formidable_admin.js
```diff
@@ -618,6 +618,13 @@ function frmAdminBuildJS() {
 	 */
 	function insertNewFieldByDragging( selectedItem, fieldButton, opts ) {
 		var fieldType = fieldButton.attr( 'id' );
+		var addBtn = fieldButton.children( '.frm_add_field' );
```
stephywells: Maybe we should only fetch this if it's a summary field?

js/formidable_admin.js
```diff
@@ -975,6 +1049,44 @@ function frmAdminBuildJS() {
 		return /\[id\]|\[key\]|\[if\s\w+\]|\[foreach\s\w+\]|\[created-at(\s*)?/g;
 	}
 
+	function isSummaryCalcBox( box ) {
+		var list = jQuery( box ).find( '.frm_code_list' );
+		return 1 === list.length && list.hasClass( 'frm_js_summary_list' );
+	}
+
+	function extractExcludedOptions( exclude ) {
+		var opts = new Array();
```
stephywells: Will you please run this file through JShint and make adjustments for anything that is new?

css/frm_admin.css
lanresmith: no problem, I'll remove the `.field_type_list::after` rule.

js/formidable_admin.js
lanresmith: @stephywells I'm not sure I understand what you mean here please

js/formidable_admin.js
lanresmith: OK

lanresmith:
@stephywells please review. Thanks!

js/formidable_admin.js
stephywells: Sorry. I mean maybe line 621 should be moved to line 625 so it's inside the `if ( 'summary' === fieldType ) {`

js/formidable_admin.js
stephywells: I'm still seeing errors with this like 'array literal notation is prefered'

If you don't have jshint set up locally, you can use it here:
https://jshint.com/

stephywells:
@lanresmith can you please review each of my comments to make sure they have been resolved?

js/formidable_admin.js
lanresmith: OK, done

js/formidable_admin.js
lanresmith: fixed

lanresmith: there are however some existing issues therein like: `The body of a for in should be wrapped in an if statement to filter unwanted properties from the prototype`. I left some (that don't really seem to be an issue) for the fear of mistakenly introducing bugs :)

lanresmith:
@stephywells kindly review please. I hope I haven't left anything out again :). Thanks!

stephywells:
I think this is all ready to merge. Once the last changes are made in Pro, we can merge both at the same time. 

Great work on this @lanresmith! It looks awesome!

cecilehabran:
Would that be possible to show the text of a dropdown field instead of the value in the summary ? Right now it displays the selected value, which can be quite ugly or unclear (like an id).

PR #129, opened 2019-10-14
Adding a `frm_pre_get_one` filter 

benjaminniess:
Hi,

Could we add this filter to the FrmForm::getOne method so we can hook on form attributes ? 

The `frm_pre_display_form` filter is not called on the `getOne` method which is called when autosaving drafts on multiple steps forms.

Thanks.

### DIFF
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -589,7 +589,7 @@ class FrmForm {
 			FrmAppHelper::unserialize_or_decode( $results->options );
 		}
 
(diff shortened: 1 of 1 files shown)


stephywells:
Hi @benjaminniess. Thanks for the PR.

If you'd like to change the hook name to something more descriptive like 'frm_form_object', I'd be happy to merge this.

benjaminniess:
Hi @stephywells ,

Thank you so much. I made the hook name change.

PR #131, opened 2019-10-17
Shows loading indicator on prev button when prev button pushed

LauraSteiner:
In a multi-page form, when the prev and next buttons are pushed, the loading icon shows on the button that was pushed.

https://github.com/Strategy11/formidable-pro/issues/2314


stephywells:
Awesome! Looks great! Thanks @LauraSteiner !

PR #136, opened 2019-12-03
fix for the pro issue #643

glopez88:
https://github.com/Strategy11/formidable-pro/issues/643

2 areas in the settings page were affected: 
Screen Shot 2019-12-04 at 2 40 39 AM
Screen Shot 2019-12-04 at 2 40 46 AM

If the total pages is less than a hundred, it will use an ordinary dropdown. But if pages are a hundred or more, it will use the autocomplete (ajax).


classes/helpers/FrmAppHelper.php
```diff
@@ -1010,12 +1010,42 @@ public static function get_pages() {
 		return get_posts( $query );
 	}
 
+    public static function maybe_autocomplete_pages_options( $args = array(), $page_id = '', $truncate = false ) {
+        $params = [ $args, $page_id, $truncate ];
+
+        $args = apply_filters( 'preformat_pages_selection_args', ...$params );
```
stephywells: Let's not use the ...
I think it may be too easy to miss and cause issues in the future.

classes/helpers/FrmAppHelper.php
```diff
@@ -1010,12 +1010,42 @@ public static function get_pages() {
 		return get_posts( $query );
 	}
 
+    public static function maybe_autocomplete_pages_options( $args = array(), $page_id = '', $truncate = false ) {
```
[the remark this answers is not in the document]
glopez88: I actually just followed the parameter signature of the 1st function for backwards compatibility but I found out that it was only called twice so I just changed it to require a single parameter

classes/helpers/FrmAppHelper.php
```diff
@@ -1010,12 +1010,42 @@ public static function get_pages() {
...
+        $params = [ $args, $page_id, $truncate ];
+
+        $args = apply_filters( 'preformat_pages_selection_args', ...$params );
+
+        $pages_count = wp_count_posts( 'page' );
+
+        if ( $pages_count->publish > 100 ) {
```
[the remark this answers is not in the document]
glopez88: 100 was stated in the issue but I changed it 25 as 20 seems too low I think

classes/controllers/FrmFormActionsController.php
```diff
@@ -548,6 +548,69 @@ public static function limit_by_type( $where ) {
 
 		return $where;
 	}
+
+    public static function page_search() {
```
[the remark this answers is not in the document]
glopez88: I'm still struggling on which controller is being by a view. Still trying to familiarize things. If there is a guide that says that this controller or class is for this section of the app only, that would be very helpful for me

css/frm_admin.css
```diff
@@ -3747,6 +3747,15 @@ h2.frm-h2,
...
 
+.frm-admin-page-formidablesettings .ui-autocomplete {
+    box-shadow: none;
+    border: 1px solid #ccc;
+}
+
+.frm-admin-page-formidablesettings .ui-autocomplete li {
+    padding: 10px;
+}
+
```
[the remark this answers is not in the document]
glopez88: the autocomplete on my end was actually showing a white background and regular hover, it could have been due to a conflict from other plugin as I did not style the autocomplete. But this is now fixed.

classes/helpers/FrmAppHelper.php
glopez88: this is noted

classes/controllers/FrmAppController.php
```diff
@@ -379,13 +379,16 @@ public static function admin_js() {
 		wp_register_style( 'formidable-admin', FrmAppHelper::plugin_url() . '/css/frm_admin.css', array(), $version );
 		wp_register_script( 'bootstrap_tooltip', FrmAppHelper::plugin_url() . '/js/bootstrap.min.js', array( 'jquery' ), '3.3.4' );
 		wp_register_style( 'formidable-grids', FrmAppHelper::plugin_url() . '/css/frm_grids.css', array(), $version );
+        wp_register_script( 'formidable-settings-autocomplete', FrmAppHelper::plugin_url() . '/js/formidable_settings.js', array('jquery', 'jquery-ui-autocomplete'), $version, true );
```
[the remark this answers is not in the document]
stephywells: I'm still seeing a few issues. Here's the most current commit:
https://travis-ci.org/Strategy11/formidable-forms/jobs/620743127

css/frm_admin.css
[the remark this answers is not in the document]
stephywells: Thanks! I'm still seeing a yellow border on hover, and the yellow text right after I hover.

classes/helpers/FrmAppHelper.php
```diff
@@ -1010,12 +1010,42 @@ public static function get_pages() {
...
+            $selected = self::get_post_param( $args['field_name'], $args['page_id'], 'absint' );
+            $title = '';
+
+            if( $selected ) {
+                $title = get_the_title($selected);
+            }
+
+            ?>
+            <input type="text"
+                placeholder="Search page..."
```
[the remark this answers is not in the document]
stephywells: Let's use esc_html_e( 'Select a Page', 'formidable' ) here

classes/helpers/FrmAppHelper.php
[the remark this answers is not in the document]
stephywells: Sounds good. Thank you!

classes/controllers/FrmFormActionsController.php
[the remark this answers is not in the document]
stephywells: I'm afraid we don't have anything like this. We've also got 10 years worth of code here, so the organization has changed over the years. Generally, you can add code near other related code. There is a lot of refactoring and new classes we could use, but we're doing it little by little.

I would also recommend this book. It was a great discovery for me, and has made a big difference in the code I write now.
https://www.investigatii.md/uploads/resurse/Clean_Code.pdf

js/formidable_admin.js
```diff
@@ -4626,6 +4626,44 @@ function frmAdminBuildJS() {
...
+					return false;
+				},
+				position: {
+					my: "left top",
+					at: "left bottom",
+					collision: "flip"
+				},
+				response: function(event, ui) {
+					if ( !ui.content.length ) {
+						var noResult = { value: "", label: "No results found" };
```
stephywells: This could use localization. 'No items found.' is a string we already have in the plugin, so it would be nice to not create an additional string to translate.

css/frm_admin.css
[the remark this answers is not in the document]
glopez88: I have fixed the issue and used id instead of class to add more weight on selector. If there's issue again (which I believe is a collision between plugins), a screenshot would be very helpful.

glopez88:
@stephywells requested changes are all done.

classes/helpers/FrmAppHelper.php
```diff
@@ -1010,12 +1010,116 @@ public static function get_pages() {
...
+		}
+	}
+
 	/**
 	 * @param array   $args
 	 * @param string  $page_id Deprecated.
 	 * @param boolean $truncate Deprecated.
 	 */
-	public static function wp_pages_dropdown( $args = array(), $page_id = '', $truncate = false ) {
+	public static function wp_pages_dropdown( $args, $page_id = '', $truncate = false ) {
```
stephywells: This used to be $args = array(). We should keep it that way for reverse compatibility.

classes/controllers/FrmHooksController.php
```diff
@@ -127,6 +127,8 @@ public static function load_admin_hooks() {
 		add_action( 'frm_before_settings', 'FrmSettingsController::license_box' );
 		add_action( 'frm_after_settings', 'FrmSettingsController::settings_cta' );
 		add_action( 'wp_ajax_frm_settings_tab', 'FrmSettingsController::load_settings_tab' );
+		add_action( 'wp_ajax_page_search', 'FrmAppHelper::page_search' );
+		add_filter( 'frm_preformat_pages_selection_args', 'FrmAppHelper::frm_preformat_selection_args', 10, 3 );
```
stephywells: Let's keep this consistent, and not use a prefix for a function inside of a class.
add_filter( 'frm_preformat_pages_selection_args', 'FrmAppHelper::preformat_selection_args', 10, 3 );

classes/controllers/FrmHooksController.php
```diff
@@ -127,6 +127,8 @@ public static function load_admin_hooks() {
 		add_action( 'frm_before_settings', 'FrmSettingsController::license_box' );
 		add_action( 'frm_after_settings', 'FrmSettingsController::settings_cta' );
 		add_action( 'wp_ajax_frm_settings_tab', 'FrmSettingsController::load_settings_tab' );
+		add_action( 'wp_ajax_page_search', 'FrmAppHelper::page_search' );
```
stephywells: We should also use a prefix for page_search so this won't be prone to conflicts. wp_ajax_frm_page_search

classes/helpers/FrmAppHelper.php
```diff
@@ -1010,12 +1010,116 @@ public static function get_pages() {
...
+		global $wpdb;
+
+		$nonce = self::get_param( 'nonce', '', 'get', 'sanitize_text_field' );
+		$term = self::get_param( 'term', '', 'get', 'sanitize_text_field' );
+
+		if ( ! wp_verify_nonce( $nonce, 'frm_ajax' ) ) {
+			wp_send_json(
+				array(
+					'status' => 'error',
+					'message' => 'Forbidden request',
```
stephywells: This would be more consistent if we use:
esc_html__( 'You do not have permission to do that', 'formidable' )

classes/helpers/FrmAppHelper.php
```diff
@@ -1010,12 +1010,116 @@ public static function get_pages() {
...
+				)
+			);
+		}
+
+		$results = array();
+		$pages = $wpdb->get_results(
+			$wpdb->prepare(
+				"SELECT
+					*
```
stephywells: Since only the ID and post_title are used, it would be best not to fetch any columns we don't need.

classes/views/frm-forms/settings-advanced.php
```diff
@@ -83,13 +83,13 @@
...
-			<span class="success_action_page_box success_action_box<?php echo esc_attr( $values['success_action'] === 'page' ? '' : ' frm_hidden' ); ?>">
+			<span class="success_action_page_box success_action_box<?php echo esc_attr( $values['success_action'] === 'page' ? '' : ' frm_hidden' ); ?> ui-front">
+
 				<?php
-				FrmAppHelper::wp_pages_dropdown(
+				FrmAppHelper::maybe_autocomplete_pages_options(
 					array(
 						'field_name'  => 'options[success_page_id]',
 						'page_id'     => isset( $values['success_page_id'] ) ? $values['success_page_id'] : '',
-						'placeholder' => __( 'Select a Page', 'formidable' ),
```
stephywells: This 'Select a Page' placeholder has gone missing.

js/formidable_admin.js
```diff
@@ -4626,6 +4626,44 @@ function frmAdminBuildJS() {
...
+						new_label = ui.item.label;
+						new_value = ui.item.value;
+					}
+
+					jQuery(this).val(new_label);
+					jQuery(this).next("input[type='hidden']").val(new_value);
+
+					return false;
+				},
+				position: {
```
stephywells: When I type a couple letters, then arrow down to select something, the post ID shows in the text field rather than the title. Then, if a post ID is included, the autocomplete options then say No items found.

I think focus: needs to be added here in order to handle that. Perhaps the autocomplete should check both the post id and title?

css/frm_admin.css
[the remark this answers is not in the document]
stephywells: Thanks, this is looking better. But I still see the yellow border and bold text when I use the arrow key to select an option.

Also, the font is different from the rest of the page. This is the normal jQuery styling that is being used.

classes/views/frm-forms/settings-advanced.php
```diff
@@ -83,13 +83,13 @@
 				<input type="text" name="options[success_url]" id="success_url" value="<?php echo esc_attr( isset( $values['success_url'] ) ? $values['success_url'] : '' ); ?>" placeholder="http://example.com" />
 			</span>
 
-			<span class="success_action_page_box success_action_box<?php echo esc_attr( $values['success_action'] === 'page' ? '' : ' frm_hidden' ); ?>">
+			<span class="success_action_page_box success_action_box<?php echo esc_attr( $values['success_action'] === 'page' ? '' : ' frm_hidden' ); ?> ui-front">
```
stephywells: Can we replace ui-front with a class in Formidable? We end up with a lot of conflicts using jQuery classes.

classes/helpers/FrmAppHelper.php
```diff
@@ -1010,12 +1010,116 @@ public static function get_pages() {
...
+	}
+
 	/**
 	 * @param array   $args
 	 * @param string  $page_id Deprecated.
 	 * @param boolean $truncate Deprecated.
 	 */
-	public static function wp_pages_dropdown( $args = array(), $page_id = '', $truncate = false ) {
+	public static function wp_pages_dropdown( $args, $page_id = '', $truncate = false ) {
 		if ( ! is_array( $args ) ) {
```
stephywells: This could use a check for reverse compatibility. What if $args is an array, but page_id and truncate are set as param 2 & 3?

classes/helpers/FrmAppHelper.php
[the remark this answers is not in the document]
stephywells: This still hasn't been adjusted.

js/formidable_admin.js
glopez88: I think it's best for autocomplete to check for title only as it's not common for user to search for page IDs instead of the title.

classes/helpers/FrmAppHelper.php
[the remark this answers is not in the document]
glopez88: this has been adjusted but note that this placeholder is for the autocomplete

classes/controllers/FrmHooksController.php
stephywells: Sorry I wasn't very clear. I meant for this to be:
add_action( 'wp_ajax_frm_page_search', 'FrmAppHelper::page_search' );

classes/helpers/FrmAppHelper.php
stephywells: Can you please change this to 'You do not have permission to do that'?

css/frm_admin.css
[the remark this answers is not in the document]
stephywells: I'm still seeing the bold text on hover and a different font used in the dropdown.

PR #140, opened 2019-12-19
Keeps WordPress sidebar full-size on larger screens

LauraSteiner:
Keeps WordPress sidebar full-size on larger screens

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -54,7 +54,7 @@ class FrmAppController {
 		}
 
 		if ( FrmAppHelper::is_full_screen() ) {
-			$classes .= apply_filters( 'frm_admin_full_screen_class', ' frm-full-screen folded' );
+			$classes .= apply_filters( 'frm_admin_full_screen_class', '' );
 		}
 
 		return $classes;
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -53,10 +53,6 @@ class FrmAppController {
 			}
 		}
 
-		if ( FrmAppHelper::is_full_screen() ) {
-			$classes .= apply_filters( 'frm_admin_full_screen_class', '' );
-		}
-
 		return $classes;
 	}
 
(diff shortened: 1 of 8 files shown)


LauraSteiner:
@stephywells,  @srwells and I are still deciding about whether there should be a cushion between the two sidebars on the right or not.   This commit has the cushion.  You can see how it looks without the cushion by changing the value of --scrollbar-cushion in frm_admin.css to 0px.   

What do you think?

In the interests of time, I wanted to get this PR to you now.  It will be very fast to remove the cushion if we don't want it.

stephywells:
@LauraSteiner @srwells I don't love the gap between the border and the edge of the page. Is it possible we can get rid of that?
Screen Shot 2019-12-30 at 8 55 58 AM

LauraSteiner:
@stephywells  I've removed the --scrollbar-cushion, which removes the gap.

stephywells:
I'm seeing an extra border on the view page. Are you seeing this too?
Screen Shot 2020-01-07 at 2 41 48 PM

LauraSteiner:
Thanks for those awesome changes, Steph!  

Yes, I see an extra border.  You're talking about the double border underneath the menu items, right?

https://cl.ly/70b663f238f8

This screenshot was taken from my Formidable sandbox site.  If we're talking about the same thing, it looks like it's in the plugin already.

LauraSteiner:
@stephywells @srwells I've pushed some code to show the form builder's right panel scrollbar all the way on the right when the screen is large so an outer scrollbar isn't needed.  When the screen is smaller, JavaScript adds a cushion so both scrollbars show.  Otherwise, the right panel scrollbar disappears (on Windows, at least).

To test, start with a screen that's short and see the two scrollbars, one for the main page and one for the form builder right panel.  Then make the screen longer and wider, as needed, the one for the form builder right panel.  That scrollbar should be all the way to the right.

The JavaScript cushion takes a moment to appear.  Do you guys think that's an issue?

There are some times when I tested this and the cushion didn't seem to adjust properly initially when I changed screen sizes.  I can't reliably replicate this.  When that happened, the inner scrollbar disappeared and only the outer scrollbar was there, which is what we have now.

What do you guys think?

LauraSteiner:
@steph Steve and I have discussed an easier solution for this.  I'll work on this and update.

LauraSteiner:
Also, yes, see that double border on the View page.  I'll work on a fix for it.

LauraSteiner:
Since the code for this was written a while ago, I did a rebase on the current master in a new branch.  I've opened a pull request on the rebase branch.

PR #141, opened 2020-01-08
Adjust base form migrator class and the JS for more flexibility

lanresmith:


classes/models/FrmFormMigrator.php
```diff
@@ -185,24 +195,89 @@ protected function prepare_fields( $fields, &$form ) {
 				continue;
 			}
 
-			$new_field                = FrmFieldsHelper::setup_new_vars( $this->convert_field_type( $type ) );
+			$new_field                = FrmFieldsHelper::setup_new_vars( $this->convert_field_type( $field ) );
```
stephywells: We'll need to do something here for reverse compatibility. Maybe something like this? (this is untested)

$new_type = $this->convert_field_type( $field );
if ( empty( $new_type ) ) {
  $new_type = $this->convert_field_type( $type );
}

classes/models/FrmFormMigrator.php
```diff
@@ -185,24 +195,89 @@ protected function prepare_fields( $fields, &$form ) {
...
 	}
 
-	protected function convert_field_type( $type ) {
-		return $type;
+	/**
+	 * @param object|array $field
+	 * @param string       $use   Which field type to prefer to consider $field as.
+	                              This also eases the recursive use of the method,
+	                              particularly the overrides in child classes, as
+	                              there will be no need to rebuild the converter
```
stephywells: A * is needed for each line here

classes/models/FrmFormMigrator.php
```diff
@@ -185,24 +195,89 @@ protected function prepare_fields( $fields, &$form ) {
...
+	                              array at usage locations.
+	 */
+	protected function convert_field_type( $field, $use = '' ) {
```
stephywells: This should be reverse compatible ($type vs $field)

classes/models/FrmFormMigrator.php
```diff
@@ -185,24 +195,89 @@ protected function prepare_fields( $fields, &$form ) {
...
+		// This pre-inc may make us skip a level/order occassionally, but that's fine.
+		// This & the 2nd increment below are here for better abstraction, the caller
+		// shouldn't have to worry about the right order when closing a section.
+		$new_field['field_order'] = ++$field_order;
+		$new_field['original']    = '';
+
+		$form['fields'][]         = $new_field;
+		$this->current_section    = null;
+
+		++$field_order;
```
stephywells: This codestyling is inconsistent

classes/models/FrmFormMigrator.php
```diff
@@ -73,7 +76,7 @@ public function import_page() {
 						</div>
 						<button type="submit" class="button button-primary button-hero">Start Import</button>
 					</form>
-					<div id="frm-importer-process" class="frm_hidden">
+					<div class="frm-importer-process frm_hidden">
```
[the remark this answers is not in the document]
lanresmith: OK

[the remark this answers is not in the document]
lanresmith: Didn't know that there is an existing published importer.

But, remember that the JS code _must_ change, else only the first
matched ID will always be selected, now that we'll have several
IDs that are exactly the same.

classes/models/FrmFormMigrator.php
lanresmith: OK

classes/models/FrmFormMigrator.php
lanresmith: OK

classes/models/FrmFormMigrator.php
lanresmith: OK

classes/models/FrmFormMigrator.php
lanresmith: Do you mind pointing out what's missing pls?

stephywells: $field_order ++; would be more consistent. Is there a reason to use ++$field_order here?

lanresmith: yes, the 1st pre-increment matters. I'll change that of the last line though.

classes/models/FrmFormMigrator.php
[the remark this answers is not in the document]
stephywells: Instead of changing the class to an ID, let's add a class. The javascript can use the class.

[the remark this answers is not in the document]
lanresmith: Maybe we should just stay with the current. Currently, this is what
we do to select the div:
`$form.next( '.frm_hidden' );`
I think that suffices too.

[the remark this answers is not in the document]
lanresmith: By the current, I mean I've reverted it to what it used to be:
`<div id="frm-importer-process" class="frm_hidden">`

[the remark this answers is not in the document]
stephywells: I think using .next( '.frm_hidden' ) will be too prone to error if anything changes in the HTML.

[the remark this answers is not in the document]
lanresmith: OK. Class added.

PR #143, opened 2020-01-29
Keeps WordPress sidebar full width on larger screens

LauraSteiner:
Keeps WordPress sidebar full width on larger screens

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -54,7 +54,7 @@ class FrmAppController {
 		}
 
 		if ( FrmAppHelper::is_full_screen() ) {
-			$classes .= apply_filters( 'frm_admin_full_screen_class', ' frm-full-screen folded' );
+			$classes .= apply_filters( 'frm_admin_full_screen_class', '' );
 		}
 
 		return $classes;
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -53,10 +53,6 @@ class FrmAppController {
 			}
 		}
 
-		if ( FrmAppHelper::is_full_screen() ) {
-			$classes .= apply_filters( 'frm_admin_full_screen_class', '' );
-		}
-
 		return $classes;
 	}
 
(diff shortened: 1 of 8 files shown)


stephywells:
@LauraSteiner I made a few changes to prevent the javascript from affecting Chrome. I was seeing some jumping in Chrome. I tested this in Chrome and Edge. Is it possible to test my changes in other browsers too?

LauraSteiner:
@stephywells Thanks!  That check for Chrome is very cool!  

I think it could be good to set the default to scrollbar-cushion to something like 7px so there's room for both scrollbars when the window is short.  If it's set to 0px, when the window is short, there's an outer scrollbar to scroll the main page, but the inner scrollbar to scroll the right form builder panel doesn't show.

In this video, you can hopefully see the scrollbar for the right form builder when the window is long enough to accommodate the full WP sidebar.

When it's too short, you can hopefully see that there's a right scrollbar for the main windows, but there's no scrollbar for the right panel of the form builder.

https://cl.ly/9442f7839852

LauraSteiner:
In Edge on Windows, the scrollbars overlap just a bit.  It looks like if the scrollbar cushion were 2px or 3px bigger, this wouldn't happen.

https://cl.ly/17624900c26a

LauraSteiner:
Vivaldi is Chromium based, but only the outer scrollbar shows and works when the cushion is set to 0 and the screen is full length.  With Chrome in the same situation, the form builder right panel shows.

LauraSteiner:
@stephywells I'm interested in hearing your thoughts on this.  

What do you think of making the default scrollbar cushion on Chrome to something like 7 - 8 px instead of 0?  Then there would be a scrollbar on the right form builder panel when the screen is shorter.

stephywells:
@LauraSteiner I'm fine with a bigger padding in the javascript. Whatever you think looks good there. You can make the adjustments you think should be made. But Chrome shouldn't be changing with javascript. Chrome is the priority and we should only use CSS there.

@srwells any thoughts on this?

PR #142, opened 2020-01-29
Warnings for reserved params in redirects, View filter values, and View Detail Page Slugs

LauraSteiner:
Shows warning popup if user uses reserved words in View filter values or detail page slug.
Shows warning message if user user uses reserved words in redirect link.
Shows explanation popup if user clicks on single use form action that's already been used
Changes alerts to styled modals based on the CTA modal.
https://github.com/Strategy11/formidable-pro/issues/2401
https://github.com/Strategy11/formidable-pro/issues/1725

PR #144, opened 2020-02-01
Adds warnings for unsafe params in redirects and Views.

LauraSteiner:
Shows warning popup if user uses reserved words as View filter values or detail page slug.
Shows warning message if user user uses reserved words as params in redirect link.
Shows explanation popup if user clicks on single use form action that's already been used
Changes admin alerts to styled info modals based on the CTA modal.
Strategy11/formidable-pro#2401
Strategy11/formidable-pro#1725


LauraSteiner:
Changes made in the other pull request

https://github.com/Strategy11/formidable-forms/pull/142

are included here.

css/frm_admin.css
```diff
@@ -14,7 +14,6 @@
 	--sidebar-hover: rgb(227, 232, 235);
 	--primary-color: rgb(65, 153, 253);
 	--primary-hover: rgb(49, 119, 199);
-	--tooltip-font-size: 14px;
```
stephywells: I'm still seeing this used in the css. Can you replace tooltip-font-size in the css with 14px?

css/frm_admin.css
```diff
@@ -5771,6 +5795,14 @@ ul .frm_col_two {
 	color: var(--orange)
 }
 
+.info-modal-inside a.frm-standard-link {
+	color: var(--primary-color);
+}
+
+.info-modal-inside a.frm-standard-link:hover {
```
stephywells: Can we combine this into one statement?
.info-modal-inside a.frm-standard-link,
.info-modal-inside a.frm-standard-link:hover{
...

css/frm_admin.css
stephywells: This is fixed.

PR #146, opened 2020-02-18
Fixes Pro/issue/2417

lanresmith:


stephywells:
@lanresmith do you think this would make more sense to fix in the PHP instead of the javascript? So if separate values are not turned on, the 'value' would use the 'label'. I'm concerned about adding a large block of code for this that could affect other fields.

lanresmith:
> @lanresmith do you think this would make more sense to fix in the PHP instead of the javascript? So if separate values are not turned on, the 'value' would use the 'label'. I'm concerned about adding a large block of code for this that could affect other fields.

@stephywells I'm not sure if that's a good idea & here's why :
The changes also rectify the `default_value` setting issue - as you might have seen from the code. Without the fix, if I add at least two new product options e.g., then the new options will all have 'New Product' as their values, if one of them is selected as default (e.g. in the case of radio/select product type, though the same holds for checkbox type too anyway), how do we know - at the PHP side - which one of them was selected?

js/formidable_admin.js
```diff
@@ -2921,12 +2926,54 @@ function frmAdminBuildJS() {
...
+		maybeUpdateFieldOptionValues();
 
 		var form = jQuery( builderForm );
 		jQuery( document.getElementById( 'frm_compact_fields' ) ).val( JSON.stringify( form.serializeArray() ) );
 		jQuery( document.getElementById( 'frm_js_build_form' ) ).submit();
 	}
 
+	function maybeUpdateFieldOptionValues() {
+		var i, value, label, match, defaultVal, separateValues,
+			labels = document.querySelectorAll( '#new_fields .frm_single_option:not([data-optkey="000"]) [name^="field_options[options_"][name$="[label]"],[name^="field_options[options_"][value="Other"]' );
```
stephywells: value="Other" isn't a static value. This is the value entered into the Other box.

lanresmith: @stephywells Well, 'other' values are not something to worry about
as far as this issue is concerned, because they're not selected when
the form is loaded - even when you manually select them (I mean :
apart from programatically doing it as I've done in this PR), although
they'll be stored in the DB as the default (or part of the default as in the
case of checkbox, which can have multiple options checked as default).

I'll try to see how better to querySelect them however.

lanresmith: Done

PR #147, opened 2020-02-20
Issue/pro/2420

LauraSteiner:
Fixes warning from FrmFieldHelper::get_unsafe_params when the URL doesn't have a query.

https://github.com/Strategy11/formidable-pro/issues/2420


stephywells:
Thanks @LauraSteiner for the fix!

PR #148, opened 2020-02-22
Switch quantity field to use checkboxes for selecting products

lanresmith:
This contains part of the back-end work

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -5053,13 +5053,11 @@ function frmAdminBuildJS() {
 
 	function updateCurrentProductFields() {
 		var cont = jQuery( this ).closest( '.frmjs_prod_field_opt' ),
-			products = cont.find( '[type="checkbox"]' ),
+			products = cont.find( '[type="checkbox"]:checked' ),
 			idsArray = [];
 
 		products.each( function() {
-			if ( this.checked ) {
-				idsArray.push( this.value );
-			}
+			idsArray.push( this.value );
 		} );
 
 		cont[0].setAttribute( 'data-frmcurrent', JSON.stringify( idsArray ) );
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1513,19 +1513,23 @@ function frmAdminBuildJS() {
 	}
 
 	function popProductFields( field ) {
-		var options = [], fields, i, checked, current,
-			fName, id;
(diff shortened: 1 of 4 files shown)


js/formidable_admin.js
```diff
@@ -5049,6 +5051,20 @@ function frmAdminBuildJS() {
...
+			products = cont.find( '[type="checkbox"]' ),
+			idsArray = [];
+
+		products.each( function() {
+			if ( this.checked ) {
+				idsArray.push( this.value );
+			}
+		} );
+
+		cont[0].setAttribute( 'data-frmcurrent', JSON.stringify( idsArray ) );
```
stephywells: @lanresmith I'm a little confused on this. Can you tell me why adding the selected boxes here is helpful instead of getting checked boxes directly?

lanresmith: Sorry, that was an oversight. Fixed.

stephywells: Sorry I wasn't very clear. I meant why use data-frmcurrent? Is there a reason this is helpful?

lanresmith: Every time a quantity field is clicked in order to edit its settings, the product checkboxes are (re)populated. It's done this way because a new product might have been added to the form, and we of course want it to show in the list of product checkboxes to be mapped to the quantity field.

We therefore need a store for each quantity field to keep track of the product fields that it's currently mapped to - this store is of course updated every time product fields are checked or unchecked while editing the quantity field - so that the next time the user clicks on the quantity field to edit its settings, while we're repopulating the product checkboxes, we'll know which ones to make checked.

stephywells: Is there a reason we can get the :checked boxes right before repopulating?

lanresmith: Done

stephywells: Looks great! That feels much cleaner. It looks like we don't need productFields in maybeSetProductField(). Right?

lanresmith: We actually still need `maybeSetProductField` because we said earlier that if there is just 1 product field in a form, upon adding the very 1st quantity field, the latter should be automatically attached to that lone product field. I just realised that that's now broken since the removal of `data-frmcurrent`. I've just fixed it.

stephywells: Thanks for fixing that! I wasn't saying to remove that function by the way. Thanks for cleaning it up!

PR #149, opened 2020-04-15
Issue/pro/2434

LauraSteiner:
For review and discussion only!


LauraSteiner:
I have a function that tests whether an option is an image option.  This allows me to change certain $atts when displaying the option in different contexts, like the entries list.

For Radio Button and Checkbox fields, I test whether image options are set for that field.  

With Dynamic fields, I couldn't sort out a reasonable way to determine whether there's an image at the end of chains of Dynamic fields of different lengths.  So I've set all Dynamic fields to be image option fields for this purpose.  

Will that interfere with anything else or cause problems?   If so, I'd appreciate some help with sorting out how to do this.

PR #150, opened 2020-04-17
Full page builder

LauraSteiner:
Expands WordPress sidebar to full size on Formidable admin pages.


LauraSteiner:
@stephywells What do you think of using the actual scrollbar width for non-Chrome browsers? 

We'd keep the set width for Chrome browsers,  with a default of 8px so there's room for both scrollbars on the right.

7px works for me in Windows Chrome.  You saw some jumping on Chrome on the Mac when the default was set to 7px in an earlier commit.  So the scrollbars could be slightly different on PC and Mac.

stephywells:
@srwells  can you review this for functionality before I get into a code review?

PR #176, opened 2020-06-26
Sets a default value for $message_placement in new.php

LauraSteiner:
Sets default value for $message_placement in new.php.

https://github.com/Strategy11/formidable-pro/issues/2522

### DIFF
--- a/classes/views/frm-entries/new.php
+++ b/classes/views/frm-entries/new.php
@@ -6,6 +6,8 @@
 	<?php
 }
 
+$message_placement = $message_placement ? $message_placement : 'before';
+
 if ( $message_placement !== 'after' ) {
 	include( FrmAppHelper::plugin_path() . '/classes/views/frm-entries/errors.php' );
 }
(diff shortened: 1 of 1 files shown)


LauraSteiner:
If you prefer, I could set a value for $message_placement before it's required in FrmProEntry.  

Note: I couldn't replicate the issue, but this should resolve it.

classes/views/frm-entries/new.php
```diff
@@ -6,6 +6,8 @@
 	<?php
 }
 
+$message_placement = $message_placement ? $message_placement : 'before';
```
stephywells, line 9: Thanks so much for working on this @LauraSteiner!

It looks like this will still cause error messages. Maybe this?
$message_placement = isset( $message_placement ) ? $message_placement : 'before';

LauraSteiner: @stephywells Of course!  Sorry, I left that out.

stephywells: Thank you! Looks good.

PR #177, opened 2020-06-27
Image options

LauraSteiner:
Adds the option of having images for Radio Buttons and Checkboxes.
https://github.com/Strategy11/formidable-pro/issues/2434


LauraSteiner:
@stephywells The "Upload image" text in the text field setting that's clicked to open the image uploader is too light a color.  I've written CSS that should override the current color, I think.  It's more specific.  But it doesn't.  

I put it in a commit:

https://github.com/Strategy11/formidable-forms/pull/177/commits/a577f293ae3c88c027c7555873fd706ce3ccbfae

PR #187, opened 2020-07-06
don't load CSS for unused custom styles

andrewklimek:
this would avoid adding CSS for unused styles.  (I'm not sure this gets re-generated when the "manage" settings are saved though.)

### DIFF
--- a/css/custom_theme.css.php
+++ b/css/custom_theme.css.php
@@ -621,8 +621,14 @@ legend.frm_hidden{
 }
 
 <?php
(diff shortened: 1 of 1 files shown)


stephywells:
Thanks for the PR @andrewklimek. We won't be able to merge this though since it would remove the styles from the generated css. The best way to exclude unused Styles would be the delete the ones you don't want. We're also working on decreasing the size of the extra styles by using CSS vars. The goal is to only set vars for the extra styles and not have anything extra.

Hope this helps! Thanks again!

PR #193, opened 2020-07-18
switch all user role dropdowns to multiselect

Crabcyborg:


classes/helpers/FrmAppHelper.php
```diff
@@ -1147,13 +1164,33 @@ public static function roles_options( $capability ) {
 			$frm_vars['editable_roles'] = $editable_roles;
 		}
 
+		if( $public === 'public' ) {
+			?>
+			<option value=""><?php esc_html_e( 'Everyone', 'formidable-pro' ); ?></option>
```
Crabcyborg, line 1169: The inspection noted that this is referring to formidable-pro. I copied this from there, but I'm moving it to formidable lite. I'm not entirely sure the best course of action in this case.

stephywells: All you need to do here is change 'formidable-pro' to 'formidable' and you'll be set

stephywells:
@Crabcyborg can you also run through the code styling issues listed here?
https://github.com/Strategy11/formidable-forms/pull/193/checks?check_run_id=883945450#step:3:236

classes/helpers/FrmAppHelper.php
Crabcyborg: Awesome, check :)

classes/helpers/FrmFormsHelper.php
```diff
@@ -1524,4 +1524,18 @@ public static function reserved_words() {
 			'year',
 		);
 	}
+
+	/**
+	 * @param object $form
+	 * @return bool
+	 */
+	public static function &is_form_visible_to_user( $form ) {
+		$visible = ! $form->logged_in || empty( $form->options['logged_in_role'] );
```
stephywells, line 1533: It seems like this isn't quite the same as the original version in user_has_permission_to_view. Without reversing the logic, I get this:

$visible = $form->logged_in && isset( $form->options['logged_in_role'] ) && ! empty( $form->options['logged_in_role'] );
if ( $visible && get_current_user_id() ) {
	$visible = ! FrmAppHelper::user_has_permission( $form->options['logged_in_role'] );
}

Crabcyborg: Since the other function is actually checking the opposite:

```
$hidden = $form->logged_in && isset( $form->options['logged_in_role'] ) && ! empty( $form->options['logged_in_role'] );
if ( $hidden && get_current_user_id() ) {
$hidden = ! FrmAppHelper::user_has_permission( $form->options['logged_in_role'] );
}
```

`isset( $form->options['logged_in_role'] ) && ! empty( $form->options['logged_in_role'] )` is identical to just ` ! empty( $form->options['logged_in_role'] )` on its own, so the above could be reduced to simply:

```
$hidden = $form->logged_in && ! empty( $form->options['logged_in_role'] );
if ( $hidden && get_current_user_id() ) {
$hidden = ! FrmAppHelper::user_has_permission( $form->options['logged_in_role'] );
}
```

And I think I am flipping everything. I just tried again, and got the same result. And my tests should be catching a problem if something was off.

I think some of the confusion might be in the name of the previous function. I actually changed it at one point, but in the spirit of maintaining backward compatibility decided not to touch it.

In my original commit I had renamed this function to properly reflect what is actually going on: https://github.com/Strategy11/formidable-forms/pull/193/commits/3742be115e1be9cd211f8475c88b9c2163fc71ec
I change it back in this commit (when I previously decided to just drop it): https://github.com/Strategy11/formidable-forms/pull/193/commits/21e235cdd02872bb16039bec754405d3371e1b85

So maybe we want to go with one of those options. I'm still playing with how much is too much vs not enough :)

stephywells: I think we need to keep isset to prevent php warnings on sites that may have not updated the form before that setting was available. Most of these user role options include a blank option that we don't want to lose. So we should check for '' instead of just empty.

Crabcyborg: I find that empty is one of those functions that is really easy to understand that gets a lot of programmers confused.

A direct quote from https://www.php.net/manual/en/function.empty.php:  `empty() does not generate a warning if the variable does not exist.`

Another direct quote:

```
 The following values are considered to be empty:

    "" (an empty string)
    0 (0 as an integer)
    0.0 (0 as a float)
    "0" (0 as a string)
    NULL
    FALSE
    array() (an empty array)
```

Basically `empty( $var )` is identical to `! isset( $var ) || ! $var`.

I believe knowing this and avoid checks that are not necessary can allow for cleaner code.

stephywells: That's cool. I thought it produced warnings. But the problem still exists that we need to check for '' here, right? Since '' means logged in users?

Crabcyborg: '' is "empty" as it is falsey, so I don't think these would have ever validated further.

It's possible that we want to treat the '' differently, yes. But what we have here is what we had before, so I'd assume this this works if it was already working.

tests/forms/test_FrmFormsHelper.php
```diff
@@ -0,0 +1,58 @@
...
+	/**
+	 * @group visibility
+	 * @covers FrmFormsHelper::is_form_visible_to_user
+	 */
+	public function test_is_form_visible_to_user() {
+		$this->assert_form_is_visible( 'administrator', 'editor', 'Administrator can view a form set to editor' );
+		$this->assert_form_is_hidden( 'editor', 'administrator', 'Editor cannot view form set to administrator' );
+		$this->assert_form_is_visible( 'editor', array( 'administrator', 'editor' ), 'Editor can view form set to both administrator and editor' );
+		$this->assert_form_is_visible( 'editor', 'loggedin', 'Editor can view form set to logged in users' );
+		$this->assert_form_is_visible( 'subscriber', 'loggedin', 'Subscriber can view form set to logged in users' );
```
stephywells, line 16: @Crabcyborg How about we change 'loggedin' here to ''? That should tell us if the changes in the function are working correctly since '' is the saved value instead of 'loggedin' in this case.

Crabcyborg: Nice catch. Changing that actually caused my tests to break (it looks like logged out users could actually view it if it was set to logged in users).

I changed the logic a bit in is_form_visible_to_user. After out discussion I believe this covers what we're actually trying to achieve and should be easier to follow :)

tests/forms/test_FrmFormsHelper.php
```diff
@@ -0,0 +1,60 @@
...
+		$this->assert_form_is_visible( 'administrator', 'editor', 'Administrator can view a form set to editor' );
+		$this->assert_form_is_hidden( 'editor', 'administrator', 'Editor cannot view form set to administrator' );
+		$this->assert_form_is_visible( 'editor', array( 'administrator', 'editor' ), 'Editor can view form set to both administrator and editor' );
+
+		// The Logged-In Users option is actually an empty string
+		$this->assert_form_is_visible( 'editor', '', 'Editor can view form set to logged in users' );
+		$this->assert_form_is_visible( 'subscriber', '', 'Subscriber can view form set to logged in users' );
+		$this->assert_form_is_visible( 'subscriber', '', 'Subscriber can view form set to logged in users' );
+		$this->assert_form_is_hidden( 'loggedout', '', 'Logged out user cannot view form set to logged in users' );
+		$this->assert_form_is_hidden( 'loggedout', '', 'Logged out user cannot view form set to editors' );
```
stephywells, line 21: Looks like 'editors" went missing here. 

I'd also love to see a test to see if it's hidden with this:
this->assert_form_is_hidden( 'editor', array( 'loggedout', 'subscriber' ), ...

Crabcyborg: For sure, I'll make some extra updates tonight Steph :)

Crabcyborg: It should be much better. I actually check the exact matches for arrays now, something that I always knew I wasn't totally doing properly but after living with this PR all week I'm getting a much better idea of the whole picture :)

I added a few tests for custom roles, as well as exact string matching.

I figured it makes sense to add an exception for administrator if there isn't an exact match, similar to Global permissions.

classes/helpers/FrmAppHelper.php
```diff
@@ -1178,13 +1186,41 @@ public static function frm_capabilities( $type = 'auto' ) {
...
+	 * @param string $role
+	 * @return bool
+	 */
+	public static function current_user_can( $role ) {
+		if ( current_user_can( 'administrator' ) ) {
+			// always grant administrator role access
+			return true;
+		}
+
+		return $role ? current_user_can( $role ) : is_user_logged_in();
```
stephywells, line 1201: @Crabcyborg Should we allow for 'loggedin', 'loggedout', and '-1' here too? That would help reduce the code in the visibility check in the other PR. Looks like there's lots of baggage here with doing things differently over the years.

tests/forms/test_FrmFormsHelper.php
stephywells: Thanks so much!

classes/helpers/FrmAppHelper.php
Crabcyborg: Sounds good to me. I moved a lot of the logic over here :)

classes/helpers/FrmAppHelper.php
```diff
@@ -1178,14 +1186,51 @@ public static function frm_capabilities( $type = 'auto' ) {
...
+	public static function current_user_can( $role ) {
+		if ( $role == '-1' ) {
+			return false;
+		}
+
+		if ( $role === 'loggedout' ) {
+			return ! is_user_logged_in();
+		}
+
+		if ( $role === 'loggedin' || ! $role ) {
```
stephywells, line 1204: @Crabcyborg Do you know if it's possible for $role to === false or 0? Perhaps we should specifically use '' here to prevent that?

Crabcyborg: @stephywells  Do you think we'd handle those values differently? If they continued after this code we'd evaluate them as actual roles, which is probably less ideal than just treating them the same as ''?

I could see them possibly being part of the `return false` condition at the beginning. I don't recall ever seeing a 0 passed though, and I don't think any of these are saving booleans.

stephywells: Okay. I'll go ahead and merge it then.

stephywells:
@Crabcyborg  I didn't check the Travis tests. It look like there are some errors:
https://travis-ci.org/github/Strategy11/formidable-forms/builds/716568908

PR #197, opened 2020-07-24
Adds a hook 

LauraSteiner:
Adds hook to FrmFieldValue so field object can be translated.

https://github.com/Strategy11/formidable-wpml/issues/47


LauraSteiner:
@stephywells I don't think the failing tests are due to the code I've added in this commit.

PR #205, opened 2020-08-06
hide protect files role toggle

Crabcyborg:
In order for the protect files roles to toggle on and off in https://github.com/Strategy11/formidable-pro/pull/2565 we'll need this branch as well


js/formidable_admin.js
```diff
@@ -6060,6 +6060,18 @@ function frmAdminBuildJS() {
 				}
 			});
 
+			//If File Protection is checked/unchecked
+			$protectFiles = document.getElementById( 'protect_files' );
+			jQuery( $protectFiles ).change( function() {
```
stephywells, line 6065: Do we need to make sure $protectFiles isn't null?

Crabcyborg: Sounds like a good idea, thanks :)

Crabcyborg: Made it a little simpler. I copied this from above, not sure why it was using getElementById first when jQuery can just handle that, so I removed the variable references and made it more anonymous.

I changed it to a `jQuery( document ).on( 'change'` listener, which is safe if it doesn't exist since it's not really targeting the element that doesn't exist.

stephywells:
Thanks! There are quite a few spots that use getElementById first since it's faster than having jQuery call it. We're trying to weed out jQuery little by little and use vanilla js as much as possible.

Crabcyborg:
Thanks for the explanation! I'll keep that in mind

PR #207, opened 2020-08-07
use strpos for partial matches

Crabcyborg:
[The first commit](https://github.com/Strategy11/formidable-forms/pull/204) was merged pretty quickly, but I had actually made a second that changes the equals to a strpos check. I think the first fix would only work is there was one item selected.

May also be nice to add unit tests around this.


stephywells:
Unit tests would be great! If you have a minute, have at it.

PR #210, opened 2020-08-10
loggedout is not a valid form visibility fix

Crabcyborg:
Follow-up from the broken tests in https://github.com/Strategy11/formidable-forms/pull/193

Took a couple of tries to get this working in travis (it was fine locally).

`Adminstrator should not be able to see a form without an exact array match`.

First thought was that the tests are checking for things that aren't valid settings, but that didn't seem to effect the test breaking.

Second thought, is that administrator might actually pass the "current_user_can" check for subscriber, so I'm switching to an exact array match.

It was still failing for custom roles though, so I fixed that as well by calling `set_user_by_role` like we do for other roles. I also think it's possible that the custom role would have originally passed an editor check, so I'm changing the test to check for exact matches instead.

Also, I had administrator spelled wrong (missing an i) in two of the error messages.

I also made the `-1` check a strict match to shut up the code styling warnings I have on now 😁

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1207,7 +1207,7 @@ class FrmAppHelper {
 	 * @return bool
 	 */
 	public static function current_user_can( $role ) {
-		if ( $role == '-1' ) {
+		if ( $role === '-1' ) {
 			return false;
 		}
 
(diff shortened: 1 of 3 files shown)


stephywells:
This looks like it should work. Thanks for fixing this!

PR #213, opened 2020-08-11
New frm class frm inline success

Crabcyborg:


classes/views/frm-entries/form.php
```diff
@@ -77,7 +77,18 @@
 	$copy_values = $values;
 	unset( $copy_values['fields'] );
 
-	FrmFormsHelper::get_custom_submit( $copy_values['submit_html'], $form, $submit, $form_action, $copy_values );
+	if ( isset( $form->options['form_class'] ) && strpos( $form->options['form_class'], 'frm_submit_success' ) !== false ) {
+		?>
+	<div class="frm_half">
```
stephywells, line 82: The frm_half classes are causing issues with my form layout. I think we'll need to remove the extra html and do this with css instead.

classes/controllers/FrmFormsController.php
```diff
@@ -1892,14 +1892,19 @@ private static function show_form_after_submit( $args ) {
...
 	 */
 	private static function message_placement( $form, $message ) {
 		$place = 'before';
-		if ( ! empty( $message ) && isset( $form->options['form_class'] ) && strpos( $form->options['form_class'], 'frm_below_success' ) !== false ) {
-			$place = 'after';
+
+		if ( $message && isset( $form->options['form_class'] ) ) {
+			if ( strpos( $form->options['form_class'], 'frm_below_success' ) !== false ) {
+				$place = 'after';
+			} elseif ( strpos( $form->options['form_class'], 'frm_submit_success' ) !== false ) {
```
stephywells, line 1905: I think .frm_inline_success would be more consistent with the naming we use for other settings.

stephywells:
I realized why my form went weird with this class. I had the frm_half class in the last field. I think we need to add "grid-column-start: 1;" for the submit button.

Also, I think it should work to either have it left aligned right next to the submit button (like to show just "Success" or something short), or right aligned like the design. Maybe the submit button should span 3 columns? Or the message could be inside the .frm_submit div and use flex instead.

Crabcyborg:
Updated.

I've gone with the second solution of using flex. By default, I have it aligning left. I'm not sure if it's best to add padding or to leave that to the user.

Screen Shot 2020-08-12 at 1 33 55 PM

I also added a new class `frm_alignright_success` that text-aligns the message to the right instead.

stephywells:
Looks like we still need a bit more css to make this look good.

When I have a multi-line result, the submit button is stretched vertically. This solves the problem for me:
.with_frm_style .frm_inline_success .frm_submit{
align-items: flex-start;
}

The message also could use some spacing when left aligned, and be vertically centered when on a single line.

Crabcyborg:
Went with align-items: center as it gets them the agree the best vertically. The biggest thing getting in the way was the inconsistent margins, so I stripped them from both but added back the original `margin: 10px 0` for frm_button_submit to the frm_submit for this class.

classes/views/frm-entries/form.php
```diff
@@ -73,11 +73,24 @@
...
+		ob_implicit_flush( false );
+		FrmFormsHelper::get_custom_submit( $copy_values['submit_html'], $form, $submit, $form_action, $copy_values );
+		$clip = ob_get_clean();
+
+		ob_start();
+		ob_implicit_flush( false );
+		include FrmAppHelper::plugin_path() . '/classes/views/frm-entries/errors.php';
+		$message = ob_get_clean();
+
+		echo str_replace( '</div>', $message . '</div>', $clip ); // WPCS: XSS ok.
```
stephywells, line 90: @Crabcyborg Can we change this up a bit so it will only insert the message once even if there is another </div>?

Crabcyborg: For sure. I wasn't totally sure about that either.

Crabcyborg: Using the same preg_replace that I found in FrmMigrate.php. Tested it a bit, looks like it only changes the last one :)

stephywells:
Nice! Thanks @Crabcyborg !

PR #226, opened 2020-08-18
do not allow captcha override

Crabcyborg:
https://github.com/Strategy11/formidable-pro/issues/2582

This _should_ be fine. replaceCheckedRecaptcha actually only gets called when captcha is actually broken (which I guess we're basically out of luck because you can't really track that in any way that isn't exploitable), or if captcha isn't actually configured properly (should_show_captcha).

As far as I understand, there is no actual "if it was already verified." logic here.

### DIFF
--- a/classes/models/fields/FrmFieldCaptcha.php
+++ b/classes/models/fields/FrmFieldCaptcha.php
@@ -64,6 +64,14 @@ class FrmFieldCaptcha extends FrmFieldType {
 		return str_replace( ' for="field_[key]"', ' for="g-recaptcha-response"', $html );
 	}
 
+	/**
+	 * @return bool
+	 */
+	protected function should_show_captcha() {
+		$frm_settings = FrmAppHelper::get_settings();
+		return ! empty( $frm_settings->pubkey );
+	}
+
 	public function front_field_input( $args, $shortcode_atts ) {
 		$frm_settings = FrmAppHelper::get_settings();
 		if ( empty( $frm_settings->pubkey ) ) {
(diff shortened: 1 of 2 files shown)


stephywells:
Thanks @Crabcyborg!

Did you test this with ajax submit, and on a multi-paged form with the recaptcha not on the last page? I think those are the two cases that we were using that extra check for.

Crabcyborg:
I'll try that next 🙂

Crabcyborg:
It looks good. The $posted_fields variable only contains and validates the values on the one page. This validation is only called when there is a captcha field being validated on the page being posted.

Crabcyborg:
When recaptcha is not on the last page, it's definitely having issues right now. I'm looking into a workaround.

Crabcyborg:
Hey @stephywells what do you think about using transients here, like I've done in my latest update?

I'm not positive that the transient option isn't also spoofable, but it would require that they're able to actually spoof the value of the transient rather than just being able to add a basic field into the form to override this.

stephywells:
Maybe we could keep the recaptcha_checked hidden field, but use better value? The replaceCheckedRecaptcha js function is triggered after an ajax call. How about we return the value from the server that we want to check for? We should do a salted value or something?

I think that would be the safest, easiest option. It looks like the transients would be messed up with multiple people on at once.

Crabcyborg:
My thought was that the nonce would be unique enough not to have any conflicts.

What would the new response value have in it? Wouldn't we still need to save this new key in the database, possibly still with a transient?

stephywells:
The reason the current nonce is exploitable, is because it's shown in the page source. I don't think the database needs to get involved. I think if we use a nonce that isn't in the page source, that would solve the problem.

Since both of the "replaceCheckedRecaptcha" functions are in the same spot, we should be able to add another parameter to pass a new nonce value. It should work to add a new nonce to the response in line 464 (formidable.js). We can return the nonce regardless of if there was a captcha. The nonce in the page source is from wp_create_nonce( 'frm_ajax' ). Let's just return one with a different name so it's still secure.

PR #230, opened 2020-08-21
More secure captcha

Crabcyborg:


classes/views/frm-entries/form.php
```diff
@@ -34,6 +34,8 @@
 <input type="hidden" name="frm_hide_fields_<?php echo esc_attr( $form->id ); ?>" id="frm_hide_fields_<?php echo esc_attr( $form->id ); ?>" value="<?php echo esc_attr( $frm_hide_fields ); ?>" />
 <input type="hidden" name="form_key" value="<?php echo esc_attr( $form->form_key ); ?>" />
 <input type="hidden" name="item_meta[0]" value="" />
+<?php FrmFieldCaptcha::render_checked_response(); ?>
```
stephywells, line 37: Can we call this using the frm_after_title hook or frm_entry_form to avoid the extra code in lite?

Crabcyborg: Awesome, changed that. Love all of these actions and filters, just getting the hang of using them properly still :)

stephywells: That sounds like a good idea to me.

stephywells: well, actually, maybe it's not worth it. We'll need to check it specifically anyway. I think it's fine as is.

PR #231, opened 2020-08-24
Xml field import fix for preexisting fields

Crabcyborg:
Working from Steph's commit [here](https://github.com/Strategy11/formidable-forms/commit/43bc0cd6f91ad3358ac4fcd014882125b57c4bdd) I have a branch that can successfully import the default calculator value in from [here](https://github.com/Strategy11/formidable-pro/issues/2566).

I don't know for sure if this encompasses the whole issue for https://github.com/Strategy11/formidable-pro/issues/1362
My issue seems mostly specific to the fact that these fields already existed and weren't being created new on import.

I have to do two replaces, one that changes their original expected id to their field key, and one that changes their field key to their new id. This way I can avoid issues of it changing my value into an id, and then changing that id into another id that wasn't the proper match.

Sounds related to https://github.com/Strategy11/formidable-pro/issues/2487 as well.

### DIFF
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -358,6 +358,8 @@ class FrmXMLHelper {
 	 * TODO: Cut down on params
 	 */
 	private static function import_xml_fields( $xml_fields, $form_id, $this_form, &$form_fields, &$imported ) {
+		global $frm_duplicate_ids;
+
 		$in_section = 0;
 
 		foreach ( $xml_fields as $field ) {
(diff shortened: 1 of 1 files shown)


stephywells:
@Crabcyborg What would happen if I import a form with similar ids to the new form? For example, what is the imported form has ids like this:
5, 6, 7, 8, 9

And the new form is:
1, 2, 3, 4, 5

Would this replace 5 with 9? I think we might need skip any fields ids that were just created.

Crabcyborg:
I might refactor it so that I can isolate the updated fields to their own array, but I do think this is safe as it's all key based.

In my case it was doing:

```
6*[15] -> 6*[daq0l] -> 6*[61]
```

Anything new, if it went through this should just change from its item key and then back again, and it wouldn't end up trying to update. I think this basically just catches any cases where the key is the same but the id is reserved by another form.

Crabcyborg:
Thanks for the feedback Steph. I've added a new variable, `$keys_by_original_field_id`. It only replaces those now, instead of whatever might be in `$frm_duplicate_ids` at the time.

PR #237, opened 2020-08-29
Add aria descriptions to multiselect

Crabcyborg:
Works to resolve https://github.com/Strategy11/formidable-pro/issues/2599

Played a bit with comboboxes and listboxes, and where I ended up feels pretty functional but doesn't use combobox or listbox as I feel this isn't consistent with those specs and does not behave as expected. I'm switching between two aria-descriptions to handle both checked/unchecked state separately for each option.

While the bootstrap multiselect is missing some `aria` support, it does [have pretty good keyboard support](http://davidstutz.de/bootstrap-multiselect/#keyboard-support) and I think I was able to make this function alright with a screen reader.

I also added connections between the buttons and the labels so you know what you're actually targeting. [A few small code updates are required in pro as well](https://github.com/Strategy11/formidable-pro/pull/2602) to actually support this for all of the inputs.

It supports additional controls than what the bootstrap documentation mentions. When in the list, a screen reader navigates the options with `ctrl + option + left/right` as well as just the up/down arrows. I tested this and it works pretty well :)


classes/views/frm-forms/accessibility.php
```diff
@@ -0,0 +1,5 @@
+<div class="hidden">
+	<div id="frm_press_space_checked"><?php echo esc_html__( 'Checked. To uncheck this option, press Space or Enter', 'formidable' ); ?></div>
```
stephywells, line 2: Instead of "echo esc_html__", we can simplify a bit with "esc_html_e"

Crabcyborg: For sure, there are a lot of these functions, hard to keep up!

classes/helpers/FrmAppHelper.php
```diff
@@ -2648,6 +2648,10 @@ public static function locales( $type = 'date' ) {
 		return $locales;
 	}
 
+	public static function accessibility() {
+		require self::plugin_path() . '/classes/views/frm-forms/accessibility.php';
```
stephywells, line 2652: Would include_once make more sense here just in case this gets called multiple times?

classes/helpers/FrmAppHelper.php
```diff
@@ -2648,6 +2648,10 @@ public static function locales( $type = 'date' ) {
 		return $locales;
 	}
 
+	public static function accessibility() {
```
stephywells, line 2651: Can you please add a doc block here? Maybe the function name could be more clear like multiselect_accessibility?

Crabcyborg:
Thanks Steph! Should look better now :)

PR #240, opened 2020-09-04
confirm backend exit

Crabcyborg:
Resolves https://github.com/Strategy11/formidable-pro/issues/2564

I looked into customizing the text in the pop up but it's a no go these days.

From [a quote I found here](https://stackoverflow.com/questions/38879742/is-it-possible-to-display-a-custom-message-in-the-beforeunload-popup#answer-38880926)

> tl;dr - You can't set custom message anymore in most modern browsers
> A quick note (since this is an old answer) - these days all major browsers don't support custom message in the beforeunload popup. There is no new way to do this. In case you still do need to support old browsers - you can find the information below.

Screen Shot 2020-09-04 at 2 41 31 PM

I don't know if we want to go above and beyond this. It seems to cover most cases at least. If you were to toggle a checkbox on and off for instance, it would still say you've made changes even if you returned to the default. Maybe we want to check for that, but it adds a lot of complexity.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -2366,12 +2366,13 @@ class FrmAppHelper {
 		wp_register_script( 'formidable_admin_global', self::plugin_url() . '/js/formidable_admin_global.js', array( 'jquery' ), $version );
 
 		$global_strings = array(
-			'updating_msg' => __( 'Please wait while your site updates.', 'formidable' ),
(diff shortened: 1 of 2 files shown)


stephywells:
I think that's fine for now to not deal with customizing the message box. We can do that down the road if it's worth it.

Can you tell me if fieldsUpdated needs to be in the PHP? Would it work the same to add it as a var at the top of the function frmAdminBuildJS() { closure?

Crabcyborg:
@stephywells thanks for the feedback! There's no need for it to be global!

PR #241, opened 2020-09-04
update condition values immediately on radio button option text update

Crabcyborg:
Working toward a solution for https://github.com/Strategy11/formidable-pro/issues/1025

I see this as multiple items. I think the most common use case is also the easiest - update it right away for single option updates.

This of course doesn't help with the batch updates, which is a bigger beast.


stephywells:
I didn't test this yet, but I'm thinking it'll need a bit more. When a form is loaded, the settings are loaded in the page source of the builder form. When a field is selected, the options are moved into the form that is submitted on save. This greatly reduces processing for large forms. It looks like this would only fix the fields that have been selected.

To see this, you might need to check the box to load the builder page with ajax, and have more than 10 fields.

Crabcyborg:
@stephywells I think it's a lot better now. I rewrote most of the logic to be in vanilla js because it could end up doing a lot of DOM searching on a big form, I wanted it to be fast considering all that's going on.

I added a call to `moveFieldSettings` for fields that get updated in the search, I tested it and it was broken before the update and works now so I think we're good for that case now :)

js/formidable_admin.js
```diff
@@ -2328,6 +2328,70 @@ function frmAdminBuildJS() {
 		return false;
 	}
 
+	function onOptionTextFocus() {
+		var input = jQuery( this );
+		input.attr( 'data-value-on-focus', input.val() );
```
stephywells, line 2333: It there a way of knowing if this is the first time the value is getting set? If so, we could skip this.

stephywells (commented):
This is looking good! Do you have any ideas on how we could also handle logic in form actions and submit button logic?

js/formidable_admin.js
Crabcyborg: True, there's no need in setting this at all if fieldsUpdated is already set.

js/formidable_admin.js
```diff
@@ -2328,6 +2328,70 @@ function frmAdminBuildJS() {
...
+			rowIndex,
+			valueSelect,
+			opt;
+
+		if ( oldValue === newValue ) {
+			return;
+		}
+
+		fieldId = jQuery( this ).closest( '.frm-type-radio, .frm-type-checkbox' ).attr( 'data-fid' );
+		fields = document.getElementById( 'frm_builder_page' ).querySelectorAll( '.frm-single-settings' );
```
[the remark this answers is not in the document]
Crabcyborg: I'll play with it some more!

js/formidable_admin.js
Crabcyborg: This comment actually confuses me. The data-value-on-focus is important on every focus. Consider this case:

1. User has an option "Apples"
2. We set "Apples" as our condition value somewhere
3. We change "Apples" to "Oranges". When we first focused, it was "" so if we don't change it to "Apples" on the next focus, it stays as "" and wouldn't change the "Apples" assignment, but a "" assignment instead.

stephywells: Sorry. I wasn't very clear. I was thinking more along the lines of:
1. User adds a checkbox option
2. User edits the checkbox option to add a label.
3. Add the logic runs, but it's a new option so it shouldn't be included in any logic.

Crabcyborg: Ha, I was thinking backwards there, sorry!

A new option still needs to be synced into the list of options if the field is selected. It's true that this can be skipped though if the field is new and has never been selected for any logic. But that might be more work than its worth to check for as it could introduce more harm than good.

js/formidable_admin.js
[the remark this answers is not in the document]
Crabcyborg: @stephywells I reduced it quite a bit, by like 12 lines of code. My fairly basic performance timing was at 3ms and dropped to 1ms with this update.

js/formidable_admin.js
stephywells: Good point. I was thinking the options were already getting updated in logic. But that's only when a new row is added.

js/formidable_admin.js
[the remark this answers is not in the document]
stephywells: Awesome!! Thanks so much!

stephywells:
This looks great! It looks ready to merge to me. But before I do, I just want to make sure this won't completely change. Do you have any ideas on how we could also handle logic in form actions and submit button logic?

Crabcyborg:
Right, sorry, I missed that part! I'll look into this now :)

Crabcyborg:
@stephywells I'm working through the PHP-side updates for submit button logic and actions. I think the only other updates we'll need in free at in [this commit](https://github.com/Strategy11/formidable-forms/pull/241/commits/a35ac7848d9d6ee47c1c41efa8e19d52ae27abdc) that maps all of the updated option values.

stephywells (commented):
This will be huge! Thanks so much!

One question... what happens if someone intends for this to be a new value? For example, someone has three options:
Red, blue, yellow

Instead of reordering, they change red to blue, and blue to red. What happens?

Crabcyborg:
@stephywells that's definitely a good call, trying to figure out all of the ways it could go wrong. I think I'll still do something on the PHP side as well.

In [this new commit](https://github.com/Strategy11/formidable-forms/pull/241/commits/53e9d27a04b0b066bf9c00699d8f72e6aa3b6e2e) I check in any other options in the same area have the same value, and aren't flagged as duplicates. If it matches this, it will continue to register as its original value and it won't map the update.

PR #246, opened 2020-09-10
confirm bulk delete

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/1034

Screen Shot 2020-09-10 at 2 15 05 PM

### DIFF
--- a/classes/helpers/FrmListHelper.php
+++ b/classes/helpers/FrmListHelper.php
@@ -398,7 +398,12 @@ class FrmListHelper {
 
 		echo "</select>\n";
 
-		submit_button( __( 'Apply', 'formidable' ), 'action', '', false, array( 'id' => "doaction$two" ) );
+		$other_attributes = array(
+			'id'                          => "doaction$two",
+			'data-frmcaution-bulk_delete' => __( 'Heads up', 'formidable' ),
+			'data-frmverify-bulk_delete'  => __( 'ALL selected entries in this form will be permanently deleted. Want to proceed?', 'formidable' ),
+		);
+		submit_button( __( 'Apply', 'formidable' ), 'action', '', false, $other_attributes );
 		echo "\n";
 	}
 
--- a/classes/helpers/FrmListHelper.php
+++ b/classes/helpers/FrmListHelper.php
@@ -398,7 +398,19 @@ class FrmListHelper {
 
 		echo "</select>\n";
 
-		echo "<a id='confirm-bulk-delete-" . esc_attr( $which ) . "' class='frm-hidden' href='confirm-bulk-delete' data-frmcaution='" . esc_html__( 'Heads up', 'formidable' ) . "' data-frmverify='" . esc_html__( 'ALL selected entries in this form will be permanently deleted. Want to proceed?', 'formidable' ) . "'></a>";
+		if ( isset( $this->_actions['bulk_delete'] ) ) {
+			if ( FrmAppHelper::is_admin_page( 'formidable' ) ) {
+				$verify = 'ALL selected forms in this form will be permanently deleted. Want to proceed?';
+			} elseif ( FrmAppHelper::is_admin_page( 'formidable-entries' ) ) {
+				$verify = 'ALL selected entries in this form will be permanently deleted. Want to proceed?';
+			} else {
+				$verify = false;
+			}
+
+			if ( $verify ) {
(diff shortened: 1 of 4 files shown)


js/formidable_admin.js
```diff
@@ -514,6 +516,33 @@ function frmAdminBuildJS() {
...
 		jQuery( '.frm_bstooltip, .frm_help' ).tooltip( );
+
+		doAction = document.getElementById( 'doaction' );
+		doAction.addEventListener( 'click', function( event ) {
+			var link;
+
+			if ( ! confirmedBulkDelete && document.getElementById( 'bulk-action-selector-top' ).value === 'bulk_delete' ) {
+				event.preventDefault();
+
+				link = document.createElement( 'a' );
```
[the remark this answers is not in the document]
Crabcyborg: Probably. I can play with it!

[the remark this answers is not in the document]
Crabcyborg: @stephywells cleaned it up a fair bit. I don't think I was properly handling the second action button either until now. Should be better :)

js/formidable_admin.js
```diff
@@ -514,6 +516,30 @@ function frmAdminBuildJS() {
 		});
 
 		jQuery( '.frm_bstooltip, .frm_help' ).tooltip( );
+
+		jQuery( document ).on( 'click', '#doaction, #doaction2', function( event ) {
```
stephywells, line 520: Do we need a page ID in here? Or maybe a different message in the link depending on which page it is? This looks like it would get triggered on other pages as well like bulk deleting forms.

js/formidable_admin.js
```diff
@@ -514,6 +516,30 @@ function frmAdminBuildJS() {
...
+			var link,
+				isTop = this.id === 'doaction',
+				suffix = isTop ? 'top' : 'bottom';
+
+			doAction = this;
+
+			if ( ! confirmedBulkDelete && document.getElementById( 'bulk-action-selector-' + suffix ).value === 'bulk_delete' ) {
+				event.preventDefault();
+				confirmedBulkDelete = false;
+				confirmLinkClick( document.getElementById( 'confirm-bulk-delete-' + suffix ) );
```
stephywells, line 530: It might cause problems if we don't check if the link exists before preventDefault. If there are pages with the WP doaction ids, this might have unintended results.

js/formidable_admin.js
```diff
@@ -514,6 +516,30 @@ function frmAdminBuildJS() {
...
+				return false;
+			}
+		});
+
+		jQuery( document ).on( 'click', '#frm-confirmed-click', function( event ) {
+			if ( this.getAttribute( 'href' ) === 'confirm-bulk-delete' ) {
+				event.preventDefault();
+				confirmedBulkDelete = true;
+				jQuery( doAction ).click();
```
stephywells, line 539: Should we check if doAction is empty first?

stephywells (changes requested):
I think we'll need a bit of graceful failing in here. And let's check this on the bulk form delete too.

Crabcyborg:
@stephywells Thanks! I looked at other doaction pages and got more context of how to implement this!

I have a separate message now for bulk deleting forms as well, so we've got another part of the app covered!
Screen Shot 2020-09-14 at 12 49 16 PM

I check everywhere if it exists first before running much of the logic.

classes/helpers/FrmListHelper.php
```diff
@@ -398,6 +398,20 @@ protected function bulk_actions( $which = '' ) {
 
 		echo "</select>\n";
 
+		if ( isset( $this->_actions['bulk_delete'] ) ) {
+			if ( FrmAppHelper::is_admin_page( 'formidable' ) ) {
+				$verify = __( 'ALL selected forms in this form will be permanently deleted. Want to proceed?', 'formidable' );
```
stephywells, line 403: This message looks like it needs some adjustment. Maybe
"ALL selected forms and their entries will be permanently deleted. Want to proceed?"

classes/helpers/FrmListHelper.php
```diff
@@ -398,6 +398,20 @@ protected function bulk_actions( $which = '' ) {
 
 		echo "</select>\n";
 
+		if ( isset( $this->_actions['bulk_delete'] ) ) {
```
stephywells, line 401: Maybe we should add a new function for this message, so each of the classes that extend it can change it?

stephywells (changes requested):
Thanks! That looks more complete now.

classes/helpers/FrmListHelper.php
Crabcyborg: Wow, I didn't realize the helpers were set up like that. Made it a lot nicer, thanks for pointing that out!

stephywells (approved):
Nice! This looks great!

PR #248, opened 2020-09-11
clear fieldsUpdated value on save

Crabcyborg:
I found one side-effect of https://github.com/Strategy11/formidable-forms/pull/240

When you save the form, it will still say you have unsaved changes. I clear the flag on save and it does the trick :)


stephywells:
Thanks for the testing!

PR #251, opened 2020-09-16
New form modal

Crabcyborg:
https://github.com/Strategy11/formidable-pro/issues/1646

I was thinking I wanted to AJAX load the content, but it wouldn't work with the implementation of accordion. I think it's probably fine loading with the rest of the content since it is a pretty commonly used item on the page and it doesn't add a ton to the load.


classes/controllers/FrmFormsController.php
```diff
@@ -77,15 +77,6 @@ public static function list_form() {
 		return self::display_forms_list( $params, $message, $errors );
 	}
 
-	/**
-	 * Choose which type of form to create
-	 *
-	 * @since 3.06
-	 */
-	public static function add_new() {
```
stephywells, line 85: We should keep this for a graceful error for reverse compatibility in case it's been used somewhere

classes/controllers/FrmFormsController.php
```diff
@@ -795,39 +786,87 @@ public static function save_per_page( $save, $option, $value ) {
...
-		$pricing = FrmAppHelper::admin_upgrade_link( 'form-templates' );
+		array_walk(
+			$custom_templates,
+			function( &$template ) {
+				$template['custom'] = true;
+			}
+		);
 
-		$categories = self::get_template_categories( $templates );
+		$categories                            = array_merge( array( 'My Templates' ), $categories );
```
stephywells, line 857: Can we make 'My Templates' translatable?

classes/controllers/FrmFormsController.php
```diff
@@ -795,39 +786,87 @@ public static function save_per_page( $save, $option, $value ) {
...
+		global $frm_license_type;
 
-		$api       = new FrmFormTemplateApi();
-		$templates = $api->get_api_info();
+		$api           = new FrmFormTemplateApi();
+		$frm_templates = $api->get_api_info();
+		$expired       = false;
+		$license_type  = '';
+		if ( isset( $frm_templates['error'] ) ) {
+			$error        = $templates['error']['message'];
```
stephywells, line 809: PHP Notice:  Undefined variable: templates in /Users/Steve/Sites/wordpress2/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php on line 809

PHP Notice:  Trying to access array offset on value of type null in /Users/Steve/Sites/wordpress2/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php on line 809

I see this when my license is expired.

classes/views/shared/new-form-overlay.php
```diff
@@ -0,0 +1,198 @@
...
+					</a>
+					<?php
+					$admin_upgrade_link = FrmAppHelper::admin_upgrade_link(
+						array(
+							'medium'  => 'new-template',
+							'content' => 'renew',
+						)
+					);
+					?>
+					<a href="<?php echo esc_url( $admin_upgrade_link ); ?>" class="button button-primary frm-button-primary" target="_blank" rel="noopener">
```
stephywells, line 163: This should link to https://formidableforms.com/account/downloads/ instead of the upgrade page

stephywells (changes requested):
this is looking really nice! Great job!

I'm seeing a couple issues though.
- On the form listings page, nothing happens when I click "trash" for a form created from a template (bulk trash works)
- After I create a form from a template, the form doesn't have any fields
- let's add a hover effect on the +, lock, and eye icons to make it clear they are clickable

classes/views/shared/new-form-overlay.php
Crabcyborg: @stephywells sorry for the confusion, I just moved the url to a variable because of a code styling error that didn't make a lot of sense (I think because it was just doing too much inside of an `href=""`, but I had used the same variable name for both of the links and kept them in the view.

It was the same url you had put. I tested it and it does go to /downloads/ still.

I called it `$admin_upgrade_link` because that's the name of the function. It was a quick fix though, I'm changing it into two variables now, $upgade_link and $renew_link, and moving them out to the controller file.

classes/controllers/FrmFormsController.php
Crabcyborg: Thanks for catching that.

classes/controllers/FrmFormsController.php
Crabcyborg: Didn't catch that it was public, thanks! I moved it to the bottom and marked it as deprecated now.

Crabcyborg:
Thanks @stephywells

>     * On the form listings page, nothing happens when I click "trash" for a form created from a template (bulk trash works)

Nice catch, fixed that :)

>     * After I create a form from a template, the form doesn't have any fields

Thanks, I made a silly mistake, can't believe I missed that! Should be good now :)

>     * let's add a hover effect on the +, lock, and eye icons to make it clear they are clickable

Done :)

>     * when I click the "add new" button to open the modal, the "cancel" label has focus on first open

I'm not sure about this one. It's been around since the beginning and I've played it with for a while and it seems that it really really wants the default focus. I've tried using tabindexes and focus() events in JavaScript, but nothing seems to take effect. I don't want to just force a focus on nothing either because then I think we'd be doing more harm than good if we messed with screen readers.

classes/views/shared/new-form-overlay.php
stephywells: Ok, I thought I saw the link to the lite-upgrade page here. Thanks!

stephywells (approved):
This is looking great. I would like to review the categories and icons before we release this. I'm thinking there are some categories that we won't list here. At this point, we need to get 4.8 out, so I don't want to merge it in right now since it isn't already in the beta.

Thanks for an awesome job on this!

classes/controllers/FrmFormsController.php
```diff
@@ -795,39 +786,93 @@ public static function save_per_page( $save, $option, $value ) {
...
+		global $frm_expired;
+		global $frm_license_type;
+
+		$api           = new FrmFormTemplateApi();
+		$frm_templates = $api->get_api_info();
+		$expired       = false;
+		$license_type  = '';
+		if ( isset( $frm_templates['error'] ) ) {
+			$error        = $frm_templates['error']['message'];
+			$error        = str_replace( 'utm_medium=addons', 'utm_medium=form-templates', $error );
```
stephywells, line 810: @Crabcyborg Scrutinizer flagged this line since $error isn't getting used. Should it be used or removed?

PR #255, opened 2020-09-23
Better email confirmation js validation (and password js validation)

Crabcyborg:
Tackling the checklist items `Email confirmation fields don't show error message` and `Validate password confirmation fields` from https://github.com/Strategy11/formidable-pro/issues/1392

It requires the message as well, which comes from pro code [here](https://github.com/Strategy11/formidable-pro/pull/2634)

Without it, it just won't display still.

This update also fixes some inconsistencies with the email validation. I tried to make the logic easier to follow/more consistent for the both fields.

Previously, if the confirm field was invalid, it wouldn't become valid if you changed the original email address only to match the confirmation address.

As for the last item on the list, I'm not familiar with the option (`Default value will NOT pass validation`). I see it's defined in FrmAppHelper.php as `'no_valid_default'  => __( 'Default value will NOT pass form validation', 'formidable' ),` but I don't see references to no_valid_default. It looks like that support doesn't exist on the PHP side either.

https://github.com/search?q=org%3AStrategy11+no_valid_default&type=Code

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -375,11 +375,11 @@ function frmFrontFormJS() {
 			return;
 		}
 
-		if ( fieldID !== strippedFieldID ) {
+		if ( fieldID !== strippedFieldID && typeof errors[ 'conf_' + strippedFieldID ] === 'undefined' ) {
 			$firstField = $fields.filter( '[name="item_meta[' + strippedFieldID + ']"]' );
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Looks great. We'll still need testing on this, but it can be done before release

PR #256, opened 2020-09-24
clear values on placeholder match only if not required

Crabcyborg:
My go at fixing https://github.com/Strategy11/formidable-pro/issues/2636

I understand that we're probably stripping the placeholder because of issues with IE interpreting placeholders as values and sending data that we don't want.

I think it makes sense to keep that if the field isn't required to avoid a ton of junk data in the database, but I don't think we should stop people if they're legitimately trying to enter data into a required field and it matches the placeholder. Of course this means that we might still lose Jim's name if it's not required, so maybe we want add an option.

I also refactored the function a bit as it was a little funny to follow. I think it's fine to just handle it in pieces, with fewer variables and more comments.

### DIFF
--- a/classes/models/FrmEntryValidate.php
+++ b/classes/models/FrmEntryValidate.php
@@ -113,21 +113,29 @@ class FrmEntryValidate {
 		$errors = apply_filters( 'frm_validate_field_entry', $errors, $posted_field, $value, $args );
 	}
 
+	/**
+	 * Set $value to an empty string if it matches the placeholder or its label
+	 *
+	 * @param object $field
+	 * @param string $value
+	 */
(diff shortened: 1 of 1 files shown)


stephywells (commented):
The reason for not allowing this was less about IE (we don't do placeholders there). The reason was because if it's an intended value, then it should be a default value instead. But since it doesn't stop people from doing it, we might as well get rid of it. It's also a bit of old baggage from before we used html placeholders.

Maybe we should just remove this completely?

Crabcyborg:
@stephywells do you think any of this is useful? Does the part about inside labels still apply, or are we dropping the entire function?

stephywells:
@Crabcyborg well, it wouldn't hurt to keep that part. I think that's fine to leave.

PR #264, opened 2020-10-07
cs fix, do not exclude extract rule

Crabcyborg:
Trying to slowly remove exclusions from our code sniffing. Extract only appears 4 times, and I figure it'd better to put the ignore on each of the 4 instances than to allow more in without the flag.

I looked at all 4 instances. I don't think any are too easy to write the extract out from.

Technically removes two rules because it seemed to have duplicates under different namespaces.

There were also duplicate exclusions for `WordPress.NamingConventions.ValidFunctionName.FunctionNameInvalid` which happens to be valid already, so I'm removing both.

### DIFF
--- a/classes/controllers/FrmStylesController.php
+++ b/classes/controllers/FrmStylesController.php
@@ -416,7 +416,7 @@ class FrmStylesController {
 	}
 
 	public static function include_style_section( $atts, $sec ) {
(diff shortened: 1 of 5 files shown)


stephywells (approved):
Thanks for checking these!

PR #265, opened 2020-10-07
better confirm field validation for embedded forms

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/1392 for embedded forms

The issue was that the name didn't match the `'conf_' + strippedFieldID` pattern in embedded fields.

I've figured out how to improve the process a bit. I removed the requirement to pass along fields, and I check for exact expected field id values. Uses fewer variables, less jQuery, and it works for embedded forms no problem.

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -151,7 +151,7 @@ function frmFrontFormJS() {
 	}
 
 	function validateForm( object ) {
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Nice improvements!

PR #266, opened 2020-10-07
dropdown types should also update conditional logic on option change

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/1025 again

Dropdowns just weren't in the scope of my update. I neglected that there were other types with options than radio and checkbox. I made it more generic, targeting the options themselves rather than specific types (it was hard coded to radios and checkboxes only, ignoring select).

I took out the hardcoding so this should catch anything else if there were another type I had missed.


stephywells (approved):
Thank you! Looks good to me.

PR #268, opened 2020-10-08
fix add on error handling

Crabcyborg:
I noticed a lot of stuff breaks when you get an add on error.

Screen Shot 2020-10-08 at 9 20 31 AM

There is no loader variable in this file outside of these two places. I think old code was left behind.

After I cleaned that up and tried removing the loading class from the button, the error message looks totally broken as well.
Screen Shot 2020-10-08 at 9 27 37 AM

I changed a lot of the styling to make it simpler. Does this look good?
Screen Shot 2020-10-08 at 9 23 19 AM

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -1731,6 +1731,10 @@ h2.frm-h2 + .howto {
 }
 
 /* Add-on tiles */
+.frm-addons .frm-card.frm-addon-not-installed {
+	position: relative;
+}
+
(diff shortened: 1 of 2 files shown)


stephywells:
@Crabcyborg is there a reason this is closed?

Crabcyborg:
@stephywells yeah I noticed your other update touches this code as well, so I made the other PR instead as some of this applies differently now.

PR #269, opened 2020-10-08
actually show error when ftp creds fail

Crabcyborg:
I'm not sure how much I totally know this whole new process or what is actually still missing in https://github.com/Strategy11/formidable-pro/issues/2649 but I see some issues at the moment with some old code.

I'd like to know, are we planning to ever actually support this ftp credentials form, or does it make sense to remove, at least temporarily for the readability of the code.

There are `loader.hide();` lines referring to a loader variable that does not exist anymore. Clearly this code is outdated.

This PR is into `develop` not master. I saw that in some of new updates the `'form' =>` key was dropped [here](https://github.com/Strategy11/formidable-forms/commit/d5ae8315f24177b5cd3fc6053da16113fd4dfa82#diff-3f0754e59a2323f0f79b3b84d882f2b2L863) in favour of `wp_send_json_error` but the js logic was checking for a `.error`, which is not what you get when you call wp_send_json_error this way.
Screen Shot 2020-10-08 at 10 07 19 AM

Does the alert look fine? It goes beyond the borders of the box its in.
Screen Shot 2020-10-08 at 10 14 13 AM

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -5276,8 +5276,8 @@ function frmAdminBuildJS() {
 			},
 			success: function( response ) {
 				// If there is a WP Error instance, output it here and quit the script.
-				if ( response.error ) {
-					addonError( response, el, button );
+				if ( ! response.success ) {
+					addonError({ error: response.data }, el, button );
 					return;
 				}
 


stephywells:
@Crabcyborg is this what you're referring to? If so, we should probably fix the php so we can use the same error responses for both.
https://github.com/Strategy11/formidable-forms/blob/develop/classes/controllers/FrmAddonsController.php#L869

Here's my code for the api looking for a response:
https://github.com/Strategy11/s11-edd/blob/master/s11EddApiConnect.php#L79

Feel free to make some changes! I haven't tested this endpoint yet.

PR #273, opened 2020-10-08
update everyone multiselect functionality for secure file roles

Crabcyborg:
@stephywells were you actually noticing this Everyone always getting selected behaviour on the protected files dropdown rather than the field visibility?

Because it seemed to be an issue here, but I was checking field visibility.


stephywells:
Yep, exactly. It was the file protection.

PR #274, opened 2020-10-08
Install Pro via API

stephywells:


Crabcyborg:
Added some logic for the WordPress form error.

From https://developer.wordpress.org/reference/functions/request_filesystem_credentials/
```
if ( $error ) {
        $error_string = __( '<strong>Error</strong>: Could not connect to the server. Please verify the settings are correct.' );
        if ( is_wp_error( $error ) ) {
            $error_string = esc_html( $error->get_error_message() );
        }
        echo '<div id="message" class="error"><p>' . $error_string . '</p></div>';
}
```

Sometimes the form echoed is actually just an error message. This explains a lot of the confusing logic that was checking for form as an error. I check the value of form now in JavaScript to distinguish this issue.

Screen Shot 2020-10-08 at 4 47 37 PM

PR #275, opened 2020-10-09
force content checkbox for view settings

Crabcyborg:
Should fix https://github.com/Strategy11/formidable-pro/issues/2633

I'm still not sure how it's getting triggered for people. There are these `hide-if-js` and `hide-if-no-js` classes that get added depending on the state of the checkbox, that might have a plugin conflict. Nat had mentioned a plugin conflict, could be the same for Kristine.

I played with unchecking the box and refreshing the page, seeing if I could recover from that. I need to do a 0 second timeout so that it happens after WordPress initializes their event listeners. Otherwise, I'd have to do some hacky interpretation of what happens when the checkbox is toggled.

But who knows. I haven't gotten enough detail from anyone experiencing the issue to know IF the checkbox is even the issue. Maybe something is just hiding the container directly with css.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -5953,7 +5953,7 @@ function frmAdminBuildJS() {
 		}
 
 		$contentCheckbox = $checkboxes.filter( '[name="frm_dyncontent-hide"]' );
-		if ( ! $contentCheckbox.is( ':checked') ) {
+		if ( ! $contentCheckbox.is( ':checked' ) ) {
 			setTimeout(
 				function() {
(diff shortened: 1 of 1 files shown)


stephywells:
@Crabcyborg were you able to see this on Kristines sandbox account?

Crabcyborg:
@stephywells the bug has nothing to do with these screen options. I'm just going to close this PR. I think we might as well remove the other line that disables the screen options checkbox as well since it's not really necessary to anything.

Still looking into what the real issue is though.

PR #278, opened 2020-10-09
fix issue with disappearing view sections

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2633

**TODO**
Keep the classes, make them possible to toggle on and off


stephywells (approved):
Let's hold this for 4.08.01

Crabcyborg:
I've spent some more time with this item, and I've updated the JavaScript for the shortcode insert so that it works from the content as well.

These are also draggable, and I think no matter what we try doing with the CSS it won't be enough. I looked over postbox.js as well and it's not very easy to prevent WordPress from doing its thing.

I think we should just leave it simple and try not messing with it, allow users to use it how they'd like just like WordPress intended with the meta boxes.

This of course is only temporary until the new Views update anyway as this will all be replaced.

stephywells:
I don't think we'll be able to do this by default since the sidebar box can no longer be clicked to insert shortcodes. Maybe we can add a lock/unlock icon that will remove the sticky styling? Maybe it could just add and remove a class with css that overrides the sticky?

PR #280, opened 2020-10-14
better action conditional logic call to action

Crabcyborg:
I noticed in the aweber notes that you can use the plugin without pro.

I tried this out and there are some UI items that show up for emails for all of the other action types as well. It looks sort of messy, since none of the other types really support Attachments.

I made it so we can extend our FrmFormActions to handle custom text but for now most of the plugins will read as `Conditional form actions` instead of `Conditional emails` which is still a nice step up. I already extended email so it still says Condtional emails :)

This update also extends `data-medium` a bit, someone could click through now on `conditional-zapier` and `conditional-aweber` rather than all being marked as `conditional-email`. I figured this was useful to distinguish.

Screen Shot 2020-10-14 at 1 07 57 PM

Otherwise the aweber plugin does work well without pro.

### DIFF
--- a/classes/models/FrmFormAction.php
+++ b/classes/models/FrmFormAction.php
@@ -834,4 +834,18 @@ class FrmFormAction {
 
 		return apply_filters( 'frm_action_triggers', $triggers );
 	}
+
+	public function render_conditional_logic_call_to_action() {
+		?>
+			<h3>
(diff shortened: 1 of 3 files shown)


stephywells (approved):

stephywells:
Nice catch! Saving this for 4.08.01.

PR #286, opened 2020-10-20
unique repeaters post check fix

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/1913

I used a global variable as the validation seems to happen with different field object instances despite validating the same field, so storing the variables on the object didn't work. I made some unit tests to reflect that (that two different field objects for the same field still track duplicates between one another, and that another unrelated field object doesn't).

### DIFF
--- a/classes/models/fields/FrmFieldType.php
+++ b/classes/models/fields/FrmFieldType.php
@@ -1082,13 +1082,70 @@ DEFAULT_HTML;
 		// Override in a child class.
 	}
 
+	/**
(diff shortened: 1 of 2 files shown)


stephywells (approved):

jairoprez:
@stephywells I did several tests based on these criteria https://github.com/Strategy11/formidable-pro/issues/1913 and as a result I can say that the Unique + repeaters conflict has been fixed

stephywells:
Awesome! Thanks so much @jairoprez !!

PR #287, opened 2020-10-22
Open an action before duplicate if it has not loaded yet

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2703

I also noticed that when the duplicate icon is clicked that it the value of `target.parentElement.className` is a [SVGAnimatedString](https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString) which does not have an indexOf function. It throws a JavaScript error, so I'm checking for a string value first.


stephywells (approved):
Nice!

PR #288, opened 2020-10-27
order radio options when syncing conditional logic

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2684

The big issue with them going to the top was because they were replacing the empty placeholder option, so I made it so they don't.

As for the ordering, I check the list of options and run it against the current conditional logic. This also handles the automatic insertion of the missing option, so it fills in a few gaps (adding it when Other is clicked, as well as when a New Option is added before the text is ever updated, and when the options are reordered).


stephywells (approved):
Looks great. We can merge once it's tested.

stephywells (changes requested):
I'm seeing an issue.

1. Create a radio field.
2. Create another field with logic based on the radio.
3. Save the form.
4. Add another radio option but don't change the name
5. See the options in the logic. It's correct.
6. Change the name of the new option
7. See the options in the logic. "New Option" still shows in addition to the updated value.

Crabcyborg:
@stephywells thanks for pointing that out, pushed a fix that check for the default value that gets cleared. It's a little different since the value on focus is "New Option" and then suddenly switches to "" on click. I'm setting the value on focus now before that clear happens as well, and also checking when I set the focus value, making sure that we only set the value on focus in other cases.

js/formidable_admin.js
```diff
@@ -2481,7 +2484,14 @@ function frmAdminBuildJS() {
 
 			logicId = row.id.split( '_' )[ 2 ];
 			valueSelect = row.querySelector( 'select[name="field_options[hide_opt_' + logicId + '][]"]' );
-			optionMatches = valueSelect.querySelectorAll( 'option[value="' + oldValue + '"]' );
+
+			console.log( 'oldValue', oldValue );
```
stephywells, line 2488: Looks like this straggler got left behind.

Crabcyborg: Aw man, thanks for catching that.

stephywells (approved):
Thank you! This needs testing and we can merge.

stephywells:
This is looking good to me!

PR #291, opened 2020-10-29
use conditional addon button on views upsell page

Crabcyborg:
- [x] Views upsell page should use the conditional Install/Activate/Upgrade link used on the Add-ons page instead of the static Upgrade link

The new upsell updates for views require some tweaks in the free plugin

- Adds views to the fallback plugin list
- Add a few helper functions to make it easier to access certain API details when just looking for a single thing
- The views upsell page now uses the same button logic as the addons list page (which is now moved into two different helper functions, one that manages the rendering and one that connects the addon to the rendering function without exposing the whole addon api)
- Do the same css adjustment for the messages warning that I did in an earlier update to add padding for errors


stephywells (approved):
Looks great! Will do some testing on this tomorrow.

PR #292, opened 2020-10-29
include beta in s11 request if views is nested

Crabcyborg:
I think this covers

- [x] Update FrmProEddController to set $beta to true when the nested views plugin is being used. This will allow us to serve updates that include the nested views for a while.

Couldn't identify a better way of handling this. FrmProEddController seems to use FrmFormApi directly.


stephywells:
@Crabcyborg sorry I wasn't very clear about this. Here's the way we did it before so it doesn't mess with other plugins:
https://github.com/Strategy11/formidable-pro/commit/200d79b62be0c4db118acbe320466d3f77442022#diff-ddc8aaf2342df7f3d0eac1e312c6716a

The main thing that is needed is $this->get_beta = true;

PR #296, opened 2020-11-04
toggle sticky customization meta box

Crabcyborg:
Made a new PR to follow up from https://github.com/Strategy11/formidable-forms/pull/278

Now there's a lock in the customization header that can toggled on and off to add these sticky effects (that are on by default).

Screen Shot 2020-11-04 at 3 18 16 PM

Also requires https://github.com/Strategy11/formidable-views/pull/34 to put the SVG html onto the page. I wonder if we want to put more into views though. What are the main reasons that the code for views is in free? Just to keep the css and js in one file? Is this just how it started and it was never changed?

Fixes https://github.com/Strategy11/formidable-pro/issues/2633


stephywells:
Yeah, it all started in one file and hasn't gotten separated. We have gotten most of the front end js separated but not the css and admin js. Maybe we can add this only in the views plugin now? Ideally, it could be done without needing updates to another plugin.

Crabcyborg:
@stephywells think I had only read the initial message before moving forward with this so I moved it all. I do like the idea of keeping views updates in views regardless. I think this will still be easier to clean up than having things in two plugins.

I was able to get everything in here only using some different css rules.

PR #297, opened 2020-11-05
embedded section field export fix

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2736

The issue here was that a field in an embedded section is really similar looking to a field in a repeater, so the check was getting both of those when it should only be getting the one.

I had to move some of the repeater logic higher up, and then I moved it into a function to made this all a little easier to follow.


stephywells:
Awesome! Is it possible to catch this with a unit test too?

stephywells:
Kristine tested this. Merging.

PR #304, opened 2020-11-12
better comment code sniffing

Crabcyborg:
Trying to make our comments more consistent similar to my update in Views. There were a few extra exceptions to add here in free, I think because we have more comments in general.

A ton of functions here were doing a `@param` flag for one variable but leaving out three others. A lot of the time the type was specified on the right side of the variable name instead of the left as well.

I also tried to go more consistent with the placement of the `@param`, vs `@return`, vs `@since`. Going off of the WordPress docs, I think it should be spaced out like:

```
/**
 * @since
 *
 * @param
 * @return
 */
```

Where `@param` and `@return` are grouped, and `@since` is always separated, and above the other two.

I also removed some commented code that seemed outdated.

Tried to guess the types are best as I could in some cases where it was less obvious.

This should all be safe. I think I may have changed a `==` into `===` on a couple string comparisons but other than that I left the code alone.


Crabcyborg:
I'm not in love with some of these changes. Nothing seems that bad in the comments here honestly. I think I'll revisit the idea.

PR #305, opened 2020-11-17
add a unit test to confirm that calculations do not get stripped when sanitized

Crabcyborg:
Adds a unit test for https://github.com/Strategy11/formidable-pro/issues/2759

I don't actually think there's an issue in free after I've isolated this. But this function is where I started, and it deserves a unit test too.


stephywells (approved):
Awesome!

PR #307, opened 2020-11-18
repeater export fix

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2766

The issue here is that the metas aren't saved in the database, so I need to look at the repeater's fields and fill in the blanks as I populate the repeater data. There's really only one place that's possible to set it, in this repeater fill function after the loop. Added several unit tests, but they're in a [separate PR in pro](https://github.com/Strategy11/formidable-pro/pull/2773)

### DIFF
--- a/classes/helpers/FrmCSVExportHelper.php
+++ b/classes/helpers/FrmCSVExportHelper.php
@@ -308,21 +308,7 @@ class FrmCSVExportHelper {
 				$entries[ self::$entry->parent_item_id ]->metas[ $meta_id ][] = $meta_value;
 			}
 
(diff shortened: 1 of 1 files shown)


jairoprez:
@Crabcyborg the code looks good and the Repeater Export issue https://github.com/Strategy11/formidable-pro/issues/2766 is fixed

jairoprez (commented):
@Crabcyborg the code looks good and the Repeater Export issue Strategy11/formidable-pro#2766 is fixed

stephywells:
Thanks so much for the fix @Crabcyborg and the testing @jairoprez !

PR #308, opened 2020-11-18
add warning when leaving form settings

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2764

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -6790,7 +6790,7 @@ function frmAdminBuildJS() {
 			});
 
 			jQuery( document ).on( 'frm-multiselect-changed', '#protect_files_role', adjustVisibilityValuesForEveryoneValues );
- 
+
 			jQuery( document ).on( 'submit', '.frm_form_settings', settingsSubmitted );
 			jQuery( document ).on( 'change', '#form_settings_page input, #form_settings_page select', fieldUpdated );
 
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -6786,6 +6791,9 @@ function frmAdminBuildJS() {
 
 			jQuery( document ).on( 'frm-multiselect-changed', '#protect_files_role', adjustVisibilityValuesForEveryoneValues );
 
+			jQuery( document ).on( 'submit', '.frm_form_settings', settingsSubmitted );
+			jQuery( document ).on( 'change', '#form_settings_page input:not(.frm-search-input), #form_settings_page select', fieldUpdated );
```
jairoprez, line 6795: Warning is working properly with input type fields but I think you forgot to cover the textarea fields https://recordit.co/nqOgv4f7IO, this should do the trick `#form_settings_page textarea`

jairoprez (commented):
@Crabcyborg This looks good to me, I found just a small detail that was overlooked and can be easily resolved I think

js/formidable_admin.js
Crabcyborg: Ohhh nice catch @jairoprez !

Crabcyborg:
@jairoprez thanks for considering the textarea. I neglected them in the original update, and added to that as well. Took a second to hunt one down, but it looks like you wouldn't get the warning after editing content for an HTML field either.

jairoprez:
@Crabcyborg Nice job! - select fields are now covered.

jairoprez:
Hey @stephywells this PR looks good to me.

stephywells:
Thanks you two! This looks great!

PR #309, opened 2020-11-20
do not disable the screen options checkboxes

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-views/issues/46

I made a PR at one point to put this back but it had other updates. Meant to make another one with just this, didn't. Missed that this is also triggered from the listing page.

There's really no need for this line. I put it in here trying to fix the disappearing view sections issue, but this wasn't the final fix.


stephywells:
Thanks @Crabcyborg !

PR #310, opened 2020-11-20
unlock all templates on free code verification

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2774

I was getting some pretty odd behaviour before also clearing the caches. I noticed that the cache key wasn't accounting for the free license either.

But the main thing is the new `updateTemplateModalFreeUrls` function that updates every free template, setting the urls that we can get. I dropped the break from the template loop and set everything to an array if it's free and has a url. Then I check that array for the specific key. Since we require these urls, there's a bit involved but nothing too crazy.

One thing though, in this new update, the license getting saved isn't base 64 encoded in the options anymore. This would mean that any previous option would no longer work. Is that fine? Do we want to revert that?

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -299,7 +299,7 @@ class FrmAppController {
 		if ( ! FrmAppHelper::pro_is_installed() ) {
 			// avoid rendering the email and code blocks for users who have upgraded or have a free license already
 			$api = new FrmFormTemplateApi();
-			if ( ! $api->get_free_license() ) {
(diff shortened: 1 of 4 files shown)


stephywells:
@Crabcyborg the license was getting saved encoded in the current version. The problem is that is was getting encoded a second time in the url, so sometimes it was encoded once, and sometimes twice. I added a workaround on the server side to cover it so we could get it added. 

Thanks for updating this! I'll take a look next week.

stephywells (approved):
Looks great!

PR #311, opened 2020-11-26
Fix issue of selected fields not updated in autoresponder when form is duplicated

jairoprez:
Fixes https://github.com/Strategy11/formidable-autoresponder/issues/12

The selected fields were not updating properly on the duplicate forms because the duplicate was being made keeping the id of the original date or time field from which the duplicate was made.

I added a filter hook which allows adding the autoresponder fields that need to be replaced by new id of the duplicate form.

I have made use of this new hook in this PR https://github.com/Strategy11/formidable-autoresponder/pull/17 that is also part of this solution.


classes/models/FrmFormAction.php
```diff
@@ -271,6 +271,10 @@ public function duplicate_one( $action, $form_id ) {
 		$action->menu_order = $form_id;
 		$switch             = $this->get_global_switch_fields();
 		foreach ( (array) $action->post_content as $key => $val ) {
+			if ( $key == 'autoresponder' ) {
```
Crabcyborg, line 274: It's best to use `===` when possible. Our phpcs rules trigger warnings otherwise.

classes/models/FrmFormAction.php
```diff
@@ -271,6 +271,10 @@ public function duplicate_one( $action, $form_id ) {
...
+                $action->post_content[ $key ] = $this->switch_autoresponder_field_ids( $action->post_content[ $key ], $val );
```
Crabcyborg, line 275: This line is really confusing me a bit. `$action->post_content[ $key ]` should be equals to `$val` all the time. They're the same variable referenced two different ways.

We can probably just pass $val only and use it for both purposes.

classes/models/FrmFormAction.php
```diff
@@ -321,6 +325,40 @@ private function duplicate_array_walk( $action, $subkey, $val ) {
...
+                $new_send_date = implode( '-', $tmp_ids );
+            } else {
+                $new_send_date = trim( FrmFieldsHelper::switch_field_ids( '[' . $old_send_date . ']' ), '[]' );
+            }
+
+            $action['send_date'] = $new_send_date;
+        }
+
+        if ( ! empty( $old_send_after_interval_field ) ) {
+            $new_send_after_interval_field = trim( FrmFieldsHelper::switch_field_ids( '[' . $old_send_after_interval_field . ']' ), '[]' );
```
Crabcyborg, line 355: `trim( FrmFieldsHelper::switch_field_ids( '[' . $variable . ']' ), '[]' );` repeats a few times. Enough that I think we should introduce a new function that prevents all of this duplication.

classes/models/FrmFormAction.php
```diff
@@ -321,6 +325,40 @@ private function duplicate_array_walk( $action, $subkey, $val ) {
...
+        );
+        $old_send_date = $val['send_date'];
+        $old_send_after_interval_field = $val['send_after_interval_field'];
+
+        if ( ! in_array( $old_send_date, $values_to_ignore ) ) {
+            $pos = strpos( $old_send_date, '-' );
+
+            if ( $pos !== false ) {
+                $pieces = explode( '-', $old_send_date );
+                $tmp_ids = array();
```
Crabcyborg, line 341: Do you think we could rename this variable to something more descriptive? It might help with readability.

classes/models/FrmFormAction.php
```diff
@@ -321,6 +325,40 @@ private function duplicate_array_walk( $action, $subkey, $val ) {
...
 
+	private function switch_autoresponder_field_ids( $action, $val ) {
+        $values_to_ignore = array(
+            'create',
+            'update',
+        );
+        $old_send_date = $val['send_date'];
+        $old_send_after_interval_field = $val['send_after_interval_field'];
+
+        if ( ! in_array( $old_send_date, $values_to_ignore ) ) {
```
Crabcyborg, line 336: Similar to using `===`, the phpcs rules prefer that we pass the `$strict = true` flag when calling in_array https://www.php.net/manual/en/function.in-array.php

classes/models/FrmFormAction.php
Crabcyborg, line 274: Instead of checking this in a loop, since the array is associative, what about just checking this before the loop once?

```
if ( isset( $action->post_content['autoresponder'] ) ) {
    $action->post_content['autoresponder'] = ...
}
```

Crabcyborg:
@jairoprez it looks when you were indenting with spaces instead of tabs. You can see it blowing up in the phpcs https://github.com/Strategy11/formidable-forms/pull/311/checks?check_run_id=1461231698#step:3:294

Crabcyborg:
@jairoprez I've tested this and my field value stuck, so that's awesome.

What do you think about adding a unit test for the new function?

classes/models/FrmFormAction.php
jairoprez: This will no longer be necessary, since I applied the change that you indicated below to do the check only once outside the loop

classes/models/FrmFormAction.php
```diff
@@ -270,6 +270,12 @@ public function duplicate_one( $action, $form_id ) {
 
 		$action->menu_order = $form_id;
 		$switch             = $this->get_global_switch_fields();
+
+		// applies only if the action contains an autoresponder
```
Crabcyborg, line 274: I feel like this comment is sort of redundant. I think we could probably even improve readability by moving this check into the function itself. We could just do something like `$action = $this->maybe_switch_autoresponder_field_ids( $action );`, keep the code that calls the function as simple as possible.

Crabcyborg:
@jairoprez I'm seeing `PHP Notice:  Undefined index: send_after_interval_field` when I duplicate my form.

classes/models/FrmFormAction.php
```diff
@@ -321,6 +327,44 @@ private function duplicate_array_walk( $action, $subkey, $val ) {
...
+		$old_send_after_interval_field = $action['send_after_interval_field'];
+
+		if ( ! in_array( $old_send_date, $values_to_ignore, true ) ) {
+			$pos = strpos( $old_send_date, '-' );
+
+			if ( $pos !== false ) {
+				$pieces = explode( '-', $old_send_date );
+				$new_field_ids = array();
+
+				$new_field_ids[] = $this->trim_and_switch_field_ids( $pieces[0] );
```
Crabcyborg, line 345: I feel like we might be able to make this cleaner still. Maybe something like..

```
foreach ( $pieces as $piece ) {
    $new_field_ids[] = $this->trim_and_switch_field_ids( $piece );
}
```

classes/models/FrmFormAction.php
```diff
@@ -321,6 +327,44 @@ private function duplicate_array_walk( $action, $subkey, $val ) {
...
+
+				$new_send_date = implode( '-', $new_field_ids );
+			} else {
+				$new_send_date = $this->trim_and_switch_field_ids( $old_send_date );
+			}
+
+			$action['send_date'] = $new_send_date;
+		}
+
+		if ( ! empty( $old_send_after_interval_field ) ) {
```
Crabcyborg, line 356: I think we might benefit to split the logic again into two functions, one that adjusts the send_date value and one that adjusts the interval field value. Since they're totally separate, it helps to keep the scope apart.

Crabcyborg:
@jairoprez There's a hook that gets called after the form is duplicated that triggers the action duplications.

`add_action( 'frm_after_duplicate_form', 'FrmFormActionsController::duplicate_form_actions', 20, 3 );`

Do you think we could use this hook and actually fix this still in the autoresponder PR using this fix, but without having to update the free plugin?

classes/models/FrmFormAction.php
```diff
@@ -321,6 +324,57 @@ private function duplicate_array_walk( $action, $subkey, $val ) {
 		return $action;
 	}
 
+	private function switch_autoresponder_field_ids( $action ) {
+		$action = $this->switch_autoresponder_send_date_field_ids( $action );
```
Crabcyborg, line 328: I think we could put the `if ( isset( $action->post_content['autoresponder'] ) ) {` check in this function and avoid it in the other 2 to avoid duplicating code.

classes/models/FrmFormAction.php
```diff
@@ -321,6 +324,57 @@ private function duplicate_array_walk( $action, $subkey, $val ) {
...
+		return $action;
+	}
+
+	private function switch_autoresponder_send_date_field_ids( $action ) {
+		if ( isset( $action->post_content['autoresponder'] ) ) {
+			$values_to_ignore = array(
+				'create',
+				'update',
+			);
+			$old_send_date = ( isset( $action->post_content['autoresponder']['send_date'] ) ) ? $action->post_content['autoresponder']['send_date'] : '';
```
Crabcyborg, line 340: This would probably be more readable if we return early when `$action->post_content['autoresponder']['send_date']` is not set.

classes/models/FrmFormAction.php
```diff
@@ -321,6 +324,57 @@ private function duplicate_array_walk( $action, $subkey, $val ) {
...
+
+			if ( ! empty( $old_send_date ) && ! in_array( $old_send_date, $values_to_ignore, true ) ) {
+				$pos = strpos( $old_send_date, '-' );
+
+				if ( $pos !== false ) {
+					$pieces = explode( '-', $old_send_date );
+					$new_field_ids = array();
```
Crabcyborg, line 347: The line below this also created a new array so this line shouldn't be necessary anymore.

classes/models/FrmFormAction.php
jairoprez: I thought about that too, but if for some reason one of the other 2 functions is used directly in the future, that could cause problems, what do you think about that?

Crabcyborg: Fair enough, yeah.

I think we'll benefit enough as long as we move it out a bit and check something like:

```
if ( ! isset( $action->post_content['autoresponder'] ) || ! isset( $action->post_content['autoresponder']['send_date'] ) ) {
    return $action;
}
```

Just trying to tuck it all away more. We can probably even make yet another function that covered this for both, like a `check_that_autoresponder_content_key_is_set( $key )` that checks both for `post_content['autoresponder']` and then also for for `post_content['autoresponder'][ $key ]`

jairoprez: I took this approach to offer the ability to call both functions at the same time from one function and also make it possible to call each function separately if needed

jairoprez: I'm not sure if adding this function `check_that_autoresponder_content_key_is_set` would be appropriate since both checks are a little different now

jairoprez:
hook



> hook



> add_action( 'frm_after_duplicate_form', 'FrmFormActionsController::duplicate_form_actions', 20, 3 );

I'm going to play with this a bit to see what we can get

jairoprez:
> hook



> add_action( 'frm_after_duplicate_form', 'FrmFormActionsController::duplicate_form_actions', 20, 3 );

Good news!, I managed to implement the solution directly in the formidable-autoresponder plugin through the hook that you suggested.

Thanks!

Crabcyborg:
> Good news!, I managed to implement the solution directly in the formidable-autoresponder plugin through the hook that you suggested.
> 
> Thanks!

@jairoprez Awesome, the other update looks great! I'll close this PR since it's no longer needed.

jairoprez:
@Crabcyborg I have reopened this PR since I have placed here a new hook in the function suggested by Steph

classes/models/FrmFormAction.php
```diff
@@ -313,6 +314,15 @@ private function duplicate_array_walk( $action, $subkey, $val ) {
 				} elseif ( $ck == $subkey && isset( $frm_duplicate_ids[ $cv ] ) ) {
 					$action[ $ck ] = $frm_duplicate_ids[ $cv ];
 				} elseif ( $ck == $subkey ) {
+					$pos = strpos( $action[ $ck ], '-' );
+
+					if ( $pos !== false ) {
+						$pieces = explode( '-', $action[ $ck ] );
+						$new_field_ids = array_map( __CLASS__ . '::trim_and_switch_field_ids', $pieces );
```
Crabcyborg, line 321: Do we actually need `trim_and_switch_field_ids` or does FrmFieldsHelper::switch_field_ids actually cover this?

Crabcyborg:
@jairo I think this all looks good to me! I left a comment in the autoresponder branch about some code that changed by accident, but everything else seems good to me!

jairoprez:
@stephywells I think this PR together with this other one https://github.com/Strategy11/formidable-autoresponder/pull/17 that is part of the same solution, are ready to be merged if there is no objection

classes/models/FrmFormAction.php
```diff
@@ -313,6 +314,15 @@ private function duplicate_array_walk( $action, $subkey, $val ) {
 				} elseif ( $ck == $subkey && isset( $frm_duplicate_ids[ $cv ] ) ) {
 					$action[ $ck ] = $frm_duplicate_ids[ $cv ];
 				} elseif ( $ck == $subkey ) {
+					$pos = strpos( $action[ $ck ], '-' );
+
+					if ( $pos !== false ) {
```
stephywells, line 319: @jairoprez It's possible this will affect existing code that isn't intended. I think we could use some unit tests here to make sure all the possible options are covered.

stephywells: Had another thought with this.... What if we create a new protected function that is called here with just this:
$action[ $ck ] = FrmFieldsHelper::switch_field_ids( $action[ $ck ] )?

Then we can add this new code in the extended class in the Autoresponder plugin. This way, we don't have to worry about issues it will create elsewhere.

jairoprez: I was seeing this more deeply https://github.com/Strategy11/formidable-forms/pull/311#discussion_r536819797 and before it made sense to me but then I was a little confused with what should be achieved there

```
$pos = strpos( $action[ $ck ], '-' );
if ( $pos !== false ) {
	$pieces = explode( '-', $action[ $ck ] );
	$new_field_ids = array_map( __CLASS__ . '::trim_and_switch_field_ids', $pieces );
	$action[ $ck ] = implode( '-', $new_field_ids );
	continue;
}
```

This code snippet that I added was to detect when the autoresponder `send_date` value comes with 2 field ids separated by a dash instead of just one field id, what happens when user choose Date + Time fields:
Screen Shot 2020-12-07 at 6 05 57 PM

The `send_date` field comes in this format 215-220 and the function FrmFieldsHelper::switch_field_ids( $action[ $ck ] ) does not support this format, it only accepts one id at a time

I don't know if the best thing would be to modify the function `FrmFieldsHelper::switch_field_ids` to support this format `123-456`

jairoprez: @stephywells Thanks for taking me to the right path at slack checkin today, with your suggestion I was able to move the logic to the autoresponder add-on, and things are simpler and more maintainable now.

classes/models/FrmFormAction.php
```diff
@@ -129,6 +129,11 @@ public function FrmFormAction( $id_base, $name, $action_options = array(), $cont
 		self::__construct( $id_base, $name, $action_options, $control_options );
 	}
 
+	public function maybe_switch_field_ids( $action ) {
+		$action = apply_filters( 'frm_maybe_switch_field_ids', $action );
+		return FrmFieldsHelper::switch_field_ids( $action );
```
stephywells, line 134: Will this cause any problems if this is run after the code in the autoresponder plugin? What would happen if the new form and the old form included some of the same field ids?

jairoprez: It would not break the duplication process of a form but the selected fields in the autoresponder settings would not be updated when the form is duplicated, maybe using `has_filter` function could help to avoid any php warnings or errors:
```
if ( has_filter( 'frm_maybe_switch_field_ids' ) ) {
    add_filter( 'frm_maybe_switch_field_ids', 'callback_fuction' );
}
```    

This function `FrmFieldsHelper::switch_field_ids` seems to handle duplicate ids quite well so there should be no problems if the new form and the old form included some of the same field ids https://github.com/Strategy11/formidable-forms/blob/master/classes/helpers/FrmFieldsHelper.php#L1268

stephywells: I'm wondering if it needs to be more like this?

```
$updated_action = apply_filters( 'frm_maybe_switch_field_ids', $action );
if ( $updated_action === $action ) {
$updated_action = FrmFieldsHelper::switch_field_ids( $action );
}
return $updated_action;
```

jairoprez: I'm going to play around with this approach you suggest - Thank you!

jairoprez:
@stephywells some new phpcs issues have been generated in some files that have not been edited in this PR, I see that it is related to a function that has been deprecated https://github.com/Strategy11/formidable-forms/pull/311/checks?check_run_id=1570300124#step:3:293

This could be solved in this PR or would it be better to create another one to specifically fix that case, what do you think about that?

stephywells (approved):
These changes look great now, and minimize the possibility of issues.

I can't merge since the phpcs is failing. We'll need a separate PR for that. Once it's merged I'll be able to merge this. Thanks!

jairoprez:
@stephywells here is the new PR with the phpcs fix: https://github.com/Strategy11/formidable-forms/pull/322 - please take a look when you get the chance.

stephywells:
@jairoprez will you post another note here once that pr is merged?

jairoprez:
@stephywells I don't think I'll make any more posts after the merge.

stephywells (approved):
Looks good now. Thanks @jairoprez !

PR #312, opened 2020-11-30
hide formidable button on form pages fix

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2780

I believe this is a good fix. I don't think we ever want to show the formidable buttons on actual formidable forms. I've moved the permission check out into another function and extended the logic. There are no changes in the original function outside of just outdenting all of the logic.

This fixes a few things. The ajax action that loads new repeater rows was always including the formidable button, even on front end forms. Front end forms were not showing the button by default, but preview forms were.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -613,14 +613,28 @@ class FrmFormsController {
 	 * @since 2.0.15
 	 */
 	public static function insert_form_button() {
-		if ( current_user_can( 'frm_view_forms' ) ) {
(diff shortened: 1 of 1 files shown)


stephywells:
@Crabcyborg would an easier and more reliable fix be to use remove_action right before a form is loaded?
remove_action( 'media_buttons', 'FrmFormsController::insert_form_button' );

Crabcyborg:
@stephywells good point, that seems to work https://github.com/Strategy11/formidable-forms/pull/313

PR #313, opened 2020-11-30
remove formidable button filter before loading a form

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2780

I'm not sure if there's a better place for this. It seems the new form / edit form stuff come through fairly differently and I wasn't finding a better place where they share a function.


stephywells (approved):
This looks good to me. It's possible it would work to include it in the file field model right before showing tiny mce. But this looks like a fine way of doing it too. Thank you!

PR #315, opened 2020-12-07
Jquery deprecation updates

Crabcyborg:
This is a pretty good first go of the jquery deprecation warnings in the free plugin. I think some other shorthand stuff might be missing, but I looked over all of the previous deprecations and I think this covers mostly everything in our code (excluding some third party stuff we're using).

- Changes `jQuery( element ).submit()` to just a new triggerSubmit function (vanilla solution).
- Added a new function to replace https://api.jquery.com/jQuery.isNumeric/ as it is deprecated since 3.3
- Changes `.submit` listeners which are now deprecated for `.on( 'submit'` that isn't (which is still jQuery), because `jQuery.fn.submit() event shorthand is deprecated`
- Changes jQuery.trim() for vanilla string.trim() as it is deprecated since 3.5 https://api.jquery.com/jQuery.trim/
- As of 3.4 pseudo selectors are all deprecated https://api.jquery.com/category/deprecated/deprecated-3.4/ including :lt, :first, and :last
- loadFields was using the :lt selector, and isFunction (which is deprecated as of 3.3 https://api.jquery.com/jquery.isfunction/), and had much more jQuery dependency than that, and I've removed all of the jQuery from the initial logic 🎉 
- As of 3.0 https://api.jquery.com/jquery.parsejson/ has been deprecated (replaced for vanilla JSON.parse)
- jQuery.fn.change() event shorthand is deprecated (replaced with .trigger( 'change' ), still jQuery for now)
- https://api.jquery.com/bind/ is deprecated as of 3.0 (replaced with .on, still jQuery for now)
- unbind is deprecated as of 3.0 (replaced with .off, still jQuery for now)
- shorthand mousedown is deprecated (replaced with .on( 'mousedown', still jQuery for now)
- shorthand click is deprecated (replaced with .trigger( 'click', and .on( 'click', still jQuery for now)
- https://api.jquery.com/size/ has been removed as of jQuery 3.0 (replaced with jQuery .length)
- andSelf is removed from jQuery 3.0 (https://api.jquery.com/category/removed/) so I replaced a call with https://api.jquery.com/add/ because I know we had a workaround for supporting andSelf/addBack, and I think this fix is less hacky.
- Replace jQuery .click/.trigger( 'click' ) with vanilla js .click where the change is simple enough.

More about the shorthand deprecations can be found here https://github.com/jquery/jquery/issues/3214 besides that I wasn't finding a lot of resources, but the warnings are all over the place.

I also made a few adjustments to move var declarations to the start of their function scopes to help reduce our number of warnings.

In an effort to reduce jQuery I made new functions triggerSubmit and triggerChange. triggerChange didn't look beautiful without jQuery (and several cases were sending jQuery objects as well), plus it's hard to test, so I'm keeping it dependent on jQuery for now, but it's better to have everything referencing one place that depends on jQuery than it is to have everything.

However, triggerSubmit seems to have a nice workable solution of inserting a hidden button and clicking it, so I've replaced our .submit() shorthands for vanilla js :)


js/formidable_admin_global.js
```diff
@@ -8,12 +8,12 @@ jQuery( document ).ready( function() {
     var deauthLink,
 		installLink = document.getElementById( 'frm_install_link' );
     if ( installLink !== null ) {
-        jQuery( installLink ).click( frmInstallPro );
+        jQuery( installLink ).on( 'click', frmInstallPro );
```
jairoprez, line 11: I think this could perhaps be more readable and consistent this way:
```
installLink = jQuery( '.frm_install_link' );
if ( installLink.length ) {
        installLink.on( 'click', frmInstallPro );
```

Crabcyborg: @jairoprez thanks for looking at this!

I think that'd be a step in the wrong direction as it would introduce more jQuery when we ideally want less. A call to `document.getElementById( 'frm_install_link' )` is faster than `installLink = jQuery( '#frm_install_link' )` would be, and if we had another good reliable way to check click without jQuery, we would be replacing this line as well.

jairoprez: You're right, it makes sense to me.

jairoprez (approved):
These changes look good to me!

js/formidable_admin.js
```diff
@@ -6795,10 +6813,8 @@ function frmAdminBuildJS() {
 			jQuery( document ).on( 'change', '#protect_files', function() {
 				if ( this.checked ) {
 					jQuery( '.hide_protect_files' ).fadeIn( 'slow' );
-					jQuery( '#edit_action' ).change();
```
Crabcyborg, line 6798: I just want to point out that I am intentionally just removing this line and a line shortly underneath it. I think I copied too much here when I added the protect files logic.

js/formidable_admin.js
```diff
@@ -6770,21 +6788,21 @@ function frmAdminBuildJS() {
...
-			$saveDraft.change( function() {
+			$saveDraft.on( 'change', function() {
 				if ( this.checked ) {
 					jQuery( '.hide_save_draft' ).fadeIn( 'slow' );
 				} else {
 					jQuery( '.hide_save_draft' ).fadeOut( 'slow' );
 				}
 			});
-			$saveDraft.change();
+			triggerChange( $saveDraft );
```
stephywells, line 6798: is this a problem to send a jquery object to this function?

stephywells (approved):
This looks good. Just making some notes for a few things we need to make sure get double tested:
- load fields in builder with ajax with form with more than 20 fields
- add two logic rows in a field, remove the first one, and add another. Do they both save correctly? Test the same process with watch lookup fields, form actions, form action logic, submit button logic, create post category & meta, order rows in a view, where rows in a view
- add a page break and collapse page 1 in builder
- save builder page when ajax save fails, and with ajax builder save turned off

Side note:
looks like some of this js should be copied into views and deprecated here (starting at line 4445)

js/formidable_admin.js
Crabcyborg: Yeah, for now. I left triggerChange jQuery dependent. If you wrap it twice, like jQuery( $saveDraft ), you still end up with the same thing as $saveDraft. In the long run I'd love to examine all of these closer but the scope of this update already feels pretty large.

stephywells: Good plan. Thank you!

PR #320, opened 2020-12-10
dynamic field in repeater export fix

Crabcyborg:
Trying to fix https://github.com/Strategy11/formidable-pro/issues/2805

This is the problem area, but it's not actually the fix I want.

I'm noticing several issues:
- I've tried exporting 2 rows with one dynamic radio option selected in each and both columns included both values.
- I tried exporting a dynamic checkbox and the two selected values in my 1 row were left as ids.
- When I exported for a set of data that sometimes included 2 columns, the 2nd column was showing the first column's result even where there was only one repeater entry

The issue has to be higher up. It's coming through wrong. Probably similar to the fix for file uploads.


Crabcyborg:
Think I've got a fix for this, but it's going into pro. Closing this PR 🙂

PR #321, opened 2020-12-15
fix select padding issue in WordPress 5.6

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2814

I listened to the TODO comment and removed the `.branch` prefixes that were avoiding the new `branch-5-6`

This will possibly mean some funny padding in WordPress < 5.3 but I imagine we're okay with that.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -7330,19 +7330,9 @@ button.frm_choose_image_box,
 	width: 100%;
 }
 
-/** TODO: After 5.5, remove the branch clases **/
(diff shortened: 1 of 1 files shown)


jairoprez (approved):
@Crabcyborg I did tests with your fix https://recordit.co/1pO1WhS3hT and the texts in the dropdowns does not cut anymore in WP version 5.6, it looks great.

stephywells (approved):
Thanks! Looks good to me.

PR #322, opened 2020-12-18
Add conditional call for libxml_disable_entity_loader function in versions prior to PHP 8

jairoprez:
In PHP 8.0 and later, PHP uses libxml versions from 2.9.0, which disabled XML EXternal Entities by default. 

`libxml_disable_entity_loader()` function has been deprecated in php 8.0 https://www.php.net/manual/en/migration80.deprecated.php and this is already causing issues in the phpcs
Screen Shot 2020-12-17 at 8 18 07 PM

On code that only runs on PHP 8.0 and later, it is now safe to remove all calls to the `libxml_disable_entity_loader()` function. For versions prior to PHP 8, a conditional call might be the best approach and that's what I did.

### DIFF
--- a/classes/controllers/FrmXMLController.php
+++ b/classes/controllers/FrmXMLController.php
@@ -10,7 +10,7 @@ class FrmXMLController {
 	}
 
 	public static function add_default_templates() {
-		if ( ! FrmXMLHelper::check_if_libxml_disable_entity_loader_exists() ) {
+		if ( FrmXMLHelper::check_if_libxml_disable_entity_loader_exists() ) {
 			// XML import is not enabled on your server
 			return;
 		}
@@ -330,7 +330,7 @@ class FrmXMLController {
 		}
 		unset( $file_type );
 
-		if ( ! FrmXMLHelper::check_if_libxml_disable_entity_loader_exists() ) {
+		if ( FrmXMLHelper::check_if_libxml_disable_entity_loader_exists() ) {
 			$errors[] = __( 'XML import is not enabled on your server with the libxml_disable_entity_loader function.', 'formidable' );
 			self::form( $errors );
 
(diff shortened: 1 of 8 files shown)


Crabcyborg (approved):
This seems good to me!

classes/controllers/FrmXMLController.php
```diff
@@ -10,13 +10,13 @@ public static function menu() {
 	}
 
 	public static function add_default_templates() {
-		if ( ! function_exists( 'libxml_disable_entity_loader' ) ) {
+		if ( ! FrmXMLHelper::check_if_libxml_disable_entity_loader_exists() ) {
```
stephywells, line 13: @jairoprez Is this tested? It looks like the will completely prevent importing in php 8+

jairoprez: @stephywells You were right, that logic is wrong, I have made some adjustments and tests and now it works as expected.

Thanks!

stephywells: Thanks @jairoprez ! Can you also add php 8.0 to the travis config so the unit tests can catch this too?

stephywells: @jairoprez this still isn't setup quite right. The 8.0 tests aren't running.

jairoprez: @stephywells you're right, I'm checking travis config file to fix it.

jairoprez: @stephywells The scope of including php 8.0 in the travis process is a bit more complex than I thought, there have been some bugs and deprecations alerts, so I have created a new PR https://github.com/Strategy11/formidable-forms/pull/323 to work with those cases and then merge it back with this PR.

PR #323, opened 2020-12-24
Fix some php8.0 deprecations and bugs

jairoprez:
Fix the depreciations and errors of php 8.0 in this PR https://github.com/Strategy11/formidable-forms/pull/322

PHP 8.0: Deprecate required parameters after optional parameters in function/method signatures so based on that i applied some tweaks in these functions:
- `FrmAppHelper::get_unique_key`
- `FrmEntryMeta::add_entry_meta`
- `FrmEntryMeta::update_entry_meta`
- `FrmEntryMeta::search_entry_metas`

Some array to string conversion notice when using `str_replace` were also fixed.

### DIFF
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -1675,12 +1675,12 @@ class FrmXMLHelper {
 	 *
 	 * @return boolean
 	 */
-	public static function maybe_libxml_disable_entity_loader( $disable ) {
+	public static function maybe_libxml_disable_entity_loader( $loader ) {
 		if ( version_compare( phpversion(), '8.0', '<' ) && function_exists( 'libxml_disable_entity_loader' ) ) {
-			$disable = libxml_disable_entity_loader( $disable ); // phpcs:disable Generic.PHP.DeprecatedFunctions.Deprecated
+			$loader = libxml_disable_entity_loader( $loader ); // phpcs:disable Generic.PHP.DeprecatedFunctions.Deprecated
 		}
 
-		return $disable;
+		return $loader;
 	}
 
 	public static function check_if_libxml_disable_entity_loader_exists() {
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -1677,7 +1677,7 @@ class FrmXMLHelper {
 	 */
 	public static function maybe_libxml_disable_entity_loader( $disable ) {
 		if ( version_compare( phpversion(), '8.0', '<' ) ) {
-			$disable = libxml_disable_entity_loader( $disable );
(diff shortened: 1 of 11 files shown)


jairoprez:
There is just one problem I couldn't solve: https://travis-ci.org/github/Strategy11/formidable-forms/jobs/751983816#L459

I was investigating and the cause is as follows: 
- PHP 8 introduces a new control structure called `match`, which makes `match` a reserved keyword in PHP 8.
- One of the dependencies of PHPUnit declares a class named `Match`, which triggers a fatal error before PHPUnit can even start to run.
- The dependency in PHPUnit has been updated to get rid of the Match classname and has released a new version.

There are more details here: https://core.trac.wordpress.org/ticket/50902 - I actually tried the solution they suggest there but it didn't give me the expected result.

Installing PHPUnit 8.x/9.x would be the ideal solution but currently WordPress only supports PHPUnit 7.X, support for PHPUnit 8.x is still under development https://core.trac.wordpress.org/ticket/46149

For now I have placed php 8 version under `allow_failures` tag in travis config file until a better solution is found.

@stephywells @Crabcyborg what thoughts do you have on this? maybe some hint of something that I missed?

stephywells:
@jairoprez you fixed the php nightly test! Looks like that is running 8.0 right now. 
https://travis-ci.org/github/Strategy11/formidable-forms/builds/752005842

does it make a difference to remove "xenial".

jairoprez:
@stephywells you're right, nightly is running php version 8.0 and I hadn't noticed, that php 8.0 version was created on Jan 7, 2020 https://travis-ci.org/github/Strategy11/formidable-forms/jobs/752113323#L441, although it is not the latest, I think it would be fine for now.

Xenial dist has been removed as it is not required.

stephywells:
Thanks @jairoprez ! Let's add 8.0 back in and not allow failures. Since nightly will change, we can't depend on it going forward.

classes/helpers/FrmXMLHelper.php
```diff
@@ -1667,5 +1667,24 @@ private static function migrate_autoresponder_to_action( $form_options, $form_id
...
 	}
+
+	/**
+	 * PHP 8 backward compatibility for the libxml_disable_entity_loader function
+	 *
+	 * @param  boolean $disable
+	 *
+	 * @return boolean
+	 */
+	public static function maybe_libxml_disable_entity_loader( $disable ) {
```
stephywells, line 1678: I'm not sure that $disable makes sense to me. Maybe $loader would make more sense?

Is boolean correct for the return value? That doesn't looks correct to me.

Has this been tested manually in php 8 too?

stephywells (approved):
Thanks for these fixes. @Crabcyborg have you had a chance to test this?

Crabcyborg:
@stephywells I just tried importing an XML file in Formidable using PHP 8 and with this update, it still appears to work, while also removing the deprecation messages.

PR #326, opened 2021-01-05
Add condition to avoid js warnings for a null property

jairoprez:
This PR should do the trick to fix the JS error reported by @altoin 
download - 2021-01-05T105636 519

https://sandbox.formidableforms.com/toyin/wp-admin/admin.php?page=formidable&frm_action=edit&id=936

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1054,8 +1054,9 @@ function frmAdminBuildJS() {
 
 		addHtmlToField = function( element ) {
 			element.classList.add( 'frm_load_now' );
-			if ( element.querySelector( '.frm_hidden_fdata' ) !== null ) {
-			    field.push( element.querySelector( '.frm_hidden_fdata' ).innerHTML );
+			var frmHiddenFdata = element.querySelector( '.frm_hidden_fdata' );
+			if ( frmHiddenFdata !== null ) {
+			    field.push( frmHiddenFdata.innerHTML );
 			}   
 		};
 
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1053,8 +1053,8 @@ function frmAdminBuildJS() {
 			field = [];
 
 		addHtmlToField = function( element ) {
-			element.classList.add( 'frm_load_now' );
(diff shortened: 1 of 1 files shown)


Crabcyborg (approved):
@jairo this should do the trick. Usually I'd avoid the repeat logic of getting the element twice like this, but I don't imagine that's too life or death here.

jairoprez:
> @Jairo this should do the trick. Usually I'd avoid the repeat logic of getting the element twice like this, but I don't imagine that's too life or death here.

Thanks @Crabcyborg - Anyway, I have applied your suggestion :zany_face:.

Crabcyborg:
@jairoprez ESLint seems a little unhappy now.

I'm seeing:

```
1059:2   error    Mixed spaces and tabs                                            no-mixed-spaces-and-tabs
1060:5   error    Trailing spaces not allowed                                      no-trailing-spaces
```

There's also the infamous `warning  All 'var' declarations must be at the top of the function scope  vars-on-top` that we break all over. I try not to introduce more of these either.

jairoprez:
> @jairoprez ESLint seems a little unhappy now.
> 
> I'm seeing:
> 
> ```
> 1059:2   error    Mixed spaces and tabs                                            no-mixed-spaces-and-tabs
> 1060:5   error    Trailing spaces not allowed                                      no-trailing-spaces
> ```

@Crabcyborg ESLint seems happy now. Thanks!

jairoprez:
@stephywells The warnings for `All 'var' declarations must be at the top of the function scope vars-on-top` have been fixed.

stephywells:
Thanks @jairoprez! Looks like this will need a bunch more testing before we can merge. In the meantime, can you submit a new pr with just the change you had before? But this time, be sure to keep the var at the top of the function. We don't need to fix all of it right now, just to the parts we make changes to.

jairoprez:
@stephywells Oh sorry, I thought I should fix them all, I have reverted the unnecessary changes in this PR.

stephywells (approved):
Looks great now! Thanks for picking this up!

PR #328, opened 2021-01-07
force jquery ui 1.11.4

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2811 since the latest jQuery ui doesn't seem to include i18n files.

I think WordPress 5.6 uses jQuery 1.12.1 by default.

Going forward we don't want everyone to need to add a workaround.

I tested this and date localizations don't work without this fix in the new WordPress.

The alternative might be trying to support 1.12.1 with the old i18n files, but I feel like we're at risk of a conflict with that. I think the best plan moving forward would be replacing datepicker with something that isn't jQuery based, possibly https://flatpickr.js.org/

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1500,6 +1500,10 @@ class FrmAppHelper {
 			$ver = $query->ver;
 		}
 
+		if ( version_compare( $ver, '1.12.1', '>=' ) ) {
+			$ver = '1.11.4';
+		}
+
 		return $ver;
 	}
 


jairoprez (approved):
This solution looks good and the possibility of replacing datepicker in the future with flatpickr or some other non-jQuery based alternative makes sense to me as a less conflicting solution.

stephywells:
The problem with this approach is that the jQuery UI script is loaded by WordPress. Forcing the version here changes the files loaded for the selected datepicker theme, although these files still exist. I'm concerned that this will cause problems going forward.

I wonder if we can find that i18n file hosted somewhere else?

Crabcyborg:
Thanks @stephywells I think I have a better solution [here](https://github.com/Strategy11/formidable-pro/pull/2836)

PR #329, opened 2021-01-12
Add welcome screen after plugin activation

jairoprez:
This page will show when the plugin is activated for the first time.
Screen Shot 2021-01-12 at 12 01 23 AM


formidable.php
```diff
@@ -26,6 +26,11 @@
 	die( 'You are not allowed to call this page directly.' );
 }
 
+/**
+ * The code that runs during plugin activation.
+ */
+register_activation_hook( __FILE__, array( 'FrmActivatorController', 'activate' ) );
```
stephywells, line 32: I think it would be a lot cleaner to do it this way instead of a whole new class.
https://github.com/Strategy11/formidable-directory/blob/master/formidable-directory.php#L77

Let's also do it the same way:
set_transient( 'frm_activation_redirect', 'formidable', 60 );

classes/controllers/FrmHooksController.php
```diff
@@ -149,6 +149,11 @@ public static function load_admin_hooks() {
 
 		add_action( 'admin_init', 'FrmUsageController::schedule_send' );
 
+		// Welcome Screen
+		add_action( 'admin_init', 'FrmWelcomeScreenController::activation_redirect' );
+		add_action( 'admin_menu', 'FrmWelcomeScreenController::screen_page' );
+		add_action( 'admin_head', 'FrmWelcomeScreenController::remove_menu' );
```
stephywells, line 155: Maybe this can be replaced with FrmWelcomeScreen::load_hooks() like like 157 below? This will keep more of the code in one place.

classes/controllers/FrmWelcomeScreenController.php
```diff
@@ -0,0 +1,30 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+
+class FrmWelcomeScreenController {
+	public static $menu_slug   = 'formidable-welcome-screen';
+	public static $option_name = 'frm_welcome_screen_activation_redirect';
+
+	public static function activation_redirect() {
```
stephywells, line 10: Please see the FrmSolution::redirect function.

stephywells (changes requested):
Thanks for working on this @jairoprez. It looks like most isn't committed yet, but I have a few requests for cleanup. Sorry if this is premature!

PR #333, opened 2021-01-18
Add welcome screen after plugin activation

jairoprez:
**Note:** this feature was being worked on in this PR https://github.com/Strategy11/formidable-forms/pull/329, but due to some git issues that couldn't be solved, the git history got corrupted and I had to create this new one.

This page will show when the plugin is activated for the first time.
104268035-5a680d80-5469-11eb-8777-03c8df74a3a7


classes/controllers/FrmWelcomeScreenController.php
```diff
@@ -0,0 +1,114 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+
+class FrmWelcomeScreenController {
+	public static $menu_slug   = 'formidable-welcome-screen';
```
Crabcyborg, line 7: When possible, it's good to default to `private`, especially for variables.

Crabcyborg:
Hey @jairoprez I get this isn't finished, I just wanted to see how it was going, and I noticed a couple of things you might not have on your list yet :)

One thing I noticed is that my fonts all have serifs.
Screen Shot 2021-01-19 at 4 05 45 PM

It looks like you're using `font-family: Open Sans;` but there's no guarantee that that font exists.

images/briefcase.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Crabcyborg, line 1: We have an `images/icons.svg`. For any small SVGs, it might make more sense to add them to that, to limit the number of requests being made.

js/formidable_admin.js
```diff
@@ -74,6 +74,10 @@ var FrmFormsConnect = window.FrmFormsConnect || ( function( document, window, $
 					$icons.hide();
 				}
 			});
+
+			jQuery( document ).on( 'click', '.frm-we-wrapper [data-location]', function() {
+				window.location = $( this ).data( 'location' );
```
Crabcyborg, line 79: I don't the `$` for jQuery woulld break here. You can probably access this without jQuery either way using something like `this.dataset.location`.

jairoprez:
@Crabcyborg Thank you so much for taking a look and leaving some comments, it is very helpful as always.

js/formidable_admin.js
jairoprez: You're right, sometimes I forget that we are trying to use jQuery as little as possible and this can be solved perfectly with vanilla JS. Thanks!

jairoprez:
@stephywells @Crabcyborg I think I completed the requirements of the welcome screen feature. Please take a look when you get a chance. Thanks!

css/welcome_screen.css
```diff
@@ -0,0 +1,382 @@
+@import url(https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300;0,400;0,600;0,700;0,800;1,300;1,400;1,600;1,700;1,800&display=swap);
```
stephywells, line 1: Great job paying attention to the details! This looks great!

Although Yaro designs with Open sans, we won't actually over ride the font in the wp admin. Let's take out all open sans references.

css/welcome_screen.css
```diff
@@ -0,0 +1,382 @@
...
+    font-style: normal;
+    font-weight: 600;
+    font-size: 30px;
+    line-height: 41px;
+
+    color: #282F36;
+}
+
+.frm-ws-block-1 p {
+    width: 538px;
```
stephywells, line 63: This looks perfect on the page. But let's slim down the unneeded css now.

width, height, font-family, font-style, font-weight can all go here. Will you review the whole file for css that isn't needed? Width and height can also cause problems with text when it's translated or for responsiveness.

css/welcome_screen.css
```diff
@@ -0,0 +1,382 @@
...
+
+body {
+    background: #F6F7FB;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+.frm-ws-wrapper .column {
```
stephywells, line 11: We can minimize conflicts by using the frm prefix on every class.

formidable.php
```diff
@@ -95,3 +95,8 @@ function frm_class_autoloader( $class_name, $filepath ) {
 		require( $filepath );
 	}
 }
+
+add_action( 'activate_formidable/formidable.php', 'frm_maybe_install' );
+function frm_maybe_install() {
+	set_transient( 'frm_activation_redirect', 'formidable-welcome-screen', 60 );
```
stephywells, line 101: Perhaps we should save a flag to know if they have already seen this page? Right now I'm taken to the welcome page after every activation.

images/icons.svg
```diff
@@ -932,5 +932,13 @@
...
...96 19.4531 18.9844 19.7679 18.9844 20.1562C18.9844 20.5446 18.6696 20.8594 18.2812 20.8594H16.875C16.4867 20.8594 16.1719 20.5446 16.1719 20.1562Z" fill="#3F4B5B"/><path fill-rule="evenodd" clip-rule="evenodd" d="M16.1719 23.9062C16.1719 23.5179 16.4867 23.2031 16.875 23.2031H18.2812C18.6696 23.2031 18.9844 23.5179 18.9844 23.9062C18.9844 24.2946 18.6696 24.6094 18.2812 24.6094H16.875C16.4867 24.6094 16.1719 24.2946 16.1719 23.9062Z" fill="#3F4B5B"/><path fill-rule="evenodd" clip-rule="evenodd" d="M19.9219 20.1562C19.9219 19.7679 20.2367 19.4531 20.625 19.4531H21.5625C21.9508 19.4531 22.2656 19.7679 22.2656 20.1562C22.2656 20.5446 21.9508 20.8594 21.5625 20.8594H20.625C20.2367 20.8594 19.9219 20.5446 19.9219 20.1562Z" fill="#7388A6"/><path fill-rule="evenodd" clip-rule="evenodd" d="M19.9219 23.9062C19.9219 23.5179 20.2367 23.2031 20.625 23.2031H21.5625C21.9508 23.2031 22.2656 23.5179 22.2656 23.9062C22.2656 24.2946 21.9508 24.6094 21.5625 24.6094H20.625C20.2367 24.6094 19.9219 24.2946 19.9219 23.9062Z" fill="#7388A6"/><path fill-rule="evenodd" clip-rule="evenodd" d="M20.625 6.09375H9.84375V9.375H20.625V6.09375ZM8.4375 4.6875V10.7812H22.0312V4.6875H8.4375Z" fill="#7388A6"/></symbol>
```
stephywells, line 935: There's this cool tool I like to use to compress svgs to optimize them. You can run the svg code through this site. Since this icons file is loaded everywhere, any bigger svgs should be either a separate file or used inline if it's not a commonly used graphic.
https://jakearchibald.github.io/svgomg/

js/formidable_admin.js
```diff
@@ -5678,6 +5682,12 @@ function frmAdminBuildJS() {
 			transitionToAddDetails( $modal, name, link, action );
 		});
 
+		jQuery( document ).on( 'click', '.frm-ws-wrapper .frm-create-form', function( event ) {
```
stephywells, line 5685: I'm not able to close the modals by clicking outside of the box like I can on other pages.

classes/controllers/FrmWelcomeScreenController.php
```diff
@@ -0,0 +1,114 @@
...
+
+		// Check if we should consider redirection.
+		if ( ! self::is_welcome_screen() ) {
+			return;
+		}
+
+		delete_transient( self::$option_name );
+
+		// Initial install.
+		wp_safe_redirect( self::settings_link() );
```
stephywells, line 53: The needs to be escaped. admin_url isn't safe.

classes/views/welcome/screen.php
```diff
@@ -0,0 +1,84 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+?>
+<div class="frm-ws-wrapper">
+	<div class="row">
```
stephywells, line 7: It would prevent some duplication by using the css grid classes that are already built in. frm_grid_container would replace row. This would also make the page responsive.

classes/views/welcome/screen.php
```diff
@@ -0,0 +1,84 @@
...
+		<div class="column">
+			<div class="frm-ws-block-1">
+				<div class="frm-ws-logo"><img src="<?php echo esc_url( FrmAppHelper::plugin_url() . '/images/logo.svg' ); ?>" alt="Logo"></div>
+				<h1><?php esc_html_e( 'Welcome to Formidable Forms!', 'formidable' ); ?></h1>
+				<p><?php esc_html_e( 'Thanks for choosing Formidable Froms - The most powerful and vesatile form builder for Wordpress', 'formidable' ); ?></p>
+				<div class="frm-ws-buttons">
+					<div class="frm-ws-btn bg-blue" data-location="https://formidableforms.com/lite-upgrade/?utm_source=WordPress&utm_medium=settings-license&utm_campaign=liteplugin">
```
stephywells, line 14: These links need to be dynamic.
FrmAppHelper::admin_upgrade_link

stephywells (changes requested):
Great job Jairo! You did a great job getting this pixel perfect. Now we need to simplify, make it responsive, and more dynamic.

jairoprez:
@stephywells Thanks so much for taking a look and leave your comments :slightly_smiling_face:, I'll be checking and working on each change.

css/welcome_screen.css
```diff
@@ -0,0 +1,277 @@
...
+    font-size: 14px;
+    line-height: 19px;
+    text-align: center;
+    color: #FFFFFF;
+    flex: none;
+    order: 0;
+    flex-grow: 0;
+    margin: 0px 10px;
+    width: 175px;
+    border-radius: nullpx;
```
Crabcyborg, line 94: Is this a typo? I don't believe `nullpx` is a thing?

css/welcome_screen.css
```diff
@@ -0,0 +1,277 @@
...
+    margin-left: 16px;
+}
+
+.frm-ws-block-1 .frm-bg-transparent span {
+    font-weight: 600;
+    font-size: 14px;
+    line-height: 19px;
+    text-align: center;
+    color: #4199FD;
+    border-radius: nullpx;
```
Crabcyborg, line 107: Looks like there are a couple of these.

jairoprez:
I think the welcome screen is now well covered in terms of responsiveness https://recordit.co/vjY2IlMJGg

classes/views/welcome/screen.php
```diff
@@ -0,0 +1,84 @@
...
+	</div>
+
+	<div class="second-section">
+		<div class="row">
+			<div class="column">
+				<div class="frm-ws-block frm-create-form">
+					<div class="icon"><?php FrmAppHelper::icon_by_class( 'frmfont frm_ws_plus_icon' ); ?></div>
+					<h3><?php esc_html_e( 'New Blank Form', 'formidable' ); ?></h3>
+					<p><?php esc_html_e( 'Create a new view from scratch', 'formidable' ); ?></p>
+					<div class="right-arrow"><?php FrmAppHelper::icon_by_class( 'frmfont frm_ws_arrow_right_icon' ); ?></div>
```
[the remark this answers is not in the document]
stephywells: Thanks for switching these out. Now they need some css to size them down to match the design. Also note that the color of the arrows is different when it's on the light colored background.

classes/views/welcome/screen.php
```diff
@@ -0,0 +1,84 @@
...
+		<div class="column">
+			<div class="frm-ws-block-1">
+				<div class="frm-ws-logo"><img src="<?php echo esc_url( FrmAppHelper::plugin_url() . '/images/logo.svg' ); ?>" alt="Logo"></div>
+				<h1><?php esc_html_e( 'Welcome to Formidable Forms!', 'formidable' ); ?></h1>
+				<p><?php esc_html_e( 'Thanks for choosing Formidable Froms - The most powerful and vesatile form builder for Wordpress', 'formidable' ); ?></p>
+				<div class="frm-ws-buttons">
+					<div class="frm-ws-btn bg-blue" data-location="https://formidableforms.com/lite-upgrade/?utm_source=WordPress&utm_medium=settings-license&utm_campaign=liteplugin">
+						<span><?php esc_html_e( 'Activate Formidable Form', 'formidable' ); ?></span>
+					</div>
+					<div class="frm-ws-btn bg-transparent" data-location="https://formidableforms.com/api-connect?utm_source=WordPress&utm_medium=connect&utm_campaign=liteplugin&v=2&siteurl=http://strategy11.test&url=http://strategy11.test/wp-json/&token=6200a7febd03e9aa56f15a7a017a3f4eeeae14f5168591a2a60f54fb3c2acdd0466629313a31fd52fff645fc9351928e4ec9c2e3354d29526f65008991d7a4d9&l=0d8e0c8ed0d356a923c9631b437c7df2">
```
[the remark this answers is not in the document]
stephywells: Thanks! Much better! There's a css issue here with the button text not centered vertically.
Screen Shot 2021-01-29 at 3 38 58 PM

Also, the "Get Formidable Now" link should open in a new window.

js/formidable_admin.js
stephywells: This is still a problem.

css/welcome_screen.css
stephywells: Most of these are updated. But I'm still seeing at least one (.second-section)

css/welcome_screen.css
stephywells: Looks great!

formidable.php
```diff
@@ -95,3 +95,10 @@ function frm_class_autoloader( $class_name, $filepath ) {
 		require( $filepath );
 	}
 }
+
+add_action( 'activate_formidable/formidable.php', 'frm_maybe_install' );
```
stephywells, line 99: This should be dynamic in case the folder name is different.

classes/controllers/FrmWelcomeScreenController.php
stephywells: This isn't done.

formidable.php
stephywells: Was this addressed?

classes/views/welcome/screen.php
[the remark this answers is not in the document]
stephywells: Nevermind. I just made a commit for this.

classes/views/welcome/screen.php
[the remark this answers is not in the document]
stephywells: Actually, since those arrows on the bottom section don't do anything, they should be removed.

classes/views/welcome/screen.php
stephywells: This isn't done

stephywells (changes requested):
Jairo, I'm seeing way too many things here that haven't been done. Please go through all my previous comments and make sure they are addressed before requesting a follow up review.

Thanks!

classes/controllers/FrmWelcomeScreenController.php
stephywells: This isn't done.

stephywells:
Jairo, I've gotten most of this done. Please pull before you make any changes.

Thanks!

js/formidable_admin.js
```diff
@@ -74,6 +74,10 @@ var FrmFormsConnect = window.FrmFormsConnect || ( function( document, window, $
 					$icons.hide();
 				}
 			});
+
+			jQuery( document ).on( 'click', '.frm-ws-wrapper [data-location]', function() {
```
stephywells, line 78: This needs to be removed since a normal link is the best option.

images/arrow-left.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="8" height="14" viewBox="0 0 8 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
stephywells, line 1: These arrow files need to be deleted.

jairoprez:
> Jairo, I'm seeing way too many things here that haven't been done. Please go through all my previous comments and make sure they are addressed before requesting a follow up review.
> 
> Thanks!

I apologize for this, I will make sure to check all comments well to prevent this from happening again.

classes/views/welcome/screen.php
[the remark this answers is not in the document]
jairoprez: @stephywells Thanks for improving this area. I noticed that these buttons don't disappear now when formidable pro is installed https://recordit.co/pTRMBVzEF7, just to confirm, is that the expected behavior on the welcome screen?

css/welcome_screen.css
jairoprez: The `.second-section` class has been replaced and the `div` that used it has been removed

formidable.php
jairoprez: Yes!, I made changes here https://github.com/Strategy11/formidable-forms/pull/333/files#diff-1850107c5088b51e8b820e88a5c01f5893ed45d2b3f38d2f317b95305ac77dadR100 and here https://github.com/Strategy11/formidable-forms/pull/333/files#diff-2c797f60db09451ddf38cc84d215362f4538597d025f8b8b75899fee2af286beR50 and that way the redirect only runs during the first activation https://recordit.co/gJRhh1Ixvm

js/formidable_admin.js
jairoprez: I have not been able to replicate this issue from my side https://recordit.co/QHPZ2YSFhg - I had mentioned that in one of the past check-ins https://strategy11.slack.com/archives/G01KDLFJTJS/p1611766980041200 - Even Mike did a test and it worked well for him too https://strategy11.slack.com/archives/G01KDLFJTJS/p1611767948044100 - at this point, I have no idea what could be the problem that is causing it to not work well for you :pensive:

classes/controllers/FrmWelcomeScreenController.php
jairoprez: Since`self::settings_link()` is a function I directly escaped the url inside the function, so it returns an escaped url, you can validate it in this commit https://github.com/Strategy11/formidable-forms/pull/333/commits/d5ecd12f3b74d71b331477153bdf376e56b00a1e - Does this approach make sense to you?

classes/views/welcome/screen.php
jairoprez: I tried to use this class when you suggest it to me but the layout was breaking, so I tried to solve it with css rules from scratch, I should have left a comment about it to clarify it. Thanks anyway for having implemented it.

jairoprez:
@stephywells I have completed the pending changes, and leave comments on others you must see and let me know if it makes sense to you:
- https://github.com/Strategy11/formidable-forms/pull/333#discussion_r567872308
- https://github.com/Strategy11/formidable-forms/pull/333#discussion_r567889440
- https://github.com/Strategy11/formidable-forms/pull/333#discussion_r567916162

js/formidable_admin.js
jairoprez: Removed!

images/arrow-left.svg
jairoprez: Removed!

formidable.php
jairoprez: Done! 101559f25e4ea98f58284e22986192ea0952d030

classes/views/welcome/screen.php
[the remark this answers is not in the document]
jairoprez: > Thanks for switching these out. Now they need some css to size them down to match the design. Also note that the color of the arrows is different when it's on the light colored background.

Done! 47910b59b19a36e47ab94c80b2054a4f73264d38

classes/views/welcome/screen.php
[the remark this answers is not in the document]
stephywells: You're totally right. Those buttons shouldn't show if it's already connected. It should behave just the same as the buttons in the global settings. When connected, we can hide the "enjoy" message with css.

formidable.php
stephywells: That looks like it will only prevent redirection within 60 seconds, right?

classes/controllers/FrmWelcomeScreenController.php
stephywells: It would be best if escaping is done last whenever possible. That change will cause double escaping for the spots that are already using it.

formidable.php
jairoprez: That's right. If this should be permanent while the plugin is installed, I think it would be better to use the Options API instead of Transients, that way some things could be simplified, for example, this 
```
function frm_maybe_install() {
	if ( get_transient( 'frm_activation_redirect' ) !== 'no' ) {
		set_transient( 'frm_activation_redirect', 'formidable-welcome-screen', 60 );
	}
}
```
could work like this
```
function frm_maybe_install() {
	add_option( 'frm_activation_redirect', 'formidable-welcome-screen', '', 'no' );
}
```

classes/controllers/FrmWelcomeScreenController.php
jairoprez: Got it! - I have made the change here.https://github.com/Strategy11/formidable-forms/pull/333/commits/0b51de437f1df1093b2e55e5c4044b3e43687a4b

classes/views/welcome/screen.php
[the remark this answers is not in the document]
jairoprez: Okay, I'll check why it doesn't work properly.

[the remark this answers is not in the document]
jairoprez: @stephywells As I could see, the license box is tied to a custom action hook https://github.com/Strategy11/formidable-forms/blob/c99336c4d30cc25102df6c7ea531764d54045233/classes/views/frm-settings/general.php#L12 

and is called https://github.com/Strategy11/formidable-forms/blob/f2b13f4f1d64c5cfd0a3c92cb97cf0085483759c/classes/controllers/FrmHooksController.php#L135 or removed https://github.com/Strategy11/formidable-forms/blob/5eacd6ad01950673ebec3506335714666fdca1cb/classes/controllers/FrmAppController.php#L338 depending on whether formidable pro is active or not.

The solution I implemented was this https://github.com/Strategy11/formidable-forms/pull/333/commits/825162b123ca81df807f89f61667fa3d65905049

This line https://github.com/Strategy11/formidable-forms/pull/333/commits/825162b123ca81df807f89f61667fa3d65905049#diff-2c797f60db09451ddf38cc84d215362f4538597d025f8b8b75899fee2af286beR25 was added by mistake, I have already removed it in this other commit https://github.com/Strategy11/formidable-forms/pull/333/commits/b10a534ad34002cf96f4d97ab174bd37633af96b

Crabcyborg:
Hey @jairoprez

I'm looking over this update now. I'm seeing tiny little icons in the new form modal.

Screen Shot 2021-02-10 at 11 26 49 AM

classes/views/welcome/screen.php
```diff
@@ -0,0 +1,82 @@
...
+				<div class="frm-icon"><?php FrmAppHelper::icon_by_class( 'frmfont frm_paperclip_icon' ); ?></div>
+				<h3><?php esc_html_e( 'File uploads', 'formidable' ); ?></h3>
+				<p><?php esc_html_e( 'Easily upload documents, files, photos, and music for user avatars, featured images, or email attachments.', 'formidable' ); ?></p>
+				<?php FrmWelcomeScreenController::upgrade_to_pro_button(); ?>
+			</div>
+			<div class="frm3">
+				<div class="frm-icon"><?php FrmAppHelper::icon_by_class( 'frmfont frm_listings_icon' ); ?></div>
+				<h3><?php esc_html_e( 'Display form data with views', 'formidable' ); ?></h3>
+				<p><?php esc_html_e( 'Now you can display form data in custom Views without any PHP.', 'formidable' ); ?></p>
+				<?php FrmAddonsController::conditional_action_button( 'views', 'views-info' ); ?>
```
Crabcyborg, line 78: I'm thinking there might be a missing element to my function here.

When views is installed, I see the button.
Screen Shot 2021-02-10 at 11 29 14 AM

The place where I put this button would only ever get rendered when views wasn't installed, so it worked out there but not always here.

You can probably get around just adding a check similar to:

`$views_installed = is_callable( 'FrmProAppHelper::views_is_installed' ) ? FrmProAppHelper::views_is_installed() : FrmAppHelper::pro_is_installed();`

jairoprez:
> Hey @jairoprez
> 
> I'm looking over this update now. I'm seeing tiny little icons in the new form modal.
> 
> Screen Shot 2021-02-10 at 11 26 49 AM

Good observation @Crabcyborg , a css was causing this issue, it is already fixed

js/formidable_admin.js
stephywells: I've got a fix for this incoming.

classes/views/welcome/screen.php
jairoprez: @Crabcyborg I encapsulated your suggestion in a new function, did manual testing and it seems to do the trick 9e437be3419a6798418fd6694eabf5abc847da1b - does it make sense to you?

images/icons.svg
stephywells: These are compressed now.

formidable.php
stephywells: This is updated. We need to use both since one is temporary and one is permanant

images/briefcase.svg
stephywells: These are now inline. If we need to reuse them, we can move them into the sprite later.

stephywells (approved):
This looks ready to merge now. @Crabcyborg @jairoprez do you see anything else that needs to change?

jairoprez:
@stephywells I pulled your last changes, tested everything and it looks good to me. Thanks for simplifying and improving the structure of the base code.

PR #334, opened 2021-01-21
add inline box shadow to autofill input

Crabcyborg:
This is a required part of the fix for https://github.com/Strategy11/formidable-pro/pull/2845

Without this, the autofilled field can't match the css check.


stephywells (commented):
Does this look weird if the field has a dark background color with white text?

Crabcyborg:
> Does this look weird if the field has a dark background color with white text?

I don't think so. I believe this style is intended to not really change much visually. I tried changing the body to black and using autofill and I didn't actually see white anywhere.

PR #335, opened 2021-01-21
Avoid XSS from data-verify / data-caution attributes

Crabcyborg:
Fixes a vulnerability Steph mentions in https://github.com/Strategy11/formidable-pro/issues/2827#issuecomment-762478526

This can be exploited when someone creates an anchor tag that matches our expected attribute.

Also fixes https://github.com/Strategy11/formidable-pro/issues/2846

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -3751,7 +3751,7 @@ function frmAdminBuildJS() {
 			if ( typeof requires === 'undefined' || requires === null || requires === '' ) {
 				requires = 'Pro';
 			}
-			jQuery( '.license-level' ).html( requires );
+			jQuery( '.license-level' ).text( requires );
 
 			// If one click upgrade, hide other content
 			addOneClickModal( this );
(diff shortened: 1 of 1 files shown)


stephywells:
This looks great! Thanks so much for solving these!

I think we're still missing part of the external image issue. If someone adds the html in an entry, and then the entry is viewed in the admin area or using a view, loading the image would be vulnerable. Did you see my note about this in the issue?

stephywells (approved):
This looks good. We just need to double check the upgrade messages that include HTML. Some include images in the upgrade message. @jairoprez can you test this in the lite version?

jairoprez (approved):
@stephywells @Crabcyborg I did some testing and it seems that we are well covered for now:
- https://recordit.co/s467N4ZSBi
- https://recordit.co/T6fWoCGCS4
- https://recordit.co/ycsDDITeAM
- https://recordit.co/omRIpz5cYF

PR #336, opened 2021-01-27
add 4.09.05 changelog

Crabcyborg:
Adding in the change log for 4.09.05.

Leaving out the copies update as it still hasn't merged, and I feel like this update is important (and the copies update isn't too necessary).

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 == Changelog ==
 = 4.09.05 =
+* Fix: Autofill was flagging form submissions as spam with Honeypot in some browsers.
 * Fix: Important security update that adds better escaping when text is used from attribute data.
 
 = 4.09.04 =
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -1,4 +1,12 @@
 == Changelog ==
+= 4.09.05 =
+* Fix: Autofill was flagging form submissions as spam with Honeypot in some browsers.
+* Fix: Repeater fields were occasionally filling in with the wrong value.
```
stephywells, line 4: @Crabcyborg I've been doing two separate changelogs. So only the lite features would be included here, and the pro features would be in the changelog on our site. 

So datepickers, repeaters, file uploads, calculations are all pro features.

For the security note, let's add a bit more info without being too specific.

Crabcyborg: Got it @stephywells what do you think about my new revision?

stephywells: Looks great! But Honeypot is in lite, so let's keep that one in there.

Crabcyborg: @stephywells I hadn't considered that! The fix was done in pro, and actually doesn't fix this issue for lite!

stephywells: Nice catch!

PR #339, opened 2021-02-02
remove jquery blur and focus shorthand

Crabcyborg:
We had a support ticket come in asking about some left behind shorthand warnings from jQuery migrate. I don't think these would ever actually be removed from jQuery, but we might as well change everything to drop the deprecation warnings.

Similar to some of the earlier deprecation updates made in https://github.com/Strategy11/formidable-forms/pull/315

### DIFF
--- a/js/bootstrap-multiselect.js
+++ b/js/bootstrap-multiselect.js
@@ -708,7 +708,7 @@
                     this.lastToggledInput = $target;
                 }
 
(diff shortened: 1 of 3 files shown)


stephywells (approved):
These look great. I'd just like a quick test on this before merging.

stephywells:
Actually... it looks safe. Merging.

PR #341, opened 2021-02-04
Maybe add wp site health before calling rest_do_request

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2865

I don't really think there's anything you can actually test for with this function. It doesn't return anything. But having the test here catches any errors, so I just bumped the assertion count.

I see that `network_upgrade_site` is called from 3 different places, and it's not very clear which one is the issue here.

I see some of the calls wrapped in conditions like `if ( ! FrmAppHelper::doing_ajax() && self::needs_update() ) {`. Maybe we need to put more of these conditions into the function itself?

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -586,7 +586,7 @@ class FrmAppController {
 		if ( ! class_exists( 'WP_Site_Health' ) ) {
 			$wp_site_health_path = ABSPATH . 'wp-admin/includes/class-wp-site-health.php';
 			if ( file_exists( $wp_site_health_path ) ) {
-				require_once ABSPATH . 'wp-admin/includes/class-wp-site-health.php';
+				require_once $wp_site_health_path;
 			}
 		}
 	}
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -562,6 +562,8 @@ class FrmAppController {
 		set_transient( 'frm_updating_api', true, MINUTE_IN_SECONDS );
 		$request = new WP_REST_Request( 'GET', '/frm-admin/v1/install' );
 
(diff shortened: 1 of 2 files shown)


classes/controllers/FrmAppController.php
```diff
@@ -576,6 +578,19 @@ public static function network_upgrade_site( $blog_id = 0 ) {
...
 
+	/**
+	 * Make sure WP_Site_Health has been included because it is required when calling rest_do_request.
+	 * Check first that the file exists because WP_Site_Health was only introduced in WordPress 5.2.
+	 */
+	private static function maybe_add_wp_site_health() {
+		if ( ! class_exists( 'WP_Site_Health' ) ) {
+			$wp_site_health_path = ABSPATH . 'wp-admin/includes/class-wp-site-health.php';
+			if ( file_exists( $wp_site_health_path ) ) {
+				require_once ABSPATH . 'wp-admin/includes/class-wp-site-health.php';
```
jairoprez, line 589: Maybe you could reuse the $wp_site_health_path variable here, since it contains the path of the file you need to include `require_once $wp_site_health_path;`

jairoprez (approved):
In general the code looks good to me :muscle:

classes/controllers/FrmAppController.php
Crabcyborg: Lol, an oversight. Thanks for catching that!

stephywells: Thanks Jairo! Is this ready to merge?

Crabcyborg: @stephywells yeah, I think so!

jairoprez: Sure!, it can be merged.

PR #342, opened 2021-02-04
do not fill frm_verify field when autofilling for free templates

Crabcyborg:
This fixes the issue with the free templates form flagging emails as invalid.


jairoprez (approved):
This solution seems good to me.

stephywells (approved):
Haha! Oops!

PR #344, opened 2021-02-15
validate message data before rendering it

Crabcyborg:
I'm working toward a solution for https://github.com/Strategy11/formidable-pro/issues/2879

This at least avoids all of the errors that happen when you load the inbox.

### DIFF
--- a/classes/models/FrmInbox.php
+++ b/classes/models/FrmInbox.php
@@ -109,6 +109,11 @@ class FrmInbox extends FrmFormApi {
 		}
 
 		$this->fill_message( $message );
+
+		if ( ! $this->validate_message_data( $message ) ) {
+			return;
+		}
+
 		$this->messages[ $message['key'] ] = $message;
 
 		$this->update_list();
--- a/classes/models/FrmInbox.php
+++ b/classes/models/FrmInbox.php
@@ -49,16 +49,31 @@ class FrmInbox extends FrmFormApi {
 	 */
 	public function set_messages() {
 		$this->messages = get_option( $this->option );
-		if ( empty( $this->messages ) ) {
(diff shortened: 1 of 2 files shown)


stephywells (commented):
It might make sense to add this check right before line 112 too:
$this->messages[ $message['key'] ] = $message;

This would prevent the message from being added, rather than adding it and then removing it later.

Crabcyborg:
> It might make sense to add this check right before line 112 too:
> $this->messages[ $message['key'] ] = $message;
> 
> This would prevent the message from being added, rather than adding it and then removing it later.

@stephywells Sorry for not getting to that already! Added now.

stephywells:
Hmm... That's weird. How did the message get added? Do you have a copy/paste of the saved array by chance?

stephywells:
Nevermind. I see it in the issue. I'm looking over that a bit more.

stephywells:
Okay, it seems like the better fix would be to remove anything with a blank key. The key should always have a value. That might be where the api error came in. So instead of checking for all of the array keys that should be filled in automatically, let's unset anything with a blank key.

Would changing
if ( $this->is_expired( $message ) || $dismissed ) {
to:
if ( empty( $k ) || $this->is_expired( $message ) || $dismissed ) {

do it?

stephywells:
And maybe this too?

	public function mark_read( $key ) {
		if ( ! $key || ! isset( $this->messages[ $key ] ) ) {

Crabcyborg:
@stephywells sounds good, starting from scratch with those. This removes the error, and it's simpler.

Closing this PR, new PR is here - https://github.com/Strategy11/formidable-forms/pull/346

PR #345, opened 2021-02-15
clean confirm message before appending text fix

Crabcyborg:
Nat just caught a bug from a recent update here - https://github.com/Strategy11/formidable-pro/issues/2878#issuecomment-779339042

The modal wasn't clearing, so the pop up could include old messages as well in it.

Best to slip this fix into the release as well 🚀

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -386,6 +386,7 @@ function frmAdminBuildJS() {
 		caution = link.getAttribute( 'data-frmcaution' );
 		verify = link.getAttribute( 'data-frmverify' );
 		$confirmMessage = jQuery( '.frm-confirm-msg' );
+		$confirmMessage.empty();
 
 		if ( caution ) {
 			frmCaution = document.createElement( 'span' );


stephywells (commented):
Are there any cases where this would be empty?

Crabcyborg:
> Are there any cases where this would be empty?

@stephywells I don't believe so. The original logic seemed to always set the HTML every time, so I think it's assumed that some of these things are not really optional attributes.

It defaults to having a "Are you sure?" text, that gets replaced every time the pop up happens, so I'm thinking that the default never gets used.

It might benefit to get cleaned up a bit more.

Crabcyborg:
@stephywells I just looked at the code that calls this, and I don't think it ever is empty.

It really only gets this far is there is a `data-frmverify` set. confirmModal is only called from confirmLinkClick which checks for this.

I looked around for `data-frmverify` and it's definitely never empty when it is set. In theory, you could set this to an empty string and end up with an empty title. That doesn't sound like an issue though.

stephywells:
Thanks for checking!

PR #346, opened 2021-02-15
check for empty message keys

Crabcyborg:
Continuing from https://github.com/Strategy11/formidable-forms/pull/344

Fixes https://github.com/Strategy11/formidable-pro/issues/2879 using Steph's suggestions instead of my original validate/filter in different parts of this code.


stephywells (approved):
This looks cleaner to me. Thanks!

PR #348, opened 2021-02-16
update 4.09.06 readme

Crabcyborg:
I'm including a few unmerged items in this list since I want to get all of these things into this release.

Also bumping the tested WordPress version. I set it to 5.6.1 since that's the latest stable version. I'm not sure if setting it to 5.7 early makes sense - technically it's what I've been mostly testing with.


stephywells:
Perfect!

PR #350, opened 2021-03-01
add -label to single option label input id in order to make it unique

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2878

I mention in my comment there more about the conflict.

This field needs a unique id. Adding `-label` makes sense to me, as the input targets the label, but maybe that's misleading, since it's not actually a label either.

I don't believe this id for the input label is actually used anywhere in particular. If it was relying on id before, it likely wouldn't have worked since it wasn't unique anyway.


jairoprez (approved):
Add `-label` to get a unique id did the trick https://recordit.co/3gdrLUwmxt 🎉

PR #351, opened 2021-03-01
check if message is array before adding it fix

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2873

I didn't realize we were initially given much of the original error from the ticket.

It's cut off, but the first error is `Illegal string offset 'key'`.

Which points out the obvious issue that `$message` was coming through as a string, not an array at all.

We already cleaned up after the fact, but this prevents the issue from happening in the first place.


stephywells (approved):
Good call. Thank you!

PR #353, opened 2021-03-02
generate a new field key if the length is over 100

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2855 and I believe it fixes https://github.com/Strategy11/formidable-pro/issues/2741 as well.

You can't duplicate a field if the key is over 100 characters because we have a size limit in our database. If the key is over 100 characters in length, I treat it like it's empty and generate a new one from scratch.

I check specifically for the `$field_key` column value as I realize this might not be limited to 100 for other columns.

The other update is just to remove a phpcs warning.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1559,7 +1559,7 @@ class FrmAppHelper {
 			$key = sanitize_key( $name );
 		}
 
-		if ( empty( $key ) ) {
+		if ( empty( $key ) || strlen( $key ) >= 100 ) {
 			$max_slug_value = pow( 36, $num_chars );
 			$min_slug_value = 37; // we want to have at least 2 characters in the slug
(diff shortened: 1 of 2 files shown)


stephywells (approved):
Weird bug.. I think we could make it smaller than 100 too, but not a big deal either way.

jairoprez (approved):
This fix is working as expected https://recordit.co/GUBjD1SvK5

Crabcyborg:
> Weird bug.. I think we could make it smaller than 100 too, but not a big deal either way.

@stephywells Yeah, I wasn't sure where to actually set the point to. I've lowered it 70 now, and I also refactored things a bit: I moved my fix into a separate function, and I moved the generate code and the code that prevents numbers and reserved strings into new functions as well.

stephywells (approved):
Nice!

PR #354, opened 2021-03-03
change deprecated jquery keyup trigger event

Crabcyborg:
Noticed another jQuery deprecation warning coming from formidable.

`jQuery.fn.keyup() event shorthand is deprecated`

Also, using `ubuntu-latest` has finally caught up with formidable lite as well and no longer works for several versions of PHP we're testing so I'm forcing ubuntu 16.04 for now because it works with all of our tests.


stephywells (approved):
Thanks Mike!

PR #355, opened 2021-03-03
update readme for 4.09.07

Crabcyborg:


stephywells (commented):
Hey @Crabcyborg Would you mind taking off the oldest version under the changelog section in the readme? I think that'll help us stay a it more consistent in our WP rankings. Also, 5.7 is coming soon and our rankings will be affected if we don't update the readme to say it works with 5.7. Have you tested that by chance?

Crabcyborg:
> Hey @Crabcyborg Would you mind taking off the oldest version under the changelog section in the readme? I think that'll help us stay a it more consistent in our WP rankings. Also, 5.7 is coming soon and our rankings will be affected if we don't update the readme to say it works with 5.7. Have you tested that by chance?

@stephywells updated!

I've been running locally on various beta versions of 5.7 for a few months now and Formidable has been running well. I could have set it earlier, but wasn't sure if we wanted to set it to versions before they were released or not.

stephywells (approved):
Perfect! I don't think too early would be a problem. But it starts counting against us as soon as it's released.

PR #357, opened 2021-03-04
Confirm before leaving a page with unsaved changes when a form action is added or removed

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2907

I usually just check for changed fields, but that doesn't work to detect new or deleted actions.

I also noticed in my testing that the same issue applied to creating the action (add an action, leave the page, no warning).


jairoprez (approved):
The code looks great to me and the enhancement is working great https://recordit.co/JvCPAknX5x 🎉

stephywells (approved):

PR #358, opened 2021-03-04
fix product price input width

Crabcyborg:
I noticed this was weird for a while, but never really knew when the issue was introduced.

The issue with the `:last-child` selector is because I introduced a hidden `optionmap` input on text focus in v4.08 (https://github.com/Strategy11/formidable-forms/commit/a35ac7848d9d6ee47c1c41efa8e19d52ae27abdc) for updating radio option conditional logic. This seems to somewhat apply to products as well, although I also noticed this small issue - https://github.com/Strategy11/formidable-pro/issues/2909

**Before**
Screen Shot 2021-03-04 at 4 15 52 PM

**After**
Screen Shot 2021-03-04 at 4 15 37 PM

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -5876,7 +5876,7 @@ tr.frm_options_heading td {
 	float: left;
 }
 
-#new_fields .frm_product_price_wrapper input:last-child {
+#new_fields .frm_product_price_wrapper input:nth-child(2) {
 	width: 37%;
 	float: right;
 }


stephywells (approved):
Thanks for fixing this!

stephywells:
Oh, does this right with the separate values box checked too?

Crabcyborg:
> Oh, does this right with the separate values box checked too?

@stephywells yeah, the separate "saved value" input is always on its own row underneath. It doesn't change much with this update.

stephywells:
Awesome! Thank you!

PR #366, opened 2021-03-10
check the actual value that would be saved to the database when checking for duplicates

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2872

The issue title is a little misleading. The hook works fine. This is really just an issue that signatures, and some other fields, don't get flagged as duplicates even when they are.

Two empty or identical signatures never get flagged as duplicated. Likely anything that might change its value afterward (files and tags appear to as well) as the comparison check is checking the before-values against after-values.

This just changes those before-values to after-values so we're not comparing two different things against each other.

An example before-value (serialized) `a:2:{s:5:"typed";s:0:"";s:6:"output";s:0:"";}`
Eventually becomes an empty string and never gets saved.

### DIFF
--- a/classes/controllers/FrmFieldsController.php
+++ b/classes/controllers/FrmFieldsController.php
@@ -583,12 +583,17 @@ class FrmFieldsController {
 		}
 	}
 
+	/**
+	 * @param array $field
+	 * @return string
+	 */
 	private static function prepare_placeholder( $field ) {
+		$placeholder          = isset( $field['placeholder'] ) ? $field['placeholder'] : '';
+		$placeholder_is_blank = empty( $placeholder ) && '0' !== $placeholder;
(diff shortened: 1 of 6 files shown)


jairoprez (approved):
@Crabcyborg You're right, the `frm_time_to_check_duplicates` hook is working properly. I did tests and your solution works excellent https://recordit.co/bcVSxkAa2Q

stephywells:
Nice find! This should help a ton.

stephywells:
I'm not sure if it's Github or my connection right now.... I'm getting errors all over the place and it's not letting me merge.

stephywells:
Hmm.... now this is saying there's a conflict. It's possible trying to merge messed it up when they failed.

Crabcyborg:
> Hmm.... now this is saying there's a conflict. It's possible trying to merge messed it up when they failed.

@stephywells ha, it seemed a little confused, but it looks like it's merged now.

PR #367, opened 2021-03-10
undefined key custom style when there are no fields fix

Crabcyborg:
I noticed something in my error log when I tried to preview a form with no fields.

```
PHP Warning:  Undefined array key "custom_style" in /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php on line 1946
```

I also refactored some other custom style code to make this more consistent across formidable. There was already a helper function in `FrmAppHelper` that had this same logic in it.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1931,7 +1931,6 @@ class FrmFormsController {
 		$description = $args['description'];
 
 		if ( empty( $args['fields'] ) ) {
-			$frm_settings = FrmAppHelper::get_settings();
 			$values       = array(
 				'custom_style' => FrmAppHelper::custom_style_value( array() ),
 			);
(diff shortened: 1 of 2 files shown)


jairoprez (approved):
Nice catch!, I did tests and the warning error was fixed: https://recordit.co/iweIrX0NYz

classes/controllers/FrmFormsController.php
```diff
@@ -1931,7 +1931,10 @@ private static function show_form_after_submit( $args ) {
 		$description = $args['description'];
 
 		if ( empty( $args['fields'] ) ) {
-			$values = array();
+			$frm_settings = FrmAppHelper::get_settings();
```
stephywells, line 1934: Is there a reason the settings are fetched right here?

Crabcyborg: @stephywells nope, nice catch! I'm getting `$frm_settings->load_style != 'none'` so originally I was using the variable in here and forgot to clean this one up.

stephywells: Looks good to me.

PR #369, opened 2021-03-17
Make icons clickable in new form modal

Crabcyborg:
I noticed in the new view modal that clicking the icons was not triggering the new form events because they were missing the `role="button"` code. Same applies to the new form modal, so I'm making an update here too.

I'm also auto-focusing the template name input if you're creating one with an empty name because I think there should be one. I've tried to type a name a few times and had to focus first. Just makes it easier.

And I'm changing a couple of label updates to use `textContent` since they come from data attributes and this could ultimately be manipulated a bit (though that would take a lot of hoops to jump through).

### DIFF
--- a/classes/views/frm-forms/list-template.php
+++ b/classes/views/frm-forms/list-template.php
@@ -34,7 +34,7 @@ if ( ! empty( $template['custom'] ) ) {
 	<div class="frm-featured-form">
 		<?php
 		if ( $render_icon ) {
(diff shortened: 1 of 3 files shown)


jairoprez (approved):
Nice catch, icons in new form modal are now clickable https://recordit.co/tomocW6UyJ 🎉

stephywells (approved):
Looks great! Thanks to both of you!

PR #373, opened 2021-03-26
Add JS code to dynamic show/hide Post Time option in pro version

truongwp:
This is related to this issue https://github.com/Strategy11/formidable-pro/issues/2943

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -4251,6 +4251,17 @@ function frmAdminBuildJS() {
 		}
 	}
 
+	function togglePostTime() {
+		/*jshint validthis:true*/
+		var v = jQuery( this ).val();
+		if ( '' === v ) {
+			jQuery( '.frm_post_time_opt' ).val( '' );
+			jQuery( '.frm_post_time_field' ).hide();
+		} else {
+			jQuery( '.frm_post_time_field' ).show();
+		}
+	}
+
 	function fillDyncontent() {
 		/*jshint validthis:true */
 		var v = jQuery( this ).val();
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -4251,6 +4251,17 @@ function frmAdminBuildJS() {
 		}
 	}
 
+	function togglePostTime() {
+		/*jshint validthis:true*/
+		var v = jQuery( this ).val();
```
stephywells, line 4256: We have been trying to use as much vanilla js as possible with new code we add. The plan is to move away from jQuery, but it'll take some time. Sorry for not mentioning this before! I'll add this in the doc here.

Would you mind updating this function to use vanilla js only? We can leave the trigger (line 6733) and plan on changing all of those out at once.

stephywells (changes requested):
This works and looks great too!

js/formidable_admin.js
truongwp: @stephywells ok. Let me update this

truongwp: @stephywells Should we use some ES6 features which are currently supported like `const` or `let` or arrow function?

Which browsers do we support?

stephywells (approved):
Looks great! Thanks so much @truongwp!

I'll go ahead and add the "needs QA" tag and request a QA review on this too.

js/formidable_admin.js
stephywells: For the back-end, we don't support old IE browsers, so it's fine to use functions that aren't supported in IE. For the front-end we still need to support IE.

On the back end we support current Chrome, Safari, and Edge.

jairoprez:
@stephywells @truongwp Hey guys, I'm not sure how to test this functionality, could you give me some hints please?

stephywells:
Sure! You'll need this PR and https://github.com/Strategy11/formidable-pro/pull/2947

Then go to a form action to create a post. These PRs add a time option for the post date/time. This needs to be tested with a time field in the form, and without a time field.

Here's more detail:
https://github.com/Strategy11/formidable-pro/issues/2943

jairoprez (approved):
The Post Time option is hidden by default and is only shown if the Post Date option is set to a specific field instead of "Date of entry submission" https://recordit.co/w3eJD9tP9c Nice job @truongwp 🎉

stephywells (approved):

PR #374, opened 2021-03-26
use unique ids when dragging a field to avoid replacement conflicts

Crabcyborg:
I was trying to push the editor to its limits to see if anything would break, and something did 😱

It turns out, that if you drag the same type of field really fast (which can happen from just barely dragging it as well, if you were clicking and moving around your mouse a lot like I was), it will break when it replaces the temporary loading spinner because the two fields that are being dragged have the same id. The value of `loadingID` is just the field type: "select", or "text", etc.

```
Uncaught TypeError: field is null
    afterAddField http://localhost:8889/wp-content/plugins/formidable/js/formidable_admin.js?ver=4.10:1259
    success http://localhost:8889/wp-content/plugins/formidable/js/formidable_admin.js?ver=4.10:1034
    jQuery 4
```

To fix this, I'm adding a function that just returns an id it didn't return before, to make sure that the value of `loadingID` is always unique.

I also bumped all of the variable declarations for this function to the top while I was at it, to remove a few of our linting warnings.

I also tested this in Chrome and there is a different error,

Screen Shot 2021-03-26 at 1 40 33 PM

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1170,7 +1170,6 @@ function frmAdminBuildJS() {
 			success: function( msg ) {
 				document.getElementById( 'frm_form_editor_container' ).classList.add( 'frm-has-fields' );
 				$newFields.append( msg );
-
 				afterAddField( msg, true );
 			},
 			error: function( jqXHR, textStatus, errorThrown ) {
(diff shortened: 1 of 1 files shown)


jairoprez (approved):
This fix seems to work fine for chrome https://recordit.co/1D2MfdEUUG and firefox https://recordit.co/S5ENC8rIpf

stephywells (approved):
Haha!! Thanks for moving your mouse really fast and looking for issues! It's great when we can find the issues before they are reported, especially when they're hard to replicate. Thank you!

PR #375, opened 2021-03-26
do not display upgrade modal if option is not disabled

Crabcyborg:
Another thing I noticed when trying to push the form editor, is that you can actually trigger the upgrade modals for fields that are not locked, by clicking in the corners where the `[data-upgrade]` listener is happening, where the inner `.frm_add_field` is not touching because of its border radius.

Screen Shot 2021-03-25 at 5 13 15 PM

To fix this, I'm checking the value of `show_upgrade` now, and actually avoiding some of the upgrade message logic and page HTML, hopefully improving performance a bit in the process for people who are on pro.

And I also refactored the JavaScript a bit, moving any variable declarations to the top of the function. And aligned some spaces in PHP.

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -1725,14 +1725,15 @@ class FrmFieldsHelper {
 		$field_label .= ' <span>' . $field_name . '</span>';
 
 		// If the individual field isn't allowed, disable it.
-		$run_filter      = true;
-		$single_no_allow = ' ';
-		$install_data    = '';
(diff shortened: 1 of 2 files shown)


jairoprez (approved):
@Crabcyborg nice catch, I'm surprised you came across a problem like this as it doesn't seem easy to spot https://recordit.co/OeE38k47Gf - your fix works quite well 🎉

stephywells (approved):
Good find! Looks good to me.

Thanks @Crabcyborg and @jairoprez !

PR #377, opened 2021-03-31
do not remove ui-state-default from field when layout options open

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2952

The issue actually begins as soon as you open the "Add Layout Classes" pop up. It removes the `ui-state-default` class, which is necessary for the event listener to register the click event.

The issue comes down to the line above the fix `classes = field.className.split( ' frmstart ' )[1].split( ' frmend ' )[0];`.

`.split( ' frmend ' )` is not finding our class because there is no leading space before our `frmend` class.

I do an additional check, for if the string begins with `'frmend '`, and set it to an empty string if it is. This is the effectively equivalent of checking at [0] after the implode, for this specific case.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -3541,7 +3541,7 @@ function frmAdminBuildJS() {
 			replace = ' frmstart  frmend ';
 			replaceWith = ' frmstart ' + replaceWith.trim() + ' frmend ';
 		} else {
-			replace = classes.trim();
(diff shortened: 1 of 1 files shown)


jairoprez (approved):
Tested and the fix seems to work correctly https://recordit.co/F6NnxPoNXX 👍

stephywells (approved):
That was fast! Thank you!

PR #378, opened 2021-04-01
remove unused var link in doaction listener

Crabcyborg:
This isn't really anything crazy. I just noticed this `var link` was totally unused in this function. My syntax highlighter has it in another color, so it stands out a bit.

Doesn't hurt to clean up 🙂.


stephywells (approved):
Thanks for keeping an eye out for cleanup!

PR #379, opened 2021-04-04
Conditional Logic adds a redundant logic option pro version fixed iss…

rejuancse:
Conditional Logic adds a redundant logic option pro version fixed issue fixed

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -3039,7 +3039,8 @@ function frmAdminBuildJS() {
 					optionMatch.textContent = expectedOption;
 				}
 
-				valueSelect.prepend( optionMatch );
+				// valueSelect.prepend( optionMatch );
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -3039,7 +3039,8 @@ function frmAdminBuildJS() {
 					optionMatch.textContent = expectedOption;
 				}
 
-				valueSelect.prepend( optionMatch );
+				// valueSelect.prepend( optionMatch );
+				valueSelect.prepend( expectedOption );
```
Crabcyborg, line 3043: @rejuancse I don't think this is right. `expectedOption` is just a string value. It isn't an element.

This is likely just going to cause a lot of bugs. The problem is a little complicated than this.

Also, we generally don't leave commented code behind. You would want to clean it up (if this was a valid fix).

PR #381, opened 2021-04-05
Improve the mail header when using mail() function

truongwp:
This fixes this pro version issue https://github.com/Strategy11/formidable-pro/issues/2825

### DIFF
--- a/classes/models/FrmEmail.php
+++ b/classes/models/FrmEmail.php
@@ -435,7 +435,9 @@ class FrmEmail {
 		$sent = wp_mail( $recipient, $subject, $this->message, $header, $this->attachments );
 
 		if ( ! $sent ) {
-			$header    = 'From: ' . $this->from . "\r\n";
+			if ( is_array( $header ) ) {
+				$header = implode( "\r\n", $header );
+			}
 			$recipient = implode( ',', (array) $recipient );
 			$sent      = mail( $recipient, $subject, $this->message, $header );
 		}


classes/models/FrmEmail.php
```diff
@@ -435,7 +435,9 @@ private function send_single( $recipient ) {
 		$sent = wp_mail( $recipient, $subject, $this->message, $header, $this->attachments );
 
 		if ( ! $sent ) {
-			$header    = 'From: ' . $this->from . "\r\n";
+			if ( is_array( $header ) ) {
+				$header = implode( "\r\n", $header );
```
stephywells, line 439: Looks like we lost the "From: " part in here.... I think we'll need to keep that. Right?

truongwp: It's included here https://github.com/Strategy11/formidable-forms/blob/master/classes/models/FrmEmail.php#L466

truongwp: It's included here https://github.com/Strategy11/formidable-forms/blob/master/classes/models/FrmEmail.php#L466

stephywells: Oh, I see now! Thanks for clearing that up for me! Looks great!

stephywells (approved):

stephywells:
@jairoprez This could use a quick test. The easiest way to test this is probably to change this line:
$sent = wp_mail( $recipient, $subject, $this->message, $header, $this->attachments );

to:
$sent = false;

Thanks for your help!

jairoprez (commented):
@stephywells Thanks for the guide to test this PR, this fix seems to be working correctly.
- test with `wp_mail()`: https://recordit.co/ysRb6c0lpY
- test with `mail()`: https://recordit.co/YYiOnrwNrD

jairoprez (approved):

PR #382, opened 2021-04-05
add description padding option to style editor

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/1114

I haven't really touched the style editor yet but it turns out that adding new rules seems pretty straight forward.


jairoprez (approved):
New padding option seems to work properly https://recordit.co/rUD5jWLkBZ

stephywells:
Nice! @jairoprez did you also test this with a non-default style? I'd just like to be sure that the other styles will override the default.

jairoprez:
> Nice! @jairoprez did you also test this with a non-default style? I'd just like to be sure that the other styles will override the default.

@stephywells Oops! :sweat_smile: I hadn't done tests on that part.

Super @Crabcyborg I did new tests based on Steph's approach and the solution doesn't work with a non-default style :pensive: https://recordit.co/PDiLsxQVKx

Crabcyborg:
> Super @Crabcyborg I did new tests based on Steph's approach and the solution doesn't work with a non-default style 😔 https://recordit.co/PDiLsxQVKx

Thanks @jairoprez this should be working now.

I missed a few small details on how this works but I think it's good now.

jairoprez:
> > Super @Crabcyborg I did new tests based on Steph's approach and the solution doesn't work with a non-default style 😔 https://recordit.co/PDiLsxQVKx
> 
> Thanks @jairoprez this should be working now.
> 
> I missed a few small details on how this works but I think it's good now.

@Crabcyborg tested and seems to work as expected now 🎉

PR #383, opened 2021-04-05
add repeat icon color to style settings

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2944 alongside https://github.com/Strategy11/formidable-pro/pull/2964

I played with the preview stuff in the editor, which seems to require single_theme. I'm a little confused by the differences between the single and custom theme files. I don't want to add too much, so I left out the changes I had in single_theme. The preview doesn't update in real time, but it does work after save, which seems to be consistent with some of the things in the style editor and not others.

### DIFF
--- a/classes/models/FrmStyle.php
+++ b/classes/models/FrmStyle.php
@@ -416,6 +416,7 @@ class FrmStyle {
 			'collapse_icon'        => '6',
 			'collapse_pos'         => 'after',
 			'repeat_icon'          => '1',
+			'repeat_icon_color'    => 'ffffff',
 
 			'submit_style'               => false,
 			'submit_font_size'           => '15px',
--- a/css/_single_theme.css.php
+++ b/css/_single_theme.css.php
@@ -444,4 +444,8 @@ if ( '' === $field_height || 'auto' === $field_height ) {
 	padding:0<?php echo esc_html( $important ); ?>;
 }
 
+.frm_repeat_sec .frm_icon_font::before {
+	color: <?php echo esc_html( $repeat_icon_color ); ?> !important;
+}
+
 <?php do_action( 'frm_output_single_style', $settings ); ?>
--- a/css/_single_theme.css.php
+++ b/css/_single_theme.css.php
@@ -444,8 +444,10 @@ if ( '' === $field_height || 'auto' === $field_height ) {
 	padding:0<?php echo esc_html( $important ); ?>;
 }
 
-.frm_repeat_sec .frm_icon_font::before {
-	color: <?php echo esc_html( $repeat_icon_color ); ?> !important;
-}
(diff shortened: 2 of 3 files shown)


jairoprez (commented):
@Crabcyborg For some strange reason, I can't see the icons on the front-end, I tried chrome and firefox, can you see them in your browser? https://recordit.co/zpg0Jvgg8d

Crabcyborg:
> @Crabcyborg For some strange reason, I can't see the icons on the front-end, I tried chrome and firefox, can you see them in your browser? https://recordit.co/zpg0Jvgg8d

@jairoprez yeah, that's pretty much the bug we're trying to fix here.

Your theme has a white background with white buttons and white icons. They're there, just white.

In the style editor you should be able to add a color to them now, to fix this yourself, as it's sort of tough to really know what color you need for your theme.

This likely isn't an issue in many themes, but it is an issue in others.

Maybe we want to give it a default color, but I was avoiding changing the default behaviour, in case that leads to unwanted colors for our users.

jairoprez:
> > @Crabcyborg For some strange reason, I can't see the icons on the front-end, I tried chrome and firefox, can you see them in your browser? https://recordit.co/zpg0Jvgg8d
> 
> @jairoprez yeah, that's pretty much the bug we're trying to fix here.
> 
> Your theme has a white background with white buttons and white icons. They're there, just white.
> 
> In the style editor you should be able to add a color to them now, to fix this yourself, as it's sort of tough to really know what color you need for your theme.
> 
> This likely isn't an issue in many themes, but it is an issue in others.
> 
> Maybe we want to give it a default color, but I was avoiding changing the default behaviour, in case that leads to unwanted colors for our users.

Hey Super @Crabcyborg thanks for clearing that up for me, you were absolutely right! - however the color settings for the repeater field icons seems to work fine for the text links with icons but it doesn't seem to take the color for the icons option without links, it remains white no matter if another color has been selected from settings https://recordit.co/0cmBawwX1O

Crabcyborg:
> Hey Super @Crabcyborg thanks for clearing that up for me, you were absolutely right! - however the color settings for the repeater field icons seems to work fine for the text links with icons but it doesn't seem to take the color for the icons option without links, it remains white no matter if another color has been selected from settings https://recordit.co/0cmBawwX1O

@jairoprez thank you for catching that! It should be working now as expected 🎉

jairoprez (approved):
Tested and is working properly now! 💪

Crabcyborg:
> Tested and is working properly now! 💪

Thank you @jairoprez!

stephywells (approved):
Nice! 

So just to clarify between the custom and single css... The single css file is repeated for each created style. This is why we started using the -var option so we can avoid adding to the single css, to keep the file size down.

PR #384, opened 2021-04-05
Do not pass the nonce referer for bulk actions

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2963

I don't think we can safely pass a nonce referer if we're passing the nonce through a `$_GET` request because the url will grow until it eventually hits a limit.

So I just don't pass it. It looks like we're not really checking for it, so everything still seems to work, just with a smaller URL.


Crabcyborg:
Closed this one because I accidentally didn't make it off of master. Easier to just remake https://github.com/Strategy11/formidable-forms/pull/385

PR #390, opened 2021-04-11
Create Name field

truongwp:
This fixes issue https://github.com/Strategy11/formidable-pro/issues/2954

Remaining things:

- [x] Focus to the first input when clicking to the primary label.
- [x] Make backend and frontend be consistent.
- [x] Make combo field work with import/export. In the export file, each sub field will be a column.


Crabcyborg:
I tried beginning with clean form that just had a new Name field, and saw several errors right away.

```
[12-Apr-2021 19:12:49 UTC] PHP Warning:  Undefined array key "first" in /var/www/src/wp-content/plugins/formidable/classes/views/frm-fields/back-end/combo-field/show-on-form-builder.php on line 37
[12-Apr-2021 19:12:49 UTC] PHP Warning:  Undefined array key "last" in /var/www/src/wp-content/plugins/formidable/classes/views/frm-fields/back-end/combo-field/show-on-form-builder.php on line 37
```

```
[12-Apr-2021 19:13:01 UTC] PHP Fatal error:  Uncaught TypeError: Cannot access offset of type string on string in /var/www/src/wp-content/plugins/formidable/classes/views/frm-fields/front-end/combo-field/combo-field.php:50
Stack trace:
#0 /var/www/src/wp-content/plugins/formidable/classes/models/fields/FrmFieldCombo.php(160): include()
#1 /var/www/src/wp-content/plugins/formidable/classes/models/fields/FrmFieldType.php(808): FrmFieldCombo->front_field_input(Array, Array)
#2 /var/www/src/wp-content/plugins/formidable/classes/models/FrmFieldFormHtml.php(346): FrmFieldType->include_front_field_input(Array, Array)
#3 /var/www/src/wp-content/plugins/formidable/classes/models/FrmFieldFormHtml.php(331): FrmFieldFormHtml->replace_input_shortcode(Array)
#4 /var/www/src/wp-content/plugins/formidable/classes/models/FrmFieldFormHtml.php(115): FrmFieldFormHtml->replace_shortcodes_with_atts()
#5 /var/www/src/wp-content/plugins/formidable/classes/models/fields/FrmFieldType.php(713): FrmFieldFormHtml->get_html()
#6 /var/www/src/wp-content/plugins/formidable/classes/models/fields/FrmFieldType.php(684): FrmFieldType->prepare_field_html(Array)
#7 /var/www/src/wp-content/plugins/formidable/classes/helpers/FrmFieldsHelper.php(342): FrmFieldType->show_field(Array)
#8 /var/www/src/wp-content/plugins/formidable/classes/views/frm-entries/form.php(47): FrmFieldsHelper::show_fields(Array, Array, Object(stdClass), 'create')
#9 /var/www/src/wp-content/plugins/formidable/classes/views/frm-entries/new.php(21): require('/var/www/src/wp...')
#10 /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php(1952): include('/var/www/src/wp...')
#11 /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php(1702): FrmFormsController::show_form_after_submit(NULL)
#12 /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php(1675): FrmFormsController::get_form_contents(Object(stdClass), true, true, Array)
#13 /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php(1639): FrmFormsController::get_form(Object(stdClass), true, true, Array)
#14 /var/www/src/wp-content/plugins/formidable/classes/views/frm-entries/direct.php(16): FrmFormsController::show_form('410', '', true, true)
#15 /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php(371): require('/var/www/src/wp...')
#16 /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php(267): FrmFormsController::load_direct_preview()
#17 /var/www/src/wp-includes/class-wp-hook.php(292): FrmFormsController::preview('')
#18 /var/www/src/wp-includes/class-wp-hook.php(316): WP_Hook->apply_filters('', Array)
#19 /var/www/src/wp-includes/plugin.php(484): WP_Hook->do_action(Array)
#20 /var/www/src/wp-admin/admin-ajax.php(187): do_action('wp_ajax_frm_for...')
#21 {main}
  thrown in /var/www/src/wp-content/plugins/formidable/classes/views/frm-fields/front-end/combo-field/combo-field.php on line 50
```

I'm going to keep looking at the rest, but leaving this here so I don't forget.

classes/models/fields/FrmFieldCombo.php
```diff
@@ -0,0 +1,232 @@
...
+	 */
+	public function validate( $args ) {
+		$errors    = isset( $args['errors'] ) ? $args['errors'] : array();
+		$blank_msg = FrmFieldsHelper::get_error_msg( $this->field, 'blank' );
+
+		$sub_fields = $this->get_processed_sub_fields();
+
+		// Validate not empty.
+		foreach ( $sub_fields as $name => $sub_field ) {
+			if ( empty( $sub_field['optional'] ) && empty( $args['value'][ $name ] ) ) {
```
Crabcyborg, line 224: It looks like the required logic is a little bit off.

A few unit tests are failing similarly [here](https://github.com/Strategy11/formidable-forms/pull/390/checks?check_run_id=2316692490):

```
1) test_FrmFieldValidate::test_not_required_fields
A field was required when it should not have been. field122-first, name, field122-last
Failed asserting that an array is empty.

/tmp/wordpress/src/wp-content/plugins/formidable/tests/fields/test_FrmFieldValidate.php:57

2) test_FrmFieldValidate::test_format_validation
number value 123 did not pass validation
Failed asserting that an array is empty.

/tmp/wordpress/src/wp-content/plugins/formidable/tests/fields/test_FrmFieldValidate.php:77

3) test_FrmFieldValidate::test_filled_required_fields
A field was required when it was not empty. field164-first, name, field164-last
Failed asserting that an array is empty.

/tmp/wordpress/src/wp-content/plugins/formidable/tests/fields/test_FrmFieldValidate.php:166
```

You should be able to submit this blank unless the "Required" is actually checked.
Screen Shot 2021-04-12 at 4 23 22 PM

classes/models/fields/FrmFieldName.php
```diff
@@ -0,0 +1,181 @@
...
+
+	/**
+	 * Gets ALL sub fields.
+	 *
+	 * @return array
+	 */
+	public function get_sub_fields() {
+		return array(
+			'first'  => array(
+				'type'     => 'text', // See supported types in classes/views/frm-fields/back-end/combo-field/show-on-form-builder.php.
```
Crabcyborg, line 27: What do you think about sharing some of this duplicate code? All three of the name fields seem pretty much the same, at least for now. The only different seems to be the value of `label`.

classes/views/frm-fields/front-end/combo-field/combo-field.php
```diff
@@ -0,0 +1,74 @@
...
+							}
+
+							$this->print_input_atts( compact( 'field', 'sub_field' ) );
+							?>
+						/>
+						<?php
+				}
+
+				if ( $sub_field['label'] && ! empty( $field[ $name . '_desc' ] ) ) {
+					echo '<div class="frm_description">' . wp_kses_post( $field[ $name . '_desc' ] ) . '</div>';
```
Crabcyborg, line 63: We have a `FrmAppHelper::kses` that I believe we generally use to cover `kses` calls.

classes/models/fields/FrmFieldCombo.php
```diff
@@ -0,0 +1,232 @@
...
+
+abstract class FrmFieldCombo extends FrmFieldType {
+
+	/**
+	 * Does the html for this field label need to include "for"?
+	 *
+	 * @var bool
+	 * @since 3.0
+	 */
+	protected $has_for_label = false;
```
Crabcyborg, line 17: I think that in this case, it does? The for should probably trigger the first name field in the set. It seems odd that it doesn't.

classes/models/fields/FrmFieldName.php
```diff
@@ -0,0 +1,181 @@
...
+	 * This should return the list of sub fields after sorting or show/hide based of some options.
+	 *
+	 * @return array
+	 */
+	protected function get_processed_sub_fields() {
+		$sub_fields  = $this->get_sub_fields();
+
+		$name_layout = FrmField::get_option( $this->field, 'name_layout' );
+
+		if ( 'last_first' === $name_layout ) {
```
Crabcyborg, line 77: I think we could probably simplify this function a bit.

I think we could probably even just `explode( $name_layout )` and then use them in the exploded order. The column class can basically just be `'frm' . ( 12 / count( $exploded ) );`.

We'll probably want to make sure it's validated, but I think it would look smaller and more consistent this way.

classes/models/fields/FrmFieldName.php
```diff
@@ -0,0 +1,181 @@
...
+	 * @return string      Most of cases, this will return string.
+	 */
+	protected function prepare_display_value( $value, $atts ) {
+		$name_layout = FrmField::get_option( $this->field, 'name_layout' );
+
+		if ( ! empty( $atts['show'] ) ) {
+			return isset( $value[ $atts['show'] ] ) ? $value[ $atts['show'] ] : '';
+		}
+
+		$value = wp_parse_args(
```
Crabcyborg, line 150: I think there's a bug here. In my testing, if I'm trying to display this from a view, `$value` is a string, formatted like "First, Middle, Last".

When that gets parsed, it looks like:

```
(
    [first] => 
    [middle] => 
    [last] => 
    [First,_Middle,_Last] => 
)
```

Not quite what we're looking for.

Crabcyborg (commented):
@truongwp this is looking pretty good!

I think we just need to work out all of the weird edge cases and fix a few bugs.

Do you think you could try added some unit tests, basically everywhere we're adding new code?

truongwp:
@Crabcyborg Thanks for testing this PR. Let me check your comments. And I will create some unit tests for new code.

classes/models/fields/FrmFieldName.php
truongwp: I didn't make it work with `string` as `$value`. Do you want to use `,` as the delimiter?

classes/models/fields/FrmFieldName.php
truongwp: Thanks for your suggestion. I think we can even rewrite that method with `explode( '_', $name_layout )`.

classes/models/fields/FrmFieldCombo.php
truongwp: Each subfield has its own label but they are screen reader text. Should we show all of them and hide the label of the field? Or should we add an option for this?

classes/models/fields/FrmFieldName.php
truongwp: It's ok. I will create a method for this.

classes/models/fields/FrmFieldCombo.php
truongwp: Thanks. I missed this case.

classes/models/fields/FrmFieldCombo.php
Crabcyborg: I'm not entirely sure about this one. But I think clicking the main label has to target our first input.

Maybe that means handling the first field's label differently from the others?

classes/models/fields/FrmFieldName.php
Crabcyborg: No, sorry, that's not what I'm asking.

If you try using the `formidable-views` plugin, you can try creating a View. Then insert a shortcode to display your Name field entries and you can see how it doesn't render anything because none of the keys are being mapped to these first/middle/last keys.

The value of `$value` is already imploded by the time it's hitting this function, as it already is.

Addresses work in views. It looks like address fields don't call `wp_parse_args` from inside of this `prepare_display_value` function but from `fill_default_atts` instead. It doesn't seem that calling it from `prepare_display_value` works in every case.

truongwp: Let me try Formidable Views. Thanks for pointing it out.

classes/models/fields/FrmFieldCombo.php
truongwp: @stephywells Can you please give your opinion? Should we make the field label be the first subfield label and hide labels of other subfields?

Crabcyborg:
@truongwp this is looking great!

Next I think we just need to catch all of our edge cases.

1. I couldn't filter for a full name when I was filtering in a view, because the data is different in the database. This is a tougher challenge, I'm not sure if we really need to address it now, but it's something to consider.

2. We need to address import. When you download your entries as CSV, and then import them again, they should be the same as before. Right now if I do this, it looks like:
Screen Shot 2021-04-16 at 10 53 21 AM

Look to `FrmProFieldAddress.php`'s `prepare_import_value` function for a good idea of what I think needs to be done here.

Crabcyborg:
I wonder if we want to have default description values as well.

In the editor, it's showing these "First" and "Last" descriptions.
Screen Shot 2021-04-16 at 11 01 06 AM

But in my form, those are not visible.
Screen Shot 2021-04-16 at 11 01 15 AM

It seems a little misleading. The same First/Last default values should probably be visible in the form.

truongwp:
> I wonder if we want to have default description values as well.
> 
> In the editor, it's showing these "First" and "Last" descriptions.
> Screen Shot 2021-04-16 at 11 01 06 AM
> 
> But in my form, those are not visible.
> Screen Shot 2021-04-16 at 11 01 15 AM
> 
> It seems a little misleading. The same First/Last default values should probably be visible in the form.

Those are not subfield descriptions. Those are subfield labels. I use `frm_description` class to maintain the styling.
In the frontend form, those label is hidden. I take it from the Address field.

classes/models/fields/FrmFieldCombo.php
stephywells: Here's the HTML for the address field (simplified a bit). 
```
<div id="field_r5eso_label" class="frm_primary_label">Address
        <span class="frm_required">*</span>
    </div>
<fieldset aria-labelledby="field_r5eso_label">
<legend class="frm_screen_reader frm_hidden">Address</legend>
<div class="frm_combo_inputs_container">
<div id="frm_field_4609-line1_container" class="frm_form_field form-field ">
	<label for="field_r5eso_line1" class="frm_screen_reader frm_hidden">Line 1	</label>
		<input type="text" id="field_r5eso_line1" value="" name="item_meta[4609][line1]" data-reqmsg="This field cannot be blank." aria-required="true" data-invmsg="Address is invalid" aria-invalid="false" autocomplete="address-line1">
	<div class="frm_description">Line 1</div></div>
<div id="frm_field_4609-line2_container" class="frm_form_field form-field ">
	<label for="field_r5eso_line2" class="frm_screen_reader frm_hidden">Line 2	</label>
		<input type="text" id="field_r5eso_line2" value="" name="item_meta[4609][line2]" data-reqmsg="This field cannot be blank." aria-required="true" data-invmsg="Address is invalid" aria-invalid="false" autocomplete="address-line2">
	</div>
</div>
</fieldset>
```

Based on my accessibility research with groups of fields, this is correct with the fieldset linked to the label with aria-labelledby. Each field can only be labeled by a single label in this type of setup.

So, I think we should exclude the main label since there isn't a field it would apply directly to.

stephywells: That said, it would still be nice as far as UX goes to make it behave like a normal label. Maybe we could add a bit of js for this? On click for that primary label, we could change the focus.

truongwp: @stephywells Thanks. I will go that way.

truongwp:
> I wonder if we want to have default description values as well.
> 
> In the editor, it's showing these "First" and "Last" descriptions.
> Screen Shot 2021-04-16 at 11 01 06 AM
> 
> But in my form, those are not visible.
> Screen Shot 2021-04-16 at 11 01 15 AM
> 
> It seems a little misleading. The same First/Last default values should probably be visible in the form.

Based on Steph's comment for the label of field and subfields. I think we should hide them in the form builder. How do you think?

Crabcyborg:
> Based on Steph's comment for the label of field and subfields. I think we should hide them in the form builder. How do you think?

@truongwp yeah, I think that sounds good! As long as it's consistent, I'm happy 🥳.

classes/models/fields/FrmFieldCombo.php
truongwp: @stephywells I am going to use `.frm_primary_label` in the selector to the field label. If users remove it in the Customize HTML option, this won't work. Is this good enough? I think we can't cover changes from users.

stephywells: Yeah, that's fine. If people change it, we can't control that. The html can be reset by clearing it out and saving, so they can get it back if needed.

classes/controllers/FrmHooksController.php
```diff
@@ -84,6 +84,9 @@ public static function load_hooks() {
 
 		add_filter( 'cron_schedules', 'FrmUsageController::add_schedules' );
 		add_action( 'formidable_send_usage', 'FrmUsageController::send_snapshot' );
+
+		// Make name field work with View.
+		add_filter( 'frm_keep_name_value_array', '__return_true' );
```
Crabcyborg, line 89: @truongwp is this actually used anymore? I don't really see it coming up anywhere else in code.

classes/helpers/FrmCSVExportHelper.php
```diff
@@ -154,9 +154,18 @@ private static function csv_headings( &$headings ) {
 				}
 
 				$fields_by_repeater_id[ $repeater_id ][] = $col;
-			} else {
-				$headings += self::field_headings( $col );
+
+				continue;
+			}
+
+			$field_type_obj = FrmFieldFactory::get_field_factory( $col );
```
Crabcyborg, line 161: It looks like `FrmField::is_combo_field( $col )` could be used here to save some logic.

classes/models/fields/FrmFieldCombo.php
```diff
@@ -0,0 +1,386 @@
...
+
+		if ( ! isset( $args['shortcode_atts'] ) ) {
+			$args['shortcode_atts'] = array();
+		}
+
+		if ( ! isset( $args['errors'] ) ) {
+			$args['errors'] = array();
+		}
+
+		if ( file_exists( FrmAppHelper::plugin_path() . "/classes/views/frm-fields/front-end/{$args['field']['type']}-field/{$args['field']['type']}-field.php" ) ) {
```
Crabcyborg, line 294: This looks a little messy. We should probably at least make the path a variable so the logic is only here once.

classes/models/fields/FrmFieldCombo.php
```diff
@@ -0,0 +1,386 @@
...
+	 *
+	 * @param array $args Arguments. Includes `field`, `sub_field`.
+	 */
+	protected function print_input_atts( $args ) {
+		$field     = $args['field'];
+		$sub_field = $args['sub_field'];
+		$atts      = array();
+
+		// Placeholder.
+		if ( in_array( 'placeholder', $sub_field['options'] ) ) {
```
Crabcyborg, line 312: I get phpcs warnings when in_array is used without the strict `true` passed in the third parameter.

I think that's generally a good idea since it keeps with the rule of using `===` over `==` whenever possible.

Crabcyborg:
@truongwp I've been playing with translating combo field's placeholders and descriptions.

Right now I have https://github.com/Strategy11/formidable-polylang/pull/18/files

```
case 'name':
    $field_data['first_placeholder']  = $field->field_options['first_placeholder'];
    $field_data['middle_placeholder'] = $field->field_options['middle_placeholder'];
    $field_data['last_placeholder']   = $field->field_options['last_placeholder'];
    $field_data['first_desc']         = $field->field_options['first_desc'];
    $field_data['middle_desc']        = $field->field_options['middle_desc'];
    $field_data['last_desc']          = $field->field_options['last_desc'];
    break;
case 'address':
case 'credit_card':
    foreach ( $field->field_options['placeholder'] as $placeholder_key => $placeholder_value ) {
        $field_data[ $placeholder_key . '_placeholder' ] = $placeholder_value;
    }
    break;
```

It looks like the name field uses first_placeholder, middle_placeholder, last_placeholder, but the other combo fields use an array 'placeholder' indexed by each key.

I'm thinking we might want to stay consistent with the other combo fields so that we can treat placeholders the same throughout?

classes/controllers/FrmHooksController.php
truongwp: Without this line, the view will implode the subfields with `,` character.

classes/helpers/FrmCSVExportHelper.php
truongwp: There are some places like this. `FrmField::is_combo_field( $col )` returns boolean value. But I also want to get the field type object. So I don't use that method here to prevent duplicated code.

Crabcyborg: Oh, I missed that! That makes sense then, thanks!

truongwp:
@Crabcyborg When checking the Address field, I also see that it does not use `placeholder` to store an array of placeholders, it uses `xxx_placeholder` instead. I thought that there might be a reason to do that. So I did the same for the combo field. 
Can you please check again and confirm that using the `placeholder` key to store the array of placeholders is safe? I see it's safe with the lite and pro, but I am not sure about other add-ons.

truongwp:
@Crabcyborg Regarding the export issue, I found that `FrmField::get_option( $field, 'id' );` does not always return the correct value. I can update it with `(array) $field;` and then `$field['id']`. But do you think we should update the `FrmField::get_option_in_object()`?

Crabcyborg:
> @Crabcyborg Regarding the export issue, I found that `FrmField::get_option( $field, 'id' );` does not always return the correct value. I can update it with `(array) $field;` and then `$field['id']`. But do you think we should update the `FrmField::get_option_in_object()`?

@truongwp right, get_option is really for the `field_options` values. id is stored directly with the `$field` and is not part of those field options.

What you're doing basically works, but I'd probably just do something like `$id = is_object( $field ) ? $field->id : $field['id'];`.

Crabcyborg:
> @Crabcyborg When checking the Address field, I also see that it does not use `placeholder` to store an array of placeholders, it uses `xxx_placeholder` instead. I thought that there might be a reason to do that. So I did the same for the combo field.
> Can you please check again and confirm that using the `placeholder` key to store the array of placeholders is safe? I see it's safe with the lite and pro, but I am not sure about other add-ons.

I think it might be mapping those placeholder keys to the object, but that's not how it looks in the database.

I just created a new address field and pulled the field_options data from the database to confirm. It looks like:

```
s:11:"placeholder";a:6:{
    s:5:"line1";s:18:"Line 1 Placeholder";
    s:5:"line2";s:18:"Line 2 Placeholder";
    s:4:"city";s:16:"City Placeholder";
    s:5:"state";s:26:"State/Province Placeholder";
    s:3:"zip";s:22:"Zip/Postal Placeholder";
    s:7:"country";s:19:"Country Placeholder";
  }
  s:12:"address_type";s:13:"international";
  s:10:"line1_desc";s:18:"Line 1 Description";
  s:10:"line2_desc";s:18:"Line 2 Description";
  s:9:"city_desc";s:16:"City Description";
  s:10:"state_desc";s:26:"State/Province Description";
  s:8:"zip_desc";s:22:"Zip/Postal Description";
  s:12:"country_desc";s:19:"Country Description";
```

description saves like x_desc but not placeholder.

truongwp:
Ok. I need to check this again. Thanks for pointing out.

truongwp:
@Crabcyborg I managed to use `placeholder` to store the array of subfield placeholders. Do you think we should do the same thing for the desc of subfields and maybe other settings in the future?

Crabcyborg:
> @Crabcyborg I managed to use `placeholder` to store the array of subfield placeholders. Do you think we should do the same thing for the desc of subfields and maybe other settings in the future?

@truongwp thanks for updating that! I think we're good to just do this with placeholders. My main focus is backward compatibility with the previous combo fields.

I think changing the descriptions should just complicate things.

I think there are a few more things we'll want to change for consistency though. I'm going to go through these fields and my polylang update again, and figure out what else we need to do here.

I think we're close though!

Crabcyborg:
@truongwp I just tested this out with my polylang stuff and it works great now. Thanks for changing the placeholders!

I just made a small update to your branch to set the description fields as translatable so that the polylang update can work the same for all types, so you'll need to do a pull as well - https://github.com/Strategy11/formidable-forms/pull/390/commits/7edee4963707ed1bd7c5c67bcf85eafc41bf5e02

Another thing, I think we should probably hide the "Middle" area unless you have the "First Middle Last" layout selected. Otherwise it's just taking up space.
Screen Shot 2021-04-27 at 11 54 24 AM

truongwp:
@Crabcyborg Thanks for your commitment. But I think we should move it to the FrmFieldCombo class and loop through the subfields to add them. How do you think?

Crabcyborg:
> @Crabcyborg Thanks for your commitment. But I think we should move it to the FrmFieldCombo class and loop through the subfields to add them. How do you think?

@truongwp oh yeah, for sure. As long as these strings are returned when translatable_strings is called on the name field, they will be translatable.

So if you want to loop through the subfields and do that in the combo field that would really clean it up and make this easier to manage.

classes/models/fields/FrmFieldCombo.php
```diff
@@ -0,0 +1,414 @@
...
+
+		return $errors;
+	}
+
+	/**
+	 * Gets export headings.
+	 *
+	 * @return array
+	 */
+	public function get_export_headings() {
```
Crabcyborg, line 383: One thing I'm noticing here is that the "Middle" column is always included even when the layout doesn't include a middle name. This probably makes sense to exclude in that case.

It also seemed odd that it was an option during the import, but I thinking fixing that here might fix it there as well.

truongwp: Ok. Since the user can choose the column to import to a field, I think we can remove the Middle column if it is unused. Let me update this.

classes/models/fields/FrmFieldCombo.php
```diff
@@ -135,8 +135,9 @@ protected function extra_field_opts() {
 	public function show_after_default( $args ) {
 		$field         = (array) $args['field'];
 		$default_value = $this->get_default_value();
+		$sub_fields    = $this->get_processed_sub_fields();
```
Crabcyborg, line 138: @truongwp I think we need to include the middle name field, but hide it if middle name isn't included in the layout options, because the layout option could be changed, and then we need to see the middle name options.

More of a JavaScript issue.

classes/models/fields/FrmFieldName.php
```diff
@@ -0,0 +1,144 @@
+<?php
+/**
+ * Name field
+ *
+ * @package Formidable
+ * @since 4.10.01
```
Crabcyborg, line 6: I see a couple of `4.10.01` and some `4.10.02`.

I plan to do 4.10.02 today and I don't intend on including this update yet.

But I think saying it will be 4.10.03 is a safe bet. Or you can set it to something like `@since x.x`. I haven't started automating that yet or anything but Steph has with BD and it sounds like our best option for making sure our `@since` statements are accurate.

truongwp: Ok. I will replace with `4.10.03`.

classes/models/fields/FrmFieldCombo.php
truongwp: If we want to use JS to show/hide the Middle name options, we will also need to do the same thing for the view in the form builder

classes/controllers/FrmHooksController.php
Crabcyborg: @truongwp ohh, I see how this works now. Since you can't really search the code for `frm_keep_name_value_array` this can sort of hard to trace. The comment is helpful, but it doesn't really get down to where in the code or why this is needed.

I think it might be worth mentioning that `FrmProContent::replace_single_shortcode` applies this filter like `'frm_keep_' . $field->type . '_value_array'` or something.

classes/models/fields/FrmFieldCombo.php
Crabcyborg: Hm, that would be pretty nice if it did. Most of our fields try to keep synced in the form builder as well. I didn't totally notice that it wasn't already doing that.

truongwp: Since we are using one view for both backend and frontend, we will need to check if is in the form builder, then print the HTML of the middle name and use JS to show/hide it. Do you think that is good or we should separate into 2 view files or use AJAX to load the middle name?

Crabcyborg: > Since we are using one view for both backend and frontend, we will need to check if is in the form builder, then print the HTML of the middle name and use JS to show/hide it. Do you think that is good or we should separate into 2 view files or use AJAX to load the middle name?

Hm, I'm not really sure. Is it possible to just render it all, then hide/show with JavaScript? What else needs to change? Just the frm4/frm6 classes?

truongwp: If First Last is selected, Middle name is not rendered. We get processed subfields, not all subfields here https://github.com/Strategy11/formidable-forms/pull/390/files#diff-8228809da5d9d1a59e3a4b1e878d137a3e66a20e32eaa914d56dc643da59486bR285

truongwp: I think we can check if is in the form builder, we will render all fields.

Crabcyborg: > I think we can check if is in the form builder, we will render all fields.

Sounds like you have a plan 🙂.

classes/controllers/FrmHooksController.php
truongwp: Sure. Let me mention that.

classes/models/fields/FrmFieldName.php
stephywells: Would 4.11 make sense for this?

Crabcyborg: @stephywells ohh, good call.

I noticed that my name field was totally broken when I switched back to master, so this isn't really backward compatible.

So changing the version probably makes some sense.

truongwp:
@Crabcyborg I have pushed a commit to fix the following issue. This issue occurs with the Address field as well.


https://user-images.githubusercontent.com/19748318/116582664-ca7aa080-a93f-11eb-8032-9246c4d95cd8.mov

Crabcyborg:
> @Crabcyborg I have pushed a commit to fix the following issue. This issue occurs with the Address field as well. I am not sure there are any issues related to this.

@truongwp oh, nice catch. That might be a css targeting issue.

I think I've seen a similar one, where the styles get lost because it's targeting :last-child and then a new child is appended temporarily.

Or even if it's something else, this would be a good thing to try fixing.

truongwp:
@Crabcyborg @stephywells I have updated the icon of the Name field. Please check when you have time.

classes/helpers/FrmCSVExportHelper.php
```diff
@@ -154,9 +154,18 @@ private static function csv_headings( &$headings ) {
 				}
 
 				$fields_by_repeater_id[ $repeater_id ][] = $col;
```
Crabcyborg, line 156: @truongwp it looks like the export gets a little funny if you try to export a name field that is inside of a repeater.

There's repeater logic above here, and your name logic is separate. I think we need to include name logic in the repeater logic though.

My export (one entry, with two repeater entries for one name field) looks like:
Screen Shot 2021-05-04 at 2 32 35 PM

Crabcyborg:
> @Crabcyborg @stephywells I have updated the icon of the Name field. Please check when you have time.

@truongwp it looks a little small. There's an svg style setting it to 20x20 but the SVG looks like it should be 24x24.

Also, Steph talked about moving it before the phone field but it looks like it's still the last one.

truongwp:
@Crabcyborg I updated the icon, position, and export columns. I am working on the import process in the pro plugin to make the import work with the combo + repeater field.

jairoprez (approved):
I did test for this new name field and overall it seems to work quite well 🎉 

Specifically these were the functions I tested:

- layout option: "first last", "last first", "first middle last".
- placeholder, default value, and description for each field.
- shortcode options: `[25]` `[25 show=first]` `[25 show=last]` `[25 show=middle]`
- form duplication with a name field
- name field with repeaters
- import/export form and entries

@truongwp @Crabcyborg Nice job! 💪 

If I missed testing any function of this new field, please just let me know.

truongwp:
@jairoprez Thanks for testing. What is the result after you importing entries? 

Currently, Importing entries with name field outside of repeater field will work with the code in this pull request https://github.com/Strategy11/formidable-pro/pull/2991
Importing entries with the name field inside the repeater field is not working correctly now. I am working on the above pull request to make it work.

jairoprez:
> @jairoprez Thanks for testing. What is the result after you importing entries?
> 
> Currently, Importing entries with name field outside of repeater field will work with the code in this pull request [Strategy11/formidable-pro#2991](https://github.com/Strategy11/formidable-pro/pull/2991)
> Importing entries with the name field inside the repeater field is not working correctly now. I am working on the above pull request to make it work.

Hey @truongwp the import of the entries worked well for me, but it was because I did it based on the xml file https://recordit.co/GM1dsvPx5q, after your comment I realized that I did not do tests using the csv file which if it is failing https://recordit.co/QMJgsY3Wur as you indicated in your comment.

I did tests importing the entries via a csv and using this PR https://github.com/Strategy11/formidable-pro/pull/2991 that you pointed out, and it worked but not 100% https://recordit.co/edqGdTd8sL since some records were not imported properly.

This is the form and entries that I am using to do the tests, maybe it can be useful for you to test with the same information as me: [name-field-form.xml.zip](https://github.com/Strategy11/formidable-forms/files/6440733/name-field-form.xml.zip)

Thanks for pointing this out regarding import, I really overlooked the csv import part, if you have any questions please just let me know.

truongwp:
@jairoprez Thank you. I will check your test.

jairoprez:
Hey @truongwp I temporarily removed the `need qa` label from this PR, when you submit new changes for testing please add it again.

truongwp:
@jairoprez Sure. I will add it again with it is ready.

truongwp:
@jairoprez I have fixed the import issue in both lite and pro version. Now it works with repeater field as well.

@Crabcyborg Do you want to take a look before I add the `needs qa` tag? I added the field key to the heading of csv field in case there are some similar headings. I checked the form from Jairo and saw that one field have Name Layout is set to empty string. So I add one more check to make sure it use the default option if the value is empty.

In this PR https://github.com/Strategy11/formidable-pro/pull/2991. What I do is replicating the structure of import values when using repeater field.

jairoprez:
> @jairoprez I have fixed the import issue in both lite and pro version. Now it works with repeater field as well.
> 
> @Crabcyborg Do you want to take a look before I add the `needs qa` tag? I added the field key to the heading of csv field in case there are some similar headings. I checked the form from Jairo and saw that one field have Name Layout is set to empty string. So I add one more check to make sure it use the default option if the value is empty.
> 
> In this PR [Strategy11/formidable-pro#2991](https://github.com/Strategy11/formidable-pro/pull/2991). What I do is replicating the structure of import values when using repeater field.

Nice job! @truongwp - import via csv is working great, with and without repeater field 💪

jairoprez (approved):

truongwp:
@jairoprez Thank you.

Crabcyborg:
@truongwp I just found a funny sort-of issue.

I was playing with using HTML (an image tag) as a name, and it was sanitized, which is great since a name field should only accept a name.

Where my issue came up is when I made this field required.

It adds these empty arrays to my meta value and validates even though the fields are all empty after the sanitizing happens.
Screen Shot 2021-05-13 at 10 46 29 AM

I've seen this come up with our checkbox fields before when you select Other and then leave the other option blank.

My solution was to check for this in the `get_value_to_save` function with a little check on the array:

```
if ( 0 === count( $value ) ) {
	$value = '';
}
```

I think we want this here as well so that the field is actually empty. Or maybe the validation that happens early would prevent this in another way, because this might not actually catch the required check either.

This is definitely not a major issue either, but I think fixing it ahead of time would be great so we can avoid it coming up as a ticket.

truongwp:
@Crabcyborg It does not help if I fill the First Name as normal string and Last Name as HTML. The `validate()` method seems to run before `get_value_to_save()` method. 

We can run `$this->sanitize_value()` before run the validate check here https://github.com/Strategy11/formidable-forms/blob/5eba8b38f5b5381fd586c0b2600633007e2b40e4/classes/models/fields/FrmFieldCombo.php#L386

How do you think?

We can even do an invalid characters validation if we want.

Crabcyborg:
@truongwp yeah I'm aware of the sort-of funny order it goes in.

That's why I mention the `FrmProEntryMeta::check_for_required_field_after_sanitizing` function.

We basically check:

validate
sanitize
(validate again for some fields that might have been cleared)

It's not perfect, but it at least catches this issue.

truongwp:
@Crabcyborg Instead of validating, then sanitizing, then validating again, why don't we call the sanitization method before all checks inside the validation? With this solution, we just need to add one line in the validate() method. And I don't see any downside of this solution versus calling validation again after sanitizing.

Crabcyborg:
@truongwp are you thinking of doing this for names only?

If we can isolate, maybe. But I think changing the order here is really likely to add a unexpected bug in.

A lot happens after validating happens. Ideally we stop if things don't validate as early as possible, even if that means validating some fields that might go through extra sanitizing.

I think this is the less common case. I don’t think  many of our fields have custom sanitizing logic which is why this check after is only happening for 3 types right now.

truongwp:
@Crabcyborg I am thinking of doing this for Combo field and its children or for Name field only. Maybe the best way is calling sanitization before validating. The sanitization method using for Name field is simple.

In the future, we can add another validation check for invalid characters. It will be clearer for the users.

Crabcyborg:
> @Crabcyborg I am thinking of doing this for Combo field and its children or for Name field only. Maybe the best way is calling sanitization before validating. The sanitization method using for Name field is simple.
> 
> In the future, we can add another validation check for invalid characters. It will be clearer for the users.

@truongwp Yeah, we

truongwp:
@Crabcyborg Ok. I will do this for the Name field only because not every child field needs this.

Isaac110:
it doesn't auto generate the descriptions for first, last and middle name.
it shows the middle field settings right after you create it, even though the middle field isn't showing.
1.if you make a name field then go into advanced than you change the middle name's placeholder or description flied setting.
2. switch the field type to "first middle last" and one of those three won't show up.

truongwp:
Hi @Isaac110. Thanks for taking a look at this.

> it doesn't auto generate the descriptions for first, last and middle name.

Yeah. We don't generate the descriptions for the subfields. But this may be a good idea. How do you think @Crabcyborg?

> it shows the middle field settings right after you create it, even though the middle field isn't showing.
1.if you make a name field then go into advanced than you change the middle name's placeholder or description flied setting.
2. switch the field type to "first middle last" and one of those three won't show up.

I am able to replicate this issue. I am fixing it now. Thanks.

truongwp:
@Isaac110  @Crabcyborg  I fixed those issues.

Crabcyborg:
> Yeah. We don't generate the default descriptions for the subfields. But this may be a good idea. How do you think @Crabcyborg?

@truongwp hm, if you add an Address field it does add default descriptions for City, State, Zip, Country but not for Lines 1 or 2.

Credit cards don't add any though.

I'm thinking we should include the defaults for names as well. What do you think @stephywells?

stephywells:
It seems confusing without those descriptions. I think those should be there since that would be the most common use case. Let's go ahead and add those by default.

Crabcyborg:
@truongwp it looks like name fields don't have a conditional logic option? Any idea why? I'm not totally sure how this is enabled, but we should be able to have conditional names.

Crabcyborg:
@truongwp it looks like the issue Toyin pointed out is a name field issue.

If I create a simple form like:
Screen Shot 2021-05-19 at 4 36 33 PM

A section with a name field where the section is conditionally hidden and the name is required.

The result
Screen Shot 2021-05-19 at 4 35 50 PM

It should allow for hidden fields to pass required validation.

stephywells:
- [ ] Let's remove the Name field from the reports page. We aren't including other combo fields there. It's defaulting to a bar graph that doesn't make sense for the field type.
- [ ] The conditional logic is missing from the name field settings. This is added with a class in Pro like this: https://github.com/Strategy11/formidable-pro/blob/master/classes/models/fields/FrmProFieldPhone.php

Crabcyborg:
Found another funny use case, text calculations.

Screen Shot 2021-05-19 at 4 43 32 PM

I think a text calculation should be able to populate with the full name value.

truongwp:
@stephywells @Crabcyborg Thanks for testing. I will check those.

I have a question about Conditional Logic. In other fields, should we show subfields in the dropdown or just the whole Name field?

Crabcyborg:
> I have a question about Conditional Logic. In other fields, should we show subfields in the dropdown or just the whole Name field?

@truongwp ohh, both would probably give us the most functionality.

I think either would at least allow for us to write the conditions we want.

Full name is likely the option users would expect to see.

stephywells:
> I have a question about Conditional Logic. In other fields, should we show subfields in the dropdown or just the whole Name field?

Let's save this part for a different release. It looks like we haven't set this up for any of the combo fields yet.

jairoprez:
I have done several tests based mainly on these criteria:

- Name with with Pro active and disabled
- Check the address field in the current version with this code in place for the name field
- Exporting from and importing into a name field when inside or outside a repeater, `csv, xml, forms, entries, views and styles`
- Test all the settings in the name field

@truongwp  the only bug I have found so far is related to the name field layout setting, specifically the `Last First layout`. The order of this layout changes to `First Last` when the form is updated https://recordit.co/wip3hGPVg9 this only happens in the form builder area.

Testing for this point is still ongoing, I'll leave a new comment when I'm done.
- Test using the name field in form actions, form success message, view...

truongwp:
@jairoprez Thanks. I have fixed the issue you mentioned.

truongwp:
@stephywells For the Stripe issue, I can't find a way to make it work without adding code to the Stripe add-on. 

In the Stripe add-on, we can check if is the Name field here, then add `show="first"` and `show="last"` to the shortcode.

Crabcyborg:
@stephywells the Stripe options are split up by two names and none of these options seem to go through much of a filtering process to be able to break the names up without writing code directly in the formidable-payments module.

Screen Shot 2021-05-21 at 9 57 28 AM

Since it is broken up into two names, I don't think anyone would just expect a full name field to work.

So I think we just need to add this update shortly and release a Stripe update.

Crabcyborg:
> @stephywells For the Stripe issue, I can't find a way to make it work without adding code to the Stripe add-on.
> 
> In the Stripe add-on, we can check if is the Name field here, then add `show="first"` and `show="last"` to the shortcode.

@truongwp thanks for looking into it!

jairoprez:
> @jairoprez Thanks. I have fixed the issue you mentioned.

I did tests and it works as expected 💪

PR #391, opened 2021-04-18
Add api_forms_loaded to frm_vars to load the main js file

truongwp:
This is the requirement for https://github.com/Strategy11/formidable-api/pull/85


Crabcyborg:
I'm going to close this one as we have a fix that works in just the API plugin 🥳.

PR #392, opened 2021-04-26
add a fallback for the_content if get_template_part does not support page

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/1880

I tried this out with Customizr only as the other plugins all seem to be paid ones that I don't have access to.

When there is no support a template part `get_template_part` actually returns `false`, so we can try calling a fallback method for rendering our form.

I skimmed over the customizr code quite a bit, and I think the right option is to just call `the_content()` to trigger the render event. I also added `get_header()` and `get_footer()` calls as well, to make it a little more page-like. But it still feels pretty ugly with Customizr.

Since this fix is so generic and simple, I think it should really work with every conflicting plugin.

Instead of nothing, I get a pretty plain, not-very-styled form, but it's better than nothing - it works!

Screen Shot 2021-04-26 at 1 46 28 PM

Maybe there's a way to get some theme styling in here, but I couldn't really figure out what hook would work.

@stephywells do you think this is an acceptable solution? I couldn't really figure out how to get better theme support without writing really custom code specific for the themes.

Without the header/footer calls, it looks more like:
Screen Shot 2021-04-26 at 1 25 30 PM

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -314,7 +314,21 @@ class FrmFormsController {
 		add_action( 'loop_no_results', 'FrmFormsController::show_page_preview' );
 		add_filter( 'is_active_sidebar', '__return_false' );
 		FrmStylesController::enqueue_css( 'enqueue', true );
-		get_template_part( 'page' );
+
+		if ( false === get_template_part( 'page' ) ) {
+			self::fallback_when_page_template_part_is_not_supported_by_theme();
+		}
+	}
+
+	/**
+	 * Not every theme supports get_template_part( 'page' ).
+	 * When this is not supported, false is returned, and we can handle a fallback.
+	 */
+	private static function fallback_when_page_template_part_is_not_supported_by_theme() {
+		if ( have_posts() ) {
+			the_post();
+			the_content();
+		}
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


stephywells (approved):
@Crabcyborg this seems like a decent fix to me. It's definitely better than a blank page. I think we should try this for a while and see what feedback we get based on the themes being used. We have usage tracking, but we hit database limits and now I can't even log in to look at it! I need to get that fixed so we can get info on the top themes people are using.

jairoprez (approved):
I did tests on the following themes:

- Customizr: https://recordit.co/rvX53ZOShR (It works well)
- Twenty Twenty: https://recordit.co/S7QAzIhQW5 (It works well)
- hacked version of Enfold: https://recordit.co/hvcJu4Pzs6 (It doesn't seem to work properly, but I'm not sure if it has something to do with it being a hacked version of the theme.)

Crabcyborg:
> I did tests on the following themes:
> 
>     * Customizr: https://recordit.co/rvX53ZOShR (It works well)
> 
>     * Twenty Twenty: https://recordit.co/S7QAzIhQW5 (It works well)
> 
>     * hacked version of Enfold: https://recordit.co/hvcJu4Pzs6 (It doesn't seem to work properly, but I'm not sure if it has something to do with it being a hacked version of the theme.)

@jairoprez ha, at least it looks like it still loads at least!

We'll try to improve this over time. Maybe we'll get a legitimate copy of Enfold at some point 😛.

Thanks for testing!

PR #393, opened 2021-04-27
Correct resource URLs when loading css via AJAX

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/2972


Crabcyborg:
@truongwp I think this is looking pretty good!

Once this saves, it looks like it's loaded through a GET request with `wp_enqueue_style( 'frm-custom-theme', admin_url( 'admin-ajax.php?action=frmpro_css' ), array(), $version );`.

Inside of `customer_theme.css.php` I see:

```
if ( ! isset( $saving ) ) {
	header( 'Content-type: text/css' );

	if ( isset( $css ) && $css ) {
		echo strip_tags( $css, 'all' ); // WPCS: XSS ok.
		die();
	}
}
```

I feel like this `ob_` stuff might get in the way here since there is a die statement. I think we need to check the `$css = get_transient( 'frmpro_css' );` value before this is called and just let it work the previous way in this case.

truongwp:
@Crabcyborg Thanks for checking. I think we still need to replace the transient or the user will need to clear the old transient to make it work.

Crabcyborg:
> @Crabcyborg Thanks for checking. I think we still need to replace the transient or the user will need to clear the old transient to make it work.

Maybe we need a migration or something.

I think this transient is really important here.

I believe the actual stylesheet is being saved as a transient, and being accessed here as an optimization so that it doesn't need to be constantly processed.

If we process the transient we're working backwards and eliminating the optimization.

So this update should:
- Make sure that the transient is set to the str_replaced output.
- Not process a str_replace on the transient.

truongwp:
@Crabcyborg I think we just need to clear that transient in migration. That can affect the first load after migration but I think that is not a big issue.

Crabcyborg:
> @Crabcyborg I think we just need to clear that transient in migration. That can affect the first load after migration but I think that is not a big issue.

@truongwp that sounds like a really good simple solution!

truongwp:
@Crabcyborg I updated this pull request. Please check when you have time.

Crabcyborg:
@truongwp this still isn't replacing the transient though.

The str_replace only happens after the include, but when there is a transient, it dies and never does the replace.

I've tested this and confirmed this is definitely still showing the `../` when I visit `/wp-admin/admin.php?page=formidable-styles` when it makes a request to `/wp-admin/admin-ajax.php?action=frmpro_css`.

I think we might want to move all of this to `FrmStyle::save_settings`. Maybe as a part of `FrmStyle::get_css_content`.

Then the migration could probably just str_replace the existing transient value if it exists instead of removing it. That or we actually call save_settings here.

truongwp:
@Crabcyborg Let me check this again. It's a bit confusing now.

truongwp:
@Crabcyborg I think we don't need to replace the transient. I made a migration here https://github.com/Strategy11/formidable-forms/pull/393/files#diff-192da826fa175610394eb9d3a6047738f05b8f4522bc216c8b8ccd43fe6b226eR283

After updating, the transient is deleted. The new CSS after replacement will be stored in the transient. Maybe the migration does not work on your side.

Crabcyborg:
@truongwp what happens when you load `/wp-admin/admin-ajax.php?action=frmpro_css` after trying to save a custom style from `/wp-admin/admin.php?page=formidable-styles`?

A new transient is just going to get made again, and when it does, this will not work again.

The migration only fixes this until you save another style and a new transient gets made because the transient still uses the `../`

truongwp:
I see that. Maybe we also need to add the replacement here https://github.com/Strategy11/formidable-forms/blob/master/classes/models/FrmStyle.php#L167

Let me check it.

truongwp:
@Crabcyborg I updated. Please check it.

classes/controllers/FrmStylesController.php
```diff
@@ -451,10 +451,40 @@ public static function load_css() {
...
+	 * Replaces relative URL with absolute URL.
+	 *
+	 * @since 4.11.03
+	 *
+	 * @param string $css CSS content.
+	 * @return string
+	 */
+	public static function replace_relative_url( $css ) {
+		$plugin_url = FrmAppHelper::plugin_url();
+		return str_replace( // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
Crabcyborg, line 473: Since you aren't echoing anymore this phpcs exception shouldn't be necessary here.

classes/controllers/FrmStylesController.php
```diff
@@ -451,10 +451,40 @@ public static function load_css() {
 	public static function load_saved_css() {
 		$css = get_transient( 'frmpro_css' );
 
+		ob_start();
 		include( FrmAppHelper::plugin_path() . '/css/custom_theme.css.php' );
+		$output = ob_get_clean();
+
+		echo self::replace_relative_url( $output );
```
Crabcyborg, line 458: You probably do need a phpcs exception here. Looks like phpcs is complaining about it.

```
FILE: ...rms/formidable-forms/classes/controllers/FrmStylesController.php
----------------------------------------------------------------------
----------------------------------------------------------------------
 458 | ERROR | All output should be run through an escaping function
     |       | (see the Security sections in the WordPress Developer
     |       | Handbooks), found 'self'.
     |       | (WordPress.Security.EscapeOutput.OutputNotEscaped)
----------------------------------------------------------------------
```

https://github.com/Strategy11/formidable-forms/pull/393/checks?check_run_id=2850176361

Crabcyborg:
> @Crabcyborg I updated. Please check it.

Thanks @truongwp other than the phpcs notes I mention, I think this is good to go 🎉.

classes/controllers/FrmStylesController.php
truongwp: Thank you. I forgot to check the build errors :D

PR #398, opened 2021-04-30
debounce change styling ajax action

Crabcyborg:
I noticed that when you start playing with a color picker in the style editors, that a ton of AJAX requests happen. Dragging around for a second triggers about 18 for me.

These are pretty heavy, and server limits cause this to start happening all out of order with a ton of delays to get the latest response.

This is the perfect opporuntity to introduce some debouncing 🏎.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -6534,6 +6534,17 @@ function frmAdminBuildJS() {
 		return object;
 	}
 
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Well that's cool. I didn't know this was an issue, or that there's a quick fix. Awesome!

PR #399, opened 2021-04-30
Guarantee unique action id

Crabcyborg:
I noticed a bug the other day when my AJAX request to create an action took a moment, and I clicked it again.

It ended up creating two action elements with the same `actionId` value. Then causes some weird behaviours, like when I go to delete the bottom action, it actually goes and deletes the top one instead.

I'm adding an extra tracked variable here that makes sure that the value returned is always higher than the value previously returned.

I also refactored the js slightly so that the variables are all in one declaration in the addFormAction function.


stephywells (approved):
Nice catch. And nice, simple fix too.

PR #400, opened 2021-05-01
Add "free template" opportunity to inbox

truongwp:
This fixes issue https://github.com/Strategy11/formidable-pro/issues/2885

### DIFF
--- a/classes/controllers/FrmInboxController.php
+++ b/classes/controllers/FrmInboxController.php
@@ -46,6 +46,7 @@ class FrmInboxController {
 	public static function inbox() {
 		FrmAppHelper::include_svg();
 		self::add_tracking_request();
+		self::add_free_template_message();
 
 		$inbox    = new FrmInbox();
 		$messages = $inbox->get_messages( 'filter' );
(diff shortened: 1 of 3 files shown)


truongwp:
@stephywells I have a commitment to do the compatibility thing. So you can take a look.

js/formidable_admin.js
```diff
@@ -6009,9 +6033,16 @@ function frmAdminBuildJS() {
 			}
 		});
 
-		urlParams = new URLSearchParams( window.location.search );
 		if ( urlParams.get( 'triggerNewFormModal' ) ) {
 			triggerNewFormModal();
+
+			if ( isShowFreeTemplatesFormFirst ) {
+				firstLockedTemplate = jQuery( 'li.frm-locked-template.frm-free-template:eq(0)' );
```
Crabcyborg, line 6040: https://api.jquery.com/eq-selector/

`As of jQuery 3.4, the :eq pseudo-class is deprecated. Remove it from your selectors and filter the results later using .eq().`

I know I had to replace some lines like this. They suggest using `.eq` now.

classes/controllers/FrmInboxController.php
```diff
@@ -99,4 +100,33 @@ private static function add_tracking_request() {
...
+			return;
+		}
+
+		$link = admin_url( 'admin.php?page=formidable&triggerNewFormModal=1&free-templates=1' );
+
+		$message = new FrmInbox();
+		$message->add_message(
+			array(
+				'key'     => 'free_templates',
+				'message' => 'Just add your email address and you \'ll get a code for 10+ free form templates.',
```
Crabcyborg, line 125: It looks like there is a space here that shouldn't be in `you'll`. But I doubt we want to use that word anyway.

I think even just `Add your email address to get a code for 10+ free form templates.` covers this with fewer words.

js/formidable_admin.js
```diff
@@ -5798,32 +5807,38 @@ function frmAdminBuildJS() {
 			setTemplateCount( $li.closest( '.accordion-section' ).get( 0 ) );
 		});
 
+		showFreeTemplatesForm = function( $el ) {
+			formContainer = document.getElementById( 'frmapi-email-form' );
```
Crabcyborg, line 5811: It doesn't look like there is actually a var declaration for this after it was copied and pasted around a bit.

Crabcyborg:
@truongwp thanks!

This looks good other than a few notes I added. I haven't tested it yet either, but I think this is probably already good to go soon.

truongwp:
@Crabcyborg Thanks for checking this. Let me check your notes.

truongwp:
@Crabcyborg I have fixed the. Please check when you have time.

Crabcyborg (approved):
I think this looks good! Thanks @truongwp.

truongwp:
@Crabcyborg @jairoprez Thank you.

PR #401, opened 2021-05-02
only remove frm_logic_class when wp_is_json_request

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3001

### DIFF
--- a/classes/controllers/FrmSimpleBlocksController.php
+++ b/classes/controllers/FrmSimpleBlocksController.php
@@ -140,6 +140,12 @@ class FrmSimpleBlocksController {
 
 		$form = FrmFormsController::get_form_shortcode( $params );
 
-		return str_replace( ' frm_logic_form ', ' ', $form ); // prevent the form from hiding
+		if ( wp_is_json_request() ) {
+			// remove fade on load when /wp-json/wp/v2/block-renderer/formidable/simple-form is called.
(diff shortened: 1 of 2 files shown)


jairoprez (commented):
Hey Super @Crabcyborg I did some testing even without the recording tool turned on so as not to affect internet speed and it seems that hidden fields are still showing in a flash before hiding when loading the page https://recordit.co/2l70L76GY1

Crabcyborg:
> Hey Super @Crabcyborg I did some testing even without the recording tool turned on so as not to affect internet speed and it seems that hidden fields are still showing in a flash before hiding when loading the page https://recordit.co/2l70L76GY1

@jairoprez hm, are you seeing the fading on any forms?

This is the default behaviour. You might not have the fade option turned on (in global settings):
Screen Shot 2021-05-03 at 9 42 30 AM

jairoprez:
> @jairoprez hm, are you seeing the fading on any forms?
> 
> This is the default behaviour. You might not have the fade option turned on (in global settings):

Ah sorry @Crabcyborg , you were right, I had the fade in option turned off :sweat_smile:, after turning it on everything worked as expected, thanks for pointing that out. https://recordit.co/rHgFoWDShR

jairoprez (approved):

stephywells (approved):
Nice solve. Looks like a great way to solve it.

PR #402, opened 2021-05-06
Add a JS Spam check

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2992

I feel like I might need to move some stuff around still, but this is working pretty good.

And there are a lot of filters here. Some of these likely need more options.

I feel like I'm probably missing some use cases too. I'm adding to the form attributes, but there's also this `! isset( $include_form_tag ) || $include_form_tag` logic that skips that, so that's probably uncovered at least.

I also refactored some of the spam checks. It doesn't make sense that we're checking one spam check, marking the error, then checking others and possibly overwriting errors, so I've put everything into `elseif`s.

**Do we need more exceptions?**
- Does this prevent submissions via API?
- Does this prevent import?

**New filters to test and document**
- frm_run_antispam (if it's on or not)
- frm_process_antispam (for post-processing the result. you could turn it off, but handle an error with custom code).
- frm_form_token_check_before_today (extra times to check for tokens before today)
- frm_form_token_check_after_today (extra times to check for tokens after today)

I've also done this for the ajax file upload [here](https://github.com/Strategy11/formidable-pro/pull/3014) but it needs more work to look good.


jairoprez (approved):
@Crabcyborg I have done tests with the new filters and they seem to work fine for me:

- `frm_run_antispam` :white_check_mark:
- `frm_process_antispam` :white_check_mark:
- `frm_form_token_check_before_today` :white_check_mark:
- `frm_form_token_check_after_today` :white_check_mark:

I'm still completing the documentation for these new hooks, when I'm done I'll share the links for you to take a look.

classes/models/FrmAntiSpam.php
```diff
@@ -0,0 +1,281 @@
...
+			return '';
+		}
+
+		// If the user is an admin, return text with a link to support.
+		// We add a space here to seperate the sentences, but outside of the localized
+		// text to avoid it being removed.
+		return ' ' . sprintf(
+			// translators: %1$s start link, %2$s end link.
+			esc_html__( 'Please check out our %1$stroubleshooting guide%2$s for details on resolving this issue.', 'formidable' ),
+			'<a href="https://formidableforms.com/knowledgebase/">',
```
stephywells, line 277: Let's point this to https://formidableforms.com/knowledgebase/add-spam-protection/

We can plan on adding extra documentation for this spam prevention option there.

stephywells (approved):
Awesome!! I think we could use some heavy QA on this.

@jairoprez can you tell us which ways you tested this? Maybe a list of what you tried would be helpful?

garretlaxton (approved):
@Crabcyborg's last push solved the previous JavaScript loading errors.

The antispam is working as should throwing the `Your entry appears to be spam!` error when the honeypot value is filled. I seen the Akismet checks added as well so I setup that plugin and the checks worked as intended.

classes/views/frm-forms/spam-settings/honeypot.php
```diff
@@ -0,0 +1,11 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+?>
+<tr>
+	<td colspan="2">
+		<input id="honeypot" type="checkbox" name="options[honeypot]" <?php checked( $values['honeypot'], 1 ); ?> value="1" />
+		<label for="honeypot"><?php esc_html_e( 'Use Honeypot to check entries for spam', 'formidable' ); ?></label>
```
stephywells, line 9: Is there a way to use the strict option without custom code? We talked about this being a dropdown with three options like:
- Off
- Basic
- Strict

classes/views/frm-forms/spam-settings/antispam.php
```diff
@@ -0,0 +1,11 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+?>
+<tr>
+	<td colspan="2">
+		<input id="antispam" type="checkbox" name="options[antispam]" <?php checked( $values['antispam'], 1 ); ?> value="1" />
+		<label for="antispam"><?php esc_html_e( 'Validate for spam with a JavaScript token', 'formidable' ); ?></label>
```
stephywells, line 9: How about something like "Check entries for spam using JavaScript"?

stephywells (commented):
Awesome! This looks great!

Crabcyborg:
I didn't want to keep updating two branches so I've made the most recent updates against the `4.11_preview` branch and not here.

I'm just going to close this branch. Any new issues can be addressed there.

But I've changed the setting to a dropdown and renamed the antispam option as Steph requested.

PR #403, opened 2021-05-06
Fix PHP notice with array_diff_assoc() when exporting form

truongwp:
This fixes issue https://github.com/Strategy11/formidable-pro/issues/2999

### DIFF
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -1202,24 +1202,11 @@ class FrmXMLHelper {
 	 * @since 3.06
 	 */
 	private static function remove_defaults( $defaults, &$saved ) {
-		// Compare the array.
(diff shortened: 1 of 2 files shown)


truongwp:
@Crabcyborg Thanks for checking. Maybe we need one more loop with this `$array_saved = array_filter( $saved, 'is_array' );`.

Or we can loop through `$defaults` to remove the same entries from `$saved`, without using `array_diff_assoc()`. 

How do you think?

Crabcyborg:
> How do you think?

@truongwp yeah, that sounds like it should do the trick!

truongwp:
@Crabcyborg I have updated it. Please check when you have time.

Crabcyborg (approved):
Thank you @truongwp this looks good to me!

jairoprez (approved):
I did some testing and PHP notice has been fixed http://recordit.co/RG7W2GbdXD 💪

PR #404, opened 2021-05-07
Also call set_value_before_save for duplicate check

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3013

These issues always mention the hook not working, but it's really just that the duplicate checks don't catch certain field types, so this is really just "duplicate checks don't work for date fields".

I missed that there are really two functions here.

`get_value_to_save` which I added in the last duplicate check, which caught signatures and some other fields.

But there is also a `set_value_before_save` that effects dates, numbers, times, credit cards.

I'm calling the new one afterward as it's meant for processing at the very end and the other function is intended for earlier in the processing.


jairoprez (approved):
Tested!, duplicate checks already work properly for date fields https://recordit.co/YQbbIPuN61 💪

stephywells (approved):
Good catch! This looks good to me.

PR #405, opened 2021-05-07
add strict option for honeypot, default to text, add new frm_process_honeypot filter

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2993

Updated functionality of `frm_run_honeypot`.
```
add_filter(
	'frm_run_honeypot',
	function( $honeypot, $atts ) {
		// default value of $honeypot is true (spam is more likely than "strict" but false positives are also lower)
		$form = $atts['form'];
		return 'strict';
	},
	10,
	2
);
```

New filter `frm_process_honeypot`.

```
add_filter(
	'frm_process_honeypot',
	function( $is_honeypot_spam, $atts ) {
		$form = $atts['form'];
		if ( $is_honeypot_spam ) {
			// handle honeypot spam.
		}
		// validate as not spam so the form still submits.
		return false;
	},
	10,
	2
);
```

### DIFF
--- a/classes/models/FrmEntryValidate.php
+++ b/classes/models/FrmEntryValidate.php
@@ -229,7 +229,7 @@ class FrmEntryValidate {
 			return;
 		}
 
-		if ( self::is_honeypot_spam() || self::is_spam_bot() ) {
+		if ( self::is_honeypot_spam( $values['form_id'] ) || self::is_spam_bot() ) {
 			$errors['spam'] = __( 'Your entry appears to be spam!', 'formidable' );
 		}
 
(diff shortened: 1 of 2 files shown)


jairoprez (approved):
The strict option seems to work fine https://recordit.co/GHwLb8e2eY

@Crabcyborg The documentation for the new `frm_process_honeypot` filter has been added here https://formidableforms.com/knowledgebase/frm_process_honeypot/ please take a look to validate if it makes sense

Crabcyborg:
@jairoprez i would not recommend returning true in an example. This is processed even when it's off.

If you return true, you would mark all entries as spam, not just the ones caught by honeypot.

jairoprez:
> @jairoprez i would not recommend returning true in an example. This is processed even when it's off.
> 
> If you return true, you would mark all entries as spam, not just the ones caught by honeypot.
> 
> You would return $is_honeypot_spam instead.

Thanks for the feedback @mike, I changed the return based on your suggestion :+1:

stephywells:
@Crabcyborg The existing Akismet setting is under "On Submit" on the general tab.
https://formidableforms.com/knowledgebase/add-spam-protection/#kb-akismet

We can add this new option with the Akismet option, plus a checkbox for the new spam option. I think that would be safer to have a way to disable the new option too.

Crabcyborg:
I've merged this pull request into the JS spam check [here](https://github.com/Strategy11/formidable-forms/pull/402).

I'm just going to close this one and merge this update with the other when the time comes.

PR #410, opened 2021-05-11
mark fields updated to 0 on publish button click

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-views/issues/116 by resetting the fieldsUpdated flag when the update button is clicked.

This was something I missed when I did https://github.com/Strategy11/formidable-forms/pull/357 as I didn't catch that the fade out logic was happening for multiple types of elements on various pages, not just actions on the settings page.

This page still doesn't really support the unsaved changes pop ups, but it might not be worth it since we're replacing this soon.


stephywells (approved):
Nice. You're right. We can make any more needed adjustments to this in 5.0

PR #417, opened 2021-05-13
Upload image button in Radio field should show only images

truongwp:
This fix https://github.com/Strategy11/formidable-pro/issues/3018

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -2159,11 +2159,14 @@ function frmAdminBuildJS() {
 		wp.media.model.settings.post.id = postID;
 
 		fileFrame = wp.media.frames.file_frame = wp.media({
-			multiple: false
+			multiple: false,
+			library: {
+				type: [ 'image' ]
+			}
 		});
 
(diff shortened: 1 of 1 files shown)


Crabcyborg (approved):
@truongwp this looks good!

I don't really see where this could have any issues.

@stephywells can you think of anything that might ever get used as a radio option image that might be actually validate here as an image?

stephywells:
Good call! I can't think of another use case. Even if someone happens to have something selected that isn't an image, this wouldn't cause any issues with existing fields.

Looks good to me! Does this need any QA?

truongwp:
If someone selects a file that isn't an image, it will show a broken image. So I think no one keeps that.

Screen Shot 2021-05-14 at 08 01 00

Crabcyborg:
> Looks good to me! Does this need any QA?

I just looked at it quickly. It appears to filter the options to images only. It looks like uploading a non-image is still possible, so we're really just preventing people from selecting invalid files.

I think this is good as it is. I'm going to merge it. Thanks @truongwp!

PR #418, opened 2021-05-13
Fix class name replace issue for fields without classes already set

Crabcyborg:
I'm starting to get into the side-by-side form editor requirements here, and I'm noticing some funny things with the current ways that the editor is adding stuff.

The replace here is not working when there are no classes because it looks like this double space check is being trimmed now.

I check first if the previous string exists, and if it doesn't, I do my replace instead to fix this issue.

There was that other layout class error happening here that I think probably broke for the same reason.


stephywells (approved):
Weird. Alright then. Good find!

PR #420, opened 2021-05-20
Update readme for 4.11

Crabcyborg:
@stephywells I know you've been updating this file as well, but I don't see those updates here. Not sure where you're updating them, if it's in SVN or something?

Also, I feel like I'm really not celebrating the name file enough here with my short little description. I'm just not sure what to say about it.


stephywells:
@Crabcyborg I added in some of my changes to the readme, but not all since they didn't seem to help. I've been making changes in SVN, and then adding them in Github once I know if they help.

I expanded the changelog a bit too. I'll need to find an article we can write about it to to give it more attention too. We should also update a spam protection article with more detail too. Then we can add them to the api-end inbox messages to let people know.

Thanks for all your work on this release! And @truongwp too!

PR #425, opened 2021-05-26
honeypot setting appears basic when off fix

Crabcyborg:
Fixes the issue Guga posted [here](https://strategy11.slack.com/archives/CJFQ599V5/p1621876733054200).

The honeypot option was working fine but the basic option would appear as selected if it was off. And then a subsequent save would likely turn it back on as a result.

My logic for the basic option was too much of a catch-all.

I don't think any of it the extra logic I had is really necessary now. The default is basic.


stephywells (approved):
Much cleaner. Since we don't have any reverse compatibility to worry about here, this should be fine.

PR #427, opened 2021-05-26
change grid column value instead of display for frm_no_grid_750

Crabcyborg:
Based off of the conversation in https://github.com/Strategy11/formidable-views/issues/111


stephywells (approved):
Perfect.

PR #428, opened 2021-05-26
add margin to wpml sub list

Crabcyborg:
The list that WPML adds to our views page looks off from everything else.

It's bothered me for a while as I've worked on views with WPML also on.

I'm just adding it to the same class definition that we use for the list above it (that toggles between All/Private) so it aligns consistently.

Before:
Screen Shot 2021-05-26 at 5 30 28 PM

After:
Screen Shot 2021-05-26 at 5 29 23 PM

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -1323,6 +1323,7 @@ a.frm_hidden,
 
 .post-type-frm_display .wrap > #posts-filter,
 .post-type-frm_display .wrap > .subsubsub,
+.post-type-frm_display .wrap > .icl_subsubsub,
 .frm_form_settings > p {
 	margin: 6px 40px;
 	padding: 0;


stephywells (approved):
Good call. Also, if all the views are private, that nav isn't needed. I'm not sure if that's worth the time right now though.

Crabcyborg:
> Good call. Also, if all the views are private, that nav isn't needed. I'm not sure if that's worth the time right now though.

@stephywells I played with that briefly but if I do hide the nav, the WPML one also never renders because it looks for our nav and tries to position underneath it.

Do you think that's an issue?

The other option seems to be hiding it with JavaScript but it feels pretty hacky and it still shows up on the load and gets hidden afterward, which probably isn't worth it.

PR #431, opened 2021-06-01
include plan=view in upgrade link for views

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3032

I think this is really all we need to do here.

The only other views upsell I can remember is for creating post actions, and it links to the views-info page that this update is targeting.
Screen Shot 2021-06-01 at 12 16 06 PM

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -87,11 +87,17 @@ class FrmAppHelper {
 		}
 
 		$anchor = '';
+		$plan   = '';
 		if ( is_array( $args ) ) {
-			$medium  = $args['medium'];
-			$content = $args['content'];
+			$medium = $args['medium'];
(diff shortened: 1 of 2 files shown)


classes/helpers/FrmAppHelper.php
```diff
@@ -111,6 +117,10 @@ public static function admin_upgrade_link( $args, $page = '' ) {
 			$query_args['f'] = $args['param'];
 		}
 
+		if ( $plan ) {
```
stephywells, line 120: Could we simplify this a bit by just doing 
```
if ( is_array( $args ) && ! empty( $args['plan'] ) ) {
$query_args['plan'] = $args['plan'];
}
```

stephywells (approved):
Looks great! Thanks Mike!

PR #435, opened 2021-06-02
Prevent adding more actions than the limit when double clicking 

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3035

Without this it wasn't terribly difficult to add 9 post actions to my form because the check has to include stuff that is in-progress.

I just add a placeholder div with the same class name used to determine count. When it is successfully inserted, I delete the placeholder.

The same applies for registration actions.

I refactored this quite a bit to support the check on both ends (the second check applies some limit classes and stuff that I don't do in the initial check). I think it's a lot easier to follow than before now.


jairoprez (approved):
@Crabcyborg I did tests and this fix works great https://recordit.co/LfaHhvEojR 🎉

stephywells (approved):
Nice catch. This looks much cleaner.

PR #440, opened 2021-06-07
addons typo when getting api info fix

Crabcyborg:
Caught this with psalm.

```
ERROR: UndefinedVariable - classes/models/FrmAddon.php:363:21 - Cannot find referenced variable $addons (see https://psalm.dev/024)
		} elseif ( isset( $addons['error'] ) ) {
```

This means that `is_expired_addon` was never really being flagged.
This might explain a lot of the weird license messages that people report.


stephywells (approved):
Doh!

PR #442, opened 2021-06-16
Update 4.11.02 readme

Crabcyborg:
There are other updates in here but I don't really think any of them are easy to describe and might not really go noticed by a user as much (https://github.com/Strategy11/formidable-forms/milestone/9?closed=1).

So I'm keeping it simple.


stephywells (approved):
Looks good. I have a few changes in SVN, but they didn't seem to make a difference. I'll add my changes back in if the rankings drop.

PR #445, opened 2021-06-17
Fix ribbon position for new add-ons

stephywells:
The "new" ribbon was hidden on the add-ons page.
Screen Shot 2021-06-17 at 9 45 08 AM

After:
Screen Shot 2021-06-17 at 9 45 46 AM


Crabcyborg (approved):
Looks good @stephywells! 🚢

PR #446, opened 2021-06-19
Fix the entries list does not show value 0 inside repeater field

truongwp:
This fixes issue https://github.com/Strategy11/formidable-pro/issues/3069


Crabcyborg (approved):
@truongwp nice! Surprised this hasn't come up as a ticket before.

PR #449, opened 2021-06-28
Deprecate the old grid css setting

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3034

Screen Shot 2021-06-28 at 12 10 35 PM

Do we want to change the wording? The double negative is a little funny.

And do we want to mention more about why we're doing this? Mention internet explorer? Should we refer to the upcoming WordPress deprecation?

### DIFF
--- a/classes/controllers/FrmInboxController.php
+++ b/classes/controllers/FrmInboxController.php
@@ -13,23 +13,9 @@ class FrmInboxController {
 	 */
 	public static function menu() {
 		$unread = self::get_notice_count();
-
-		$frm_settings = FrmAppHelper::get_settings();
-		if ( $frm_settings->old_css ) {
-			$inbox = new FrmInbox();
-			$inbox->add_message(
-				array(
-					'key'     => 'old_css',
-					'subject' => 'The option to not use CSS Grids for form layouts is being removed.',
-					'message' => 'New formidable features will require the CSS Grid. We recommend enabling the CSS Grid in Global Settings and testing that your forms still look good with updated layouts.',
-					'cta'     => '<a href="' . esc_url( admin_url( 'admin.php?page=formidable-settings' ) ) . '" class="button-primary frm-button-primary">' . esc_html__( 'Go to Global Settings', 'formidable' ) . '</a>',
-					'icon'    => 'frm_report_problem_icon',
-					'type'    => 'news',
-				)
-			);
-		}
-
 		add_submenu_page( 'formidable', 'Formidable | ' . __( 'Inbox', 'formidable' ), __( 'Inbox', 'formidable' ) . $unread, 'frm_change_settings', 'formidable-inbox', 'FrmInboxController::inbox' );
+
+		self::maybe_add_old_css_message();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


classes/views/frm-settings/general.php
```diff
@@ -31,13 +31,15 @@
...
-		<input type="checkbox" id="frm_old_css" name="frm_old_css" value="1" <?php checked( $frm_settings->old_css, 1 ); ?> />
-		<?php esc_html_e( 'Do not use CSS Grids for form layouts', 'formidable' ); ?>
-		<span class="frm_help frm_icon_font frm_tooltip_icon" title="<?php esc_attr_e( 'Form layouts built using CSS grids that are not fully supported by older browsers like Internet Explorer. Leave this box unchecked for your layouts to look best in current browsers, but show in a single column in older browsers.', 'formidable' ); ?>"></span>
-	</label>
-</p>
+<?php if ( $frm_settings->old_css ) { ?>
+	<p>
+		<label for="frm_old_css">
+			<input type="checkbox" id="frm_old_css" name="frm_old_css" value="1" <?php checked( $frm_settings->old_css, 1 ); ?> />
+			<?php esc_html_e( 'Do not use CSS Grids for form layouts', 'formidable' ); ?>
```
stephywells, line 38: Let's add a warning here that once this is turned off, it can't be turned back on.

Warning: this setting will be removed when it is turned off.

classes/controllers/FrmInboxController.php
```diff
@@ -14,6 +14,21 @@ class FrmInboxController {
 	public static function menu() {
 		$unread = self::get_notice_count();
 
+		$frm_settings = FrmAppHelper::get_settings();
```
stephywells, line 17: Can we add this new code into a separate function?

classes/controllers/FrmInboxController.php
```diff
@@ -14,6 +14,21 @@ class FrmInboxController {
...
+		if ( $frm_settings->old_css ) {
+			$inbox = new FrmInbox();
+			$inbox->add_message(
+				array(
+					'key'     => 'old_css',
+					'subject' => 'The option to not use CSS Grids for form layouts is being removed.',
```
stephywells, line 23: How about:
Subject: The option to disable CSS Grids for form layouts is being removed.
Message: We've got some awesome form layout features coming soon. These new layouts will require CSS grids, and will not be supported in Internet Explorer. If a visitor views your forms in IE, a single field will show in each row.
We recommend enabling CSS Grids in Global Settings and then checking your form layouts.

stephywells (changes requested):

classes/views/frm-settings/general.php
Crabcyborg: @stephywells how do you picture this?

I played with adding a warning in here. Putting it above/below it isn't very clear which setting it's for. Inline seems okay though.

Screen Shot 2021-06-29 at 4 27 32 PM

What do you think?

stephywells: That looks fine to me. I was thinking just plan text or italic, but yours looks better.

PR #452, opened 2021-06-29
Fix name field not working in the From and To options of Send mail action

truongwp:
This fixes https://github.com/Strategy11/formidable-pro/issues/3083


Crabcyborg (approved):
Looks good!

PR #453, opened 2021-06-30
update readme and changelog

Crabcyborg:
Think this is it. Not sure if we want to mention the deprecation at all? Doesn't seem significant.


stephywells (approved):
Thanks Mike!

PR #454, opened 2021-06-30
Make an ajax request for menu order field

Crabcyborg:
Testing the new menu_order setting, I feel like it has to be dynamic like the post_parent options.

I refactored the post_parent logic as well since these both use almost entirely the same logic.

I wonder if we might want to use one action instead, but that might be weird with backward compatibility and having this in free and pro.

Also requires the AJAX action logic in the pro plugin [here](https://github.com/Strategy11/formidable-pro/pull/3089).

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -4381,73 +4381,47 @@ function frmAdminBuildJS() {
 
 		// Get new post parent option.
 		if ( postParentField ) {
-			const postParentOpt     = postParentField.querySelector( '.frm_autocomplete_value_input' ) || postParentField.querySelector( 'select' );
-			const postParentOptName = postParentOpt.getAttribute( 'name' );
-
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -4380,38 +4381,47 @@ function frmAdminBuildJS() {
...
+
+				// Post type is not a page.
+				if ( '0' === response ) {
+					field.classList.add( 'frm_hidden' );
+					field.value = '';
+					return;
+				}
+
+				field.classList.remove( 'frm_hidden' );
+				// The replaced string is declared in FrmProFormActionController::ajax_get_post_menu_order_option() in the pro version.
```
stephywells, line 4419: Since we don't need autocomplete, or a refresh of the options in the order dropdown, maybe we could return here if '1' === response?

PR #456, opened 2021-07-05
Put back the watch lookup option when switching away from single line text

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2214

Pretty simple one. There was logic to hide it, and no logic to put it back again.


stephywells (approved):
Looks perfect

PR #457, opened 2021-07-07
Save and display reCaptcha_v3 scores on form entries in admin

dmchale:
This PR hopes to take a baby step toward providing more admin control over reCaptcha submissions with different scores using the v3 API

When submitting a form with reCaptcha to the google API, a "score" param is returned in the response. This is meant to give developers more control over when they would like to qualify submissions as good/bad on their own terms, but currently Formidable (like the vast majority of plugins) does not utilize this ability and treats reCaptcha usage as a simple on/off switch.

This PR will save the returned score to the global $frm_vars variable and if found, saves that value to the meta for the captcha field_id on the submission entry. When viewing individual entries (eg: /wp-admin/admin.php?page=formidable-entries&frm_action=show&id=## ), a new line for "reCaptcha Score" is displayed under the Entry Details heading in the sidebar.

Any configurations using "checkbox v2" or "invisible v2" reCaptcha keys will not be affected in any way. 

This PR will only save the information on future submissions; all historical submissions ignored this part of the API response. 

I'm happy to discuss the specific implementation and make any changes necessary, but I hope you'll consider this update!

### DIFF
--- a/classes/models/FrmEntryMeta.php
+++ b/classes/models/FrmEntryMeta.php
@@ -87,18 +87,39 @@ class FrmEntryMeta {
 	}
 
 	/**
+	 * @param $entry
+	 *
+	 * @return mixed|null
+	 */
+	public static function get_entry_recaptcha_score( $entry ) {
+		foreach ( $entry->metas as $field_id => $value ) {
+			$field_obj = FrmFieldFactory::get_field_object( $field_id );
+			if ( 'captcha' == $field_obj->type ) {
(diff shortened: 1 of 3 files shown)


Crabcyborg:
Thank you @dmchale.

As of 5.0.15 we now have support for v3 reCAPTCHA. It's a little different from what you have, but it's similar. Our update includes a new threshold slider as well so an error is flagged based off of score.

PR #459, opened 2021-07-09
fix autocomplete styling

stephywells:
Fixes https://github.com/Strategy11/formidable-pro/issues/2959

Here's what it looks like now:
Screen Shot 2021-07-09 at 2 27 57 PM


Crabcyborg (approved):
Thanks @stephywells looks good to me!

PR #460, opened 2021-07-12
Use email from from value when reply to value is a shortcode that returns empty

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2823

Quite simple. It was checking if the reply to was empty and then using the from name. But it wasn't check if the processed result was empty. So I flipped the logic a bit. First, we check if it's set, process it. Then if it's empty (which could have been true before, but it also matters if it is empty after), we use the from name.

### DIFF
--- a/classes/models/FrmEmail.php
+++ b/classes/models/FrmEmail.php
@@ -180,11 +180,14 @@ class FrmEmail {
 	private function set_reply_to( $user_id_args ) {
 		$this->reply_to = trim( $this->settings['reply_to'] );
 
-		if ( empty( $this->reply_to ) ) {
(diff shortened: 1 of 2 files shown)


stephywells (approved):
Nice, simple, quick fix for an annoying issue!

garretlaxton (approved):
When the email field is left blank, the reply-to defaults to the _from_ email.

Crabcyborg:
> When the email field is left blank, the reply-to defaults to the _from_ email.

Thank you for testing @garretlaxton sounds like we're all good to merge here :)

PR #461, opened 2021-07-12
Automatically focus on the first error input

Crabcyborg:
I did some research on this topic. There's more we could probably do to make this even more accessible but I think this is a good start.

Fixes https://github.com/Strategy11/formidable-pro/issues/2897

I found some information on the issue at https://www.w3schools.com/accessibility/accessibility_errors.php

There's a section that reads:
> Move focus
> This is more important for server-side validations, compared to client-side validation. When the user submits a form, the focus moves to the first invalid field.
> In this example, all three fields have been invalid. The focus was moved to the first field.

So this checks out as a good best practice and it does make a form field easier to use when you mess it up, even if you're not blind 🥳.

I checked all of the many ways that errors are added that I can think of.

- A default form submission.
- With AJAX submit turned on.
- With JavaScript validation turned on.
- With multiple pages.
- A required name field (it autofocused the missing last name as expected)

Everything uses the same basic function that doesn't have terribly complicated logic so it should be good at catching every type of error.

@garretlaxton with this one, have fun really trying to trigger errors in as many ways as you can for whatever form configurations you can, see if I'm missing anything here.

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1007,7 +1007,7 @@ function frmFrontFormJS() {
 	}
 
 	function checkForErrorsAndMaybeSetFocus() {
-		var errors, element;
+		var errors, element, timeoutCallback;
 
 		errors = document.querySelectorAll( '.frm_form_field .frm_error' );
 		if ( ! errors.length ) {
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Nice Mike! Thank you for adding this!

garretlaxton (approved):
Text fields are focused when an error occurs as expected. Specifically, focused in order. 

Paragraph fields don't get focused when an error occurs. Not sure if that the intention or not.

Crabcyborg:
> Paragraph fields don't get focused when an error occurs. Not sure if that the intention or not.

Thanks for checking that @garretlaxton.

I did some extra updates. It looks like dropdowns, paragraphs, and rich text types all needed some extra focus logic.

If you could try those out again and let me know how it goes for you, that'd be appreciated 🚀

garretlaxton:
@Crabcyborg The new changes now focus paragraphs, dropdowns, and rich text fields when an error occurs as intended! The only other field I noticed was the signature. Not sure if that's even possible or wanted though...

Crabcyborg:
Thank you @garretlaxton!.

I think we're probably fine to ignore signatures for now, but that's a fair point.

PR #462, opened 2021-07-14
Replace field ids on unserialized grid view content

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-views/issues/191

I figure we can squeeze this into the 4.11.04 release since it's important 🚢.

Looks like I also added some commits for 4.11.04 in here by accident but I don't see any reason to bother taking them out at this point.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -11,7 +11,7 @@ class FrmAppHelper {
 	/**
 	 * @since 2.0
 	 */
-	public static $plug_version = '4.11.03';
+	public static $plug_version = '4.11.04';
 
 	/**
 	 * @since 1.07.02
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
Importing grid views onto other sites no longer shows the objects in the content. 

Something I'm just curious about, why do we use `serialize`, `unserialize()`, and `maybe_unserialize()` instead of `json_encode` and `json_decode`?

stephywells:
@Crabcyborg I think where this is still pretty new, it would be best to switch to json since it's so much less error prone. The FrmAppHelper::unserialize_or_decode functions are designed for moving away from serialized data. We did this with the form actions to make it more reliable.

garretlaxton:
I only pointed it out as these older functions can lead to security issues (PHP Object Injection) if user input is passed through. I usually see `json_encode()` and `json_decode()` used more commonly when looking through other plugins, so I was also just curious.

Crabcyborg:
Thank you @garretlaxton and @stephywells.

I'm going to close this then and work on changing the encoding for the future. We should be able to support the new format at least without requiring an update in this plugin this way, though this old format will always have this issue.

stephywells:
I think that's not a problem. We'll just need to update our templates. Thanks!

PR #467, opened 2021-07-23
Prevent integers from being switched to a string

stephywells:
This prevents integers from being switched to a string when content is encoded for a json post


Crabcyborg (approved):
Thanks @stephywells looks good!

PR #468, opened 2021-07-23
Stop using serialization for views in import

Crabcyborg:
Export/import should now all be JSON based for views.

Should be tested with https://github.com/Strategy11/formidable-views/pull/210

### DIFF
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -778,6 +778,10 @@ class FrmXMLHelper {
 				$post_id = wp_insert_post( $post );
 				self::maybe_update_custom_css( $custom_css );
 			} else {
+				if ( $post['post_type'] === 'frm_display' ) {
+					$post['post_content'] = FrmAppHelper::maybe_json_decode( $post['post_content'] );
+					$post['post_content'] = FrmAppHelper::prepare_and_encode( $post['post_content'] );
+				}
 				// Create/update post now
 				$post_id = wp_insert_post( $post );
 			}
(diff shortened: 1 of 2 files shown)


stephywells (approved):
This looks good! Let's get more testing on this.

garretlaxton (commented):
Exports seems fine. However, when I import the view I get the following error: `Warning: Invalid argument supplied for foreach() in /wp-content/plugins/formidable-forms-issue_2823/classes/helpers/FrmAppHelper.php on line 2266`

The import still gets uploaded, but I get JSON in my grid view content boxes.
Screenshot 2021-07-23 at 16-27-08 Contact Us Grid – AWPCP TESTING

Here's my export I'm using, I changed it to a .txt since GH doesn't allow XML.
[contact-us-form.txt](https://github.com/Strategy11/formidable-forms/files/6871089/contact-us-form.txt)

Crabcyborg:
Thanks @garretlaxton I'll look into it!

Crabcyborg:
@garretlaxton I've tried importing your XML file and it seems to work fine on my end - I'm not seeing the JSON.

But I don't think any of my recent updates would have had any effect, since I haven't touched how the data is handled on load, and your data loaded fine for me.

I'm a little stumped if there are still issues 😕.

garretlaxton:
@Crabcyborg Hmm.. It seems I did something wrong here. 

I just imported this again on a fresh install and it worked for me. I guess I was missing something. Sorry about that!

Crabcyborg:
Thank you for double checking @garretlaxton!

PR #469, opened 2021-07-26
Allow for larger save/preview form button labels before wrapping

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2276

I tested using "Nieuwe toevoegen" for the "Add New" button (Dutch), it looks like this was improved earlier and doesn't look so bad.
Screen Shot 2021-07-26 at 10 54 17 AM

But the Aktualisieren (Update) and Vorschau (Preview) example (German) was still an issue. Now it looks like:
Screen Shot 2021-07-26 at 10 32 49 AM

And I tested it in Dutch as well
Screen Shot 2021-07-26 at 10 55 12 AM

I tried using the rest of the site in plugin in Dutch for a moment and nothing else really seemed off 🕺.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -430,7 +430,8 @@ ul.frm_form_nav > li {
 #frm-publishing {
 	float: right;
 	margin-top: 15px;
-	width: 225px;
+	min-width: 225px;
+    max-width: 300px;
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Perfect!

PR #472, opened 2021-07-28
Use a div for heads up caution in confirmModal

Crabcyborg:
I came across this weird styling testing the side-by-side fields. Changing it so it displays as block instead of inline is really as we need to do here.

**Before**
Screen Shot 2021-07-28 at 2 09 19 PM

**After**
Screen Shot 2021-07-28 at 2 06 22 PM

I looked for other references to `data-frmcaution` and it looks like this "Heads up" it all it's ever used for at the moment.


stephywells (approved):
Thanks for fixing that!

PR #475, opened 2021-07-29
Include addons in wp search

stephywells:
This helps people find our add-ons easier.
Screen Shot 2021-07-29 at 5 21 08 PM

I'm updating the API end to include more search terms. Right now, the name of the add-on has to be exact in order to show up. For example "paypal standard" brings up our add-on and "paypal" doesn't. This part isn't included in the code here.


garretlaxton (approved):
Addons are searchable through the WP plugin repository when the exact addon name is searched.

The only addon that didn't show results was the Legacy Views. I figure this is since we have the new Visual Views.

Crabcyborg (approved):
Looks good to me!

PR #479, opened 2021-08-09
Add padding to calculations before sanitizing happens

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/2977

I'm just normalizing the expression a bit before it goes through the other sanitizing that was already there.

`5<10`, `5<=10`, get stripped. `5 < 10`, `5 <= 10` don't. Making sure that the spaces are always there before the sanitizing happens prevents the final value getting saved as just a `5`.

To avoid the preg_replace when it would have no effect, I've also added a small strpos check first so it only processes if it sees any `<` characters.

I also moved the calc sanitizing to its own function, and added unit tests to cover a lot similar patterns.

It looks like workflows are also breaking because of the ubuntu version. I tried playing with that briefly but it looks like there are other issues too. I think a new version of composer might be causing issues.

### DIFF
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -267,16 +267,30 @@ class FrmForm {
 	private static function sanitize_field_opt( $opt, &$value ) {
 		if ( is_string( $value ) ) {
 			if ( $opt === 'calc' ) {
-				$allow = array( '<= ', ' >=' ); // Allow <= and >=
-				$temp  = array( '< = ', ' > =' );
-				$value = str_replace( $allow, $temp, $value );
-				$value = strip_tags( $value );
-				$value = str_replace( $temp, $allow, $value );
+				$value = self::sanitize_calc( $value );
 			} else {
 				$value = FrmAppHelper::kses( $value, 'all' );
 			}
 			$value = trim( $value );
 		}
+	}
+
+	/**
+	 * @param string $value
+	 * @return string
+	 */
+	private static function sanitize_calc( $value ) {
+		$allow = array( '<= ', ' >=' ); // Allow <= and >=
+		$temp  = array( '< = ', ' > =' );
+		$value = str_replace( $allow, $temp, $value );
+
+		if ( false !== strpos( $value, '<' ) ) {
+			$value = preg_replace( '/(\d)<(\d)/', '$1 < $2', $value );
+		}
+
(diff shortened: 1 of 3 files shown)


classes/models/FrmForm.php
```diff
@@ -267,18 +267,30 @@ public static function update_fields( $id, $values ) {
...
 		}
 	}
 
+	/**
+	 * @param string $value
+	 * @return string
+	 */
+	private static function sanitize_calc( $value ) {
+		if ( false !== strpos( $value, '<' ) ) {
+			$value = preg_replace( '/(\d)( ){0,1}<(=){0,1}( ){0,1}(\d)/', '$1 <$3 $5', $value );
```
stephywells, line 284: Would you mind adding a note here that says what the regex is doing?

classes/models/FrmForm.php
```diff
@@ -267,18 +267,30 @@ public static function update_fields( $id, $values ) {
...
 			$value = trim( $value );
 		}
 	}
 
+	/**
+	 * @param string $value
+	 * @return string
+	 */
+	private static function sanitize_calc( $value ) {
+		if ( false !== strpos( $value, '<' ) ) {
```
stephywells, line 283: Is this only a problem with <? Or is it also >?

stephywells (approved):
Looks good! Just a couple small questions.

classes/models/FrmForm.php
Crabcyborg: It's only an issue with <. I added unit tests to check for the > stuff, including >= and they're all good.

But < and <= both have this issue.

classes/models/FrmForm.php
Crabcyborg: @stephywells I added a ton of comments, moved the replace to another function, and added specific unit tests to cover it, just to make this all a little more clear.

garretlaxton (approved):
Using < or <= in a calculation now shows the full value with spaces.
Screenshot 2021-08-09 at 15-24-34 Formidable Forms ‹ Formi QA — WordPress

I also tried to escape the script tag with no luck, so sanitization looks good!

stephywells (approved):

classes/models/FrmForm.php
stephywells: Looks great! Thanks for adding that!

PR #482, opened 2021-08-11
Update readme for 5.0

Crabcyborg:
Question - How should I mention in the changelog that the required WordPress version has been bumped up to 5.0?

I'm also not sure if we want to mention the updates to the search? If we do, what should I put?


stephywells (commented):
Looks good! Let’s be sure to add the min WP requirement in the change log too.

PR #483, opened 2021-08-12
Fix issue with duplicating long form keys

Crabcyborg:
I'm not totally sure what willed me to test this but duplicating long form keys totally breaks, similar to the long field key duplicating issue I fixed a while ago.

Screen Shot 2021-08-12 at 10 54 26 AM

This double checks that a key is long before making it longer, if we're extending a key that was reserved in the database.


stephywells (approved):
Good call! Thanks for checking that!

PR #484, opened 2021-08-12
Fix issues with toggling pages

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3121

Page toggle logic had to change a bit since it uses nextUntil but I have the pages wrapped like everything else for consistency.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -4039,15 +4039,14 @@ function frmAdminBuildJS() {
 
 	// Find all fields in a page and hide/show them
 	function toggleCollapsePage( field ) {
-		var toCollapse = getAllFieldsForPage( field.parent().closest( 'li.frm_field_box' ).next() );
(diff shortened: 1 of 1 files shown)


stephywells (commented):
This works and looks great. I am seeing one js error when I try to collapse a page with no fields.
formidable_admin.js?ver=5.0:4066
Uncaught TypeError: Cannot read property 'querySelector' of null

Crabcyborg:
> Uncaught TypeError: Cannot read property 'querySelector' of null

Thanks @stephywells nice catch! I've pushed a check for the initial value of the first wrapper as well.

stephywells (approved):
Perfect!

garretlaxton (approved):
Collapsing the first page no longer collapses all the pages.

Screenshot 2021-08-12 at 14-33-57 Formidable Forms ‹ Formi QA — WordPress

I tested this with 8 pages like the ticket and all pages collapse individually.

Crabcyborg:
Thank you for testing @garretlaxton and @stephywells!

PR #485, opened 2021-08-12
Add search to fields list

stephywells:
a few small styling and icon updates

This is just a bit more to make the 5.0 release a little more different from 4.x.


Crabcyborg:
@stephywells I'm not entirely sure what this update is supposed to look like but it seems off to me.

Screen Shot 2021-08-13 at 10 55 42 AM

Having no hover state anymore makes it unclear that my group controls have anything to do with my group.

And the change to selected field groups seems a little odd.
Screen Shot 2021-08-13 at 10 57 22 AM

stephywells:
Yep, for sure! I just noticed that too. I was only looking at single fields on that change. That's updated now.

Crabcyborg (approved):
Thank you, looks good!

PR #488, opened 2021-08-16
Fix compare for update unit test

Crabcyborg:
I made a release and it didn't go live because of this failing unit test that seemed to happen when I bumped the version.

It looks like it has specific tests for going up to the next major version? It seems that it breaks now that we're at 5.0.

This update makes it pass. I'm targeting the 6.0 major version now instead, which I think is what needs to be done here? I set the 5.0 test to use the current db version. I think that's what that's intended for?


stephywells (approved):
Thanks!

PR #491, opened 2021-08-24
Add duplicate/delete options to field groups

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3132

I moved the dropdown options for fields out of the HTML/PHP for each field, which should help to optimize long forms by removing tons of HTML on load 🥳.

For now these just use the individual field hooks. In the future I'd love to make this a little optimized but this is much less error prone as it is.

### DIFF
--- a/classes/views/frm-forms/add_field.php
+++ b/classes/views/frm-forms/add_field.php
@@ -3,8 +3,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 	die( 'You are not allowed to call this page directly.' );
 }
 ?>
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
I duplicated a bunch of rows using different advanced field options. They all stay the same when duplicated. Deleting also works great!

Screenshot 2021-08-24 at 16-24-00 Formidable Forms ‹ Formi QA — WordPress

I think it would be cool if you didn't have to click update when changing a field option before you duplicate a field or row. My 2nd text row is missing the field descriptions because I didn't click update before duplicating. Just my little feedback! (:

Crabcyborg:
Thanks @garretlaxton that looks like a pretty separate item. I agree that seems really odd.

I've logged that as a new issue https://github.com/Strategy11/formidable-pro/issues/3142

It looks like fixing that might not be too easy.

I also pushed a small update that allows for the multiselect delete option in the corner if you have 1 group selected but that group has multiple fields in it.

stephywells (approved):
Cool! Thanks Mike!

PR #492, opened 2021-08-25
Add chinese translations

Crabcyborg:


Crabcyborg:
> We don't actually need this. Once a translation is 100% on wp.org, the version in our plugin is no longer used. We can close this pr without merging.

Oh yeah? Should we delete all of our old translation files then? They're way out of date (3 years ago).

stephywells:
We can only delete the ones for the languages listed here:
Screen Shot 2021-08-25 at 9 10 50 AM

I forgot that the language files aren't getting updated automatically with releases. We're doing this automatically in BD with this:
https://github.com/Strategy11/business-directory-plugin/blob/master/Gruntfile.js#L47-L79

We're also updating the version number in more places so we can use @ since x.x in the code:
https://github.com/Strategy11/business-directory-plugin/blob/master/Gruntfile.js#L114-L145

Crabcyborg:
Thanks for clearing that up @stephywells!

Closing this.

PR #497, opened 2021-08-30
Duplicate some unsaved field keys

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3142 at least for most keys.

I wanted to give this a go. It isn't too tricky to check both sets of inputs for inconsistencies.

Right now there are likely some keys that don't copy over because I'm intentionally using an allowlist to avoid some surprises or unexpected behaviour issuess. Right now the things that copy even if the changes in the original field were unsaved are:

- Name
- Required
- Unique
- Read only
- Default value
- Placeholder
- Description
- Size
- Max
- Format
- Before/after
- Separate values
- Options for radio/checkbox/dropdown fields

Works when duplicating fields individually or when duplicating groups. It looks like there might still be some gaps with duplicating sections though.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -3104,7 +3104,7 @@ function frmAdminBuildJS() {
 		$newRow.on(
 			'frm_added_duplicated_field_to_row',
 			function( event, args ) {
-				var $duplicatedFields, index, $injectTarget;
+				var $duplicatedFields, index;
 
 				originalFieldIdByDuplicatedFieldId[ jQuery( args.duplicatedFieldHtml ).attr( 'data-fid' ) ] = args.originalFieldId;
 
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
Duplicating works well with advanced field options! 

Something kind of unrelated I noticed with page breaks is the Button Label doesn't get updated in the builder until the form is saved. It duplicates fine, but it just doesn't get updated in the builder to see. 

Screenshot 2021-08-30 at 15-21-52 Formidable Forms ‹ Formi QA — WordPress

It also might be a good idea to add a label for the back button as well?

Screenshot 2021-08-30 at 15-26-41 Formidable Forms ‹ Formi QA — WordPress

Crabcyborg:
@garretlaxton thank you for testing!

The previous button is actually customized from the form settings under styling/buttons. I get that it's a little odd that they're in two places, but it is technically supported somewhere.
Screen Shot 2021-08-30 at 4 40 19 PM

stephywells (approved):
Awesome! Thanks so much! I thought most of this already worked.

Issue #125, opened 2019-07-16
Zero value ('0') in 'other' field of checkbox is not sent in email notification


lanresmith:
Although the value is properly saved into the DB, it doesn't seem to be sent in the email notification.

@stephywells can you please confirm this?

PR #498, opened 2021-08-30
Add elementor widget

Crabcyborg:
This has the same options as the Gutenberg block, more or less, I think.

Screen Shot 2021-08-30 at 4 00 58 PM
Screen Shot 2021-08-30 at 4 01 04 PM
Screen Shot 2021-08-30 at 4 01 02 PM


garretlaxton (commented):
The options don't seem to work as the title & description should be hidden right?
Screenshot 2021-08-30 at 16-47-12 Elementor Elementor Page #125

I'm not sure if I'm doing something wrong, or if I need Elementor Pro for this?

Crabcyborg:
@garretlaxton thanks for testing. I'm not sure what's up. It seems to work as expected on my end.

I don't think I've had pro active during most of this testing either 😕.

Is it always broken? Is it just a default thing?

garretlaxton:
The settings don't seem to be affected when changed. 

https://www.loom.com/share/b2660164c1134e72827d9c3a78d5c7d7

Crabcyborg:
Thanks @garretlaxton it works for me locally but as soon as I tried it on the qa site I'm seeing what you are.

What I'm seeing on my end,
https://recordit.co/BJlvXOhbk5

Now that I can replicate this somewhere I'm going to keep figuring this out.

Crabcyborg:
@garretlaxton it should work now after the latest changes. It does for me on the qa site with the 5.0.03b version I've uploaded.

It looks like the shortcode function was pretty picky about the 1/0 I was passing it. It looks like booleans work better.

Crabcyborg:
Ha it doesn't do a ton.

There's some info on it here.
https://formidableforms.com/knowledgebase/why-doesnt-my-form-look-pretty/

classes/widgets/FrmElementorWidget.php
```diff
@@ -0,0 +1,102 @@
...
+
+if ( class_exists( '\Elementor\Widget_Base' ) ) {
+	class FrmElementorWidget extends \Elementor\Widget_Base {
+
+		public function get_name() {
+			return 'formidable';
+		}
+
+		public function get_title() {
+			return __( 'Formidable Forms', 'formidable' );
```
stephywells, line 14: Do you think we should whitelabel this with FrmAppHelper::get_menu_name()?

classes/widgets/FrmElementorWidget.php
```diff
@@ -0,0 +1,102 @@
...
+				array(
+					'label' => $title,
+					'type'  => \Elementor\Controls_Manager::SWITCHER,
+				)
+			);
+		}
+
+		private function get_form_options() {
+			$query   = array();
+			$where   = apply_filters( 'frm_forms_dropdown', $query, 'form' );
```
stephywells, line 72: Do you think it would be worth moving this to a function that we could reuse, and refactor the other similar code?

stephywells (approved):
This looks great! Feel free to merge as is, or make changes if you'd like. My comments are more of questions than requests.

classes/widgets/FrmElementorWidget.php
Crabcyborg: Good call, the icons all come through pretty easily! 🥳

Screen Shot 2021-09-01 at 11 29 00 AM
Screen Shot 2021-09-01 at 11 28 46 AM
Screen Shot 2021-09-01 at 11 28 24 AM

classes/widgets/FrmElementorWidget.php
Crabcyborg: @stephywells initially I was trying to use FrmFormsHelper::forms_dropdown but custom fields seem more complicated with Elementor, so I went with a more supported type and pulled out the options logic.

It looks like this filter is also called in FrmFormsController::list_templates and FrmFormsHelper::form_switcher.

They're all a little different. I kept mine simple by removing whatever I didn't need.

I don't intend on really doing this again. It was more because I needed the options isolated from the dropdown for Elementor specifically.

I agree some merging might help trim this down though. But I think for saving time, I'd rather leave that for later.

stephywells: Sounds good. Thanks for checking!

classes/widgets/FrmElementorWidget.php
Crabcyborg: Another update, I tried testing this out with the Simple Blocks Gutenberg stuff, and it looks like the second icon was breaking.

Screen Shot 2021-09-01 at 12 00 26 PM

I've moved my settings check out of the widget into the helper and used it to fix the bug there. It looks like the white label icon filter is not really safe because it always switches to an SVG which doesn't work in either of these cases.

stephywells: Awesome! Thank you!

PR #499, opened 2021-08-31
Update previews for next button and slider labels

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3153

Also requires https://github.com/Strategy11/formidable-pro/pull/3160 for some of the fix.

Applies to the next button for page break fields and slider labels, and the add/remove labels for repeater fields in the other pull request.

### DIFF
--- a/classes/views/frm-fields/back-end/number-range.php
+++ b/classes/views/frm-fields/back-end/number-range.php
@@ -9,10 +9,10 @@ if ( ! defined( 'ABSPATH' ) ) {
 	</label>
 	<span class="frm_grid_container">
 		<span class="frm5 frm_form_field frm-range-min">
-			<input type="text" name="field_options[minnum_<?php echo absint( $field['id'] ); ?>]" value="<?php echo esc_attr( $field['minnum'] ); ?>" />
+			<input type="text" name="field_options[minnum_<?php echo absint( $field['id'] ); ?>]" value="<?php echo esc_attr( $field['minnum'] ); ?>" data-changeme="field_<?php echo esc_attr( $field['field_key'] ); ?>" data-changeatt="min" />
 		</span>
 		<span class="frm5 frm_last frm_form_field">
-			<input type="text" name="field_options[maxnum_<?php echo absint( $field['id'] ); ?>]" value="<?php echo esc_attr( $field['maxnum'] ); ?>" />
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
These fields update in the builder now. 

Screenshot 2021-08-31 at 16-54-17 Formidable Forms ‹ Formi QA — WordPress

The slider also seems to act this way as well. If I change the range from `0 - 100` down to `0 - 10` the slider value should update to `5`. You can also set a `Default Value` and the slider will update, but the slider value does not.

Screenshot 2021-08-31 at 17-03-14 Formidable Forms ‹ Formi QA — WordPress

Not sure if you wanna tackle this while you're at it!

Crabcyborg:
Thanks @garretlaxton why not, it was a quick change update!

I also took out the logic for add/remove repeater labels in exchange for https://github.com/Strategy11/formidable-pro/pull/3160. There is "live changes" logic that I can hook into the simplify that one.

I tried that for the others but had less luck. I think this is the simplest solution to all of these issues.

stephywells (approved):
Nice update!

Crabcyborg:
@garretlaxton thanks for the other note. It looks like changing the min/max atts live was easy too, and fixing that issue with the blank value appearing blank in the label was also a pretty easy fix.

I think it's all good now. If you want to test one last time, I think it's good to merge now 🥳

garretlaxton:
@Crabcyborg The default value looks good. it's no longer blank when changed.

The number range still acts a little funny for me. It seems the slider moves all the way to the end & the value label doesn't change. The slider should be in the middle with a value of 5.
Screenshot 2021-09-02 at 13-23-34 Formidable Forms ‹ Formi QA — WordPress

Crabcyborg:
Thanks @garretlaxton it definitely was more off than I initially realized.

It looks like when you set a ranges value to empty, it doesn't actually set the midpoint as expected either, so I ended up breaking out my slider logic a bit and added some custom checks and a new midpoint function on the JavaScript side for showing a more accurate preview.

It also looks like `setAttribute( 'value', value )` doesn't work as well on ranges as `.value = value` so I changed that too.

It should be a lot better now.

garretlaxton (approved):
Looks good now, nice!

Screenshot 2021-09-02 at 15-42-03 Formidable Forms ‹ Formi QA — WordPress

PR #500, opened 2021-09-02
Remove captcha settings placeholders

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3162

I feel like there isn't a lot of reason to have a placeholder on these at all so I'm just removing them. I agree the placeholder we have is just doing more harm than good.

Easy peasy

**Before**
Screen Shot 2021-09-02 at 11 15 34 AM

**After**
Screen Shot 2021-09-02 at 11 15 26 AM


stephywells (approved):
Good call!

PR #501, opened 2021-09-03
Fix search filter is cut if half in Entries page when no results #3169

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3169

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -5344,7 +5344,7 @@ span.howto {
 	line-height: 200%;
 }
 
-.formidable_page_formidable-entries .no-items {
+#form_entries_page #the-list .no-items {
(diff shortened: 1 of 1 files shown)


Crabcyborg:
Thanks @truongwp this looks good but it also doesn't always work.

I happen to be using a whitelabel icon right now so I caught an issue with this class early - it isn't always used. Instead, on my page I have a class `my-plugin_page_formidable-entries` instead because my plugin's name is "My Plugin" not formidable.

I think `.frm-admin-page-entries .frm_list_entry_page` or some variation of that might be safer.

truongwp:
@Crabcyborg Thank you. I updated the CSS selector. Please check.

Crabcyborg (approved):
Looks good, thanks!

PR #502, opened 2021-09-07
Allow custom HTML for error messages when validate with JavaScript is on

Crabcyborg:
Related to https://github.com/Strategy11/formidable-pro/issues/2360 and includes part of a fix for that as well.

Also see https://github.com/Strategy11/formidable-pro/pull/3177 for a related fix (submit with AJAX).

I noticed while working on that issue that we have all of these gaps as well with our JavaScript validation, which I got confused with this at first since it's the same set of issues.

I tried to make it as performant as possible by storing the form object as a variable when we initially check for the js validation in the global var, by only checking if there are required/invalid errors embedded in the HTML, and by ignoring the default HTML so only custom HTML is added to the form.

This update adds support for custom HTML from the "Customize HTML" part of settings but also via the frm_before_replace_shortcodes filter.

_With the JavaScript validation, the frm_before_replace_shortcodes filter will always pass an empty array for $errors, which isn't something we can workaround. If you notice this during testing, nothing can really be done about it. It might make sense to mention this limitation in our documentation though._

### DIFF
--- a/classes/controllers/FrmFieldsController.php
+++ b/classes/controllers/FrmFieldsController.php
@@ -706,39 +706,48 @@ class FrmFieldsController {
 	 * @param array $add_html
 	 */
 	private static function maybe_add_error_html_for_js_validation( $field, array &$add_html ) {
-		if ( ! array_key_exists( 'custom_html', $field ) ) {
+		$form = FrmForm::getOne( $field['form_id'] );
+		if ( empty( $form->options['js_validate'] ) ) {
 			return;
+		}
+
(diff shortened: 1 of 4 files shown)


garretlaxton (approved):
Adding a class to the error div tag works as should with JS. 

Screenshot 2021-09-07 at 16-54-39 hi – Formi QA

stephywells (approved):
This looks good. I'm not sure how feasible it is to pull out the js validation pieces to the pro version, so I'm good with having it here.

PR #504, opened 2021-09-08
Add style settings action and pro tooltip

Crabcyborg:
Screen Shot 2021-09-08 at 12 32 03 PM

### DIFF
--- a/classes/helpers/FrmTipsHelper.php
+++ b/classes/helpers/FrmTipsHelper.php
@@ -325,6 +325,20 @@ class FrmTipsHelper {
 		return $tip;
 	}
 
+	public static function get_bg_image_style_settings_tip() {
+		$tips = array(
+			array(
+				'link' => array(
+					'content' => 'style',
+					'param'   => 'bg-image-style-settings',
+				),
+				'tip'  => __( 'Want to add a background image?', 'formidable' ),
+				'call' => __( 'Upgrade to Pro.', 'formidable' ),
+			),
+		);
+		return $tips;
+	}
+
 	public static function get_random_tip( $tips ) {
 		$random = rand( 0, count( $tips ) - 1 );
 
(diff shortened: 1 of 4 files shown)


classes/views/styles/_general.php
```diff
@@ -42,6 +42,14 @@
 	<input type="text" name="<?php echo esc_attr( $frm_style->get_field_name( 'fieldset_bg_color' ) ); ?>" id="frm_fieldset_bg_color" class="hex" value="<?php echo esc_attr( $style->post_content['fieldset_bg_color'] ); ?>" size="4" />
 </p>
 
+<?php
+if ( FrmAppHelper::pro_is_installed() ) {
```
stephywells, line 46: Is it worth including the pro version check here? Or can be include the hook either way?

css/frm_admin.css
```diff
@@ -2415,7 +2415,7 @@ li[data-preview] .frm-hover-icons .frm-preview-form,
 	height: 50px;
 }
 
-.frm-right-panel .inside a.frm_pro_tip,
+.frm-right-panel .accordion-section-content .inside a.frm_pro_tip,
```
stephywells, line 2418: This would look a bit better with the white background used on .frm_field_list .frm_pro_tip

classes/helpers/FrmTipsHelper.php
```diff
@@ -318,6 +318,20 @@ public static function get_banner_tip() {
 		return $tip;
 	}
 
+	public static function get_bg_image_style_settings_tip() {
```
stephywells, line 321: Maybe we should add this to the get_styling_tip function instead of adding it in a new spot? That will rotate the tips randomly.

Crabcyborg: @stephywells are you thinking we remove it from the sidebar entirely then? I wouldn't want the same tip in two places. I just thought it would make sense to include it in the same spot as the feature, especially since it would probably be the first place you'd look to set a background image and you might just think we don't support the feature if it's just not there.

Would we prefer a placeholder upload that triggered a pop up like I did with the AJAX submit?

If I put frm_noallow on the button it looks like
Screen Shot 2021-09-22 at 12 40 49 PM

stephywells: Yes! Love it. I think it was just the style of the tip that I didn't like.

stephywells (approved):
Awesome! Looks great!

PR #509, opened 2021-09-09
Add a grayed out submit with AJAX option with tooltip

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3150

I bumped the old tooltip up into the other section, at the bottom, then added a new one that is more in-position and targeted with a placeholder checkbox.

**Before**
Screen Shot 2021-09-09 at 4 38 40 PM

**After**
Screen Shot 2021-09-13 at 12 06 32 PM
Screen Shot 2021-09-13 at 12 06 35 PM

Adding a couple generic classes because I keep wanting an easier way to set some basic styles like tachyons/tailwind. Adding just a few, that I've wanted pretty often.

### DIFF
--- a/classes/views/frm-forms/settings-advanced.php
+++ b/classes/views/frm-forms/settings-advanced.php
@@ -141,11 +141,10 @@ if ( ! defined( 'ABSPATH' ) ) {
 	<?php if ( ! FrmAppHelper::pro_is_installed() ) { ?>
 		<tr>
 			<td>
-				<label>
+				<label data-upgrade="<?php esc_attr_e( 'AJAX Form Submissions', 'formidable' ); ?>" data-medium="ajax" data-message="<?php esc_attr_e( 'Want to submit forms without reloading the page?', 'formidable' ); ?>">
 					<input type="checkbox" disabled="disabled" />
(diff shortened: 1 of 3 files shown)


classes/views/frm-forms/settings-advanced.php
```diff
@@ -136,6 +138,18 @@
 		</td>
 	</tr>
 	<?php do_action( 'frm_add_form_ajax_options', $values ); ?>
+	<?php if ( ! FrmAppHelper::pro_is_installed() ) { ?>
+		<tr>
+			<td>
+				<label>
+					<input type="checkbox" disabled="disabled" />
```
stephywells, line 145: Maybe this would be better if we remove the Ajax tip and show the message when the box is checked? It would be nice to have something happen when the box is checked whether we keep the extra tip or not.

stephywells (commented):
Nice! Great addition!

classes/views/frm-forms/settings-advanced.php
Crabcyborg: Thanks @stephywells I made some adjustments so the label triggers the upgrade pop up instead!

stephywells (approved):
Perfect! Thank you!

PR #511, opened 2021-09-13
Add save and reload to settings page after plugin activation

Crabcyborg:
https://recordit.co/5zjic0q2aC

Setting up the landing page plugin, I find the flow really jarring that I have to activate it and then close the pop up and save myself.

I've added a check, if we're activating a plugin from the settings page, I've added an option to trigger the save button from the pop up.

This update also removed some duplicate strings and moves some var declarations to the top of the function.

We should probably do this on other pages too but settings seems to cover most of these.


stephywells (approved):
Awesome! This has been bugging me for a long time. Great quick and simple solution.

PR #512, opened 2021-09-13
Fix inbox visibility assignment issue

Crabcyborg:
I was testing messages and noticed that there's data coming from the API but I never see the message.

It looks like the who value I'm seeing is an array with an everyone option:

```
[who] => Array
        (
            [0] => everyone
        )
```

I'm adding to the check so that passes. It seems to expect an 'all' message instead, but this must have changed at some point.

With this update I actually see the new message.
Screen Shot 2021-09-13 at 2 47 29 PM

### DIFF
--- a/classes/models/FrmInbox.php
+++ b/classes/models/FrmInbox.php
@@ -160,13 +160,19 @@ class FrmInbox extends FrmFormApi {
 
 	/**
 	 * Show different messages for different accounts.
+	 *
+	 * @param array $message
+	 * @return bool
 	 */
 	private function is_for_user( $message ) {
 		if ( ! isset( $message['who'] ) || $message['who'] === 'all' ) {
 			return true;
 		}
-
-		return in_array( $this->get_user_type(), (array) $message['who'] );
+		$who = (array) $message['who'];
+		if ( in_array( 'everyone', $who, true ) ) {
+			return true;
+		}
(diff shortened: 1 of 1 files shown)


classes/models/FrmInbox.php
```diff
@@ -160,13 +160,19 @@ private function is_expired( $message ) {
...
+	 * @return bool
 	 */
 	private function is_for_user( $message ) {
 		if ( ! isset( $message['who'] ) || $message['who'] === 'all' ) {
 			return true;
 		}
-
-		return in_array( $this->get_user_type(), (array) $message['who'] );
+		$who = (array) $message['who'];
+		if ( in_array( 'everyone', $who, true ) ) {
```
stephywells, line 172: Oops! Thanks for catching this @Crabcyborg !!!! I did this when I created a form for the inbox notices:
https://formidableforms.com/wp-admin/admin.php?page=formidable&frm_action=edit&id=235

It's returning 'all' in all array now. Would you mind updating this to allow for 'all' as a string, or in the array?

Crabcyborg: @stephywells got it, made that change, looks good!

stephywells (approved):
Thanks so much Mike! I added 'everyone' back in too to cover the notices that are already cached.

PR #514, opened 2021-09-15
New inbox banners

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3022

The Inbox API can now pass along two new keys, emoji and banner.

Screen Shot 2021-09-15 at 12 27 11 PM

For now any testing has to be done with fake data.

I was injecting into the current active message with:

```
$message['emoji']  = '🕺';
$message['banner'] = 'Multi-column forms are easier to create with the all-new form layout tool';
```

### DIFF
--- a/classes/models/FrmInbox.php
+++ b/classes/models/FrmInbox.php
@@ -76,7 +76,6 @@ class FrmInbox extends FrmFormApi {
 		}
 
 		foreach ( $api as $message ) {
-			$message['banner'] = 'here';
 			$this->add_message( $message );
 		}
 	}
--- a/classes/models/FrmInbox.php
+++ b/classes/models/FrmInbox.php
@@ -76,8 +76,6 @@ class FrmInbox extends FrmFormApi {
 		}
 
 		foreach ( $api as $message ) {
-			$message['emoji']  = '🥳';
-			$message['banner'] = 'Fake message'; // TODO remove this.
 			$this->add_message( $message );
 		}
 	}
--- a/classes/models/FrmInbox.php
+++ b/classes/models/FrmInbox.php
@@ -76,6 +76,7 @@ class FrmInbox extends FrmFormApi {
 		}
 
 		foreach ( $api as $message ) {
+			$message['banner'] = 'here';
 			$this->add_message( $message );
 		}
 	}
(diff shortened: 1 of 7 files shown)


stephywells:
This will be so nice!!!

Are you seeing this in the banner?

No link showing:
Screen Shot 2021-09-15 at 1 13 15 PM

On hover the button is messy:
Screen Shot 2021-09-15 at 1 13 21 PM

classes/models/FrmInbox.php
```diff
@@ -288,4 +293,33 @@ public function remove( $key ) {
 	private function update_list() {
 		update_option( $this->option, $this->messages, 'no' );
 	}
+
+	public static function maybe_show_banner() {
+		if ( empty( self::$banner_messages ) ) {
+			return;
+		}
+		$message = reset( self::$banner_messages );
```
stephywells, line 301: This might be a little more useful if we switch to the most recent message instead of the oldest.

classes/views/frm-forms/list.php
```diff
@@ -2,6 +2,7 @@
 if ( ! defined( 'ABSPATH' ) ) {
 	die( 'You are not allowed to call this page directly.' );
 }
+FrmInbox::maybe_show_banner();
```
stephywells, line 5: Would it work to include this in the admin_header file instead of just on the forms page?

stephywells (commented):
Awesome job on this! The only issue I'm seeing, is when the CTA uses the secondary classes instead of primary.

classes/views/frm-forms/list.php
Crabcyborg: Sounds good Steph, I moved that to the header.

classes/models/FrmInbox.php
Crabcyborg: Ohh, good call!

Crabcyborg:
> Are you seeing this in the banner?

@stephywells it looks like button-secondary has a lot of conflicting styles.

I've pushed an update to the banner view that replaces button-secondary classes with button-primary ones in the cta so the buttons are more consistent.

stephywells (approved):
Looks perfect! Now we just need to figure out how to save an emoji in a field. Looks like the db isn't set up for it.

PR #517, opened 2021-09-16
Update readme for 5.0.04

Crabcyborg:
@stephywells what do you think about the filters? There's a ton of them, I'm not sure if we want to list them?

Should I just go with something like "Added several new filters required to support our new surveys add on" or something instead?


stephywells:
Sure, I'm fine with that. It would still be nice to copy the list of hooks into the list of hooks that need docs. But it's probably not all that helpful to list them all here.

PR #518, opened 2021-09-16
Add frm_after_duplicate_field hook

truongwp:
This is required for the Surveys add-on

### DIFF
--- a/classes/controllers/FrmFieldsController.php
+++ b/classes/controllers/FrmFieldsController.php
@@ -127,6 +127,19 @@ class FrmFieldsController {
 		$values = apply_filters( 'frm_prepare_single_field_for_duplication', $values );
 
 		$field_id = FrmField::create( $values );
+
+		/**
+		 * Fires after duplicating a field.
+		 *
+		 * @since 5.0.04
+		 *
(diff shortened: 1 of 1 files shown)


classes/controllers/FrmFieldsController.php
```diff
@@ -127,6 +127,19 @@ public static function duplicate() {
...
+		 * Fires after duplicating a field.
+		 *
+		 * @since 5.0.04
+		 *
+		 * @param int    $field_id   New field ID.
+		 * @param array  $values     Values before inserting.
+		 * @param object $copy_field Copy field data.
+		 * @param int    $form_id    Form ID.
+		 */
+		do_action( 'frm_after_duplicate_field', $field_id, $values, $copy_field, $form_id );
```
stephywells, line 141: Let's combine these params into an array to make it easier to use.

stephywells (changes requested):

classes/controllers/FrmFieldsController.php
truongwp: I did it.

stephywells (approved):
Thank you!!

PR #521, opened 2021-09-16
Refactor to make duplicating single field code usable

stephywells:
This makes it easier to use the existing code to duplicate a single field.


Crabcyborg (approved):
Looks good!

PR #522, opened 2021-09-17
Overwrite fieldset padding on admin

Crabcyborg:
If I set my padding to 150px for example:

**Before**
Screen Shot 2021-09-17 at 1 58 21 PM

**After**
Screen Shot 2021-09-17 at 1 58 02 PM

I added a little more top padding than the default. It looks nicer I think spaced from the "Add new entry" header a bit. It used to look more like:
Screen Shot 2021-09-17 at 1 58 57 PM

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -26,6 +26,10 @@
 	--biggest-sidebar: 450px;
 }
 
+.with_frm_style .frm_form_fields > fieldset {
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Perfect!

Do we need to override the form width too? Or was that just for small screens?

Crabcyborg:
I think width is fine. It was just the horizontal padding that made the width small.

PR #523, opened 2021-09-20
Do not wrap html around validation pattern

Crabcyborg:
https://secure.helpscout.net/conversation/1634066257/81156/
And now also https://secure.helpscout.net/conversation/1636090425/81257?folderId=2445391
And looks like a third time now as well https://secure.helpscout.net/conversation/1636180030/81261?folderId=2445391

Looks like my custom error html update had a funny side effect I wasn't expecting.

`getFieldValidationMessage` is actually called in one place to get the pattern. It's not always a "validation message" as the function might lead you to think.

So the custom error html (which wasn't actually custom, it turns out that the error html sometimes doesn't include an id and I wasn't checking for that), was applying over the pattern so the pattern would break.

I also updated the default check so it would include these missing id cases better too. If that check was better, this customer wouldn't have been effected.

_This bug would effect any pattern checks with JavaScript validation on for a field that either had custom error HTML or is missing the id from the default. It's common enough that I think an early patch release might make some sense._

### DIFF
--- a/classes/controllers/FrmFieldsController.php
+++ b/classes/controllers/FrmFieldsController.php
@@ -750,9 +750,13 @@ class FrmFieldsController {
 			return false;
 		}
 
-		$error_body = substr( $custom_html, $start + 10, $end - $start - 10 );
-		if ( '<div class="frm_error" id="frm_error_field_[key]">[error]</div>' === $error_body ) {
-			// no custom HTML if the default is detected, so do nothing special.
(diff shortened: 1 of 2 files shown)


stephywells (approved):
Oooh... How soon do we need to release an update for this?

Crabcyborg:
> Oooh... How soon do we need to release an update for this?

I was thinking I'd make one today, along with a couple of the other small updated we've rounded up.

It's too common to let it go very long I think.

PR #524, opened 2021-09-20
More accessible field dropdowns

Crabcyborg:
Trying to make field dropdowns more accessible

**Related to** 
https://github.com/Strategy11/formidable-pro/issues/3172
https://github.com/Strategy11/formidable-pro/issues/3207

I can use the keyboard to delete and duplicate fields now when following VoiceOver instructions.

Tried a few things to make this nicer (giving the list an aria-label, setting a default aria-expanded=false, changing the li elements to use anchor which focus better).

### DIFF
--- a/classes/views/frm-forms/add_field.php
+++ b/classes/views/frm-forms/add_field.php
@@ -26,7 +26,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 		</a>
 
 		<div class="dropdown">
-			<a href="#" class="frm_bstooltip frm-hover-icon frm-dropdown-toggle dropdown-toggle" title="<?php esc_attr_e( 'More Options', 'formidable' ); ?>" data-toggle="dropdown" data-container="body">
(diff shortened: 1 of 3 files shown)


stephywells (approved):
Nice! Does this need extra QA?

Crabcyborg:
I think it's good, at least good enough for now. But it wouldn't hurt.

@garretlaxton do you have any available time now to go over this update yourself, test it for accessibility? I plan to do a small release shortly, and any extra testing on this would be appreciated.

Crabcyborg:
I'm just merging this as I'm preparing a release. Still feel free to still chime in @garretlaxton I can always slip in something extra in another update 🙂.

garretlaxton:
Never used a screen reader before, that was pretty neat. 

The changes worked great for me. I was able to navigate the page using my arrow keys to delete and duplicate fields. Same with required fields on the frontend, the voice told me it is a required field that needs to be edited.

Crabcyborg:
> Never used a screen reader before, that was pretty neat.
> 
> The changes worked great for me. I was able to navigate the page using my arrow keys to delete and duplicate fields. Same with required fields on the frontend, the voice told me it is a required field that needs to be edited.

Thank you for testing @garretlaxton!

They're pretty interesting, yeah. A totally different way of using the web for sure.

PR #525, opened 2021-09-20
Remove support for old grid css

Crabcyborg:
Also some pulled from pro https://github.com/Strategy11/formidable-pro/pull/3208

I think it's been long enough since we added the warning.


stephywells (changes requested):
Good call! Looks like this also needs to be removed from FrmUsage

stephywells:
Also, what are your thoughts on frm_old_grids.css? Should we drop that file too?

Crabcyborg:
> Good call! Looks like this also needs to be removed from FrmUsage

@stephywells I figured it might be safer to leave the defaults there, so had left the default for FrmUsage as well, but that one probably isn't as important.

I've removed the FrmUsage default now.

Crabcyborg:
> Also, what are your thoughts on frm_old_grids.css? Should we drop that file too?

Oops, forgot about that!

I was thinking about this least week. I've added a comment to the css file, and removed the css from it.

Just in case people expect the file to exist, I'm leaving it there.

stephywells (approved):
Looks perfect!

PR #528, opened 2021-09-21
Allow custom columns in CSV export

Crabcyborg:
This update adds basic support for downloading a custom subset of the CSV data.

So far there is no UI for this but it already works really well and could probably be useful enough for many purposes as is.

@stephywells would it make sense to put a popup in the free plugin, or to make that a pro enhancement?

An example url
`http://localhost:8889/wp-admin/admin-ajax.php?frm_action=0&action=frm_entries_csv&form=111&columns=992,994,created_at,updated_at`


stephywells:
@Crabcyborg would it be difficult to add to Pro? I think it would make the most sense to include it in Pro if possible. Then we can add the interface for it at some point.

PR #531, opened 2021-09-22
Add aria required to required radio and checkbox fields

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3207, at least how it should. We use `aria-required`, not `required` but this covers the issue.

@stephywells any idea why radio and checkbox were ever excluded here? Was it just an oversight or something?

It removed them from this exception and they work well with the `aria-required` attributes added.

### DIFF
--- a/classes/controllers/FrmFieldsController.php
+++ b/classes/controllers/FrmFieldsController.php
@@ -764,7 +764,7 @@ class FrmFieldsController {
 	}
 
 	/**
-	 * If 'required' is added to a conditionall hidden field, the form won't
+	 * If 'required' is added to a conditionally hidden field, the form won't
 	 * submit in many browsers. Check to make sure the javascript to conditionally
 	 * remove it is present if needed.
 	 *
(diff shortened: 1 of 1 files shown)


stephywells:
@Crabcyborg radio and checkboxes were excluded, because it was making every box in the field required. Has that changed?

Crabcyborg:
> @Crabcyborg radio and checkboxes were excluded, because it was making every box in the field required. Has that changed?

Since we're using aria-required nothing is required anyway. This really just lets the screen reader point out that these are in fact required.

Every field should appear as required but it looks like the `required` attribute even works fine if we set it to every single one (https://stackoverflow.com/questions/8287779/how-to-use-the-required-attribute-with-a-radio-input-field).

I played with it with JavaScript validation on and the radio and checkbox had no issues switching between aria-invalid=true/false while I was trying submitting empty/filling them in.

stephywells:
Awesome! It should be great then!

stephywells (approved):

Crabcyborg:
@garretlaxton could I get some extra eyes on this, to make sure that there aren't any issues with required radio/checkboxes submitting with this update?

garretlaxton (approved):
Tested the usual things that break these fields such as page breaks, repeater, visibility options, submit through ajax, and validate through JS. Didn't have any issues while testing these required fields.

Crabcyborg:
Awesome, thank you @garretlaxton!

PR #534, opened 2021-09-23
Make name fields more accessible

Crabcyborg:
After updating the `aria-required` attributes for radios/checkboxes (https://github.com/Strategy11/formidable-forms/pull/531) I thought I'd check our other fields.

Some of the new fields have missing attributes. It looks like they're missing this call to the `frm_field_input_html` action which is pretty important.

It makes it more extendable, adds support for the `frm_field_classes` filter, adds `aria-invalid` when JavaScript validation is turned on, adds the missing `aria-required` attribute, adds the JavaScript validation messages that are missing.

@truongwp do you see any issues with this update? Do you think anything could go wrong here?

**Before**
frm_blank_field data-sub-field-name=first event

**After**
data-sub-field-name=first

**Before (JavaScript validation on in error state)**
Required Name

**After (JavaScript validation on in error state)**
Required Name

### DIFF
--- a/classes/models/fields/FrmFieldCombo.php
+++ b/classes/models/fields/FrmFieldCombo.php
@@ -357,10 +357,10 @@ class FrmFieldCombo extends FrmFieldType {
 
 		if ( ! empty( $sub_field['optional'] ) ) {
 			$classes .= ' frm_optional';
-		} else {
-			$field['default_value'] = ''; // fake it to avoid printing frm-val attribute.
-			do_action( 'frm_field_input_html', $field );
 		}
+
+		$field['default_value'] = ''; // fake it to avoid printing frm-val attribute.
+		do_action( 'frm_field_input_html', $field );
 
 		if ( $classes ) {
 			$atts[] = 'class="' . esc_attr( $classes ) . '"';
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Good find! Let's make sure this gets tested with address fields too. I'm not sure if it affects them or not.

Crabcyborg:
> Good find! Let's make sure this gets tested with address fields too. I'm not sure if it affects them or not.

I tested address/credit card fields already since they're the other combo fields. They both seem fine. I actually got the action code here from the address code, and it helped to understand how the aria-required is set up here.

Crabcyborg:
@garretlaxton name fields should now work better with screen readers when you mark them as required.

This hook really adds quite a bit, so we could also test the JavaScript validation more, and maybe the frm_field_classes filter, and probably some other stuff.

truongwp:
@Crabcyborg I tested the functionality and checked with WAVE tool. It works great. The code also looks good. Thank you.

truongwp (approved):

garretlaxton (approved):
When the file upload field is required, the screen reader doesn't mention this.

I also found another bug related to the survey module which I'll log now.

PR #535, opened 2021-09-23
Tweak field dropdowns

Crabcyborg:
Following up on the other field dropdown accessibility update, I missed a couple details.

- It looks like I had messed up the styling on the Delete Group/Duplicate Group options because it uses a different style definition.
- It also looks like clicking just the edge of a Delete Field/Duplicate Field/Field Settings option wasn't triggering it, only if you clicked the text.
- It looks like it was still difficult to Delete/Duplicate a Field Group using a screenreader.

So I've made them use the same style class now with a new `frm_more_options_li` class, added a tabindex and focus() when the dropdown opens to help imrprove accessibility.

I also added in a fix that was on my todo list for side-by-side field enhancements, to delete tooltips when a field is deleted. I don't think there's a very effective way to target a specific tooltip, but deleting all of them doesn't seem to have a very negative effect in my experience since they will just get generated again by bootstrap.

I feel way better about them now. Hopefully eliminating most of the weird Delete/Duplicate related issues.

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -21,7 +21,7 @@ jobs:
           - php: 7.3
             wordpress: master
             phpunit: 6.5.14
-          - php: 7.0.33
+          - php: 7.1
             wordpress: 5.8
             phpunit: 6.5.14
 
--- a/classes/views/frm-forms/add_field.php
+++ b/classes/views/frm-forms/add_field.php
@@ -21,7 +21,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 			<a href="#" class="frm-collapse-section frm-hover-icon frm_icon_font frm_arrowdown6_icon" title="<?php esc_attr_e( 'Expand/Collapse Section', 'formidable' ); ?>"></a>
 		<?php } ?>
 
-		<a href="#" class="frm_bstooltip frm-move frm-hover-icon" title="<?php esc_attr_e( 'Move Field', 'formidable' ); ?>" data-container="body">
+		<a href="#" class="frm_bstooltip frm-move frm-hover-icon" title="<?php esc_attr_e( 'Move Field', 'formidable' ); ?>" data-container="body" aria-label="<?php esc_attr_e( 'Move Field', 'formidable' ); ?>">
 			<?php FrmAppHelper::icon_by_class( 'frm_icon_font frm_thick_move_icon' ); ?>
 		</a>
 
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -6762,23 +6762,35 @@ ul .frm_col_two {
 	font-size: 15px;
 }
 
-#frm_field_group_controls .frm-dropdown-menu .frm_dropdown_li,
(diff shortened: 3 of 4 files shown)


garretlaxton (approved):
This is definitely a lot cleaner now. 

I do see that the `ROW LAYOUT` dropdown doesn't work with the screen reader.
Screenshot 2021-09-23 at 15-05-21 Formidable Forms ‹ Formi QA — WordPress

Crabcyborg:
Good catch, I'll give that a go, thanks @garretlaxton!

garretlaxton:
Thanks champ! The screen reader also doesn't work too well with the `move field`. It doesn't say anything about moving the field. It also acts this way in the group section as well.

Crabcyborg:
Yeah I'm okay neglecting the moving for now. It isn't really supported without a mouse yet.

Thanks!

Crabcyborg:
@garretlaxton you can give it another go now I made a few improvements. It should now be possible to select one of the row layout options, or set a custom layout, or break the row, using only the keyboard.

I added a label to the move options so it's easier to identify with the keys without seeing what it is but it doesn't do anything if you try to trigger a click event.

garretlaxton:
@Crabcyborg I'm not able to use my keyboard to navigate through the row layout options. 

Also, if you click `row layout` or `more options` for a group then close the dropdown, it's quite hard to open the dropdown again. This only happens when the screen reader is on. I originally thought this was just a screen reader issue, but this works fine with the single field `more options` button.

And one little thing is the row layout & move field group buttons don't have a label when you hover them. When hovering those buttons you should have a label appear.

Crabcyborg:
Thanks @garretlaxton I pushed another update.

I was testing a lot with ctrl + option + arrow keys which gives pretty good control sometimes when tabbing doesn't. But using just tab, it was hard to select the field group options, you're right. I gave it a group role and a tabindex now and it seems to take focus properly now.

And good catch with the missing tooltips. Definitely an inconsistency, those have been added now!

I couldn't replicate anything with requiring a double click. It seems to work pretty well on my end. Maybe this update helped though. If there are more issues though, I think this update is probably enough for now. Anything else, probably, we can log for the future I think.

garretlaxton:
@Crabcyborg I use the arrow keys as well, I wasn't aware I could use the tab key. I'm still having trouble using the arrow keys with just the row layout, the tab key works fine though.

Here is a quick clip of what I mean by the group buttons are difficult to click: https://www.loom.com/share/dc3c589b864446c0a3a5f7da76312f09
You can see me single-click the more options button in the email field easily, but the group buttons have to be clicked a few times. This only happens when the screen reader is on.

Crabcyborg:
Hm, thanks Garret. The arrow keys work for me in that pop up as well.

I've using VoiceOver on my mac. I feel like what you're experiencing might just be differences between screen readers. It doesn't look like the dropdown is focused before you clicked on it, which would explain things. Similar to the row layout option, I seem to sometimes be on the group holding the options first, then I need to tab into the group, to the first option, before I click it.

PR #536, opened 2021-09-23
Add data-requires to actions with Business/Elite in the category list

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3228

Super easy, the api data already includes a categories array that includes "Business" and "Elite" so the package can be really easily determined already 🚀.

Screen Shot 2021-09-23 at 1 04 40 PM
Screen Shot 2021-09-23 at 1 04 54 PM
Screen Shot 2021-09-23 at 1 04 58 PM

Screen Shot 2021-09-23 at 7 01 06 PM

Screen Shot 2021-09-23 at 7 00 47 PM

Screen Shot 2021-09-23 at 7 00 54 PM

### DIFF
--- a/classes/controllers/FrmFormActionsController.php
+++ b/classes/controllers/FrmFormActionsController.php
@@ -252,7 +252,7 @@ class FrmFormActionsController {
 		if ( ! isset( $upgrading['categories'] ) || ! is_array( $upgrading['categories'] ) ) {
 			return false;
 		}
-		$plans = array( 'Business', 'Elite', 'Personal' );
+		$plans = array( 'Business', 'Elite', 'Personal', 'Plus' );
 		foreach ( $upgrading['categories'] as $category ) {
 			if ( in_array( $category, $plans, true ) ) {
 				return $category;
--- a/classes/controllers/FrmFormActionsController.php
+++ b/classes/controllers/FrmFormActionsController.php
@@ -252,10 +252,13 @@ class FrmFormActionsController {
 		if ( ! isset( $upgrading['categories'] ) || ! is_array( $upgrading['categories'] ) ) {
 			return false;
 		}
-		$plans = array( 'Business', 'Elite' );
+		$plans = array( 'Business', 'Elite', 'Personal' );
 		foreach ( $upgrading['categories'] as $category ) {
 			if ( in_array( $category, $plans, true ) ) {
 				return $category;
+			}
+			if ( 'Creator' === $category ) {
+				return 'Personal';
 			}
 		}
 		return false;
--- a/classes/controllers/FrmFormActionsController.php
+++ b/classes/controllers/FrmFormActionsController.php
@@ -252,13 +252,14 @@ class FrmFormActionsController {
 		if ( ! isset( $upgrading['categories'] ) || ! is_array( $upgrading['categories'] ) ) {
 			return false;
 		}
-		$plans = array( 'Business', 'Elite', 'Personal', 'Plus' );
+		$plans      = array( 'Business', 'Elite', 'Basic' );
+		$plus_plans = array( 'Creator', 'Personal', 'Plus' );
 		foreach ( $upgrading['categories'] as $category ) {
 			if ( in_array( $category, $plans, true ) ) {
 				return $category;
 			}
-			if ( 'Creator' === $category ) {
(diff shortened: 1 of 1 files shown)


classes/controllers/FrmFormActionsController.php
```diff
@@ -234,7 +239,26 @@ public static function show_action_icon_link( $action_control, $allowed ) {
...
+	 * @since 5.0.06
+	 *
+	 * @param array $upgrading
+	 * @return string|false
+	 */
+	private static function action_requires( $upgrading ) {
+		if ( ! isset( $upgrading['categories'] ) || ! is_array( $upgrading['categories'] ) ) {
+			return false;
+		}
+		$plans = array( 'Business', 'Elite' );
```
stephywells, line 255: What happens with Business+Views or something else with +Views?

stephywells (commented):
Nice!! 'Personal' is the same as 'Plus':
https://github.com/Strategy11/formidable-forms/blob/master/classes/helpers/FrmFormsHelper.php#L1394

But it looks like we need to get rid of the creator tag. Want to update that on our site?

classes/controllers/FrmFormActionsController.php
Crabcyborg: These are just categories with action data. I don't think any actions require views?

stephywells: The Category on the Views is Basic+Views, and Directories are Business+Views. Can we handle these too?

Crabcyborg: > Nice!! 'Personal' is the same as 'Plus':
> https://github.com/Strategy11/formidable-forms/blob/master/classes/helpers/FrmFormsHelper.php#L1394
> 
> But it looks like we need to get rid of the creator tag. Want to update that on our site?

So right now there's no sign of a "Personal" or "Plus" on this information at all, or Basic. I'm really asking about Creator because that's all there is here.

I looked in our github for Creator references and it seems sort of off too, like it's filling in as Basic but Basic licenses don't get Aweber but it's flagged as Creator.

It's tough to follow.

Where would I update that on our site? I'm still pretty unsure how that all works.

stephywells: I just took another look at this. Creator is Personal now. Basic doesn't include any add-ons (except Formidable Pro)
https://formidableforms.com/wp-admin/edit.php?s&post_status=all&post_type=download&action=-1&download_category=basic&filter_action=Filter&paged=1&action2=-1

We've ended up with different category names for the templates though. It would be nice to consolidate them. It just gets a bit tricky when it comes to coordinating the plugin code with the categories on the products.

So there are categories for:
- add-ons (Basic, Creator, Business, Elite, Basic+Views, Business+Views)
- form templates (free, Basic, Personal, Business, Elite)
- style templates (Basic, Personal, Business, Elite)
- view templates (Personal+Views, Business+Views, Elite+Views)

If we add Personal everywhere that Creator is set now, Would that get weird in the plugin? Would it show both tags?

Crabcyborg: Thanks!

So I double checked our pricing and everything, and it looks like `Creator` and `Personal` should really both appear as `Plus`. So I've set up a `$plus_plans` check that returns `Plus` if any Creator, Personal, or Plus is found.

And I set it to also check for Basic if we want to start including that category as well but I saw nothing currently that had it. If it finds nothing, it will fallback to Pro.

stephywells: That sounds Perfect!

stephywells: Is this new function a duplicate of get_plan_required?

Crabcyborg: It's similar.

`get_plan_required` actually unsets categories, which I'd rather avoid, and with the inconsistencies between categories everywhere it does have different rules.

Like I don't see Creator mentioned in this function and it's likely not necessary there. And I don't think "free" is ever important with the actions check.

We'd probably benefit to consolidate better in the future but I think it's fine for now.

stephywells: Sounds good to me.

stephywells (approved):

PR #539, opened 2021-09-24
Unset margin on block select caret svg

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3190

The rule above this new rule:
```
.frm-block-intro-content svg {
	margin-top: 32px;
	margin-bottom: 12px;
}
```

Is causing issues. I'm undoing that and it seems to do the trick.

Screen Shot 2021-09-24 at 11 00 48 AM

Screen Shot 2021-09-24 at 10 59 09 AM

### DIFF
--- a/css/frm_blocks.css
+++ b/css/frm_blocks.css
@@ -28,6 +28,10 @@
 	color: #4d4d4d;
 }
 
+.frm-block-intro-content select + span svg {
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Nice! Looks great!

PR #540, opened 2021-09-24
Fix issues with failing unit tests

Crabcyborg:
Trying to capture 2 pretty big issues with our unit tests here.

A lot of tests fail with this `The following file is not importing correctly: https://s3.amazonaws.com/fp.strategy11.com/images/knowledgebase/global-settings_enter-license1.png` message lately. I think, because we have more tests than before, that we're hitting rate limits trying to get files from our CDN.

These are grabbed every time setUp is called which can be quite often. I tested running the pro unit tests locally and logging how many times this avoided a download and I got _186 hits_. That should hopefully help speed up unit tests too if we don't have to download files from an external server as often.

```
1) test_FrmProFieldRange
The following file is not importing correctly: https://s3.amazonaws.com/fp.strategy11.com/images/knowledgebase/global-settings_enter-license1.png
Failed asserting that false is true.

/tmp/wordpress/src/wp-content/plugins/formidable/tests/base/FrmUnitTest.php:193
/tmp/wordpress/src/wp-content/plugins/formidable/tests/base/FrmUnitTest.php:87
/tmp/wordpress/src/wp-content/plugins/formidable/tests/base/FrmUnitTest.php:23
/tmp/wordpress/tests/phpunit/includes/abstract-testcase.php:76
```

And the other big issue:

```
Error: The PHPUnit Polyfills library is a requirement for running the WP test suite. 
```

Which introduced new issues:

```
PHP Fatal error:  Declaration of FrmUnitTest::setUp() must be compatible with Yoast\PHPUnitPolyfills\TestCases\TestCase::setUp(): void
```

The addition of the `: void` piece is sort of annoying. It means that we need to drop tests in PHP 5.6 because 5.6 doesn't support that feature. I also think we'll likely have to stop running PHPUnit against any free branches but master for a while because of these breaking changes with WordPress/PHPUnit requirements.

And because WordPress < 5.8 doesn't support any of this either, I had to drop our WordPress 4.9 compatibility tests too.

Unfortunately this means there's less coverage for a lot of backwards compatibility stuff but I don't think there's any way around this.

```
Parse error: syntax error, unexpected ':', expecting ';' or '{' in /tmp/wordpress/src/wp-content/plugins/formidable/tests/base/FrmUnitTest.php on line 42
```

And then after all of that, a lot of `assertContains`/`assertNotContains` checks broke because they're only supposed to get called on arrays but they're being used on strings in a few places.
```
14) test_FrmStylesController::test_front_head
TypeError: Argument 2 passed to PHPUnit\Framework\Assert::assertContains() must be iterable, string given, called in /tmp/wordpress/src/wp-content/plugins/formidable/tests/styles/test_FrmStylesController.php on line 31
```

And it looks like I did break a unit test with name fields when I moved the class. I've updated the test so the action gets called, and I moved the class in the check because it happens earlier in the string now and there are extra spaces added to the HTML than there were before.

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -22,7 +22,7 @@ jobs:
             wordpress: master
             phpunit: 6.5.14
           - php: 7.0.33
-            wordpress: 4.9
+            wordpress: 5.8
             phpunit: 6.5.14
 
     name: PHP ${{ matrix.php }} tests in WP ${{ matrix.wordpress }}
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -24,9 +24,6 @@ jobs:
           - php: 7.0.33
             wordpress: 4.9
             phpunit: 6.5.14
-          - php: 5.6.20
-            wordpress: 4.7
-            phpunit: 5.7.27
 
     name: PHP ${{ matrix.php }} tests in WP ${{ matrix.wordpress }}
     steps:
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -16,10 +16,10 @@ jobs:
       matrix:
         include:
           - php: 7.4
-            wordpress: 5.8
+            wordpress: master
             phpunit: 6.5.14
           - php: 7.3
-            wordpress: 5.8
+            wordpress: master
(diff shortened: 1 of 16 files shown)


stephywells (approved):
Awesome!! Thanks for getting all of this updated! Looks great!

PR #541, opened 2021-09-27
Add frm_fields_to_validate hook

truongwp:
This is required by Surveys add-on


Crabcyborg (approved):
Thanks @truongwp! This seems like a good solution since we can't really determine this with the other `frm_posted_field_ids` filter ahead of time.

Issue #551, opened 2021-10-04
Akismet API requests are made even if Akismet checking is disabled for the form


cfinke:
On line 258 of `FrmEntryValidate.php`:

`
} elseif ( self::is_akismet_spam( $values ) && self::is_akismet_enabled_for_user( $values['form_id'] ) ) {
`

Because `is_akismet_spam()` is called before `is_akismet_enabled_for_user()`, the API request will be made even if the `Use Akismet to check entries for spam for ...` setting is set to `no one`.

Switching the order of these two function calls should fix the bug.

Crabcyborg:
Thank you @cfinke you're totally right that the order of those conditions is backwards from what it should be. I've made a pull request for this https://github.com/Strategy11/formidable-forms/pull/552

Should be in next release.

PR #552, opened 2021-10-05
Check akismet setting before calling akismet

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-forms/issues/551

Also dropping a redundant isset here.


stephywells (approved):
Good call!

PR #554, opened 2021-10-05
Check unfiltered html permission before allowing unfiltered form/field data

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3250 though there maybe other places to filter as well that I missed.

For now I'm filtering field label, description, and custom_html.
As well as form name, description, submit text, success message, before html, after html, and submit html.

This is only if:
- The DISALLOW_UNFILTERED_HTML constant is on
or 
- The user logged in does not have the unfiltered_html permission

A super admin shouldn't notice a difference unless the constant was on.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -782,18 +782,18 @@ class FrmAppHelper {
 				'xlink:href' => true,
 			),
 			'ul'         => $allow_class,
-			'label' => array(
-				'for'   => true,
+			'label'      => array(
+				'for'    => true,
 				'class' => true,
 				'id'    => true,
 			),
-			'button' => array(
+			'button'     => array(
 				'class' => true,
 				'type'  => true,
 			),
-			'legend' => array(
+			'legend'     => array(
 				'class' => true,
-			)
+			),
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


stephywells (approved):
Thanks so much Mike! This looks like a great solution!

Crabcyborg:
@garretlaxton let me know if you can find anything else that I missed! I plan to merge this shortly.

Crabcyborg:
@stephywells looks like kses strips some default HTML.

I've added a few new exceptions to fix this (allow legend+class, label+class,for,id, button+class,type, and adding aria-hidden to span. Any others you can think of?

garretlaxton (approved):
Looks good here. There are definitely a few more spots we could use this throughout the plugin. I'll start logging those this week. Fields look fine as of now.
Screenshot 2021-10-05 at 15-25-37 Formidable Forms ‹ Formi QA — WordPress

stephywells:
> @stephywells looks like kses strips some default HTML.
> 
> I've added a few new exceptions to fix this (allow legend+class, label+class,for,id, button+class,type, and adding aria-hidden to span. Any others you can think of?

Thanks for catching that! I've been finding these as I try things and they don't work. There's likely more that will come up, but this looks great for now.

PR #563, opened 2021-10-06
Stop calling files remotely

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3254

Just bringing files from our remote urls locally. I think this covers everything.


stephywells (approved):
Perfect!

PR #564, opened 2021-10-06
Remove old load_wp function

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3255 by removing code we don't need any longer.


stephywells (approved):
Perfect! Let's be sure to include this in the readme to let people know who might be using that code example.

PR #568, opened 2021-10-07
Remove jquery placeholder from combined js code

Crabcyborg:
This line confuses me a bit but I noticed a warning in a log and looked wanted to fix this.

Level All

It doesn't look like this file exists anywhere, and I really couldn't even find a tag when it ever did.


stephywells (approved):
Nice! That got missed from the days before browser placeholders. Thank you!

PR #569, opened 2021-10-07
Add psalm code analysis workflow

Crabcyborg:
Finally setting this up automatically.

I've been playing with this locally for a while and have fixed some of the issues as I've played with it.

This pull request gets us to a passing state. I'm suppressing some stuff but not too much. We can continue to improve it by suppressing less over time.


.github/workflows/psalm.yml
```diff
@@ -0,0 +1,15 @@
+on: [push]
```
stephywells, line 1: Can we switch this to only run when in a PR so we save some minutes?

classes/helpers/FrmListHelper.php
```diff
@@ -176,7 +176,7 @@ public function prepare_items() {
 	/**
 	 * @since 3.0
 	 */
-	protected function get_param( $args ) {
+	protected static function get_param( $args ) {
```
stephywells, line 179: Do you think it would be better to switch all of the self::get_param() function to $this->get_param()?

stephywells (changes requested):
Cool! This looks great!

PR #572, opened 2021-10-11
Quizzes add-on v2

truongwp:
New hooks:

`frm_entries_show_args`
`frm_formatted_entry_values_content`
`frm_entry_formatter_format`
`frm_prepend_and_or_where`


Crabcyborg (approved):
Looks good to me! Thanks @truongwp

classes/views/frm-entries/show.php
```diff
@@ -38,21 +38,31 @@
...
+
+					/**
+					 * Allows modifying the arguments when showing entry in the Entries page.
+					 *
+					 * @since 5.0.16
+					 *
+					 * @param array $show_args The arguments.
+					 * @param array $args      Includes `form`.
+					 */
+					$show_args = apply_filters( 'frm_entries_show_args', $show_args, compact( 'form' ) );
```
Crabcyborg, line 63: @truongwp it looks like `frm_entries_show_args` is never actually used in quizzes? Is it actually required?

This seems like it could be pretty useful either way, I'm just trying to figure out if it's necessary. I've already made the release with it so it's probably going to stay now regardless.

truongwp: I had used it but it's not required for Quizzes now.

PR #573, opened 2021-10-12
Get less data from database when loading pages dropdown fix

Crabcyborg:
Related to https://secure.helpscout.net/conversation/1656485914/82787?folderId=4324209

A user with 1 form that has no entries is hitting memory usage limits in settings because of this call to `get_pages`.

They only have 30 pages so it isn't trying to use autocomplete. However, I think some of these pages have a ton of data in their post content and all of that is being loaded for every page, which is using up a ton of memory.

Since the other function is public I can't really just replace it without risking breaking something, so I added an extra function in addition that only gets the id and title using our FrmDb code instead. It takes memory usage down a ton.

Since this can be so useful for improving performance I made it public so if someone is using the other function, they have the option to possibly move to a faster option (unless they need post_content).

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1034,6 +1034,7 @@ class FrmAppHelper {
 
 	/**
 	 * Gets all post from a specific post type.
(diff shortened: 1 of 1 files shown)


stephywells (approved):
This looks great! Good call on making it public. I think might have other other get_pages calls around that don't need all the data.

PR #575, opened 2021-10-14
Set antispam token on new form creation

Crabcyborg:
Part of https://github.com/Strategy11/formidable-pro/issues/3260

Because of caching-related issues I can never just turn this on and change the actual default with likely breaking a ton of forms. I have a few other items in my list to help promote this feature for those old forms.

But this part is easy and makes a big difference to change the default on new forms.


stephywells (approved):
Great call!

PR #579, opened 2021-10-18
Name field is not used as commenter name in Akismet spam check

truongwp:
This fixes https://github.com/Strategy11/formidable-pro/issues/3265

### DIFF
--- a/classes/models/FrmEntryValidate.php
+++ b/classes/models/FrmEntryValidate.php
@@ -521,7 +521,7 @@ class FrmEntryValidate {
 	 * @return bool
 	 */
 	private static function is_akismet_guest_info_value( $key, $value, $field_id, $name_field_ids ) {
-		if ( ! $value ) {
+		if ( ! $value || is_numeric( $value ) ) {
 			return false;
 		}
 
@@ -537,7 +537,7 @@ class FrmEntryValidate {
 					// If there is name field in the form, we should always use it as author name.
 					return in_array( $field_id, $name_field_ids, true );
 				}
-				return ! is_numeric( $value ) && strlen( $value ) < 200;
+				return strlen( $value ) < 200;
 		}
 
 		return false;
--- a/classes/models/FrmEntryValidate.php
+++ b/classes/models/FrmEntryValidate.php
@@ -535,7 +535,11 @@ class FrmEntryValidate {
 				return 0 === strpos( $value, 'http' );
 
 			case 'comment_author':
-				return in_array( $field_id, $name_field_ids, true ) || ! is_numeric( $value ) && strlen( $value ) < 200;
+				if ( $name_field_ids ) {
+					// If there is name field in the form, we should always use it as author name.
+					return in_array( $field_id, $name_field_ids, true );
+				}
+				return ! is_numeric( $value ) && strlen( $value ) < 200;
 		}
 
 		return false;
--- a/classes/models/FrmEntryValidate.php
+++ b/classes/models/FrmEntryValidate.php
@@ -391,8 +391,12 @@ class FrmEntryValidate {
 	 */
 	private static function parse_akismet_array( &$datas, $values ) {
 		self::add_site_info_to_akismet( $datas );
-		self::add_user_info_to_akismet( $datas, $values );
 		self::add_server_values_to_akismet( $datas );
+
+		$form_ids           = self::get_all_form_ids_and_flatten_meta( $values );
+		$values['form_ids'] = $form_ids;
+
+		self::add_user_info_to_akismet( $datas, $values );
 		self::add_comment_content_to_akismet( $datas, $values );
 	}
 
@@ -441,31 +445,64 @@ class FrmEntryValidate {
 			$values = array_filter( $values );
 
 			$datas['frm_duplicated'] = array();
+
+			// Get comment author email.
 			foreach ( $values as $index => $value ) {
-				// If is a name array, convert to string.
 				if ( is_array( $value ) ) {
-					if ( ! is_numeric( $index ) ) {
-						continue;
+					foreach ( $value as $sub_value ) {
+						if ( strpos( $sub_value, '@' ) && is_email( $sub_value ) ) {
+							$datas['comment_author_email'] = $sub_value;
+							$datas['frm_duplicated'][]     = $index;
+							break 2;
+						}
 					}
-
-					if ( 'name' !== FrmField::get_type( $index ) ) {
-						continue;
-					}
-
-					$value = implode( ' ', $value );
+					continue;
 				}
 
-				if ( ! is_array( $value ) ) {
-					if ( $datas['comment_author_email'] == '' && strpos( $value, '@' ) && is_email( $value ) ) {
-						$datas['comment_author_email'] = $value;
-						$datas['frm_duplicated'][] = $index;
-					} elseif ( $datas['comment_author_url'] == '' && strpos( $value, 'http' ) === 0 ) {
-						$datas['comment_author_url'] = $value;
-						$datas['frm_duplicated'][] = $index;
-					} elseif ( $datas['comment_author'] == '' && ! is_numeric( $value ) && strlen( $value ) < 200 ) {
-						$datas['comment_author'] = $value;
-						$datas['frm_duplicated'][] = $index;
+				if ( strpos( $value, '@' ) && is_email( $value ) ) {
+					$datas['comment_author_email'] = $value;
+					$datas['frm_duplicated'][]     = $index;
+					break;
+				}
+			}
+
+			// Get comment author URL.
+			foreach ( $values as $index => $value ) {
+				if ( is_array( $value ) ) {
+					foreach ( $value as $sub_value ) {
+						if ( strpos( $sub_value, 'http' ) === 0 ) {
+							$datas['comment_author_url'] = $sub_value;
+							$datas['frm_duplicated'][]   = $index;
+							break 2;
+						}
 					}
+					continue;
+				}
+
+				if ( strpos( $value, 'http' ) === 0 ) {
+					$datas['comment_author_url'] = $value;
+					$datas['frm_duplicated'][]   = $index;
+					break;
+				}
+			}
+
+			// Get comment author name.
+			foreach ( $values as $index => $value ) {
+				if ( is_array( $value ) ) {
+					foreach ( $value as $sub_value ) {
+						if ( ! is_numeric( $sub_value ) && strlen( $sub_value ) < 200 ) {
+							$datas['comment_author']   = $sub_value;
+							$datas['frm_duplicated'][] = $index;
+							break 2;
+						}
+					}
+					continue;
+				}
+
+				if ( ! is_numeric( $value ) && strlen( $value ) < 200 ) {
+					$datas['comment_author']   = $value;
+					$datas['frm_duplicated'][] = $index;
+					break;
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


classes/models/FrmEntryValidate.php
```diff
@@ -440,7 +440,20 @@ private static function get_spam_check_user_info( $values ) {
...
 			$values = array_filter( $values );
-			foreach ( $values as $value ) {
+			foreach ( $values as $field_id => $value ) {
+				// If is a name array, convert to string.
+				if ( is_array( $value ) ) {
+					if ( ! is_numeric( $field_id ) ) {
+						continue;
+					}
+
+					if ( 'name' !== FrmField::get_type( $field_id ) ) {
```
Crabcyborg, line 450: @truongwp I think in this case we should probably avoid more database stuff.

If you look at the logic beneath we're just trying to guess what field is a name right now from all meta. So we're going through all meta again.

I think, in this case, we should go with a similar update to the other one. I think we should check `$values['form_id'] for a type='name'` before this loop. Then we could use our `$name_field_id`, if we have one to confirm if the meta is a name or not.

Names are actually really important here. We really should always set `comment_author` if the data is a name field. Before, this was a lot harder to determine, and we should of course fallback to the other option if we can't find a name field as well.

Crabcyborg: Another thought. This brings us right back to the other issue with embedded/repeater forms possibly not getting caught.

I think we should probably pull those form ids out before both of these functions are called, then pass those form ids as arguments so we don't have to check for the form ids twice.

Crabcyborg: I'm also thinking we should break this loop if we find a name, url, and email early. This could be doing way more than necessary. If we even stopped checking for names after the first name, that would be great.

truongwp: @Crabcyborg Why don't we just check if the value is an array contains `first` and `last`, that's name field? It's already been in my mind. We will loop through repeater/embedded to guess the name field too.

truongwp:
I think that we should run `get_all_form_ids_and_flatten_meta()` before `get_spam_check_user_info()`, so we don't need to loop through the embedded|repeater value again. How do you think?

truongwp:
@Crabcyborg Thank you. I also feel like we need to improve that code. I thought that splitting into 3 loops could reduce the number of checks. But look like I was wrong. Let me update the code.

truongwp:
@Crabcyborg I updated the code. Please check if it works as you want. I will clean and fix phpcs errors later.

classes/models/FrmEntryValidate.php
```diff
@@ -530,6 +507,47 @@ private static function get_spam_check_user_info_for_guest( $values ) {
...
+	 * @param array  $values   The values array.
+	 * @return bool Return `true` if email, url and name are found.
+	 */
+	private static function found_akismet_spam_check_value( $name, &$datas, $value, $field_id, $values ) {
+		switch ( $name ) {
+			case 'comment_author_email':
+				if ( strpos( $value, '@' ) && is_email( $value ) ) {
+					$datas[ $name ]            = $value;
+					$datas['frm_duplicated'][] = $field_id;
+				}
```
Crabcyborg, line 528: This would still check `! empty( $datas['comment_author_email'] ) && ! empty( $datas['comment_author_url'] ) && ! empty( $datas['comment_author'] );` on every iteration, even when nothing is found. I think it would make sense if these if statements had an `else` at least, so we could `return false` as soon as we do the check and it is not a match.

classes/models/FrmEntryValidate.php
```diff
@@ -470,58 +470,35 @@ private static function get_spam_check_user_info_for_guest( $values ) {
...
-						}
-						continue;
+					if ( self::found_akismet_spam_check_value( 'comment_author_url', $datas, $sub_value, $index, $values ) ) {
+						break 2;
 					}
 
-					if ( ! $datas['comment_author'] && self::maybe_comment_author_name( $sub_value ) ) {
-						$datas['comment_author']   = $sub_value;
-						$datas['frm_duplicated'][] = $index;
+					if ( self::found_akismet_spam_check_value( 'comment_author', $datas, $sub_value, $index, $values ) ) {
```
Crabcyborg, line 483: I think we should really keep the `! $datas['comment_author']` checks (and all of the other keys) in here. Once we find one, we do not need to do any more checks at all. It looks like that check is gone now, so if we found a name, then found some comment, I think the comment would overwrite the name.

classes/models/FrmEntryValidate.php
```diff
@@ -530,6 +507,47 @@ private static function get_spam_check_user_info_for_guest( $values ) {
...
+
+			case 'comment_author_url':
+				if ( 0 === strpos( $value, 'http' ) ) {
+					$datas[ $name ]            = $value;
+					$datas['frm_duplicated'][] = $field_id;
+				}
+				break;
+
+			case 'comment_author':
+				if ( isset( $values['name_field_ids'] ) && in_array( $field_id, $values['name_field_ids'], true ) ||
```
Crabcyborg, line 539: I think we should be stronger on this check. We don't really want `! is_numeric( $value ) && strlen( $value ) < 200` ever getting used if there are `name_field_ids`, I don't think. That's really just good for when there are no name fields.

I think we might want to do something like.

```
$is_name_field = false;
if ( empty(  $values['name_field_ids'] ) ) {
    $is_name_field = ! is_numeric( $value ) && strlen( $value ) < 200;
} else {
    $is_name_field = in_array( $field_id, $values['name_field_ids'], true ) && $value;
}
```

truongwp:
@Crabcyborg Sorry. I am a bit confused. Why don't we use three loops, and exclude the field value if it's found in the previous loop? I think it would be faster because it reduces "found email", "found URL" and "found name" checks.

Crabcyborg:
> @Crabcyborg Sorry. I am a bit confused. Why don't we use three loops, and exclude the field value if it's found in the previous loop? I think it would be faster because it reduces "found email", "found URL" and "found name" checks.

@truongwp if you for example had a form with 1000 items, and everything was a number, you would have to loop through and do 3000 checks.

This data could be anything, it's tough to know because it's based off of form data set by someone. Doing 3000 iterations isn't ideal when we could guarantee that we're only ever doing 1000.

truongwp:
@Crabcyborg If we use just one loop, there would be more checks each time. So I am not sure there would be less checks then using 3 loops.

Crabcyborg:
I just really don't like the idea of looping those nested arrays every time. That's a mess. It's not just 3 loops. It could be 300 loops. Everything is x3.

classes/models/FrmEntryValidate.php
```diff
@@ -418,48 +422,168 @@ private static function add_user_info_to_akismet( &$datas, $values ) {
...
-					} elseif ( $datas['comment_author_url'] == '' && strpos( $value, 'http' ) === 0 ) {
-						$datas['comment_author_url'] = $value;
-						$datas['frm_duplicated'][] = $index;
-					} elseif ( $datas['comment_author'] == '' && ! is_numeric( $value ) && strlen( $value ) < 200 ) {
-						$datas['comment_author'] = $value;
-						$datas['frm_duplicated'][] = $index;
+		foreach ( $values as $index => $value ) {
+			if ( is_array( $value ) ) {
+				foreach ( $value as $sub_value ) {
+					if ( self::found_akismet_spam_check_value( 'comment_author_email', $datas, $sub_value, $index, $values ) ) {
```
Crabcyborg, line 475: At this point we could probably reduce code more. It's pretty generic if we keep it all in the function. We could  loop through `array( 'comment_author_email', 'comment_author_url', 'comment_author' )`.

Crabcyborg: Oh, I got an idea.

What about a variable like `$missing_keys = array( 'comment_author_email', 'comment_author_url', 'comment_author' );`. We could loop that so and avoid the extra checks for if the value is set or not. Everytime we find one, we remove the key from `$missing_keys`.

truongwp:
@Crabcyborg I have added a recursive method to reduce the duplicated code and improve the checks. Please check it.

classes/models/FrmEntryValidate.php
```diff
@@ -418,46 +422,123 @@ private static function add_user_info_to_akismet( &$datas, $values ) {
...
+
+		switch ( $key ) {
+			case 'comment_author_email':
+				return strpos( $value, '@' ) && is_email( $value );
+
+			case 'comment_author_url':
+				return 0 === strpos( $value, 'http' );
+
+			case 'comment_author':
+				return in_array( $field_id, $name_field_ids, true ) || ! is_numeric( $value ) && strlen( $value ) < 200;
```
Crabcyborg, line 538: This still would match `! is_numeric( $value ) && strlen( $value ) < 200;` first if it came across a short comment before it matched a name field.

In my example I was trying to avoid that. If an open text field is right before the name, the name would never get picked. We should always use the name, as long as its set.

classes/models/FrmEntryValidate.php
```diff
@@ -418,46 +422,123 @@ private static function add_user_info_to_akismet( &$datas, $values ) {
...
+	 *
+	 * @since 5.0.13
+	 *
+	 * @param array    $datas        Guess data.
+	 * @param array    $values       The values.
+	 * @param int|null $custom_index Custom index (or field ID).
+	 */
+	private static function recursive_add_akismet_guess_info( &$datas, $values, $custom_index = null ) {
+		foreach ( $values as $index => $value ) {
+			if ( $datas['comment_author_email'] && $datas['comment_author_url'] && $datas['comment_author'] ) {
```
Crabcyborg, line 490: I really want to get rid of this triple check. It's still here. It seems like it's even checking more times than it was before.

I'm still thinking something like `$data['missing_keys']` would be really useful here. We can do one check to check for all three of those things at once, and you can really simplify the line of code underneath:

```
foreach ( array( 'comment_author_email', 'comment_author_url', 'comment_author' ) as $key ) {
				if ( $datas[ $key ] ) {
					continue;
				}
```

Should really just be:

```
foreach ( $datas['missing_keys'] as $key ) {
```

Crabcyborg:
> @Crabcyborg I have added a recursive method to reduce the duplicated code and improve the checks. Please check it.

Thanks @truongwp I added some notes.

This change did cut back on a lot of repetition, which is great. Thanks for that!

classes/models/FrmEntryValidate.php
truongwp: Oh. I got your idea. Let me update it.

classes/models/FrmEntryValidate.php
```diff
@@ -418,46 +422,125 @@ private static function add_user_info_to_akismet( &$datas, $values ) {
...
+	 * @since 5.0.13
+	 *
+	 * @param string $key            Guess info key.
+	 * @param string $value          Value to check.
+	 * @param int    $field_id       Field ID.
+	 * @param array  $name_field_ids Name field IDs.
+	 * @return bool
+	 */
+	private static function is_akismet_guess_info_value( $key, $value, $field_id, $name_field_ids ) {
+		if ( ! $value ) {
```
Crabcyborg, line 524: I'm thinking we could continue here as well if value is numeric since a number wouldn't match email, url, or name.

classes/models/FrmEntryValidate.php
```diff
@@ -418,46 +422,125 @@ private static function add_user_info_to_akismet( &$datas, $values ) {
...
+			'frm_duplicated'       => array(),
+		);
+
+		if ( isset( $values['item_meta'] ) ) {
+			$values = $values['item_meta'];
+		}
+
+		$values = array_filter( $values );
+
+		self::recursive_add_akismet_guess_info( $datas, $values );
```
Crabcyborg, line 473: I'm pretty sure this is supposed to be guest info? Since this is part of get_spam_check_user_info_for_**guest**

truongwp: Oh. That's my typo. Thank you :D

truongwp:
@Crabcyborg Thanks for checking. I updated it.

Crabcyborg:
> @Crabcyborg Thanks for checking. I updated it.

Perfect, thanks @truongwp!

PR #583, opened 2021-10-20
Stop trying to validate rich text buttons

Crabcyborg:
I noticed that because of this `.frm_required_field:visible input` selector every single rich text button (bold, italic, etc) is checked.

Screen Shot 2021-10-20 at 11 39 16 AM

I'm just skipping these earlier rather than calling checkRequiredField on something that is not a field.

Also cleaning up the validate function. It takes fieldId as the first parameter but never uses it. It looks like it's inaccessible outside of this file so removing it should be safe.

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -201,7 +201,7 @@ function frmFrontFormJS() {
 	 *
 	 * @param {object} element
 	 * @param {string} targetClass
-	 * @returns bool
+	 * @returns {boolean}
 	 */
 	function hasClass( element, targetClass ) {
 		var className = ' ' + element.className + ' ';
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -208,7 +208,7 @@ function frmFrontFormJS() {
 		return -1 !== className.indexOf( ' ' + targetClass + ' ' );
 	}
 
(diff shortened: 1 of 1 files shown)


js/formidable.js
```diff
@@ -160,6 +160,10 @@ function frmFrontFormJS() {
 		).filter( ':not(.frm_optional)' );
 		if ( requiredFields.length ) {
 			for ( r = 0, rl = requiredFields.length; r < rl; r++ ) {
+				if ( requiredFields[r].classList.contains( 'ed_button' ) ) {
```
stephywells, line 163: We've still been trying not to break js on the front-end for IE. Here's what we've been doing instead:
https://github.com/Strategy11/formidable-forms/blob/a4787922030019d8c264f95c1fd65f25400f16c8/js/formidable.js#L273-L274

Crabcyborg: Updated that!

I ended up calling it `hasClass` and just calling the variable `element` as it could really be used for any class check. I replaced the obvious spots where this was beneficial, all just inside of this required check function.

stephywells (approved):
Looks great! Thank you!

PR #604, opened 2021-11-02
Add frm_inline_form to forms with inline submit class (if missing)

Crabcyborg:
Talked about this in slack the other day.  It came up in Helpscout the other week. As I went into this code, it becomes pretty clear what is missing (the frm_inline_form class). We could probably make this more clear just with documentation, but it's also something we can just check for and add automatically 🕺

When `frm_inline_submit` is added in custom HTML for a form, the frm_inline_form class will be added automatically (if it isn't already set), because without it it does nothing.

**Before**
Screen Shot 2021-11-02 at 2 15 23 PM

**After**
Screen Shot 2021-11-02 at 2 15 00 PM


stephywells (approved):
Nice fix for this!

PR #601, opened 2021-10-29
Reduce server calls for license checks

stephywells:
Any ideas on a better way of doing this?

If server response isn't array, don't keep checking. We're getting a lot of repeat pings from the same sites. The inbox fixes were part of it, but I think it's a problem with version checks too.

This can be replicated by commenting out "return addons" on line 92 in FrmFormApi to skip the cache:
```
		$addons = $this->get_cached();
		if ( is_array( $addons ) ) {
			return $addons;
		}
```

This change uses thee same instance, so we know if the url has already been pinged on that page load. If the url has been called, we know the cache isn't working. This saves about 8 pings per page load in this scenario. 

The incoming pro version PR saves 7 more.

This log is edited a bit since this is public:

```
User-Agents grouped with IP: non-HIT's, sorted by most up top:
awk: warning: regexp escape sequence `\"' is not a known regexp operator
  16590 35.20.1.23 "form_action_automation/2.04; https://samesite.com"
  16583 35.20.1.23 "authorizenet_aim/2.01; https://samesite.com"
```
Many of these requests have been rate limited but it might be worth blocking the IP or making the rate limiting much stricter.

These are the types of requests we're seeing:

```
35.20.1.23 - - [28/Oct/2021:17:04:51 +0000] formidableforms.com "GET /wp-json/s11edd/v1/updates/?l=ZDdlOTU3ZTI0MTExNmNjNmFhMzFmMmE%3D HTTP/1.1" 200 31308 "formidable-pro/5.0.09; https://samesite.com" EXPIRED 1.756
35.20.1.23 - - [28/Oct/2021:17:04:51 +0000] formidableforms.com "GET /wp-json/s11edd/v1/updates/?l=ZDdlOTU3ZTI0MTExNmNjNmFhMzFmMmE%3D HTTP/1.1" 200 31308 "formidable-pro/5.0.09; https://samesite.com" EXPIRED 1.933
35.20.1.23 - - [28/Oct/2021:17:04:52 +0000] formidableforms.com "GET /wp-json/s11edd/v1/updates/?l=ZDdlOTU3ZTI0MTExNmNjNmFhMzFmMmE%3D HTTP/1.1" 200 31321 "formidable-pro/5.0.09; https://samesite.com" EXPIRED 2.101
[...]

PR #605, opened 2021-11-02
Make inbox messages static

Crabcyborg:
Following up from https://github.com/Strategy11/formidable-forms/pull/601

### DIFF
--- a/classes/models/FrmFormApi.php
+++ b/classes/models/FrmFormApi.php
@@ -81,29 +81,26 @@ class FrmFormApi {
 				'user-agent' => $agent . '; ' . get_bloginfo( 'url' ),
 			)
 		);
+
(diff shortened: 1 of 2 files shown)


stephywells (approved):
This looks great for inbox notices. Do you think we should include this for form templates and plugin update checks?
https://github.com/Strategy11/formidable-forms/pull/601/commits/84061883bc34fa8fcf96e076c5a3a3b09c873a26

Crabcyborg:
Right, thanks @stephywells

I think that other update would let stuff slip through the cracks. I think an invalid error response would empty end returning the old cached value if it wasn't an array for whatever reason.

I added this commit to cover it https://github.com/Strategy11/formidable-forms/pull/605/commits/5e46e54d65a758336af60830a45e1ad8e007b7ec

stephywells (approved):
Thanks! Looks perfect.

PR #606, opened 2021-11-03
Fix issue with some add-ons not showing the plugin info

stephywells:
This should fix the issue with any add-ons with a different name for the main file as the plugin folder. This was written assuming the plugin folder name would always have a file with the same name to go with it.

There is a separate issue in the signature plugin too. PR incoming for that in the add-on.


Crabcyborg (approved):
Thanks @stephywells I didn't really know how this worked. Good to know!

PR #608, opened 2021-11-03
Use "Bottom Margin" instead of "Margin" to describe margin-bottom sty…

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3293 as discussed by just renaming the setting.


stephywells (approved):
Perfect!

PR #609, opened 2021-11-04
Validate form exists when trying to export csv

Crabcyborg:
Tried exporting a CSV for a form that wasn't on my server and it results in two errors when there is no form found (since it's trying to find things with no id).

```
PHP Warning:  Attempt to read property "id" on null in /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmXMLController.php on line 570
[04-Nov-2021 13:38:46 UTC] PHP Notice:  wpdb::prepare was called <strong>incorrectly</strong>. The query argument of wpdb::prepare() must have a placeholder. Please see <a href="https://wordpress.org/support/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 3.9.0.) in /var/www/src/wp-includes/functions.php on line 5663
```

Just making this safer by checking if no form is found and exiting early.

### DIFF
--- a/classes/controllers/FrmXMLController.php
+++ b/classes/controllers/FrmXMLController.php
@@ -566,9 +566,14 @@ class FrmXMLController {
 
 		global $wpdb;
 
-		$form    = FrmForm::getOne( $form_id );
-		$form_id = $form->id;
+		$form = FrmForm::getOne( $form_id );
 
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Good find!

PR #610, opened 2021-11-05
Update readme 5.0.12

Crabcyborg:


stephywells (approved):
Looks great! It might be nice if the api requests were in one line instead of two, but it's not a big deal either way.

Crabcyborg:
Thanks @stephywells I was thinking the same thing. Updated that.

PR #612, opened 2021-11-05
Phpcs cleanup (mostly rule consistency)

Crabcyborg:
This update really just changes the deprecated `WPCS` exceptions to `phpcs:ignore` exceptions, and moves all exceptions out from the phpcs XML.

And some pretty minor code cleanup. There isn't a lot of crazy stuff happening in this pull request (it's more of a foundation that I'm working off of).

PR #613, opened 2021-11-05
New function FrmAppHelper::maybe_kses

Crabcyborg:
Branching off of https://github.com/Strategy11/formidable-forms/pull/612

Working on https://github.com/Strategy11/formidable-pro/issues/3253

Now when the DISALLOW_UNFILTERED_HTML constant is set kses will run on more HTML.

This adds front end kses filter for:
- Success messages
- "Before fields" html
- "After fields" html
- Custom button HTML
- Redirect message (when headers have already been sent).
- Fields names.
- Field descriptions.
- Custom field html.

I also found some links we were echoing without any kses. I wrapped `FrmAppHelper::kses( $link, array( 'a' ) );` around those.

Also improves this line
```
formidable/classes/models/FrmStyle.php:74: $new_instance['post_content'] = isset( $_POST['frm_style_setting']['post_content'] ) ? $_POST['frm_style_setting']['post_content'] : ''; // WPCS: sanitization ok.
```

I tested all of these changes. I believe there's more (I think I'm just getting started really) but I'm doing this in manageable portions. So future updates to come. I could use your help @garretlaxton determining where else we can add more filtering.


garretlaxton (approved):
These changes work great. 

I can probably find some locations grepping for $_POST, $_GET, etc... Should I create a new issue or log them [here](https://github.com/Strategy11/formidable-pro/issues/3253)?

Crabcyborg:
@garretlaxton it's probably easiest to just add scripts into everything (without the constant of course) and then turn on the constant.

I haven't been thorough yet. There are probably obvious ones still.

Crabcyborg:
- [ ] FrmAppHelper::icon_by_class echos `$icon` without any filtering/escaping. I tried kses on it but I one icon includes `style="--primary-hover:rgb(49, 119, 199)"` which seems to always get stripped by kses.
- [ ] FrmFormsHelper::get_custom_submit needs to be filtered.

classes/helpers/FrmAppHelper.php
```diff
@@ -2971,12 +2972,21 @@ private static function get_images_dropdown_option_html_attrs( $option, $args )
...
 		return current_user_can( 'unfiltered_html' );
 	}
 
+	/**
+	 * @since 5.0.13
+	 *
+	 * @return bool
+	 */
+	private static function should_never_allow_unfiltered_html() {
+		return defined( 'DISALLOW_UNFILTERED_HTML' ) && DISALLOW_UNFILTERED_HTML;
```
stephywells, line 2987: I'm thinking we should add a filter here. It's possible people would want unfiltered HTML in the forms, but not everywhere else.

Crabcyborg: I was thinking that earlier, it makes a lot of sense.

Added a commit with a new frm_disallow_unfiltered_html filter https://github.com/Strategy11/formidable-forms/pull/613/commits/9a62a0bb6a8e428d6c99df281906af8ae61594f0

stephywells: Thank you! Looks perfect!

Can you think of anything else to test with this?

Crabcyborg: I just tried to add scripts into any data I could find (in the database) and found one more in custom field HTML.

I haven't found anything else yet. Unless @garretlaxton finds anything else though I think the free plugin is covered at least.

stephywells (approved):
Awesome!! Thanks so much for solving all of this. Hopefully we don't get a big surge in support.

garretlaxton:
The Default Value doesn't seem to have a filter applied. The script tags never get stripped, it is sanitized so it doesn't execute but something to consider. 
Screenshot 2021-11-08 at 16-02-38 Formidable Forms ‹ Formidable Forms — WordPress

Also, for the Field Label & Description in the builder, if you input `` the javascript will execute client-side. If you update the form it will strip the JS, but the HTML img tag is still present.

Crabcyborg:
> Also, for the Field Label & Description in the builder, if you input `` the javascript will execute client-side. If you update the form it will strip the JS, but the HTML img tag is still present.

Thanks @garretlaxton since that would only effect the user triggering JavaScript on themselves it isn't anymore harmful than them running JavaScript on themselves in any other way. This kses update does escape that properly on the front end, so I think we're covered there.

As for the default values, I think we're probably safer leaving it as is since they are escaped and I wouldn't want to break anyone's default value logic, which could be fairly complex and might have issues with the filtering.

Crabcyborg:
> Awesome!! Thanks so much for solving all of this. Hopefully we don't get a big surge in support.

@stephywells I don't think people paid a lot of attention in the first update that added the initial kses calls. I don't think people use the constant very often or have unprivileged users making forms.

I'm realizing that the old maybe_filter_options call on submit button HTML was stripping a ton of stuff and I haven't really heard any of this come up in a ticket. button_action, back_hook, draft_hook, formnovalidate="formnovalidate", come other things, all stripped out.

I've added a new commit, a new function kses_submit_button, that runs kses and puts some stuff back if it was there initially. I think this is really the last thing I wasn't running kses on because I noticed it was breaking my submit buttons.
https://github.com/Strategy11/formidable-forms/pull/613/commits/f6ecbbda9acfc445bcce9266316983ef7d8a23d4

classes/helpers/FrmAppHelper.php
```diff
@@ -644,6 +644,31 @@ public static function kses( $value, $allowed = array() ) {
...
+	 *
+	 * @since 5.0.13
+	 *
+	 * @param string $html
+	 * @return string
+	 */
+	public static function kses_submit_button( $html ) {
+		$previously_included_button_action = false !== strpos( $html, '[button_action]' );
+		$previously_included_back_hook     = false !== strpos( $html, '[back_hook]' );
+		$previously_included_draft_hook    = false !== strpos( $html, '[draft_hook]' );
```
stephywells, line 658: These vars seem wordy and hard to read. Maybe $included_draft_hook would be a bit more readable?

classes/helpers/FrmAppHelper.php
```diff
@@ -644,6 +644,31 @@ public static function kses( $value, $allowed = array() ) {
...
+		$html                              = self::kses( $html, 'all' );
+		if ( $previously_included_button_action ) {
+			$html = str_replace( 'class="frm_button_submit" type="submit"', 'class="frm_button_submit" type="submit" [button_action]', $html );
```
stephywells, line 661: This is the only line that seems a bit finicky. I think adding classes to the submit button happens fairly often. I wonder if adding a different placeholder before the kses would work. Something like this?
```
$html = str_replace( '[button_action]', 'data-temp=""', $html );
$html = self::kses( $html, 'all' ); // (Should this use the new button kses function?)
$html = str_replace( 'data-template=""', [button_action], $html );
```

stephywells (changes requested):
Hmm... surprised that hasn't caused big issues. Glad you found that!!

classes/helpers/FrmAppHelper.php
Crabcyborg: Thanks @stephywells I definitely should have considered that already.

Rather than having to add a placeholder that would need a special kses exception I went with a more complex regex replace. I added an extra test with extra submit button classes as well.

classes/helpers/FrmAppHelper.php
stephywells: So... the tricky part about the custom HTML, is that it's changed over the years. Sometimes we've done migrations, and sometimes we haven't. Here's on version that's on some older forms:

```
<div class="frm_submit">
[if back_button]<input type="button" value="[back_label]" name="frm_prev_page" formnovalidate="formnovalidate" class="frm_prev_page" [back_hook] />[/if back_button]
<input type="submit" value="[button_label]" [button_action] />
<img class="frm_ajax_loading" src="[frmurl]/images/ajax_loader.gif" alt="Sending" style="visibility:hidden;" />
[if save_draft]<a class="frm_save_draft" [draft_hook]>[draft_label]</a>[/if save_draft]
</div>
```

In this case, we'd still lose the button_action. Any ideas on what would be more reliable?

classes/helpers/FrmAppHelper.php
stephywells: So... the tricky part about the custom HTML, is that it's changed over the years. Sometimes we've done migrations, and sometimes we haven't. Here's one version that's on some older forms:

```
<div class="frm_submit">
[if back_button]<input type="button" value="[back_label]" name="frm_prev_page" formnovalidate="formnovalidate" class="frm_prev_page" [back_hook] />[/if back_button]
<input type="submit" value="[button_label]" [button_action] />
<img class="frm_ajax_loading" src="[frmurl]/images/ajax_loader.gif" alt="Sending" style="visibility:hidden;" />
[if save_draft]<a class="frm_save_draft" [draft_hook]>[draft_label]</a>[/if save_draft]
</div>
```

In this case, we'd still lose the button_action. Any ideas on what would be more reliable?

Crabcyborg: I can add some extra checks to also support that pattern. That shouldn't be too difficult.

Crabcyborg: @stephywells I've updated it. It looks like prior to this we weren't allowing `input` tags at all with our kses calls so I had to add quite a few exceptions for this.

It also looks like `visibility:hidden` is stripped by default so I had to add a new filter for that but I have it keeping that entire string now.

https://github.com/Strategy11/formidable-forms/pull/613/commits/92859497ead342787fae315f9439552c534c59b3

classes/helpers/FrmAppHelper.php
```diff
@@ -787,12 +834,21 @@ private static function safe_html() {
...
-				'type'  => true,
+				'class'          => true,
+				'type'           => true,
+				'name'           => true,
+				'formnovalidate' => true,
 			),
 			'legend'     => array(
 				'class' => true,
 			),
+			'input'      => array(
```
stephywells, line 845: Hmm... I wonder if we shouldn't consider this one safe for everyone. Maybe this should only be added in the cases where we intend to have inputs?

Crabcyborg: @stephywells good call.

I've moved a couple of the extra tags that are really only necessary for the submit button (the input tag, the formnovalidate and name attributes in buttons, the style tag in images) into a filter that gets added and removed from kses_submit_button.

https://github.com/Strategy11/formidable-forms/pull/613/commits/79db96ed4bd2b3117315c621a61631c0572e28d4

classes/helpers/FrmAppHelper.php
```diff
@@ -644,6 +644,72 @@ public static function kses( $value, $allowed = array() ) {
...
+		$included_back_hook     = false !== strpos( $html, '[back_hook]' );
+		$included_draft_hook    = false !== strpos( $html, '[draft_hook]' );
+		add_filter( 'safe_style_css', 'FrmAppHelper::allow_visibility_style' );
+		add_filter( 'frm_striphtml_allowed_tags', 'FrmAppHelper::add_allowed_submit_button_tags' );
+		$html = self::kses( $html, 'all' );
+		remove_filter( 'safe_style_css', 'FrmAppHelper::allow_visibility_style' );
+		remove_filter( 'frm_striphtml_allowed_tags', 'FrmAppHelper::add_allowed_submit_button_tags' );
+		if ( $included_button_action ) {
+			if ( false !== strpos( $html, '<input type="submit"' ) ) {
+				$pattern = '/(<input type="submit")([^>]*)(\/>)/';
```
stephywells, line 666: Do we need the class here too? It's possible people have added an extra submit. We do get questions about how to have 2 submit buttons that do different things...

classes/helpers/FrmAppHelper.php
stephywells: Nice! That looks great!

classes/helpers/FrmAppHelper.php
Crabcyborg: @stephywells the input has no class, at least not in your example? I only do once replace though. I believe anything with more than one button_action totally doesn't work anyway since we explode on '[button_action]' and explode `$button_parts[0]` and `$button_parts[1]`.

stephywells: Oh, got it. It looks great then!

PR #615, opened 2021-11-08
Add backslashed if conditional to switch_field_ids function

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3303

The issue isn't too crazy. It just has to replace `[\/if 4938]`. Since the post content can be in JSON objects now it can be backslashed.


stephywells (commented):
Nice! Looks like we need to do the same for foreach?

Crabcyborg:
> Nice! Looks like we need to do the same for foreach?

Right, my bad! Updated that.

stephywells (approved):
Perfect! Thank you!!

garretlaxton (approved):
Fix works great when exported/imported to another site, ID matches the correct field.
Screenshot 2021-11-08 at 19-34-30 ‹ Formi QA — WordPress

PR #616, opened 2021-11-09
Only check Akismet spam if form is submitted

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3278

### DIFF
--- a/classes/models/FrmEntryValidate.php
+++ b/classes/models/FrmEntryValidate.php
@@ -251,11 +251,22 @@ class FrmEntryValidate {
 		$antispam_check = self::is_antispam_check( $values['form_id'] );
 		if ( is_string( $antispam_check ) ) {
 			$errors['spam'] = $antispam_check;
-		} elseif ( self::is_honeypot_spam( $values ) || self::is_spam_bot() ) {
+			return;
+		}
+		if ( self::is_honeypot_spam( $values ) || self::is_spam_bot() ) {
 			$errors['spam'] = __( 'Your entry appears to be spam!', 'formidable' );
-		} elseif ( self::blacklist_check( $values ) ) {
+			return;
+		}
+		if ( self::blacklist_check( $values ) ) {
 			$errors['spam'] = __( 'Your entry appears to be blocked spam!', 'formidable' );
-		} elseif ( self::is_akismet_enabled_for_user( $values['form_id'] ) && self::is_akismet_spam( $values ) ) {
+			return;
+		}
+
+		if ( isset( $values[ 'frm_page_order_' . $values['form_id'] ] ) ) {
+			return; // Only check Akismet spam if form is submitted in multi-page form.
+		}
+
+		if ( self::is_akismet_enabled_for_user( $values['form_id'] ) && self::is_akismet_spam( $values ) ) {
 			$errors['spam'] = __( 'Your entry appears to be spam!', 'formidable' );
 		}
 	}
--- a/classes/models/FrmEntryValidate.php
+++ b/classes/models/FrmEntryValidate.php
@@ -251,19 +251,15 @@ class FrmEntryValidate {
 		$antispam_check = self::is_antispam_check( $values['form_id'] );
 		if ( is_string( $antispam_check ) ) {
 			$errors['spam'] = $antispam_check;
-			return;
-		}
-		if ( self::is_honeypot_spam( $values ) || self::is_spam_bot() ) {
+		} elseif ( self::is_honeypot_spam( $values ) || self::is_spam_bot() ) {
 			$errors['spam'] = __( 'Your entry appears to be spam!', 'formidable' );
-			return;
-		}
-		if ( self::blacklist_check( $values ) ) {
+		} elseif ( self::blacklist_check( $values ) ) {
 			$errors['spam'] = __( 'Your entry appears to be blocked spam!', 'formidable' );
-			return;
 		}
 
-		if ( isset( $values[ 'frm_page_order_' . $values['form_id'] ] ) ) {
-			return; // Only check Akismet spam if form is submitted in multi-page form.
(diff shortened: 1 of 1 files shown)


classes/models/FrmEntryValidate.php
```diff
@@ -251,11 +251,22 @@ public static function spam_check( $exclude, $values, &$errors ) {
 		$antispam_check = self::is_antispam_check( $values['form_id'] );
 		if ( is_string( $antispam_check ) ) {
 			$errors['spam'] = $antispam_check;
-		} elseif ( self::is_honeypot_spam( $values ) || self::is_spam_bot() ) {
+			return;
```
Crabcyborg, line 254: This looks like a lot of returns. I think the old `elseif` stuff might be nicer.

We could just check `if ( isset( $errors['spam'] ) )` and return early before the Akismet check.

Crabcyborg (approved):
Thanks @truongwp!

The one comment I left is really just a code styling thing. This looks good. I tested it, it looks like it works as expected.

classes/models/FrmEntryValidate.php
```diff
@@ -251,11 +251,22 @@ public static function spam_check( $exclude, $values, &$errors ) {
...
-		} elseif ( self::blacklist_check( $values ) ) {
+			return;
+		}
+		if ( self::blacklist_check( $values ) ) {
 			$errors['spam'] = __( 'Your entry appears to be blocked spam!', 'formidable' );
-		} elseif ( self::is_akismet_enabled_for_user( $values['form_id'] ) && self::is_akismet_spam( $values ) ) {
+			return;
+		}
+
+		if ( isset( $values[ 'frm_page_order_' . $values['form_id'] ] ) ) {
```
Crabcyborg, line 265: @truongwp one more thought. Maybe we should have a check that the form actually has page breaks? This could be pretty easy to fake. I don't want people easily getting around Akismet.

Crabcyborg: It looks like we have these `$frm_vars['next_page']` and `frm_vars['prev_page']` variables. Can these be used here?

classes/models/FrmEntryValidate.php
Crabcyborg: I'm okay with returning, but I think we can reduce it to one return instead of three.

classes/models/FrmEntryValidate.php
truongwp: I found the method `FrmProFormsHelper::is_final_page()` uses `$frm_vars['next_page']` to check. But when I log its returned value, it's always `true` for every pages. So I think it's not reliable to check.

classes/models/FrmEntryValidate.php
truongwp: Ok. I updated it.

classes/models/FrmEntryValidate.php
Crabcyborg: Hm, if it's callable and it returns true, that's still would avoid any issues in the free plugin ever thinking there were multiple pages. It might be good enough.

truongwp: Ok. I added pro installed check.

truongwp: Should we check pro is connected?

Crabcyborg: I don't think that's important. We really need some sort of check that the form has page breaks first.

It's probably fine to add a call here for `$page_breaks = FrmField::get_all_types_in_form( $form_id, 'break' );` if pro is installed. It gets called quite often in pro already and should be cached.

truongwp: Thank you. I don't want an extra DB call. But since it's cached, I'll go that way.

classes/models/FrmEntryValidate.php
```diff
@@ -255,7 +255,14 @@ public static function spam_check( $exclude, $values, &$errors ) {
 			$errors['spam'] = __( 'Your entry appears to be spam!', 'formidable' );
 		} elseif ( self::blacklist_check( $values ) ) {
 			$errors['spam'] = __( 'Your entry appears to be blocked spam!', 'formidable' );
-		} elseif ( self::is_akismet_enabled_for_user( $values['form_id'] ) && self::is_akismet_spam( $values ) ) {
+		}
+
+		// Don't do Akismet check if form has break fields and current request is not a form submission (final page).
+		if ( isset( $errors['spam'] ) || FrmAppHelper::pro_is_installed() && isset( $values[ 'frm_page_order_' . $values['form_id'] ] ) && FrmField::get_all_types_in_form( $values['form_id'], 'break' ) ) {
```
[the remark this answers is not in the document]
truongwp: I did it. Thank you.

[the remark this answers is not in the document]
Crabcyborg: Awesome, thank you! I'll merge this soon, after I test it a bit more.

Crabcyborg:
@truongwp Testing this it looks like it's getting past this check still if I go to the previous page from the last page.

My `$_POST` data looks something like:
```
Array
(
    [frm_action] => create
    [form_id] => 157
    [frm_hide_fields_157] => 
    [form_key] => pageswakismet
    [item_meta] => Array
        (
            [0] => 
            [654] => First
            [656] => Last
        )

    [frm_submit_entry_157] => ...
    [_wp_http_referer] => ...
    [frm_next_page] => 2
    [item_key] => 8i1v1
    [frm_verify] => 
    [frm_state] => ...
    [antispam_token] =>...
)
```

I think this `frm_next_page` variable might need to be part of the check.

truongwp:
@Crabcyborg You're right. When clicking on the Next button, `frm_page_order_{form_id}` is set. When clicking on the Previous button, `frm_next_page` is not empty. I added the check.

Crabcyborg:
Thank you @truongwp it works great now!

PR #617, opened 2021-11-10
New function FrmAppHelper::kses_icon

Crabcyborg:

Adding filtering for this `echo $icon` line is a little tricky and requires some additional filters that kses calls to allow for some exceptions in the style attribute values.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -929,7 +929,7 @@ class FrmAppHelper {
 		} elseif ( 0 === strpos( $value, 'rgb(' ) ) {
 			$match = preg_match( '/^rgb\((\d{1,3}),\s*(\d{1,3}),\s*(\d{1,3})\)$/', $value );
 		} elseif ( 0 === strpos( $value, '#' ) ) {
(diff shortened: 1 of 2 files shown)


classes/helpers/FrmAppHelper.php
```diff
@@ -866,12 +868,93 @@ public static function icon_by_class( $class, $atts = array() ) {
...
+	}
+
+	/**
+	 * @since 5.0.13
+	 *
+	 * @param string $value
+	 * @return bool
+	 */
+	private static function is_a_valid_color( $value ) {
+		if ( 0 === strpos( $value, 'rgb' ) ) {
```
stephywells, line 926: This seems like a long way of doing this. Do you think regex would work for this too?
rgb\((?:\s*\d+\s*,){2}\s*[\d]+\)
rgba\((\s*\d+\s*,){3}[\d\.]+\)
(I didn't test these)

Crabcyborg: @stephywells thanks, I changed those out for regex. Makes it smaller at least.

stephywells (approved):
This looks good to me. I'm still not entirely sure how much the valid color matters since this is already escaped. Can you think of any examples that could cause a problem?

stephywells:
Then it looks perfect. Thanks Mike!!

PR #618, opened 2021-11-10
Remove phpcs ignore from FrmFieldCombo

Crabcyborg:

The line
```
formidable/classes/models/fields/FrmFieldCombo.php:377: echo implode( ' ', $atts ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```

Was brought up. And I'm trying to use phpcs:ignore as little as possible now.

Getting around this is pretty easy by just doing the escaping in the loop instead. I also got rid of `$atts` as the only other use for it was for placeholder which appears to be safe just passing into `$field` and letting `do_action( 'frm_field_input_html', $field );` handle it.

Requesting your review on this @truongwp since you wrote it initially. See any issues with this update?

### DIFF
--- a/classes/models/fields/FrmFieldCombo.php
+++ b/classes/models/fields/FrmFieldCombo.php
@@ -339,13 +339,12 @@ class FrmFieldCombo extends FrmFieldType {
 	protected function print_input_atts( $args ) {
 		$field     = $args['field'];
 		$sub_field = $args['sub_field'];
-		$atts      = array();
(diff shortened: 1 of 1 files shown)


truongwp (approved):
Your changes are good. I checked the code and it will print the placeholder correctly. 
Thank you.

Crabcyborg:
Thanks @truongwp

PR #621, opened 2021-11-11
Add echo params to array_to_html_params and js_redirect functions

Crabcyborg:
Related to https://github.com/Strategy11/formidable-pro/issues/3253

I'm trying to remove phpcs:ignore lines wherever possible. In both of these cases (array_to_html_params and js_redirect), if the function handles the echo there is no need for any phpcs:ignore line, as long as the logic is duplicated in a way that doesn't involve echoing an unescaped variable.

So with both of these I've added a new `$echo` param. I also found a line in a combo field view where this function was able to replace some logic.

This update removes 6 phpcs:ignore comments.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1022,23 +1022,36 @@ class FrmAppHelper {
 	 * @return string|void
 	 */
 	public static function array_to_html_params( $atts, $echo = false ) {
-		if ( ! $atts ) {
-			if ( ! $echo ) {
-				return '';
+		$callback = function() use ( $atts ) {
+			if ( $atts ) {
+				foreach ( $atts as $key => $value ) {
+					echo ' ' . esc_attr( $key ) . '="' . esc_attr( $value ) . '"';
+				}
 			}
-			return;
+		};
+		return self::clip( $callback, $echo );
+	}
+
+	/**
+	 * Call an echo function and either echo it or return the result as a string.
+	 *
+	 * @since 5.0.13
+	 *
+	 * @param Closure $echo_function
+	 * @param bool    $echo
+	 * @return string|void
+	 */
+	public static function clip( $echo_function, $echo = false ) {
+		if ( ! $echo ) {
+			ob_start();
 		}
+
+		$echo_function();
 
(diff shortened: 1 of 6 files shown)


classes/helpers/FrmAppHelper.php
```diff
@@ -1015,17 +1015,31 @@ public static function include_svg() {
...
+		if ( ! $echo ) {
+			$html = '';
 			foreach ( $atts as $key => $value ) {
 				$html .= ' ' . esc_attr( $key ) . '="' . esc_attr( $value ) . '"';
 			}
+			return $html;
+		}
+
+		foreach ( $atts as $key => $value ) {
+			echo ' ' . esc_attr( $key ) . '="' . esc_attr( $value ) . '"';
```
stephywells, line 1041: This looks like duplication. Could we keep the old logic and add
```
if ( ! $echo ) {
return $html;
}
echo $html;
```

stephywells (changes requested):
Great call! Is there a good way of doing this without duplication?

classes/helpers/FrmAppHelper.php
Crabcyborg: I don't think it's possible without exceptions. I prefer avoiding the exception. These functions are never likely to change and they're small so I think the duplication is ultimately worth it in this case.

stephywells: I'm fine with whatever you think is best. But just throwing one more option out there...

```
if ( ! $echo ) {
ob_start();
}
echo stuff;
if ( ! $echo ) {
$return = ob_get_contents();
ob_end_clean();
return $return;
}
```

This is still quite a few lines, so it may not be worth it...

Crabcyborg: @stephywells Yeah that was the other option I had considered and it seemed like a lot.

But I abstracted this a bit to a new `FrmAppHelper::clip` function that basically is what you pasted there exactly, and it takes a function that does the echo. This way I can use this chunk of code in both functions 🥳

stephywells: That looks great! Thanks for trying that route.

stephywells (approved):

PR #623, opened 2021-11-12
Deprecate backend dropdown view

Crabcyborg:
Removes two phpcs:ignore comments by removing the code entirely since it isn't used anymore.

Tested with a snippet:

```
add_action(
	'plugins_loaded',
	function() {
		$field_object = FrmField::getOne( 326 );
		$field        = FrmFieldsHelper::setup_edit_vars( $field_object );
		$field_name   = 'item_meta[326]';
		$view = FrmAppHelper::plugin_path() . '/classes/views/frm-fields/back-end/dropdown-field.php';
		require $view;
		die();
	}
);
```

It renders fine by the back end view file, after I made sure that `$read_only` and `$html_id` are actually set. Otherwise, the require was throwing errors since the two views are fairly different.

### DIFF
--- a/classes/views/frm-fields/back-end/dropdown-field.php
+++ b/classes/views/frm-fields/back-end/dropdown-field.php
@@ -3,36 +3,15 @@ if ( ! defined( 'ABSPATH' ) ) {
 	die( 'You are not allowed to call this page directly.' );
 }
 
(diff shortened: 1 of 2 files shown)


stephywells (approved):
Perfect! Thanks so much!

PR #625, opened 2021-11-17
Fix error on blacklist check

truongwp:
This fixes https://github.com/Strategy11/formidable-pro/issues/3313


Crabcyborg (approved):
Thank you @truongwp! Looks great!

PR #631, opened 2021-11-18
Try removing yoast from phpunit-polyfills tool name

Crabcyborg:
Looks like the polyfills aren't working again.


Crabcyborg:
Those didn't work. I'm not really sure still why they're suddenly broken.

PR #639, opened 2021-11-19
Add array_separator option to FrmEntryFormatter

Crabcyborg:
This update makes it possible to change the default `, ` separation that gets applied to array type values (including checkboxes) when calling `FrmEntryFormatter::flatten_array` which is used when showing entries as a table (like in an email default message).

Initially I went with a filter but this really makes more sense as a shortcode option like the other options are in here (like "line_breaks").

I tested that this works in an email, and that it works in a view by setting up a tiny little `[my_frm_entry]` shortcode for testing that was loading a specific entry I was testing with:

```
add_shortcode( 'my_frm_entry', 'show_entry' );
function show_entry() {
	return FrmEntriesController::show_entry_shortcode(
		array(
			'id' => 56220,
			'array_separator' => '<br/>',
		)
	);
}
```

In an email action
```
[default-message array_separator="<br/>"]
```

Screen Shot 2021-11-19 at 2 12 27 PM


garretlaxton (approved):
I don't think the Likert scale is an array type, but I tested it anyway haha. I also tested dropdown multi-select which works great.
Screenshot 2021-11-20 at 09-06-51 test – Formidable Forms

stephywells (approved):
Looks perfect!

PR #640, opened 2021-11-22
Support rgba in adjust brightness function

Crabcyborg:
https://secure.helpscout.net/conversation/1704392934/86079?folderId=2445391

It looks like border colors are the only setting that calls this function. It's used to calculate the color of an alt table row (zebra striped).

I missed this when I added rgba support to styles. Now I'm checking for rgba, and adjusting it if I find it.


stephywells (approved):
Looks good! Thank you!

PR #641, opened 2021-11-29
Use css grid for spam settings and add tooltips

Crabcyborg:
The other day a ticket came up asking about the difference between honeypot and the antispam token (https://secure.helpscout.net/conversation/1709526153/86563).

We have enough information on this in our documentation at https://formidableforms.com/knowledgebase/add-spam-protection/. But if people don't check the documentation, this can be confusing. It isn't very clear what the new JavaScript token option actually is from a glance, or that it's even a token at all.

I've just added tooltips to make the plugin more educational on the topic as it isn't very straight forward. Hoping this can help avoid future tickets like that one. I also mention that file uploads can be protected with the antispam token if pro is installed only as it isn't applicable for free use, which is somewhat related to https://github.com/Strategy11/formidable-pro/issues/3260

And while I was at it, these spam settings were all in a table, which was complicating things. I never liked how the dropdowns were 100%, they need much less space. The easiest way to enforce this is using the grid classes like some of the other settings on this page are.

I ended up moving around the markup quite a bit in this process. The Akismet label wasn't in a label tag before, so now it's more accessible as well 🥳. I also added a `data-container="body"` to the JavaScript validation to avoid an issue with the tooltip losing opacity when I disable the JavaScript validation option when chat forms are active.

**Before**
Screen Shot 2021-11-29 at 10 34 01 AM

**After**
Screen Shot 2021-11-29 at 10 31 35 AM
Screen Shot 2021-11-29 at 10 31 43 AM
Screen Shot 2021-11-29 at 10 31 47 AM

### DIFF
--- a/classes/views/frm-forms/settings-advanced.php
+++ b/classes/views/frm-forms/settings-advanced.php
@@ -109,19 +109,17 @@ if ( ! defined( 'ABSPATH' ) ) {
 	</label>
 </p>
 
-<table class="form-table">
-	<tr>
-		<td colspan="2">
-			<label for="no_save" class="frm_inline_block">
-				<input type="checkbox" name="options[no_save]" id="no_save" value="1" <?php checked( $values['no_save'], 1 ); ?> />
-				<?php esc_html_e( 'Do not store entries submitted from this form', 'formidable' ); ?>
-			</label>
-		</td>
-	</tr>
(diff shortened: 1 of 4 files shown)


stephywells (approved):
This looks great. Thank you!

PR #643, opened 2021-11-30
Use tiny mce for html field descriptions

Crabcyborg:
Figured I'd give a go at this Sprint item https://app.asana.com/0/1201149924118706/1201160282718496

HTML fields are a free plugin thing and nothing here requires the paid plugin so I put everything here. If we want to move it and call it a more premium feature of HTML fields, let me know!

I also needed to update `should_insert_form_popup` a bit to include the formidable builder page now.

Most of this is JavaScript copied from pro and views code. All of the weird tiny mce issues I had I've had before so things weren't too hard to figure out.

I'm also not sure if we want to address the styling at all? There isn't a lot of space here for a wysiwyg so it gets kind of squished and take up more vertical space than it really needs.
Screen Shot 2021-11-30 at 3 19 20 PM

Or if we want to remove any of the buttons? It might be that we want to keep this a bit more simple.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -651,9 +651,11 @@ class FrmFormsController {
 		}
 	}
 
+	/**
+	 * @return void
+	 */
 	public static function insert_form_popup() {
-		$page = basename( FrmAppHelper::get_server_value( 'PHP_SELF' ) );
-		if ( ! in_array( $page, array( 'post.php', 'page.php', 'page-new.php', 'post-new.php' ) ) ) {
+		if ( ! self::should_insert_form_popup() ) {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


stephywells (approved):
Nice job! It looks like we could keep the icons to two rows if we remove the Read More icon. 

Let's go ahead and keep this in lite. I think not being able to easy add images might cause a lot of drop off for lite users.

Thanks so much1

Crabcyborg:
Removed the read more icon, and yes, it goes down to 2 rows!

Screen Shot 2021-11-30 at 4 20 56 PM

stephywells:
That looks great now!

garretlaxton:
This looks great. Something I did notice is when I duplicate the HTML field, the wysiwyg buttons are missing until you update the page.
Screenshot 2021-11-30 at 16-49-23 Formidable Forms ‹ Formidable Forms — WordPress

Crabcyborg:
Thanks @garretlaxton that turned out to be a pretty big issue. It also effected any new HTML fields that were added just by inserting a field as well.

I also caught that if you add an HTML field and didn't already have another HTML field on the form, that it wasn't working.

I made a pretty big update now that should cover all of this. So if you can give it another round of testing, see if you can break it with duplicating now at all, it'd be appreciated!

garretlaxton (approved):
Looks good! Also tested with import & export and a few other ways.

PR #646, opened 2021-12-02
Remove phpcs exceptions from textarea show_on_form_builder function

Crabcyborg:
Looking for more opportunities to improve on https://github.com/Strategy11/formidable-pro/issues/3253

This update removes two `phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped` comments. It turns out the first one was there for no reason and was passing without the comment.

I broke it out into multiple echo statements. The big echo was making it difficult to output the style without the need for an exception, and it wasn't really escaping in the right place as it was happening after `width:`, and the semi colon is after it closes the style attribute, which looks like a bug.

Instead of setting `$default_value`, I echo at the same time that we call `FrmAppHelper::esc_textarea` so it's more clear that escaping is happening here. That's the only exception that can't be removed.

### DIFF
--- a/classes/models/fields/FrmFieldTextarea.php
+++ b/classes/models/fields/FrmFieldTextarea.php
@@ -31,18 +31,23 @@ class FrmFieldTextarea extends FrmFieldType {
 	 * @param string $name
 	 */
 	public function show_on_form_builder( $name = '' ) {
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Looks great! Thank you!!

PR #647, opened 2021-12-03
Remove phpcs exceptions in FrmListHelper

Crabcyborg:

This update removes 4 more phpcs exceptions by replacing echo statements that aren't easily escaped with calls to `FrmAppHelper::array_to_html_params` that handles the escaping more nicely.

I also cleaned up the part that was checking `} elseif ( method_exists( $this, 'column_' . $column_name ) ) {`. The `else` condition directly under it was doing the exact same thing, just not calling the method. It makes more sense to just put this if in with the else to get rid of duplicate code.

### DIFF
--- a/classes/helpers/FrmListHelper.php
+++ b/classes/helpers/FrmListHelper.php
@@ -391,9 +391,17 @@ class FrmListHelper {
 		echo "<option value='-1' selected='selected'>" . esc_attr__( 'Bulk Actions', 'formidable' ) . "</option>\n";
 
 		foreach ( $this->_actions as $name => $title ) {
-			$class = 'edit' == $name ? ' class="hide-if-no-js"' : '';
+			echo "\t<option ";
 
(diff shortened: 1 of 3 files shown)


classes/helpers/FrmListHelper.php
```diff
@@ -391,9 +391,17 @@ protected function bulk_actions( $which = '' ) {
 		echo "<option value='-1' selected='selected'>" . esc_attr__( 'Bulk Actions', 'formidable' ) . "</option>\n";
 
 		foreach ( $this->_actions as $name => $title ) {
-			$class = 'edit' == $name ? ' class="hide-if-no-js"' : '';
+			echo "\t<option ";
```
stephywells, line 394: This might be  bit more readable if this is moved down to line 402 where the other echoing starts

stephywells (approved):
Nice! This looks good to me

classes/helpers/FrmListHelper.php
Crabcyborg: Good point. I moved that down.

PR #652, opened 2021-12-13
reCaptcha v3

Crabcyborg:
Screen Shot 2021-12-13 at 2 04 52 PM
_Grammar is a bit off in screenshot but I pushed a tweak to fix that._

Screen Shot 2021-12-13 at 2 40 50 PM
_I've also fixed the capitalization of captcha so it's "reCAPTCHA" now_

### DIFF
--- a/classes/helpers/FrmEntriesHelper.php
+++ b/classes/helpers/FrmEntriesHelper.php
@@ -694,4 +694,28 @@ class FrmEntriesHelper {
 
 		return apply_filters( 'frm_entry_actions_dropdown', $actions, compact( 'id', 'entry' ) );
 	}
+
+	/**
+	 * @since 5.0.15
+	 *
+	 * @param string|int $entry_id
+	 * @return void
+	 */
+	public static function maybe_render_captcha_score( $entry_id ) {
+		$query                 = array(
+			'item_id'  => (int) $entry_id,
+			'field_id' => 0,
+		);
+		$metas_without_a_field = (array) FrmEntryMeta::getAll( $query, ' ORDER BY it.created_at DESC', '', true );
+		foreach ( $metas_without_a_field as $meta ) {
+			if ( ! empty( $meta->meta_value['captcha_score'] ) ) {
+				echo '<div class="misc-pub-section">';
+				FrmAppHelper::icon_by_class( 'frm_icon_font frm_shield_check_icon', array( 'aria-hidden' => 'true' ) );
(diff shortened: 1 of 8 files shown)


classes/helpers/FrmEntriesHelper.php
```diff
@@ -694,4 +694,28 @@ private static function get_action_links( $id, $entry ) {
...
+	/**
+	 * @since 5.0.15
+	 *
+	 * @param string|int $entry_id
+	 * @return void
+	 */
+	public static function maybe_render_captcha_score( $entry_id ) {
+		$query                 = array(
+			'item_id'  => (int) $entry_id,
+			'field_id' => 0,
```
stephywells, line 707: Would it be more or less complicated to use the recaptcha field id here?

stephywells (commented):
I'm not seeing the threshold getting used anywhere. Is there a file missing from this commit?

classes/helpers/FrmEntriesHelper.php
Crabcyborg: We could, I just recall talking about doing it this way.  To render the score it would require an extra check for the captcha field first. But we probably can.

Crabcyborg:
> I'm not seeing the threshold getting used anywhere. Is there a file missing from this commit?

Check out https://github.com/Strategy11/formidable-forms/blob/ad192abb5be922d39108ffe7b12a407ab9e6310a/classes/models/fields/FrmFieldCaptcha.php#L164

classes/helpers/FrmEntriesHelper.php
stephywells: It's not a big deal either way. It looks good to me!

stephywells (approved):
Oh, right! I suppose I was looking for javascript changes. Awesome job on this!!

garretlaxton (approved):
V3 Captcha works great. There are still a few places where `reCAPTCHA` is spelled `reCaptcha`. Not sure if we want to be consistent with that.

Crabcyborg:
Good point @garretlaxton. I looked around the code and it seems to mostly just be the section name in the sidebar/the page title. I've changed that to be more consistent.

garretlaxton:
The last place I see this is the multiple captcha option.
Screenshot 2021-12-14 at 17-30-00 Formidable Global Settings ‹ Formidable Forms — WordPress

PR #655, opened 2021-12-15
Add option to generate csv export file

Crabcyborg:
This updates does a few things:

1. Adds support for a new `'mode' => 'file'` argument that will change the behavior so it generates a CSV file instead of printing one.
2. frm_notification_attachment now passes a `$args['action_id']` value.
3. `FrmCSVExportHelper` now passes `context` and `meta` into a lot of filters include frm_csv_sep, frm_csv_line_break, frm_csv_date_format, frm_csv_column_sep, frm_csv_format, frm_csv_field_columns, frm_csv_filename, and frm_csv_columns. Meta in this case will include `action_id` passed from `FrmProNotification` which generates the CSV file.


stephywells (approved):
This looks great!!

PR #656, opened 2021-12-16
Landing page placeholders

Crabcyborg:
@stephywells I think this should do the trick.

I added some extra logic if it can't find the plugin in the API but it can find it installed locally already so we can still get this working before it's available in the API.

We could also add a `data-requires`. I think I can just hard code that. What level of pro license is this going to be available for?


stephywells (approved):
This looks great! Let's include this in the Plus plan. That's "creator" in the plugin code.

I'd like the tabs that aren't allowed to be selectable and include more info. But we can add that separately.

Thank you!

PR #657, opened 2021-12-17
Fixes for white styles on dark bg

stephywells:
- allow for more transparent bg colors
- On focus, use the color set for input fields

These came up while working on this style template:
https://sandbox.formidableforms.com/demos/wp-admin/admin-ajax.php?action=frm_forms_preview&form=contact-us

Before:
Read only fields and fields with focus had a white background and black text

After:
Transparent bg is allowed for read only and fields with focus.
The active input color now overrides the theme.

This is all tested and pretty minor changes.


Crabcyborg (approved):
Looks good!

PR #658, opened 2021-12-17
Add sanitize_url=1 to redirect url shortcodes on save

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/830

This update does a few things:

1. When you click to insert a field shortcode, if the field is not a URL type, the sanitize_url=1 option is injected automatically.
2. On save, if there are shortcodes detected in the success_url URL params, I add in the `sanitize_url=1` option. This way they get caught even if you didn't use the pop up.
3. Added support for sanitize_url and sanitize shortcode options when replacing shortcodes in the free plugin.

**Before saving**
Screen Shot 2021-12-17 at 1 18 27 PM

**After saving**
Screen Shot 2021-12-17 at 1 18 31 PM

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -746,6 +746,7 @@ class FrmFieldsHelper {
 			$replace_with  = self::get_value_for_shortcode( $atts );
 
 			if ( $replace_with !== null ) {
+				$replace_with = self::trigger_shortcode_atts( $replace_with, $atts );
 				self::sanitize_embedded_shortcodes( compact( 'entry' ), $replace_with );
 				$content = str_replace( $shortcodes[0][ $short_key ], $replace_with, $content );
 			}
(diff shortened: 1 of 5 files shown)


stephywells:
Yep, you're totally right. The redirect option used to be in Pro, and was moved to lite. Looks like we missed that part. That's no problem at all to move sanitize_url to lite.

One more thing... I'm concerned about this possible causing issues with reverse compatibility. For example, what if a full url is in a hidden field or something?

I wonder if it would be best to include sanitize_url=1 at the time a field is added to the box. That would solve the problem if people are making changes or new forms, but wouldn't change the behavior of existing redirects.

What do you think?

Crabcyborg:
@stephywells are you just thinking that it would get added when you click the option from the pop up?

I hadn't considered people pulling the full URL but that makes sense that might not not be something we'd always want.

What if we check the rest of the string and only automatically add this to shortcodes that appear in the parameters? If the string is only a shortcode, we could leave it alone.

stephywells:
Great idea! That sounds like a better option. 

Let's also make sure this doesn't get added to an [else]. I'm not sure if that would cause a problem or not.

garretlaxton (approved):
Looks good! Now automatically adds `sanitize_url=1` when using a shortcode to redirect.
Screenshot 2021-12-22 at 02-54-22 Formidable Forms ‹ Formidable Forms — WordPress

stephywells (approved):
Nice! This looks great!!!

PR #659, opened 2021-12-18
Add Auto Height option for the text field

burhandodhy:
This PR is linked to https://github.com/Strategy11/formidable-pro/pull/3351 that adds the `Auto Height` field for TextArea

### DIFF
--- a/classes/models/fields/FrmFieldTextarea.php
+++ b/classes/models/fields/FrmFieldTextarea.php
@@ -65,9 +65,10 @@ class FrmFieldTextarea extends FrmFieldType {
 		$input_html = $this->get_field_input_html_hook( $this->field );
 		$this->add_aria_description( $args, $input_html );
 		$rows = ( $this->field['max'] ) ? 'rows="' . esc_attr( $this->field['max'] ) . '" ' : '';
+		$auto_height = ( $this->field['auto_height'] ) ? 'data-auto-height="' . esc_attr( $this->field['auto_height'] ) . '" ' : '';
 
 		return '<textarea name="' . esc_attr( $args['field_name'] ) . '" id="' . esc_attr( $args['html_id'] ) . '" ' .
-			$rows . $input_html . '>' .
+			$rows . $input_html . $auto_height . '>' .
 			FrmAppHelper::esc_textarea( $this->field['value'] ) .
 			'</textarea>';
 	}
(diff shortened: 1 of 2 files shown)


classes/models/fields/FrmFieldTextarea.php
```diff
@@ -65,9 +65,10 @@ public function front_field_input( $args, $shortcode_atts ) {
 		$input_html = $this->get_field_input_html_hook( $this->field );
 		$this->add_aria_description( $args, $input_html );
 		$rows = ( $this->field['max'] ) ? 'rows="' . esc_attr( $this->field['max'] ) . '" ' : '';
+		$auto_height = ( $this->field['auto_height'] ) ? 'data-auto-height="' . esc_attr( $this->field['auto_height'] ) . '" ' : '';
```
Crabcyborg, line 68: If this is a pro feature I think we should try to get as much of this into the pro plugin as we can. I think it'd be great if the free plugin doesn't ever reference auto_height directly.

We inherit this model in the pro plugin with `FrmProFieldTextarea` so maybe we want to overwrite `front_field_input`.

I'm thinking we could probably even call the parent function and then inject the HTML into the string so we re-use as most of the free function as possible still.

classes/views/frm-fields/back-end/settings.php
```diff
@@ -58,6 +58,14 @@
...
 
+			if ( $display['auto_height'] ) {
+				?>
+				<label for="frm_auto_height_field_<?php echo esc_attr( $field['id'] ); ?>" class="frm_inline_label frm_help" title="<?php esc_attr_e( 'Auto Height: Automatically expand the height of the field when the text reaches the maximum rows', 'formidable' ); ?>" >
+					<input type="checkbox" id="frm_auto_height_field_<?php echo esc_attr( $field['id'] ); ?>" name="field_options[auto_height_<?php echo esc_attr( $field['id'] ); ?>]" value="1" <?php checked( $field['auto_height'], 1 ); ?>/>
+					<?php esc_html_e( 'Auto Height', 'formidable' ); ?>
+				</label>
+				<?php
+			}
 			do_action( 'frm_field_options_form_top', $field, $display, $values );
```
Crabcyborg, line 69: It looks like we're calling an action here, `frm_field_options_form_top`. That should be easy to hook into to echo this setting from pro instead.

PR #660, opened 2021-12-21
Add float none to radio and checkbox inputs

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/1731135096/88134/

A customer is using `The H-Code premium WordPress theme from Themeforest.`.

The default styles show a radio input something like
Screen Shot 2021-12-21 at 9 01 47 AM

I looked at a demo of H-Code and added a radio input to the page to see what styles were on it, and found:
Screen Shot 2021-12-21 at 9 00 22 AM

So I'm fighting this so the default is `float: none` instead of the `float: left` that's messing it up.

I didn't have H-Theme but I replicated this by putting that CSS in my theme CSS, and then fixed it with this new style rule:
Screen Shot 2021-12-21 at 9 03 45 AM


stephywells (approved):
This looks perfect!

PR #661, opened 2021-12-21
Do not set data-message if it is empty

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3354

It looks like back in January 2020 there was a commit https://github.com/Strategy11/formidable-forms/commit/81d916fa94cd6b8555a903f4d6af6fcdddd4f52d

That added the `data-message` in always. Before, it was never there. So it looks like it was broken one way (it was never appearing), and then broken another way (always appearing).

That line of code is really long. I broke it up into an array and I'm calling `FrmAppHelper::array_to_html_params` now. I've kept the JavaScript the same.

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -1852,8 +1852,22 @@ class FrmFieldsHelper {
 			}
 		}
 
+		$li_params = array(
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Wow. That's been in there for a long time. Weird that we didn't notice it sooner.

Thanks for finding the source and cleaning up after me. Sorry about that!

PR #662, opened 2021-12-22
Radio and Checkbox styling

stephywells:
This started as just fixing a few conflicts with 2021 theme, but it ended up being easier to add our own radio/checkbox styling then to override everything back to the browser default.

- Fixes https://github.com/Strategy11/formidable-pro/issues/2201
- Fixes https://github.com/Strategy11/formidable-pro/issues/2840
- Removes weird outline styling for fields and submit button in 2021
- Accept a few more colors as transparent so they don't cause css errors when empty

I tested the new styling in a few browsers, but a second look would be nice. The appearance in IE isn't the best, but at this point I don't think it's a huge deal.


garretlaxton (approved):
Tested on Firefox, Chrome, and Safari on my linux VM. Styling looks good on my end.  As for IE, there are many issues I saw, but I don't think it's worth fixing as WordPress itself doesn't really support IE.

Crabcyborg (approved):
Looks good!

PR #663, opened 2021-12-28
Update readme for 5.0.16

Crabcyborg:
@stephywells right now this is scheduled for tomorrow.

I know you mentioned releasing Landing pages on the 4th though, which would probably require a core release next week.

Do you think we should hold off on this release for a week, or do a release both weeks, or do landing pages the next week on the 11th?

Also, do you think we should include anything in the change log for https://github.com/Strategy11/formidable-forms/pull/657? I wasn't sure what to put for that.


stephywells:
@Crabcyborg Would it be possible to add the PR for the preview button, but maybe only show the full page info if that add-on is included in the API response? Would that be possible, so it would show up in the form pretty quickly after the release?

For https://github.com/Strategy11/formidable-forms/pull/657 we could include something like:
Allow more colors in the styler to be transparent

PR #664, opened 2021-12-28
Do not skip required inline datepicker in js required check

Crabcyborg:
Part of https://github.com/Strategy11/formidable-dates/pull/68

On its own it will validate and properly prevent the page from submitting but requires https://github.com/Strategy11/formidable-dates/pull/68 for an error message to be displayed.

One funny thing I was also noticing was that the error border styling happens on the month/year pickers.

Screen Shot 2021-12-28 at 1 08 39 PM

We have this style set
Screen Shot 2021-12-28 at 1 09 04 PM

So I added some `:not` rules to the theming so it avoids those.


stephywells (approved):
Nice fix!

PR #667, opened 2022-01-04
Drop codemirror files

Crabcyborg:
Since we're officially only supporting v5.0+ of WordPress and Codemirror was added to WordPress core as of v4.9 (since November 2017) I think we're good to go here with removing Codemirror.

I didn't really know any of this but was looking at our Codemirror code and noticed this. Great opportunity to drop almost 9000 lines of code that we don't need.

After ripping all of this out I see no issues with updating custom CSS which I believe is all this is used for.

**One other thing** - A call to wp_enqueue_code_editor can return false so even prior to this update it's possible that you come to this page and never see the code editor. I was seeing this on a customer's site running on WordPress v5.8.2. The function immediately begins with:

```
if ( is_user_logged_in() && 'false' === wp_get_current_user()->syntax_highlighting ) {
	return false;
}
```

and then also has:
```
if ( empty( $settings ) || empty( $settings['codemirror'] ) ) {
	return false;
}
```

`$settings` is filtered so it could be unset in many cases.

I wonder if we want to show an error or something that code mirror couldn't be loaded in these cases, rather than showing nothing? We could probably offer a troubleshooting guide or something.


stephywells (approved):
This looks good to me. If Codemirror isn't loaded, does the textarea show instead? If not, I think that would be fine as a fallback.

Crabcyborg:
Thanks @stephywells that cleared up some details on this issue's history, I wasn't totally aware that were was already code here trying to fallback to a textarea that just isn't working for every use case.

It looks like there was an update made for that to fix an old issue back in 2018 https://github.com/Strategy11/formidable-pro/issues/2038

I changed that check now for `syntax_highlighting` so it relies more on the results of what WordPress returns as `$settings` so that it covers both of these early return conditions instead of just the one.

PR #669, opened 2022-01-10
Conversational forms placeholders

Crabcyborg:
I just used _Plus_ for license like landing pages for now, not sure if we're going with a different pricing plan though.

I refactored some of my landing page stuff so it will work better for new plugins in the future.

It also looks like I mixed up the `data-medium` value for landing page upgrades when I was removing duplicate code so the settings option was also passing `landing-preview` when my intention was to have that link pointing to `landing` and the preview link point to `landing-preview` only. That now works.

Screen Shot 2022-01-10 at 11 54 51 AM
Screen Shot 2022-01-10 at 12 11 06 PM


classes/controllers/FrmFormsController.php
```diff
@@ -1123,6 +1123,19 @@ private static function get_settings_tabs( $values ) {
...
 			),
+			'chat'        => array(
+				'name'       => __( 'Conversational Forms', 'formidable' ),
+				'icon'       => 'frm_icon_font frm_chat_forms_icon',
+				'html_class' => 'frm_show_upgrade frm_noallow',
+				'data'       => FrmAppHelper::get_upgrade_data_params(
+					'chat',
+					array(
+						'upgrade'  => __( 'Conversational Forms', 'formidable' ),
+						'requires' => 'Plus',
```
stephywells, line 1134: Let's use Business for this one.

stephywells (changes requested):
Is there a way to get the "requires" from the API response instead of hardcoding it in so many places? That makes it pretty hard to make changes easily if we move things around.

Crabcyborg:
> Is there a way to get the "requires" from the API response instead of hardcoding it in so many places? That makes it pretty hard to make changes easily if we move things around.

@stephywells yeah it looks like it's totally possible to do what we do with the category checks here. I'll try to clean up some of the duplicate category logic here. I think I could probably normalize the categories and always set a `requires` in some automatic way. I'm playing with this now.

Crabcyborg:
@stephywells I just pushed a few more updates into this:

Dropped `action_requires` I introduced into 5.0.06 for `FrmFormsHelper::get_plan_required` as the differences between the two functions does seem to be pretty negligible. It doesn't seem to really be an issue that the category gets unset after one call because it never really calls the function again on the same link variable.

I had to add "Creator" to the list for this to work with the current categories I saw on landing pages.

I fixed this up where I hard coded it for Auto-responders as well.

stephywells (approved):
Awesome! This looks great, and simplified the plans too. Thank you!!

PR #672, opened 2022-01-10
Small ui tweaks

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3379

**Updates**
- The pop ups are slightly shorter now
- No more underline on some of the menu options (wasn't intentional, it got added automatically when they were made into anchors for accessibility)
- Add more (and more consistent) margin-right value to every svg so the text is farther from the icon. I had this on the custom layout/break into rows options and didn't have the same padding on the menu options.
- Change out the gear icon. I had one in the custom layout options that wasn't being used for the field settings.
- Add some margin after the custom layout title to pad it from the subtitle.
- Use the proper trash icon
- Add a new copy icon and use that
- Change some padding. List options are bigger now, the padding on the custom layout options is slightly smaller.

Screen Shot 2022-01-10 at 4 14 21 PM

Screen Shot 2022-01-10 at 4 14 17 PM

Screen Shot 2022-01-10 at 4 14 11 PM

Screen Shot 2022-01-10 at 4 14 09 PM


js/formidable_admin.js
```diff
@@ -2051,13 +2051,13 @@ function frmAdminBuildJS() {
 	}
 
 	function getDeleteActionOption( isFieldGroup ) {
-		var option = { class: 'frm_delete', icon: 'frm_delete_icon' };
+		var option = { class: 'frm_delete', icon: 'frm_trash_icon' };
```
stephywells, line 2054: How does it look if we replace these three icons in the svg file instead of adding new ones? Are there any spots where it looks weird?

Crabcyborg: @stephywells I see it in several places

**Deleting an action**
_Before_
Screen Shot 2022-01-11 at 12 41 56 PM

_After_
Screen Shot 2022-01-11 at 12 42 39 PM

**Deleting a style**
_Before_
Screen Shot 2022-01-11 at 12 43 15 PM

_After_
Screen Shot 2022-01-11 at 12 43 28 PM

**Delete an entry**
_Before_
Screen Shot 2022-01-11 at 12 44 28 PM

_After_
Screen Shot 2022-01-11 at 12 49 14 PM

I think there's more as well but this is what I've been able to find so far.

Crabcyborg: I see it defined as well in this `FrmFormsHelper::actions_dropdown` function but that never gets called anywhere.

Should we deprecate this function? I don't think it's used.

stephywells: Thanks Mike for going through all of that. I'm not loving a lot of those.

@semelyak it seems to me like these icons might be too thin to work well in a lot of spots. Is it possible to adjust these new icons so they can be used in more places? Also, it would be great if the new icon can match the sizing of the old one. There's extra whitespace around it, making it smaller than the existing one.

PR #678, opened 2022-01-24
Add frm_entries_column_value filter

truongwp:
This is required for Quizzes

### DIFF
--- a/classes/helpers/FrmEntriesListHelper.php
+++ b/classes/helpers/FrmEntriesListHelper.php
@@ -291,6 +291,16 @@ class FrmEntriesListHelper extends FrmListHelper {
 				if ( $val === false ) {
 					$this->get_column_value( $item, $val );
 				}
+
+				/**
+				 * Allows changing entries list column value.
+				 *
+				 * @since 5.0.18
+				 *
+				 * @param mixed $val Column value.
+				 * @param array $args Contains `item` and `col_name`.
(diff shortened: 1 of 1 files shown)


classes/helpers/FrmEntriesListHelper.php
```diff
@@ -291,6 +291,16 @@ private function column_value( $item ) {
...
+
+				/**
+				 * Allows changing entries list column value.
+				 *
+				 * @since 5.0.18
+				 *
+				 * @param mixed $val Column value.
+				 * @param array $args Contains `item` and `col_name`.
+				 */
+				$val = apply_filters( 'frm_entries_column_value', $val, compact( 'item', 'col_name' ) );
```
Crabcyborg, line 303: @truongwp I see there's already a `frm_entries_{$col_name}_value` filter above this. Is it not possible to just use that?

truongwp: @Crabcyborg `$col_name` is the field key, we don't know what it is to call in the hook name.

Crabcyborg (approved):

classes/helpers/FrmEntriesListHelper.php
Crabcyborg: Sounds good! If we can't easily use the other filter this one is fine.

PR #679, opened 2022-01-24
Listen for checkbox captcha

Crabcyborg:
I noticed that a reCAPTCHA checkbox is not triggering auto-advance, but it could.

What's stopping me currently is that I can't really hook into the success event without this update.

This update adds a new triggerEvent function so we can trigger events without needing to use jQuery.

I'm not sure if it's worth adding this for checkbox reCAPTCHAs though, or if everyone is just going to use v3 from now on.

What do you think @stephywells?

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1103,6 +1103,24 @@ function frmFrontFormJS() {
 		} while ( element.previousSibling );
 	}
 
+	/**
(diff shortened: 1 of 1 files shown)


Crabcyborg:
Thanks for the input. I'll just close this as it is an outdated feature.

PR #681, opened 2022-01-26
Add role alert in default html

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3238

- Errors in new fields now include role="alert" so they get more screen reader attention on error.
- Old fields will inject role="alert" when it replaces the [error] shortcode even if it is missing from the custom HTML.
- JavaScript side validation now includes aria-describedby on an input on error. This was happening already with PHP-side validation, but not JavaScript validation.

There are also some small tweaks in pro https://github.com/Strategy11/formidable-pro/pull/3406 to add role="alert" to total fields by default as well.

### DIFF
--- a/classes/models/FrmFieldFormHtml.php
+++ b/classes/models/FrmFieldFormHtml.php
@@ -254,8 +254,8 @@ class FrmFieldFormHtml {
 			return false;
 		}
 
-		$end = strpos( $html, '[/if error]' );
-		if ( false === $end || $end < $start ) {
+		$end = strpos( $html, '[/if error]', $start );
+		if ( false === $end ) {
 			return false;
 		}
 
--- a/classes/models/FrmFieldFormHtml.php
+++ b/classes/models/FrmFieldFormHtml.php
@@ -232,10 +232,35 @@ class FrmFieldFormHtml {
 		$error = isset( $this->pass_args['errors'][ 'field' . $this->field_id ] ) ? $this->pass_args['errors'][ 'field' . $this->field_id ] : false;
 
 		if ( ! empty( $error ) && false === strpos( $this->html, 'role="alert"' ) ) {
-			$this->html = str_replace( 'class="frm_error', 'role="alert" class="frm_error', $this->html );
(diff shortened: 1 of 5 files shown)


classes/models/FrmFieldFormHtml.php
```diff
@@ -230,6 +230,11 @@ private function add_element_id( $param, $id ) {
 	private function replace_error_shortcode() {
 		$this->maybe_add_error_id();
 		$error = isset( $this->pass_args['errors'][ 'field' . $this->field_id ] ) ? $this->pass_args['errors'][ 'field' . $this->field_id ] : false;
+
+		if ( ! empty( $error ) && false === strpos( $this->html, 'role="alert"' ) ) {
```
stephywells, line 234: Do you think this might be safer to check for ' role=' in case someone has added something else? I'm not sure what they might add though.

Crabcyborg: I think we'd still want something with role="alert". I think most other possible role values, things like role="group" and role="region", which will expect an alert as well.

stephywells: Yeah, I'm just wondering if role was already in the HTML, would it be a problem if we ended up with two?

Crabcyborg: I just pushed a commit that adds some extra checks so that isn't possible anymore.

https://github.com/Strategy11/formidable-forms/pull/681/commits/56c5a7add27161ae15edb1a77fb06f7d24711f61

stephywells: Looks great! I'm not sure if it would matter. But this covers it just in case.

stephywells (approved):

garretlaxton (approved):
Looks good! Here's what I see when I get an error on my form.
`<div class="frm_error" role="alert" id="frm_error_field_3hspm">This field cannot be blank.</div>`

I also see the aria-describedby tag when JS validation is enabled.

Crabcyborg:
Thanks for testing @garretlaxton!! 🥳

PR #684, opened 2022-01-28
New function postToAjaxUrl

Crabcyborg:
This is part of https://github.com/Strategy11/formidable-pro/issues/3396

This covers every non-deprecated call that uses frm_js.ajax_url in the free plugin, which is only 3 places so this wasn't too difficult.

I've moved the call to `jQuery.ajax` into the function and cleaned this all up a bit. The shared details (that it's a POST request, that it uses the ajax url, have all been moved into the function.

The new function is pretty straight forward. The first parameter is `form`, the element of the `<form>` tag which needs to be passed to every call to this function to determine whether or not the target form is an API form or not.

In my testing, a form loaded with the API includes an action attribute that uses the ajax url, just pointed to the preview page instead. So it's easy to parse this URL and use the AJAX url already available in the HTML.

Otherwise we just fallback to the frm_js.ajax_url so it functions the same as it always did if it's not loaded by the API add on.

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -684,9 +684,9 @@ function frmFrontFormJS() {
 		var ajaxUrl, action, ajaxParams;
 
 		ajaxUrl = frm_js.ajax_url;
+		action = form.getAttribute( 'action' );
 
-		if ( form.getAttribute( 'action' ) ) {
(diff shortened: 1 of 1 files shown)


stephywells (approved):
This looks like a great approach. Nice!

PR #685, opened 2022-01-28
Replace shortcodes column with embed icon and modal

Crabcyborg:
For now I've opted out of showing the form key variant for simplicity. I haven't seen any update from Yaro yet, but we could add more examples later.

This update also adds a little bit of vertical adjustment to the modal. I noticed the WP admin bar was getting in the way of the top of the new form modal. It looks nicer now.

**Completed**
- [x] Use the third icon from the list I posted for copy.
- [x] Add missing icons from design
- [x] Add copy icon/copy functionality
- [x] Support education link like in design
- [x] I haven't added anything to the modal footer yet
- [x] I think I'm picking textareas at an arbitrary length. I should be more accurate here - what can fit one line?
- [x] Test accessibility
- [x] Add required validation to the existing page dropdown

**How it looks by default**
Screen Shot 2022-02-02 at 12 15 14 PM

**With the API add on filter**
Screen Shot 2022-02-02 at 12 14 27 PM


stephywells:
It looks like the "how to use shortcodes" link has focus when I click the Embed shortcode.
Screen Shot 2022-02-02 at 10 41 10 AM

And instead of adding another new icon for copy, can we use the one that's already there?

Other than that, this looks like a great start!

It would be super cool if we could include this part too, to make it easy to add a form to a page. I'm not sure how much extra time that would add right now.
https://www.figma.com/file/pz62UBLsxjYqLo4pz2qp67/Formidable_embed-form?node-id=103%3A934

Crabcyborg:
Thanks @stephywells I didn't notice that page. That looks like an easy enough add!

As for the auto-focus, the focusing happens pretty automatically. The link is just the first thing with a tabindex. What are you thinking should get the first focus instead? The first copy trigger? I tried to avoid tabbing into the inputs directly because readonly inputs don't really get detected as readonly by screen readers, even with aria-readonly set.

As for the icon, I'm just struggling with all of the conflicting icons right now. Which one of these other icons do we want exactly?

Screen Shot 2022-02-02 at 1 53 27 PM
Screen Shot 2022-02-02 at 1 53 41 PM
Screen Shot 2022-02-02 at 1 54 02 PM
Screen Shot 2022-02-02 at 1 54 16 PM

Our "copy" icons are a little out of hand. So I agree adding a 5th one probably isn't a good call 😛.

stephywells:
Yep, let's go with the third one. Let's try really hard in the future not to add new icons unless we have to. It looks like we need to get rid of those first two and use the second set instead. I suppose we could deprecate the old ones in the icon_by_class() function.

So, for this one, let's either the third icon. We can replace it with the new icon if we have a solid version too, and if it's the same proportions. And the SVg color can be set in the css so the color matches the design.

Crabcyborg:
@stephywells I've got another question.

For the first option on this page, insert into an existing page, how do you envision that working? It seems like the most complicated one to implement with the least likely chance that someone would be looking for it.

Screen Shot 2022-02-02 at 2 36 45 PM

I'm thinking we probably just ignore that option for now, but I wanted to run that by you first.

I've pushed a rough commit that adds the first page to the modal, that will generate a page with the shortcode in it now. I need to clean this all up though - I'm seeing the Done button is getting cut off, and I should add a back option to the modal, and I need to add the icons to the options. I also need to address some of your other comments still.

stephywells:
@Crabcyborg WPForms is doing a really good job with this. Check out the Embed process:
https://qa.formidableforms.com/competitors/wp-admin/admin.php?page=wpforms-builder&view=fields&form_id=4

css/frm_admin.css
```diff
@@ -2359,6 +2359,103 @@ li[data-preview] .frm-hover-icons .frm-preview-form,
...
+	top: 8px;
+}
+
+.frm_embed_form_content > div {
+	position: relative;
+}
+
+.frm_embed_form_content svg[id^="frm_copy_embed_"] {
+	position: absolute;
+	bottom: 25px;
```
stephywells, line 2409: This looks more centered to me with 30px instead of 25.

css/frm_admin.css
```diff
@@ -2359,6 +2359,103 @@ li[data-preview] .frm-hover-icons .frm-preview-form,
...
+	justify-content: center;
+}
+
+#frm_form_embed_modal .frm_embed_example {
+	background: #F6F7FB;
+	border: 1px solid #E3E8EB;
+	border-radius: 6px;
+	color: rgba(40, 47, 54, 0.85);
+	font-size: 14px;
+	padding: 10px;
```
stephywells, line 2386: This works a bit better for me so the text doesn't overlap the icon:
padding: 10px 30px 10px 10px

stephywells (changes requested):
Nice job on this! It looks great.

When I choose an existing page, nothing comes up when I start typing into the page name box. Is this supposed to be an autocomplete box?

It would also be great to add this embed button at the top of the form settings page. We can remove the shortcode section from that page also.

images/icons.svg
```diff
@@ -920,5 +920,6 @@
...
 <symbol id="frm_thick_more_vert_icon" viewBox="0 0 20 20"><path d="M10 10.834a.833.833 0 100-1.667.833.833 0 000 1.667zM10 5a.833.833 0 100-1.667A.833.833 0 0010 5zM10 16.667A.833.833 0 1010 15a.833.833 0 000 1.667z" fill="#9EA9B8" stroke="#9EA9B8" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/></symbol>
 <symbol id="frm_chat_forms_icon" viewBox="0 0 21 21" fill="none" stroke="currentColor"><path d="M19.991 13.914a2 2 0 0 1-2 2h-12l-4 4v-16a2 2 0 0 1 2-2h14a2 2 0 0 1 2 2v10Z" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/></symbol>
+<symbol id="frm_embed_icon" viewBox="0 0 20 20" fill="none"><path d="M13.3335 15L18.3335 10L13.3335 5" stroke="currentColor" stroke-width="1.3" stroke-linecap="round" stroke-linejoin="round"/><path d="M6.6665 5L1.6665 10L6.6665 15" stroke="currentColor" stroke-width="1.3" stroke-linecap="round" stroke-linejoin="round"/></symbol>
```
stephywells, line 923: It looks like we could remove the existing frm_code_icon icon for the html field with this one. This new icon looks fine for the HTML field in the form builder.

garretlaxton:
Only thing I see is what Steph mentioned. When selecting an existing page I am only taken to edit the page with no form embedded.

It may be helpful to add a `target="_blank"` parameter to the _How to use shortcodes in WordPress_ link so it opens in a new tab rather leaving the page.

css/frm_admin.css
Crabcyborg: Thanks @stephywells. The design actually has them closer to the top, I just changed it to bottom when I was trying to make labels optional before. I've seen the copy for GitHub is always in the top right, so I've updated it to be there instead. With the other padding update, this doesn't get in the way of the example anymore (sort of another reason I was using the bottom corner instead - less likely to be in the way).

Screen Shot 2022-02-04 at 10 23 18 AM

Crabcyborg:
> It would also be great to add this embed button at the top of the form settings page. We can remove the shortcode section from that page also.

@stephywells 

Are you picturing just adding another new button?
Screen Shot 2022-02-04 at 10 14 52 AM

Or doing a split button and moving the preview button to a modal?
screen_shot_2022-02-04_at_10 19 29_am

images/icons.svg
Crabcyborg: Thanks @stephywells that works/looks fine, I considered replacing it initially but wasn't sure.

Screen Shot 2022-02-04 at 10 31 14 AM

Crabcyborg:
@stephywells I've added a new filter so the landing page can add an icon as well https://github.com/Strategy11/formidable-landing/pull/14

I've also fixed issues with the autocomplete and updated the icon, and added the new embed button to the settings and form builder pages.

Crabcyborg:
> How hard would it be to add a bit of guidance on the edit page? I'm not sure if we could automatically open the block panel and search for "Formidable"?

@stephywells it seems pretty tough. I played with it in the console at leas and triggering the search with JavaScript seems to not work.

It's easy to popup with insert pop up by triggering a click on the "+" toggle.
```
document.querySelector( '.block-editor-inserter__toggle' ).click();
```

And then setting the input value is not problem.
```
document.querySelector( '.block-editor-inserter__quick-inserter input[type="search"]' ).value = 'Formidable';
```

But I can't get it to actually search. I can't tell which event is triggering it. I tried triggering a few different events on different elements with different types (change, click, focus, keyup, keydown, input, keypress) but none of them seem to make any difference.

We could add an action for this instead and forcefully inject HTML into their page. It's much riskier to do that with a page that already has content in it, but it's possible. I just avoided it because WP Forms doesn't do anything else.

Crabcyborg:
@stephywells sorry about that, I deleted my last comment because I found another option that I thought was better.

I made a new commit
https://github.com/Strategy11/formidable-forms/pull/685/commits/2555960ce05e9c345a891715eeb4f998849bac39

I found out how to programmatically insert the block JavaScript side on load when we're passing a `$_GET['frmForm']` value to the Gutenberg page.

I didn't notice the pop up in WP Forms at first, thanks for pointing that out. It's pretty easy to miss.

garretlaxton:
I am able to see the block when selecting an existing page, but when I save or preview the page the form doesn't save.
Screenshot 2022-02-07 at 19-16-42 Edit Page “test save” ‹ Formidable Forms — WordPress
Screenshot 2022-02-07 at 19-17-05 test save – Formidable Forms

Going back to the edit the page shows this in the block.
Screenshot 2022-02-07 at 19-19-58 Edit Page “test save” ‹ Formidable Forms — WordPress

Crabcyborg:
Thanks for testing that @garretlaxton! It'll have to see what's causing that.

Crabcyborg:
@garretlaxton that issue should be fixed now.

It turns out that it's pretty picky about type. I was passing the formId as an integer but it needed a string.

https://github.com/Strategy11/formidable-forms/pull/685/commits/16108367fe2ac8d117f2b351185957c07f78f404

classes/controllers/FrmAppController.php
```diff
@@ -511,6 +523,68 @@ public static function admin_js() {
...
+								closeListener();
+								requestAnimationFrame( () => injectFormidableBlock() );
+							}
+						}
+					);
+				}
+
+				document.addEventListener( 'DOMContentLoaded', handleDomReady );
+
+				const injectFormidableBlock = () => {
```
stephywells, line 576: This works great now! Awesome!

When I choose a page with a lot of content, I can't see that a form got added. Is it possible to scroll to the new form block too?

Crabcyborg: @stephywells I just pushed a new commit that auto-scrolls after a new block is detected https://github.com/Strategy11/formidable-forms/pull/685/commits/1c81b0fcb18ffb21dfeef971bbc62a0b20a814a8

🥳

garretlaxton (approved):

stephywells (approved):
This is great! Thanks so much Mike!

PR #686, opened 2022-02-02
Only include class on dropdown option if other option

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/1775519188/91384/

The other trigger has a special class "frm_other_trigger" but every other option in the dropdown is showing `class=""` which is a lot of unnecessary HTML getting printed to the page. 

**Before**
image

**After**
Screen Shot 2022-02-02 at 9 02 08 AM


stephywells (approved):
Good call!

PR #689, opened 2022-02-03
New filter frm saved css

Crabcyborg:
I need to add this new filter so that the API add on can overwrite the path to the fonts and use an action instead.

Because of CORS blocking the fonts need to include the `Access-Control-Allow-Origin` header but in order to do that and have it work on both Apache and Nginx (since htaccess doesn't work for both) I need to do this with PHP and serve the file afterward.

Related branch https://github.com/Strategy11/formidable-api/pull/103


stephywells (approved):
Looks great!

PR #691, opened 2022-02-08
Use wp-block-editor, not wp-editor

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3134

I googled for the issue and found a Yoast thread with the same issue https://github.com/Yoast/wordpress-seo/pull/17999

The fix is pretty simple. It looks like `wp-editor` is deprecated and we're supposed to use `wp-block-editor` which has all of the same functions/classes included that we were using. I tested the form block out in a widget and in the page builder. No warning anymore.


Crabcyborg:
> Awesome. Is there anything we need to do for reverse compatibility?

@stephywells I think there could probably be issues with versions of WordPress that don't include the block-editor. It isn't very clear in the documentation when that was added.

I tried switching between WordPress branches and it looks like version 5.0 and 5.1 did not include the block editor. But 5.2+ do.

Do you think it's worth trying to do the version check and loading one or the other? Or would we want to change our WordPress target to 5.2 to keep this simple?

stephywells:
I think it's fine to set the min version to 5.2. I think WordPress will warn people or not allow the update. But the new  feature warrants a "v5.2"

PR #693, opened 2022-02-09
Menu position should always be an integer

Crabcyborg:
I've switched to using the WordPress 6.0.0 beta locally.

It's logging a new deprecation.

> PHP Notice:  Function add_menu_page was called <strong>incorrectly</strong>. The seventh parameter passed to <code>add_menu_page()</code> should be an integer representing menu position. Please see <a href="https://wordpress.org/support/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 6.0.0.) in /var/www/src/wp-includes/functions.php on line 5768

I'm not totally sure why these are strings at the moment or why they are using decimals. But it looks like WordPress wants integers (non-decimal values), so I rounded down to 29 instead. Same position as before (under comments).

I looked at the last commit that changed this three years ago. It mentions that it should be "below comments" so I think the rounded value is probably fine. https://github.com/Strategy11/formidable-forms/commit/2239413a30f390b6ea85f3b6a2ecacfd4f570d6d

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -15,8 +15,11 @@ class FrmAppController {
 		add_menu_page( 'Formidable', $menu_name, 'frm_view_forms', 'formidable', 'FrmFormsController::route', self::menu_icon(), self::get_menu_position() );
 	}
 
+	/**
+	 * @return int
+	 */
 	private static function get_menu_position() {
-		return apply_filters( 'frm_menu_position', '29.3' );
+		return (int) apply_filters( 'frm_menu_position', 29 );
 	}
 
 	/**


stephywells:
I'm not sure if this is an issue anymore, but this was a decimal because if another nav had the same, exact order, one of the nav items wouldn't show up. Maybe this has been fixed now?

Crabcyborg:
@stephywells thanks for clearing that up. I just tried adding a second duplicate icon with the same 29 and it appears to display both without any issue!

Screen Shot 2022-02-09 at 3 04 40 PM

stephywells (approved):
Perfect! Thanks for checking that!

PR #697, opened 2022-02-15
Do not call json decode on null fix

Crabcyborg:
Noticed a PHP8+ deprecated message in this function. I'm not totally positive how I ended up with a null, I was playing with a few things.

> PHP Deprecated:  json_decode(): Passing null to parameter #1 ($json) of type string is deprecated in /var/www/src/wp-content/plugins/formidable/classes/helpers/FrmAppHelper.php on line 2571


stephywells (approved):
Thanks!

PR #699, opened 2022-02-16
Add a PHPStan workflow

Crabcyborg:
Fixes two remaining issues. I'm hiding the rest at the moment with the goal of removing exceptions over time. But it is running at the max level already aside from those exceptions.

Runs in 7-11 seconds, significantly the fastest workflow which is nice for a few reasons - catching issues faster, using less of our GitHub allowance.

 ------ -----------------------------------
  Line   classes/helpers/FrmListHelper.php
 ------ -----------------------------------
  1156   Undefined variable: $data
 ------ -----------------------------------

I trace this error back to 6 years ago when Jamie copied the file from WordPress and missed the undefined variable https://github.com/Strategy11/formidable-forms/commit/14289c2500f3a358abbb5fbc2370da6f6bd4cb33

It's trying to call a function that starts with `_column` which we never do, so this code was never getting called. It looks like it's used in two WordPress functions `_column_blogs` and `_column_title` but since we are not extending WP_List_Table this is just extra code with an undefined variable in it. So I'm removing the condition and block of code entirely.

 ------ ----------------------------------------------------------------------------------------------------------------------------
  Line   classes/models/FrmEntryFormatter.php
 ------ ----------------------------------------------------------------------------------------------------------------------------
  711    Parameter #1 $field_value of method FrmEntryFormatter::package_value_args() expects FrmFieldValue, FrmProFieldValue given.
 ------ ----------------------------------------------------------------------------------------------------------------------------

Just required a comment update. A FrmProFieldValue value is a FrmFieldValue value anyway, because of inheritance. Nothing crazy.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -26,6 +26,7 @@ package.json export-ignore
 phpcs.xml export-ignore
 phpunit.xml export-ignore
 psalm.stubs.php export-ignore
+phpstan.stubs.php export-ignore
 psalm.xml export-ignore
 readme.md export-ignore
 tests export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -26,8 +26,9 @@ package.json export-ignore
 phpcs.xml export-ignore
 phpunit.xml export-ignore
 psalm.stubs.php export-ignore
-phpstan.stubs.php export-ignore
 psalm.xml export-ignore
+phpstan.stubs.php export-ignore
+phpstan.neon export-ignore
 readme.md export-ignore
 tests export-ignore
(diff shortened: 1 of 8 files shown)


stephywells (approved):
Nice! Does that mean we can remove any workflows? Does this replace one?

Crabcyborg:
> Nice! Does that mean we can remove any workflows? Does this replace one?

@stephywells I don't think so, no. PHPStan, Psalm, and PHPCS all have unique strengths and all catch unique issues that the others don't. This is really just a complimentary third option that will help avoid new bugs automatically.

PR #705, opened 2022-02-21
Show addon install popup when clicking to locked template

truongwp:
Todo:

- [ ] Apply this for templates in Quizzes categories, not Contact form.
- [ ] Update addon install URL.
- [ ] Maybe remove or change `ADDON` text in the locked template.

What I did:
- Check if template requires an addon and that addon is not installed, add `'addon' => 'slug'` to the template array.
- If template contains `addon`, show it as locked and add `data-addon="slug"` to the `<li> element.
- When showing `upgrade` page of new form modal, check if `<li>` contains `data-addon`, change all text/image/url of upgrade content into addon install content. The addon install content is registered in global JS variable.


stephywells:
@truongwp lets start with just what is needed for quizzes here. Do we need anything besides the added category in the list, and the template Id? I think if it's just that, we can release it before quizzes even. 

Let's hold off on the rest of this so we can do it in a way that works more universally.

truongwp:
@stephywells Do you mean we just need the quiz template and the category, we don't need the install Quizzes popup?

stephywells:
@truongwp yes, exactly. For this PR, let's do it the same way the registration template works right now.

truongwp:
@stephywells This is the code set icon for Quizzes category: https://github.com/Strategy11/formidable-forms/pull/705/files#diff-c3f25890259e06951f6e9c7f66838c489ff9c9e0462fd21978a96d8a3e869159R1283

Look like the category for template is set outside plugin, we will get them via API.

Do we need to add Quizzes here? https://github.com/Strategy11/formidable-forms/blob/master/classes/controllers/FrmAddonsController.php#L118

stephywells:
@truongwp Yes, exactly. So we only need that line, and the line that adds the template id. But we should use a singular term so it matches the others (Quiz instead of Quizzes).

The id for this template will be 28109851.

> Do we need to add Quizzes here? https://github.com/Strategy11/formidable-forms/blob/master/classes/controllers/FrmAddonsController.php#L118
Sure! That would be great!

Crabcyborg:
I'm going to close this for now for a couple reasons.

1. It's so old that a lot needs to change now. There are merge conflicts, and this updates old removed code since this was part of the old new form modal which has been replaced by the form templates page.
2. There are JS errors when you're using an Elite license because some of these expected elements don't exist.
3. We have a new `used_addons` key that we can use instead of checking for specific hard coded categories. We'll need to adjust this logic to use that instead.

I figured out how to trigger this so I could get a screenshot of what it looked like.

Screen Shot 2024-12-04 at 9 59 25 AM

PR #707, opened 2022-02-21
Fix issue with checkbox styling in enfold

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3397

Took a while figuring out how to reliably replicate this without enfold. I ended up copying the CSS from their demo into a blank HTML file and pulled in a new API form 😂.

The issue is actually pretty simple, just annoying to debug. Enfold includes this rule on checkboxes, `#top input[type="checkbox"]{display:inline}`. Because of the `#top` reference it beats our CSS rule and forces the checkbox to inline instead of inline-block. Since it's targeting by id, I think our best bet here is to just use `!important`.

### DIFF
--- a/css/custom_theme.css.php
+++ b/css/custom_theme.css.php
@@ -791,7 +791,7 @@ a.frm_save_draft{
 	appearance: none;
 	background-color: var(--bg-color);
 	flex: none;
-	display:inline-block;
+	display:inline-block !important;
 	margin: 0 5px 0 0;
 	color: var(--border-color);
 	width: 18px;


css/custom_theme.css.php
```diff
@@ -791,7 +791,7 @@
 	appearance: none;
 	background-color: var(--bg-color);
 	flex: none;
-	display:inline-block;
+	display:inline-block !important;
```
stephywells, line 794: Usually, we use <?php echo esc_html( $important ); ?>; so the important can be disabled. Would that do the job?

Crabcyborg: @stephywells ha, the PHP code in your comment disappeared, I was confused for a second until I edited your comment and saw the code there.

With all of the styles being applied to a checkbox now I really think we want to force `!important` here no matter what the setting is set to. Inline styling for this checkbox completely breaks it, and if someone didn't have the important setting on, then they'd be stuck with it broken still.

This isn't the first time we're setting it no matter what. If I search for `!important` in this file, I get 43 matches.

stephywells: Oops! Sorry about that!

That sounds like a good plan.

PR #708, opened 2022-02-21
Update to Bootstrap 4.6.1

Crabcyborg:
**I considered updating to Bootstrap 5 and Popper 2 but so far Bootstrap Multiselect does not work with Bootstrap 5. Here is an open issue https://github.com/davidstutz/bootstrap-multiselect/issues/1226**. With Bootstrap 5 they add back support for the `<li>` tags so this solution could possibly just be temporary, and might be removable once we migrate to Bootstrap 5.

Fixes https://github.com/Strategy11/formidable-pro/issues/3402.

Requires https://github.com/Strategy11/formidable-pro/pull/3438 for a few pro features to work. If `dropdown-menu` is missing from a list, Popper throws errors.

I believe this also fixes https://github.com/Strategy11/formidable-pro/issues/3176. It's most likely that conflicts have all been because the other plugin is loading Bootstrap 4.

Bootstrap is used in a few places (all back end).

**Multiselect dropdowns**
Screen Shot 2022-02-21 at 4 31 07 PM

**Preview dropdowns**
Screen Shot 2022-02-21 at 4 31 25 PM

**Field dropdowns**
Screen Shot 2022-02-21 at 4 31 40 PM

**Field group dropdowns**
Screen Shot 2022-02-21 at 4 32 01 PM

**Tooltips**
Screen Shot 2022-02-21 at 4 32 18 PM
Screen Shot 2022-02-21 at 4 32 15 PM

**Create a form from a template dropdown**
Screen Shot 2022-02-21 at 4 51 12 PM

**Form search dropdown**
Screen Shot 2022-02-21 at 4 56 08 PM


garretlaxton (approved):
Looks good, tested all the mentioned fields in the backend and all still work as should.

classes/helpers/FrmFormsHelper.php
```diff
@@ -134,9 +134,9 @@ public static function form_switcher( $selected = false ) {
 					<?php FrmAppHelper::icon_by_class( 'frmfont frm_arrowdown4_icon', array( 'aria-hidden' => 'true' ) ); ?>
 				</h1>
 			</a>
-			<ul class="frm-dropdown-menu frm-on-top frm-inline-modal frm_code_list frm-full-hover" role="menu" aria-labelledby="frm-navbarDrop">
+			<div class="dropdown-menu frm-dropdown-menu frm-on-top frm-inline-modal frm_code_list frm-full-hover" role="menu" aria-labelledby="frm-navbarDrop">
```
stephywells, line 137: Is it possible to use the frm-dropdown-menu class instead of adding dropdown-menu? We had to prefix this to avoid conflicts. 

Also, can you help me understand why we need to switch from ul to div? I'm concerned about the reverse compatibility with any of our add-ons that are using multiselect.

Crabcyborg: @stephywells I don't believe it's required to change them all to divs. The main things that were always causing breaking issues were not having the `dropdown-menu` class and not having the `dropdown-item` classes.

But there's no reason for the `ul`. The bootstrap multiselect add on at least made that transition from 3 to 4 (https://davidstutz.github.io/bootstrap-multiselect/#migration). I'm all for it because it isn't as easy to work with ul/li elements. But I can probably put that part back.

The thing with not having the `dropdown-menu` class is that everything completely breaks with bootstrap and popper. I think it's a hard requirement. I'm looking at the source code for bootstrap and I see a couple of lines that hint to that:

`var SELECTOR_MENU = '.dropdown-menu';`

````
if (parent) {
          this._menu = parent.querySelector(SELECTOR_MENU);
 }
 ```

Crabcyborg: @stephywells I figured out a way to get this to work without `dropdown-menu` by overwriting the `_getParentFromElement` function in bootstrap. https://github.com/Strategy11/formidable-forms/pull/708/commits/24f8e865b9868830cd65afe0497d6ec01ded269c

It adds the class just for the split moment where bootstrap needs it, then removes it again.

Crabcyborg:
@garretlaxton I've made some pretty big changes so I may have introduced a bug since last time, if you could give this another run through.

stephywells (approved):
Great idea! This is an awesome way of making it reverse compatible. We can phase this out too once we update it in any add-ons and add a console message for deprecation.

garretlaxton (approved):
Looks good still!

PR #709, opened 2022-02-22
Add Quiz template category and ID

truongwp:
Separate from #705 

How can I see the template in the list? And now we have 7 items the top section of modal, we need to add or remove one item.

### DIFF
--- a/classes/views/frm-forms/list-templates.php
+++ b/classes/views/frm-forms/list-templates.php
@@ -15,7 +15,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 		</div>
 	</li><?php
 	$render_icon = true;
-	foreach ( array( 20872734, 20874748, 20882522, 20874739 ) as $template ) {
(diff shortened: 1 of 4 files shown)


classes/views/frm-forms/list-templates.php
```diff
@@ -15,7 +15,7 @@
 		</div>
 	</li><?php
 	$render_icon = true;
-	foreach ( array( 20872734, 20874748, 20882522, 20874739 ) as $template ) {
+	foreach ( array( 28109851, 20872734, 20874748, 20882522, 20874739 ) as $template ) {
```
stephywells, line 18: Let's move the quiz to the end of this array. How about we add the survey template too so we'll have an even number. 

The survey Id is 20908981. Let's put it before the quiz template.

stephywells (changes requested):

stephywells (approved):
Thanks! Looks perfect.

PR #710, opened 2022-02-22
Set current screen on preview

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3436

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -270,8 +270,12 @@ class FrmFormsController {
 		echo self::page_preview(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 	}
 
+	/**
+	 * @return void
+	 */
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Nice, simple fix! It seems like that was in there at some point and we removed it for a conflict. But I'm not finding it, so I might be making it up.

Crabcyborg:
@stephywells it looks like it's really a common pattern.

I noticed next that it was also logging `PHP Warning:  Undefined global variable $hook_suffix in /var/www/src/wp-admin/includes/class-wp-screen.php on line 223`. I looked up the other references to `global $hook_suffix` and this pattern has appeared in a couple of other places before, just not the preview page logic.

I've refactored it now so all of those similar cases use the same new function https://github.com/Strategy11/formidable-forms/pull/710/commits/311ac68ab9b19133e094709c7b432a03558d748a.

PR #717, opened 2022-02-25
Remove margin-bottom of combo child fields

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3374

### DIFF
--- a/css/_single_theme.css.php
+++ b/css/_single_theme.css.php
@@ -73,10 +73,6 @@ $arrow_icons = FrmStylesHelper::arrow_icons();
 .<?php echo esc_html( $style_class ); ?> .form-field{
 	margin-bottom:<?php echo esc_html( $field_margin . $important ); ?>;
 }
-
-.<?php echo esc_html( $style_class ); ?> .frm_combo_inputs_container > .form-field{
-	margin-bottom:0;
-}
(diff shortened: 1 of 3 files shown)


css/_single_theme.css.php
```diff
@@ -73,6 +73,10 @@
 .<?php echo esc_html( $style_class ); ?> .form-field{
 	margin-bottom:<?php echo esc_html( $field_margin . $important ); ?>;
 }
+
+.<?php echo esc_html( $style_class ); ?> .frm_combo_inputs_container > .form-field{
```
Crabcyborg, line 77: @truongwp the CSS is built using two CSS files: custom_theme.css.php and this file. This file adds the rule for each style in the database, but that isn't necessary for this style at all (as it isn't a custom setting).

We should move it to custom_theme.css.php instead.

truongwp: I think that is the right place to add that CSS. The `margin-bottom` above it causes this issue. If that `if` block is not run, we don't need this CSS too.

I found that I'm missing `$important` too.

truongwp: I got it. It's updated now.

Crabcyborg (approved):

PR #719, opened 2022-03-02
Allow name fields to be the entry name

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3434

This doesn't apply to old entries unless the user updates them.
Besides the `name` and `text` fields, we can add support for more field types. I'm not sure if we should do that.

### DIFF
--- a/classes/models/FrmEntryValidate.php
+++ b/classes/models/FrmEntryValidate.php
@@ -148,21 +148,6 @@ class FrmEntryValidate {
 			$item_name = $value;
 		}
 
-		/**
-		 * Allows modifying the item name before it's saved to database.
-		 *
-		 * @since 5.2.02
-		 *
-		 * @param string|false $item_name The item name.
-		 * @param array        $args      {
-		 *     The arguments.
-		 *
-		 *     @type object $field The current processing field.
-		 *     @type mixed  $value The value of current processing field.
-		 * }
-		 */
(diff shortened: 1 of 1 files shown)


Crabcyborg:
Thanks @truongwp, I think we could probably improve on this more. I'm concerned that a form with no text or name fields would end up calling this function and filter a lot, possibly taking a performance hit. But leaving it open makes sense so it's possible to filter.

I think we should probably only call the filter once with all posted fields at the end, after validation but before it saves it. We could track some variable here, maybe with a global, that catches something like "first available item name field" and use that value before passing it to the filter, catching text and name fields by default.

Right now the place where this logic is is a little odd as well. If it's after `if ( $posted_field->required == '1' && FrmAppHelper::is_empty_value( $value ) ) {` then that means we're calling this every time there is an empty but optional field, which isn't necessary. We should isolate the empty check or something.

truongwp:
Thanks, @Crabcyborg. Now I think we don't need that new filter. Users can hook into `frm_pre_create_entry` to change the entry name.
Look like calling this new method (without filter) doesn't slow down much, so I think we don't need to move it at the end of processing.

Crabcyborg:
@truongwp sounds good!

truongwp:
I updated the code. Please check it @Crabcyborg.

Crabcyborg (approved):

PR #721, opened 2022-03-03
Remove unused function FrmFormsController::init_modal

Crabcyborg:
I noticed we call these two lines of code in a few places and I was refactoring it to one function.

Then I noticed one of those functions is private and never gets called so I'm removing it.


stephywells (approved):
Good call!

PR #723, opened 2022-03-04
Set a string hook suffix in set_current_screen_and_hook_suffix to avo…

Crabcyborg:
…id a PHP 8.1 null deprecation issue

> [04-Mar-2022 17:33:44 UTC] PHP Deprecated:  substr(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/src/wp-admin/includes/class-wp-screen.php on line 231

Just avoiding a deprecation. The hook suffix not only needs to exist, it can't be null.


stephywells (approved):
Great

PR #728, opened 2022-03-08
Use NEW tag for new templates

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3454


Crabcyborg (approved):
Thanks @truongwp!

PR #733, opened 2022-03-09
Fix Duplicator Pro conflict

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3459

They use Popper 2 but Bootstrap 4 uses Popper 1. De-registering it so it registers our popper js instead fixes the issue.

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -459,9 +459,9 @@ class FrmAppController {
 			$dependecies[] = 'wp-color-picker';
 		}
 
+		self::register_popper1();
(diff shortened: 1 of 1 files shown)


stephywells (changes requested):
This looks like a great suggestion. I'm thinking this might break their stuff though. Maybe we can only deregister on Formidable pages?

Crabcyborg:
> This looks like a great suggestion. I'm thinking this might break their stuff though. Maybe we can only deregister on Formidable pages?

Thanks @stephywells good point. I've updated it so it only calls this when it is enqueuing `formidable_admin.js` now. I moved the register popper code a bit so it could quickly register popper 1 after de-registering popper 2.

stephywells (commented):
Thank you!

PR #738, opened 2022-03-15
Allow for custom field key separators

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/1814896451/94441

> If you could add "___" to me before each random string, it would solve all my problems.

It seems like the easiest solution to their issue. Seems like a nice enhancement to add as well.

Example use.
```
function use_custom_field_key_separator() {
	return '___';
}
add_filter( 'frm_unique_field_key_separator', 'use_custom_field_key_separator' );
```

Also supports `add_filter( 'frm_unique_form_key_separator', 'use_custom_form_key_separator' );`, etc. I figure passing the column in the filter name makes the most sense in this case.

But I'm open for suggestions to a different solution as well.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1816,8 +1816,19 @@ class FrmAppHelper {
 
 		if ( $key_check || is_numeric( $key_check ) ) {
 			$key = self::maybe_truncate_key_before_appending( $column, $key );
+
+			/**
+			 * Allow for a custom separator between the attempted key and the generated suffix.
+			 *
+			 * @since 5.2.03
+			 *
+			 * @param string $separator. Default empty.
+			 * @param string $key the key without the added suffix.
+			 */
+			$separator = apply_filters( 'frm_unique_' . $column . '_separator', '', $key );
+
 			// Create a unique field id if it has already been used.
-			$key = $key . substr( md5( microtime() . rand() ), 0, 10 );
+			$key = $key . $separator . substr( md5( microtime() . rand() ), 0, 10 );
 		}
 
 		return $key;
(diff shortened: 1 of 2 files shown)


stephywells (approved):
Wow. This one little line can solve all his problems, huh? Impressive.

I think the hook name looks good and ready to merge.

I just got to thinking... We did it this way to avoid a loop for multiple database calls checking if the key had already been used. The performance was an issue. But it might work to get all keys that start with the key we're checking and loop that way. I'm not sure if it's worth changing or not, but that might help to get more compact field keys. A duplicate of a duplicate gets really long.

Crabcyborg:
> But it might work to get all keys that start with the key we're checking and loop that way.

@stephywells I just pushed a new update that does it this way. It wasn't too complicated, and shouldn't be much slower really than the other check.

I also added some unit tests, including some for the new separator filter.

classes/helpers/FrmAppHelper.php
```diff
@@ -1805,19 +1805,35 @@ public static function get_unique_key( $name, $table_name, $column, $id = 0, $nu
...
+			 *
+			 * @since 5.2.03
+			 *
+			 * @param string $separator. Default empty.
+			 * @param string $key the key without the added suffix.
+			 */
+			$separator = apply_filters( 'frm_unique_' . $column . '_separator', '', $key );
+
+			do {
+				$suffix    = substr( md5( microtime() . rand() ), 0, 5 );
```
stephywells, line 1832: Awesome! That's sure to be unique now. With this improvement, we can just increment the number here rather than using md5. So, after duplication the key would go from 'first-name' to 'first-name1', and another duplication would change 'first-name1' to 'first-name11' or 'first-name2'

stephywells (changes requested):

PR #739, opened 2022-03-22
Do not load css late if all in one seo is active

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3479

I can't really figure out why All in One SEO is causing this. But as long as `wp_print_styles` does not get called, it seems to work.

@stephywells you may understand this better. There might be a more robust solution. I didn't really get to the core of what's causing this.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -2250,18 +2250,32 @@ class FrmFormsController {
 	 * to prevent a flash of unstyled form.
 	 *
 	 * @since 4.01
+	 *
+	 * @return void
 	 */
 	private static function load_late_css() {
 		$frm_settings = FrmAppHelper::get_settings();
-		$late_css = $frm_settings->load_style === 'dynamic';
-		if ( ! $late_css ) {
+		$late_css     = $frm_settings->load_style === 'dynamic';
+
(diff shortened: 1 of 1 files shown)


stephywells:
I just talked with Brohas about this. He says this is a bug that will be fixed in their next release. The problem is that the content is getting run in the meta description, so the form thinks the style has already been loaded. 

This hook will prevent it for now too:
https://aioseo.com/docs/aioseo_conflicting_shortcodes/

I’m fine with either fix. Whichever you think is better is great.

Crabcyborg:
@stephywells how does the hook work exactly? 

I tried adding a quick snippet:

```
add_filter(
	'aioseo_conflicting_shortcodes',
	function( $conflicting_shortcodes ) {
		$conflicting_shortcodes = array_merge(
			$conflicting_shortcodes,
			array(
				'Formidable' => '[formidable]',
			)
		);
		return $conflicting_shortcodes;
	}
);
```

It doesn't appear to have any effect. The shortcode seems fine, it just seems to be something about how styles are being loaded.

stephywells:
I think that's fine if we hold off on it since they have a fix coming. Does it solve the problem to add a custom meta description instead of letting it pull from the post content?

PR #740, opened 2022-03-22
Add border box box sizing to admin css buttons

Crabcyborg:
I'm not totally sure when this actually was introduced. It looks like it was still happening when when I was checking out v5.0. I think it may have changed to a button when I updated bootstrap multiselect.

```
.frm-fields button.btn {
     padding: 6px 9px;
}
```

I was looking at a quiz action and noticed the multiselect dropdown is exceeding it's space, overflowing past the action container.

**Before**
Screen Shot 2022-03-22 at 2 20 04 PM
Screen Shot 2022-03-22 at 2 27 44 PM

**After**
Screen Shot 2022-03-22 at 2 36 36 PM

I also dropped a `vertical-align` style because that doesn't work with `display: block`. A warning gets flagged on my IDE about it.
Screen Shot 2022-03-22 at 2 24 46 PM

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -3042,7 +3042,10 @@ a.frm_option_icon:hover::before {
 
 .frm-btn-group.btn-group, .frm-btn-group.btn-group-vertical {
 	display: block;
-	vertical-align: middle;
+}
+
+.frm-btn-group.btn-group button {
+	box-sizing: border-box;
 }
 
 .frm_scale {
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -3004,8 +3004,9 @@ a.frm_option_icon:hover::before {
 	min-width: 40px;
 }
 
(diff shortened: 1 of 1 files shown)


stephywells (commented):
It looks like it's still a bit taller than the text field. How hard is it to get those to match up?

Crabcyborg:
Just went back to this issue noticing it was suddenly gone. I'm realizing I haven't noticed this before because it's another conflict with All in One SEO.

I'll see what I can do about the extra height.

Crabcyborg:
@stephywells updated it!

After a bit more investigation, this come from All in One SEO's multiselect styling:

```
.multiselect {
    box-sizing: content-box;
    min-height: 40px;
}
```

I'm basically just undoing both of those rules, inside of a `.frm-btn-group` so it shouldn't effect All in One SEO.

It looks normal now.
Screen Shot 2022-03-24 at 9 33 01 AM

PR #743, opened 2022-03-24
Maybe match exploded arrays to values

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/1803938901/93581/

Trying to set a single checkbox item by a URL param is not working if the option includes a comma. It seems to be breaking it up.

So it's comparing an exploded array to a string, but they're the same value.
image

Screen Shot 2022-03-24 at 4 25 29 PM

http://docker.for.mac.localhost:8889/wp-admin/admin-ajax.php?action=frm_forms_preview&form=p9qnt&resources[]=Streaming%20Services%20(Remote%20hits,%20Feed%20ingest,%20LiveU,%20Tentpole%20Events)

@stephywells do you think this makes sense? It's a little funny of a solution but it does solve the issue.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1622,12 +1622,23 @@ class FrmAppHelper {
 		}
 	}
 
+	/**
+	 * Check if current value matches.
+	 *
+	 * @param array  $values
+	 * @param string $current
+	 * @return bool
+	 */
 	public static function check_selected( $values, $current ) {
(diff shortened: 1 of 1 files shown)


stephywells (commented):
I'm thinking we can catch it earlier. It looks like the issue is in FrmProFieldsHelper::maybe_force_array

Maybe the part that checks if it matches one of the options isn't working quite right?
in_array( $value, (array) $field->options )

PR #744, opened 2022-03-25
Exclude field options from Akismet comment data

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3482

In the previous version, we get the field IDs of choice fields that don't have `Other` option. In this version, I get the field ID and options of all choice fields, then loop through to check.


Crabcyborg (approved):

PR #746, opened 2022-03-28
Remove Quiz Score field from the field lists

truongwp:


Crabcyborg (approved):
Thanks for doing this @truongwp!

PR #753, opened 2022-04-19
Add hooks for accessibility features

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3507

This update adds two new simple filters that allow you to turn off specific accessibility features (as requested by this user).

```
add_filter( 'frm_focus_first_error', '__return_false' );
add_filter( 'frm_include_alert_role_on_field_errors', '__return_false' );
```

It should work with both JavaScript validation and without. If the role="alert" is included in custom HTML, it will be included anyway but this will prevent any forced insertion.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -2706,17 +2706,19 @@ class FrmAppHelper {
 		$ajax_url = apply_filters( 'frm_ajax_url', $ajax_url );
 
 		$script_strings = array(
-			'ajax_url'     => $ajax_url,
-			'images_url'   => self::plugin_url() . '/images',
-			'loading'      => __( 'Loading&hellip;', 'formidable' ),
-			'remove'       => __( 'Remove', 'formidable' ),
-			'offset'       => apply_filters( 'frm_scroll_offset', 4 ),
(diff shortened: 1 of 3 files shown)


stephywells (approved):
This looks good. But can you tell me why this was requested? I don't understand the reason for not wanting to focus on the first error. Was there a problem?

Crabcyborg:
@stephywells they're actually in favour of the functionality but requested this because they have a custom accessibility implementation and this was conflicting with it.

garretlaxton:
Works good, but when I enable the screen reader I noticed the first field with an error is still being focused. 

Here's a quick demonstration: https://www.loom.com/share/c2edff8403764b6fbdb7d42697842561
This is with JS validation off, it works fine when enabled.

Crabcyborg:
@garretlaxton any chance this may be a caching issue? It seems like I can replicate this for a moment until I clear my cache and then it never happens again.

https://recordit.co/vsFz3FpSqj

Crabcyborg:
Thanks for clearing that up @garretlaxton.

I think that auto-focusing is specific to your screen reader. It looks like my screen reader works differently. This update is not really screen reader dependent, and I think this should be fine as it is. Preventing your screen reader's default behavior might not go well if we tried.

PR #761, opened 2022-05-09
Fix incorrect step validation with float value #760

### LINKED ISSUE #760: validate_step - PHP rounding errors
https://github.com/Strategy11/formidable-forms/blob/5e0364eaccd5360fff482fdda200cfd74632294e/classes/models/fields/FrmFieldNumber.php#L98-L103

PHP rounding errors can cause issues with this.
Could you perhaps round this value to a level of precision that would flag errors, but ignore rounding errors?

Example:
```php
$value = (float) 68.93;
$step = (float) 0.01;
$div = 68.93 / 0.01;

// ( floor( $div ) === $div ) evaluates to false
// number_format((float) $div, 20, '.', '') evaluates to 6893.00000000000090949470
```

So even though `floor($div)` and `$div` show the same value when you print them, the stored value of $div is different

I've added a temporary workaround of this:
```php
$value = (float) $args['value'];
$step  = (float) $step;
$div   = $value / $step;
     ######$div   = round( $div, 5 );
if ( floor( $div ) === $div ) {
	return;
}
```

truongwp:
Fixes #760

### DIFF
--- a/tests/fields/test_FrmFieldNumber.php
+++ b/tests/fields/test_FrmFieldNumber.php
@@ -7,15 +7,15 @@ class test_FrmFieldNumber extends FrmUnitTest {
 	public function test_check_value_is_valid_with_step() {
 		$number_field = new FrmFieldNumber();
 
-		$this->assertTrue( $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 9, 3 ) ) );
-		$this->assertTrue( $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 9.0, 3 ) ) );
-		$this->assertTrue( $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 9, 3.0 ) ) );
-		$this->assertTrue( $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 68.93, 0.01 ) ) );
-		$this->assertTrue( $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( '68.93', '0.01' ) ) );
+		$this->assertEquals( 0, $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 9, 3 ) ) );
+		$this->assertEquals( 0, $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 9.0, 3 ) ) );
+		$this->assertEquals( 0, $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 9, 3.0 ) ) );
+		$this->assertEquals( 0, $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 68.93, 0.01 ) ) );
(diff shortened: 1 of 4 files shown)


tests/fields/test_FrmFieldNumber.php
```diff
@@ -0,0 +1,21 @@
...
+		$this->assertTrue( $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 9.0, 3 ) ) );
+		$this->assertTrue( $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 9, 3.0 ) ) );
+		$this->assertTrue( $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 68.93, 0.01 ) ) );
+		$this->assertTrue( $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( '68.93', '0.01' ) ) );
+
+		$this->assertEquals( 0, $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 1, 3 ) ) );
+		$this->assertEquals( 4, $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 9, 2 ) ) );
+		$this->assertEquals( 3446, $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( 68.93, 0.02 ) ) );
+		$this->assertEquals( 3446, $this->run_private_method( array( $number_field, 'check_value_is_valid_with_step' ), array( '68.93', '0.02' ) ) );
```
Crabcyborg, line 19: @truongwp thanks for working on a fix for this.

I think the update looks good but it's a bit tough to follow. These unit tests cover a specific portion of the logic. I think it's might be nice to add a comment here or something that explains this a bit. It's a little confusing at first glance how `68.93` and `0.02` would ever expect a result of `3446` (though I understand that it's just the result of`68.93 / 0.02`.

Could you add tests for the whole `validate_step` function call as well? I think that might be a bit easier to follow what this function is covering.

truongwp: @Crabcyborg Thanks for checking. I updated `check_value_is_valid_with_step()` method and its unit tests to make it easier to follow.

Crabcyborg (approved):
Thanks @truongwp! Looks much more readable now.

garretlaxton:
I'm not sure if this is expected behavior, but when I have a calculation and add a decimal place there is no step option in the textbox.
Screenshot 2022-05-10 at 13-40-39 Formidable Forms ‹ Formidable Forms — WordPress
Screenshot 2022-05-10 at 13-40-57 Sample Page – Formidable Forms

Crabcyborg:
@garretlaxton Interesting observation. I think that's likely intended as a field with a calculation set probably wouldn't actually use the step value. When you have the calculation on, the decimal places setting seems more important than the step.

I think this is fine for now. I'm not really sure if it makes sense touching that.

garretlaxton:
@Crabcyborg Yeah, this is the only thing I noticed. Just figured it was worth mentioning.

Crabcyborg:
Thanks @garretlaxton! Never hurts to mention anything off. I wasn't aware of this myself but I imagine it could be intentional.

PR #764, opened 2022-05-11
Fix Deadline Calculator template does not work

truongwp:
Related issue: https://github.com/Strategy11/formidable-pro/issues/3513

I want to fix the Default value that is not updated in another PR because this PR is enough to fix the reported issue, and the Default value issue is a bigger different story.

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -1347,6 +1347,10 @@ class FrmFieldsHelper {
 			$replace_with[] = '[' . $new . ']';
 			$replace[]      = '[' . $old . ' ';
 			$replace_with[] = '[' . $new . ' ';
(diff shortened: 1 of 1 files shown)


Crabcyborg (approved):
Thanks @truongwp! I think the default value issue just threw me off and I missed that the issue was about the success message at first 😅.

This works great.

I'm less sure about the default value thing now that I look at it. I looked at the original XML and the short codes appear to be pointing to themselves. I don't really know why.

Screen Shot 2022-05-11 at 10 21 32 AM
Screen Shot 2022-05-11 at 10 21 26 AM

truongwp:
@Crabcyborg Thanks for checking. I also see the shortcode in Default value in that XML is pointless. But in some other cases, it needs to be updated correctly.

PR #765, opened 2022-05-12
Replace field ID in fields after duplicating

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3539

### DIFF
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -144,12 +144,23 @@ class FrmForm {
 	 * @param int $form_id Form ID.
 	 */
 	private static function switch_field_ids_in_fields( $form_id ) {
-		$fields = FrmField::getAll(
+		global $wpdb;
+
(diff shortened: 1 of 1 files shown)


classes/models/FrmForm.php
```diff
@@ -132,6 +132,64 @@ public static function after_duplicate( $form_id, $values ) {
...
+
+	/**
+	 * Switches field ID in fields.
+	 *
+	 * @since 5.2.08
+	 *
+	 * @param int $form_id Form ID.
+	 */
+	private static function switch_field_ids_in_fields( $form_id ) {
+		$fields = FrmField::getAll(
```
Crabcyborg, line 147: @truongwp would it be smaller if we used `FrmField::get_results` and only got the `id, default_value, field_options` columns? It might take the size down a bit at least.

truongwp: Thanks. I updated the code. Please check it.

PR #766, opened 2022-05-12
Add frm_new_form_values filter

truongwp:
This is required for https://github.com/Strategy11/formidable-pro/pull/3543

### New hook:

`frm_new_form_values`: Allows changing form values before creating from the modal.

**Parameters:**
- $values (array): Form values.

**Example:**

Enable `Allow logged-in users to save drafts` by default for new created forms:

```
add_filter( 'frm_new_form_values', function( $values) {
    $values['options']['save_draft'] = 1;
    return $values;
} );
```

### Changed hook:

`frm_form_options_before_update`: The third parameter is added to determine form is updating or creating.

- $update (bool): Is `true` if is updating, `false` if is creating.


Crabcyborg:
@truongwp do you think it would make sense to use this filter instead of adding this new one? The example there on how to set the protect_files option to 1 by default is really similar to your example here. https://formidableforms.com/knowledgebase/frm_pro_default_form_settings/

truongwp:
@Crabcyborg I think that makes sense. Let me update the code using that filter.

truongwp:
@Crabcyborg That doesn't work as we want. That filter will enable Start over for all old forms.

Crabcyborg:
Thanks @truongwp. This is probably fine then. I think the landing page add on might use the other filter - do you think it would make more sense to use this new one here as well?

truongwp:
@Crabcyborg If we need a new setting that is enabled for new forms but disabled for old forms, we should use this new one.

PR #768, opened 2022-05-23
close open custom color parenthesis.

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3483

**Missing**
- [x] Unit tests
- [x] Expected value unit tests
- [x] Make the function private
- [x] ctype_space is not guaranteed to exist.
- [x] Function is public.
- [x] Needs cleaning up and more comments.
- [x] Simplify double not `( ! strpos( $color_val, '#' ) !== 0 )` line.
- [x] Reduce `if ( '' === trim( $value ) || '' === $value ) {` as it's repeated twice.
- [x] Comments to describe where code is handling r, g, b values and alpha.


classes/models/FrmStyle.php
```diff
@@ -88,7 +88,15 @@ public function update( $id = 'default' ) {
 				}
 
 				if ( $this->is_color( $setting ) ) {
-					$new_instance['post_content'][ $setting ] = str_replace( '#', '', $new_instance['post_content'][ $setting ] );
+					$color_val = $new_instance['post_content'][ $setting ];
+					if ( $color_val !== '' && ( ! strpos( $color_val, '#' ) !== 0 ) ) {
+						// if invalid rgba value is entered
+						if ( ! ( preg_match( '/rgba\((\s*\d+\s*,){3}[\d\.]+\)/', $color_val, $matches ) === 1 ||
+								preg_match( '/rgb\((?:\s*\d+\s*,){2}\s*[\d]+\)/', $color_val, $matches ) === 1 ) ) {
```
Crabcyborg, line 95: Any if statement that breaks into multiple lines is probably more readable if it's moved into a function. The fact that it's all wrapped in a `!` is barely noticeable which could lead to bugs down the road. This also seems like nearly the same line of code twice, so it might be nicer written as a loop instead.

classes/models/FrmStyle.php
```diff
@@ -109,6 +117,57 @@ public function update( $id = 'default' ) {
...
+				return;
+			}
+			$length_of_color_codes = strpos( $color_val, '(' );
+			$new_color_values      = array();
+
+			// replace empty values by 0 or 1 (if alpha position).
+			foreach ( explode( ',', $color_rgba ) as $index => $value ) {
+				if ( '' === $value ) {
+					if ( $index === $length_of_color_codes - 1 ) {
+						4 === $length_of_color_codes ? $new_color_values[] = 1 : $new_color_values[] = 0;
```
Crabcyborg, line 142: @AbdiTolesa using ternary operators like this doesn't look very readable. You're adding values to the same array on both sides so this could be reduced down to:

```
$new_color_values[] = 4 === $length_of_color_codes ? 1 : 0;
```

classes/models/FrmStyle.php
```diff
@@ -109,6 +117,57 @@ public function update( $id = 'default' ) {
...
+					} else {
+						$new_color_values[] = 0;
+					}
+				} else {
+					$new_color_values[] = $value;
```
Crabcyborg, line 147: Instead of burying this case way at the bottom I think it might be nicer to just do an early continue.

Something like:

```
if ( '' !== $value ) {
    $new_color_values[] = $value;
    continue;
}
```

It breaks down the logic more. This is really separated from the if condition that defines it.

classes/models/FrmStyle.php
```diff
@@ -109,6 +117,57 @@ public function update( $id = 'default' ) {
...
+			}
+
+			// add more 0s and 1 (if alpha position) if required.
+			$missing_values = $length_of_color_codes - count( $new_color_values );
+			if ( $missing_values > 1 ) {
+				$insert_values = array_fill( 0, $missing_values - 1, 0 );
+				array_push( $insert_values, 1 );
+			} elseif ( $missing_values === 1 ) {
+				$insert_values = array( 1 );
+			}
```
Crabcyborg, line 158: It does look like `$insert_values` would be undefined in the unhandled else condition here. We want to avoid it breaking the next line.

classes/models/FrmStyle.php
```diff
@@ -109,6 +114,66 @@ public function update( $id = 'default' ) {
 		return $action_ids;
 	}
 
+	/**
+	 * Sanitize custom color values and convert it to valid one filling missing values.
+	 *
+	 * @param string $color_val, The color value, by reference.
+	 */
+	public function maybe_sanitize_rgba_value( &$color_val ) {
+		$patters = array( '/rgba\((\s*\d+\s*,){3}[\d\.]+\)/', '/rgb\((\s*\d+\s*,){2}\s*[\d]+\)/' );
```
Crabcyborg, line 123: It looks like there is a typo here. `$patterns` with the n would be a better variable name but I imagine that's just missing by mistake.

classes/models/FrmStyle.php
```diff
@@ -109,6 +114,66 @@ public function update( $id = 'default' ) {
...
+			$length_of_color_codes = strpos( $color_val, '(' );
+			$new_color_values      = array();
+
+			// replace empty values by 0 or 1 (if alpha position).
+			foreach ( explode( ',', $color_rgba ) as $index => $value ) {
+				if ( '' !== $value ) {
+					$new_color_values[] = $value;
+					continue;
+				}
+				if ( '' === $value ) {
```
Crabcyborg, line 148: There isn't any need for this check because the check above is continuing if it isn't `''`. It's guaranteed that it is an empty string at this point already without the check.

tests/styles/test_FrmStyle.php
```diff
@@ -0,0 +1,25 @@
...
+ * @group styles
+ */
+class test_FrmStyle extends FrmUnitTest {
+	/**
+	 * @covers maybe_sanitize_rgba_value
+	 */
+	public function test_maybe_sanitize_rgba_value() {
+		$frm_style            = new FrmStyle();
+		$invalid_color_values = array( 'rgba(45, 45, 45,', 'rgba(,,,1)', 'rgba(45, 45', 'rgba(45,,,0.5)', 'rgb(45,,45)', 'rgb(,,)' );
+		$patters              = array( '/rgba\((\s*\d+\s*,){3}[\d\.]+\)/', '/rgb\((\s*\d+\s*,){2}\s*[\d]+\)/' );
```
Crabcyborg, line 13: It looks like `$patterns` is missing an n here as well.

tests/styles/test_FrmStyle.php
```diff
@@ -0,0 +1,25 @@
...
+
+		foreach ( $invalid_color_values as $color_val ) {
+			$frm_style->maybe_sanitize_rgba_value( $color_val );
+
+			if ( strpos( $color_val, '(' ) === 4 ) {
+				$this->assertEquals( 1, preg_match( $patters[0], $color_val ) );
+			} elseif ( strpos( $color_val, '(' ) === 3 ) {
+				$this->assertEquals( 1, preg_match( $patters[1], $color_val ) );
```
Crabcyborg, line 21: @AbdiTolesa could we add more to these asserts? It would be nice to have `$invalid_color_values` mapped to the expected sanitized value.

Something like:
```
$invalid_color_values = array(
    'rgba(45, 45, 45,' => 'rgba(45, 45, 45, 1)',
    ...
);
```

The regex is an alright check but it's less easy to read.

Crabcyborg (commented):
Thanks for updating this @AbdiTolesa I have a few comments. Mostly smaller things. The bigger thing is that I'd like to see a few more tests.

classes/models/FrmStyle.php
```diff
@@ -109,6 +114,66 @@ public function update( $id = 'default' ) {
 		return $action_ids;
 	}
 
+	/**
+	 * Sanitize custom color values and convert it to valid one filling missing values.
+	 *
+	 * @param string $color_val, The color value, by reference.
+	 */
+	public function maybe_sanitize_rgba_value( &$color_val ) {
```
Crabcyborg, line 122: @AbdiTolesa similar to what I mentioned in the section update (https://github.com/Strategy11/formidable-pro/pull/3575#discussion_r883691708), it would be great if we could make this function private. Could you change that here too?

AbdiTolesa: @Crabcyborg Sure thing

tests/styles/test_FrmStyle.php
AbdiTolesa: @Crabcyborg I have covered this as well.

tests/styles/test_FrmStyle.php
AbdiTolesa: @Crabcyborg Thanks for catching this, it is corrected now.

classes/models/FrmStyle.php
AbdiTolesa: @Crabcyborg That's true, I have removed that already in the latest code pushed.

classes/models/FrmStyle.php
```diff
@@ -109,6 +114,66 @@ public function update( $id = 'default' ) {
...
+		}
+
+		if ( 'rgb' === substr( $color_val, 0, 3 ) ) {
+			if ( substr( $color_val, -1 ) !== ')' ) {
+				$color_val = $color_val .= ')';
+			}
+			if ( preg_match( '/\((.*?)\)/', $color_val, $match ) === 1 ) {
+				$color_rgba = $match[1];
+			} else {
+				return;
```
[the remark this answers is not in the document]
AbdiTolesa: @Crabcyborg This makes sense and done.

classes/models/FrmStyle.php
```diff
@@ -88,7 +88,12 @@ public function update( $id = 'default' ) {
 				}
 
 				if ( $this->is_color( $setting ) ) {
-					$new_instance['post_content'][ $setting ] = str_replace( '#', '', $new_instance['post_content'][ $setting ] );
+					$color_val = $new_instance['post_content'][ $setting ];
+					if ( $color_val !== '' && ( ! strpos( $color_val, '#' ) !== 0 ) ) {
```
Crabcyborg, line 92: @AbdiTolesa this check is a double not. It's a lot simpler to take

```
( ! strpos( $color_val, '#' ) !== 0 )
```

down to

```
strpos( $color_val, '#' ) === 0
```

Crabcyborg, line 92: The last comment I left might be wrong so I deleted it. This line is confusing either way. I think we want to change it to `strpos( $col_val, '#' ) === false`. We want to avoid the double not.

classes/models/FrmStyle.php
Crabcyborg: It looks like you updated the unit test but the function itself is still public.

classes/models/FrmStyle.php
```diff
@@ -109,6 +114,75 @@ public function update( $id = 'default' ) {
...
+			}
+			$color_rgba = $match[1];
+
+			$length_of_color_codes = strpos( $color_val, '(' );
+			$new_color_values      = array();
+
+			// replace empty values by 0 or 1 (if alpha position).
+			foreach ( explode( ',', $color_rgba ) as $index => $value ) {
+				$new_value = null;
+				if ( ctype_space( $value ) || '' === $value ) {
```
Crabcyborg, line 145: `ctype_space` is not guaranteed to actually exist as it requires the ctype extension.

Symfony has polyfills for ctype though https://github.com/symfony/polyfill-ctype/blob/main/Ctype.php

```

    /**
     * Returns TRUE if every character in text creates some sort of white space, FALSE otherwise. Besides the blank character this also includes tab, vertical tab, line feed, carriage return and form feed characters.
     *
     * @see https://php.net/ctype-space
     *
     * @param mixed $text
     *
     * @return bool
     */
    public static function ctype_space($text)
    {
        $text = self::convert_int_to_char_for_ctype($text, __FUNCTION__);

        return \is_string($text) && '' !== $text && !preg_match('/[^\s]/', $text);
    }
```

I did something similar with ctype_xdigit https://github.com/Strategy11/formidable-forms/blob/bd934e3bb3d1e05db290f59f224d6de7219d8f59/classes/helpers/FrmAppHelper.php#L3372

But maybe there's an easier check that doesn't require ctype functions so we can avoid the polyfill.

classes/models/FrmStyle.php
```diff
@@ -109,6 +114,75 @@ public function update( $id = 'default' ) {
...
+	 */
+	public function maybe_sanitize_rgba_value( &$color_val ) {
+		$patterns = array( '/rgba\((\s*\d+\s*,){3}[[0-1]\.]+\)/', '/rgb\((\s*\d+\s*,){2}\s*[\d]+\)/' );
+		foreach ( $patterns as $pattern ) {
+			if ( preg_match( $pattern, $color_val ) === 1 ) {
+				return;
+			}
+		}
+
+		if ( 'rgb' === substr( $color_val, 0, 3 ) ) {
```
Crabcyborg, line 130: @AbdiTolesa it looks like everything is wrapped in this but this should really be an early return.

Crabcyborg: I'm thinking this check should be before the other pattern checks. It's a quick general "does this start with rgb"?

classes/models/FrmStyle.php
```diff
@@ -109,6 +114,75 @@ public function update( $id = 'default' ) {
...
+			if ( substr( $color_val, -1 ) !== ')' ) {
+				$color_val = $color_val .= ')';
+			}
+			if ( preg_match( '/\((.*?)\)/', $color_val, $match ) !== 1 ) {
```
Crabcyborg, line 134: It would be nice to add more comments here. What exactly is this regex checking?

tests/styles/test_FrmStyle.php
Crabcyborg: Thanks for adding those @AbdiTolesa the tests look good. What do you think about adding some valid color value tests too? It's nice to confirm that nothing is changing as well.

classes/models/FrmStyle.php
AbdiTolesa: @Crabcyborg I have replaced it with a simpler solution.

tests/styles/test_FrmStyle.php
AbdiTolesa: @Crabcyborg this is a good idea and I have done that.

classes/models/FrmStyle.php
```diff
@@ -110,18 +116,72 @@ public function update( $id = 'default' ) {
...
-	 * Unslash everything in post_content but custom_css
+	 * Sanitize custom color values and convert it to valid one filling missing values.
 	 *
-	 * @since 5.0.13
-	 *
-	 * @param array $settings
-	 * @return array
+	 * @param string $color_val, The color value, by reference.
 	 */
-	private function unslash_post_content( $settings ) {
```
Crabcyborg, line 120: @AbdiTolesa what happened here to unslash post content? It looks like shouldn't have been removed.

AbdiTolesa: @Crabcyborg I think I deleted it by accident. It is restored now.

classes/models/FrmStyle.php
```diff
@@ -109,6 +115,75 @@ public function update( $id = 'default' ) {
...
+		}
+
+		$color_rgba            = substr( $color_val, strpos( $color_val, '(' ) + 1, strlen( $color_val ) - strpos( $color_val, '(' ) - 2 );
+		$length_of_color_codes = strpos( $color_val, '(' );
+		$new_color_values      = array();
+
+		// replace empty values by 0 or 1 (if alpha position).
+		foreach ( explode( ',', $color_rgba ) as $index => $value ) {
+			$new_value = null;
+			if ( '' === trim( $value ) || '' === $value ) {
```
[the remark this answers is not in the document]
AbdiTolesa: @Crabcyborg Yes that's right, the $new_value = 0 is only needed in the first if block. I have reviewed the code again, thanks.

Crabcyborg:
@garretlaxton I'm planning on making a core release today if possible.

Right now there's nothing new in the lite plugin. I think this would be a good update to include. If you could help test it out shortly that would be great.

Before this update, it's possible to save a weird custom style color value like `rgba(1,2` and when it's broken like that it can really break all of the styling because it expects the closing bracket that never gets added. This update checks for those weird gaps and changes that color for example into `rgba(1,2,0,1)` instead.

garretlaxton:
@Crabcyborg This looks great! Anytime the bracket is left open it is automatically closed on update and adds a `0` to any missing decimal.

PR #769, opened 2022-05-24
Css tweaks

stephywells:
- Use the same border on "no fields" section as views (grey instead of blue)
- Remove the form description field from the new form modal
- Move css for modal from Views
- remove the header border on the modal
- Vertically position text correctly inside buttons in headings
- Vertically position carrot correctly in template accordion


Crabcyborg (commented):
@stephywells it seems unrelated to this update but it seems that I can't create a new form anymore from the modal. I noticed it testing this. Any idea why this error would come up? It seems that `get_form_actions` is returning an array for me which isn't expected on the other end.

> [24-May-2022 18:47:54 UTC] PHP Fatal error:  Uncaught Error: Call to a member function create() on array in /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php:96
> Stack trace:
> #0 /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php(582): FrmFormsController::create_default_email_action(Object(stdClass))
> #1 /var/www/src/wp-includes/class-wp-hook.php(307): FrmFormsController::build_new_form('')
> #2 /var/www/src/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters(NULL, Array)
> #3 /var/www/src/wp-includes/plugin.php(476): WP_Hook->do_action(Array)
> #4 /var/www/src/wp-admin/admin-ajax.php(187): do_action('wp_ajax_frm_ins...')
> #5 {main}
>   thrown in /var/www/src/wp-content/plugins/formidable/classes/controllers/FrmFormsController.php on line 96

stephywells:
Thanks @Crabcyborg. Did you clear your js cache? Removing the description was causing a js error for me too. But it should be covered here.

stephywells:
actually... that's not the error I was seeing. I'm taking another look now.

Crabcyborg:
I think I caught the issue. It looks like I introduced this trying to work on the API embed update. It looks like adding an action for `frm_registered_form_actions` can break the whole form submit process.

PR #771, opened 2022-05-26
Remove unexpected add new button on import page

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3583

This came up when I refactored the add new button code. I missed that the import button calls this code without any of the expected attributes.

My solution is to just check if none of the attributes are set. If all of those are missing it expects the function to do nothing.

The alternative is trying to do something like.

```
if ( ! $atts['import_link'] ) {
    FrmAppHelper::add_new_item_link( $atts );
}
```

in the admin-header.php, or to check against `$atts['import_link']` specifically in this function but that introduces another option that isn't really related to the function.

But that would only catch the import button. This may effect other calls to `add_new_item_link`. As it's a public function I wouldn't be surprised if someone was using it in a custom flow so fixing it in-function probably makes more sense.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1102,6 +1102,11 @@ class FrmAppHelper {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Nice! Thank you!

PR #773, opened 2022-05-26
Move unlocked application templates to the front

Crabcyborg:
Following up from a comment the other day.

> How complicated would it be to move allowed applications to the front of the list, and alphabetic each group separately?

Screen Shot 2022-05-26 at 2 45 42 PM


stephywells (approved):
Awesome!! Looks great

PR #774, opened 2022-05-27
Make applications more responsive, and fix position of upgrade button…

Crabcyborg:
…, and rename formidable applications category

I played with the widths a bit and added a 3 column view.

This update also renames the category and fixes the upgrade button position. I don't really know when it broke but I just made an issue for it today https://github.com/Strategy11/formidable-pro/issues/3589

Screen Shot 2022-05-27 at 10 58 08 AM

IVz4O7YJg6

There's also a small fix in here for phpstan.

Error: Result of static method FrmFormsController::list_form() (void) is used.
 ------ ------------------------------------------------------------------- 
  Line   classes/controllers/FrmFormsController.php                         
 ------ ------------------------------------------------------------------- 
  1609   Result of static method FrmFormsController::list_form() (void) is  
         used.                                                              
 ------ ------------------------------------------------------------------- 

I might have just missed this when I added a comment. Easy fix though to separate the two lines so it's easier to tell that the function isn't actually returning anything to pass forward.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1606,7 +1606,8 @@ class FrmFormsController {
 				if ( strpos( $action, 'bulk_' ) === 0 ) {
 					FrmAppHelper::remove_get_action();
 
-					return self::list_form();
+					self::list_form();
+					return;
 				}
 
 				self::display_forms_list();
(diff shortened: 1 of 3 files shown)


stephywells (approved):
Nice job!

PR #777, opened 2022-05-27
Add filter for gmt offset to avoid fatal error in php8

Crabcyborg:
Fixes https://secure.helpscout.net/conversation/1900124057/100035/

It happens when a call to `current_time( 'timestamp' )` is made if the `gmt_offset` WordPress setting is an empty string.

It's failing on this WordPress code.

```
time() + (int) ( get_option( 'gmt_offset' ) * HOUR_IN_SECONDS );
```

Technically this is more or a WordPress bug and we could probably remove this in the future. But it's causing fatal errors as soon as you load the admin pages if gmt_offset is set to an empty string. It's possible to fix on our end with this filter, and that probably makes sense while the issue exists in WordPress.

I made it public because it looks like pro uses `current_time( 'timestamp' )` as well so I think I may add an update there as well.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -4,9 +4,14 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 class FrmAppHelper {
-	public static $db_version = 98; // Version of the database we are moving to.
+	public static $db_version     = 98; // Version of the database we are moving to.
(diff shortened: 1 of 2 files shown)


stephywells (approved):
Looks like a good approach. Do we need any extra doc comment to mention that it's a WP bug?

Crabcyborg:
Thanks @stephywells good point, I added more comments!

I also fixed the return line. I had the type wrong.

PR #779, opened 2022-05-30
fix view activation info are being cut.

AbdiTolesa:
[issue #3552](https://github.com/Strategy11/formidable-pro/issues/3552)

### DIFF
--- a/classes/models/fields/FrmFieldType.php
+++ b/classes/models/fields/FrmFieldType.php
@@ -1056,16 +1056,20 @@ DEFAULT_HTML;
 	 * @since 3.0
 	 */
 	protected function add_aria_description( $args, &$input_html ) {
-		$describedby = '';
-		if ( $this->get_field_column( 'description' ) != '' ) {
-			$describedby = 'frm_desc_' . $args['html_id'];
+		$aria_describedby_exists = preg_match( '/aria-describedby=\"[^\"]*\"/', $input_html, $match ) === 1;
+		$describedby             = isset( $match[0] ) ? substr( $match[0], 0, -1 ) : '';
+
+		if ( $this->get_field_column( 'description' ) !== '' ) {
+			$describedby .= ' frm_desc_' . $args['html_id'];
 		}
 
 		if ( isset( $args['errors'][ 'field' . $args['field_id'] ] ) ) {
 			$describedby .= ' frm_error_' . $args['html_id'];
 		}
 
-		if ( ! empty( $describedby ) ) {
+		if ( $aria_describedby_exists ) {
+			$input_html = preg_replace( '/aria-describedby=\"[^\"]*\"/', $describedby . '"', $input_html );
+		} elseif ( ! empty( $describedby ) ) {
 			$input_html .= ' aria-describedby="' . esc_attr( trim( $describedby ) ) . '"';
 		}
 	}


css/frm_admin.css
```diff
@@ -500,6 +500,7 @@ ul.frm_form_nav > li:last-of-type {
 	flex-direction: column;
 	overflow: hidden;
 	padding-top: 0;
+    overflow: visible;
```
Crabcyborg, line 503: @AbdiTolesa it looks like you're indenting with spaces here? You can see how it isn't aligning properly with the other properties in Git? It looks like there's some other space indenting in this file, but the majority of it is tabs. We should stick to tabs.

AbdiTolesa: @Crabcyborg Sorry I forgot to use tabs here. It is corrected now.

Crabcyborg (commented):
@AbdiTolesa 

Thanks for working on this. I have a couple notes.

1. I think we want to move this. `.frm_wrap .frm_page_container` is used on a lot of pages, definitely not just this one. The other pages handle the scrolling within the inner containers instead, which explains this rule. I think we want to keep the hidden overflow there. Could you remove this rule and try to add the scroll to the child element instead? This is in the views-info.php file. For now you can probably even just use inline styles as this page isn't really critical.

2. It looks like there's already an overflow rule here. We wouldn't want to have `overflow: hidden` only 2 lines above this line in the same definition.

3. That second part doesn't really matter. Neither did the space thing, just trying to point that out. It's great to look at your update in GitHub and do your own review first to see if anything looks off. Also, I'm not sure when you are actually ready for a review or not. Could you start requesting my review when something is ready? And link the issue? If you type it out as "Fixes #..." it will automatically sync it most of the time.

But thanks for looking into this. This points out what the weird part is. It looks like `frm_page_container` generally explains the page to have self contained scrolling.

AbdiTolesa:
@Crabcyborg Ok I will request a review once I'm done with my updates going forward. I overlooked the existing style thinking this just needs a quick fix as my mind was caught up trying to improve some logic in the other issue, I know that sounds weird. I wasn't sure that this is the best way to do this but was expecting your feedback to know the caveats of overriding the common classes styles. I'm working to do it the recommended way now.

AbdiTolesa:
> Thanks @AbdiTolesa understandable. I didn't really know what was causing it either, so the commit you pushed helped explain the issue a bit more.
> 
> I just think we need to take this a step further and avoid breaking anything.

@Crabcyborg Agreed 100%. I was trying to add the issue under development tab but this issue is from the pro repo so it cannot be found here.

classes/views/shared/views-info.php
```diff
@@ -4,7 +4,7 @@
 }
 ?>
 <div class="frm_wrap">
-	<div class="frm_page_container">
+	<div class="frm_page_container" style="overflow: scroll;">
```
Crabcyborg, line 7: @AbdiTolesa I'm thinking it might be nicer if the top bar stays on scroll like it does on other page container pages?

I'm also seeing a horizontal scrollbar which we don't want. I think we might just want to add a rule for `overflow-y`.

And using `scroll` means that it will always show a scroll bar even when there isn't a need for one. Generally it's good to prefer `auto`.

AbdiTolesa: @Crabcyborg Understood and changes are made.

Crabcyborg: @AbdiTolesa I'm still thinking this should be in a child or something rather than the frm_page_container itself. The other comments weren't really as important as the first line.

You can see on another page that the top bar stays and the container content scrolls.

0qNwM6PtWU

On your page you're scrolling the whole container. I think we need a child that fits the parent that holds the content that has the overflow styling.
0uNxWUm0CY

Crabcyborg:
@AbdiTolesa I'm just going to close this pull request. It looks like the other update got pulled in here. There isn't much in here, so starting from scratch makes more sense at this point.

If you haven't caught my last comment yet, it's still applicable. I'm hoping we can move this scrolling into a child div so the top bar stays as you scroll like the other pages. https://github.com/Strategy11/formidable-forms/pull/779#discussion_r885023312

PR #780, opened 2022-05-31
fix view activation info are being cut.

AbdiTolesa:
issue [#3552](https://github.com/Strategy11/formidable-pro/issues/3552)


Crabcyborg:
Thank you @AbdiTolesa this is looking closer.

One thing I'm seeing is that you have a conflict with this view. You should try to pull the latest version of master on a regular basis so it's up to date to avoid that from happening. That update isn't brand new so it could have been easily avoided.

The video is now pointing to https://www.youtube.com/embed/pmYbQ79wonQ.

The update is from 16 days ago so I imagine you might not ever be pulling updates from origin.
https://github.com/Strategy11/formidable-forms/commit/2d6c6df41857c6e0293c36205258c1088c47ed7a

I'm also seeing that when the scroll bar is fully scrolled, that their isn't any padding on the bottom of the container. I think we want a little bit of white space there so it isn't right against the bottom.

Screen Shot 2022-06-01 at 9 27 30 AM

Crabcyborg:
@AbdiTolesa sorry I don't follow. That shouldn't have changed. I think the only thing that changed is the link to the video.

AbdiTolesa:
> @AbdiTolesa sorry I don't follow. That shouldn't have changed. I think the only thing that changed is the link to the video.
> 
> If you're just asking about how to resolve the conflict, that's up to you. It might be easier to just start over again or you can just take your changes and use the new embed link. I think that's all that changed.

@Crabcyborg sorry for confusing you, I think I sort of accidentally clicked on accept both local and remote changes in my VS Code while trying to resolve the conflict. I have added a padding now as well.

PR #784, opened 2022-06-07
redirect page to form edit page after form is duplicated.

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3161

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1629,7 +1629,7 @@ class FrmFormsController {
 
 				$message = FrmAppHelper::get_param( 'message' );
 				if ( 'form_duplicate_error' === $message ) {
-					self::display_forms_list( array(), __( 'There was a problem duplicating the form', 'formidable' ) );
+					self::display_forms_list( array(), '', array( __( 'There was a problem duplicating the form', 'formidable' ) ) );
 					return;
 				}
 
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -235,7 +235,7 @@ class FrmFormsController {
 
 		$params  = FrmForm::list_page_params();
 		$form    = FrmForm::duplicate( $params['id'], $params['template'], true );
-		$message = 'form_duplicated';
+		$message = 'form_duplicate_error';
 		$nonce   = FrmAppHelper::simple_get( '_wpnonce' );
 
 		$frm_settings = FrmAppHelper::get_settings();
@@ -244,13 +244,15 @@ class FrmFormsController {
 		}
 
 		if ( $form ) {
+			$message = 'form_duplicated';
 			$url = admin_url( 'admin.php?page=formidable&frm_action=edit&id=' . absint( $form ) );
-			$url .= '&message=' . $message;
-
-			wp_safe_redirect( $url );
-			exit();
+		} else {
+			$url = admin_url( 'admin.php?page=formidable' );
 		}
-		self::display_forms_list( $params, __( 'There was a problem creating the new template.', 'formidable' ) );
+		$url .= '&message=' . $message;
+
+		wp_safe_redirect( $url );
+		exit();
 	}
 
 	/**
@@ -1625,7 +1627,14 @@ class FrmFormsController {
 					return;
 				}
 
+				$message = FrmAppHelper::get_param( 'message' );
+				if ( 'form_duplicate_error' === $message ) {
+					self::display_forms_list( array(), __( 'There was a problem duplicating the form', 'formidable' ) );
+					return;
+				}
+
 				self::display_forms_list();
+
 				return;
 		}
 	}
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -235,7 +235,6 @@ class FrmFormsController {
 
 		$params  = FrmForm::list_page_params();
 		$form    = FrmForm::duplicate( $params['id'], $params['template'], true );
-		$message = 'form_duplicate_error';
 		$nonce   = FrmAppHelper::simple_get( '_wpnonce' );
 
 		$frm_settings = FrmAppHelper::get_settings();
@@ -243,12 +242,14 @@ class FrmFormsController {
 			wp_die( esc_html( $frm_settings->admin_permission ) );
 		}
 
+		$url = admin_url( 'admin.php?page=formidable' );
+		$message = 'form_duplicate_error';
+
 		if ( $form ) {
-			$message = 'form_duplicated';
 			$url = admin_url( 'admin.php?page=formidable&frm_action=edit&id=' . absint( $form ) );
-		} else {
-			$url = admin_url( 'admin.php?page=formidable' );
+			$message = 'form_duplicated';
 		}
+
 		$url .= '&message=' . $message;
 
 		wp_safe_redirect( $url );
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -628,7 +628,7 @@ class FrmFormsController {
 			}
 
 			$response = array(
-				'redirect' => admin_url( 'admin.php?page=formidable&frm_action=duplicate&id=' . $new_form_id ),
+				'redirect' => admin_url( 'admin.php?page=formidable&frm_action=duplicate&id=' . $new_form_id ) . '&_wpnonce=' . wp_create_nonce(),
 			);
 		}
 
@@ -991,7 +991,7 @@ class FrmFormsController {
 				'name'        => $template->name,
 				'key'         => $template->form_key,
 				'description' => $template->description,
-				'url'         => admin_url( 'admin.php?page=formidable&frm_action=duplicate&id=' . absint( $template->id ) ),
+				'url'         => wp_nonce_url( admin_url( 'admin.php?page=formidable&frm_action=duplicate&id=' . absint( $template->id ) ) ),
 				'released'    => $template->created_at,
 				'installed'   => 1,
 			);
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -232,17 +232,16 @@ class FrmFormsController {
 
 	public static function duplicate() {
 		FrmAppHelper::permission_check( 'frm_edit_forms' );
+		$nonce = FrmAppHelper::simple_get( '_wpnonce' );
 
-		$params  = FrmForm::list_page_params();
-		$form    = FrmForm::duplicate( $params['id'], $params['template'], true );
-		$nonce   = FrmAppHelper::simple_get( '_wpnonce' );
-
-		$frm_settings = FrmAppHelper::get_settings();
 		if ( ! wp_verify_nonce( $nonce ) ) {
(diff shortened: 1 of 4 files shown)


classes/views/shared/errors.php
```diff
@@ -3,6 +3,10 @@
 	die( 'You are not allowed to call this page directly.' );
 }
 
+if ( '' !== FrmAppHelper::get_param( 'message', '', 'get' ) ) {
```
Crabcyborg, line 6: @AbdiTolesa this view gets used for a lot of error messages. This check could too easily get in the way and cause issues. I don't think we should really be adding anything to this file.

Crabcyborg, line 6: @AbdiTolesa this view gets used for a lot of error messages. This check could too easily get in the way and cause issues. I don't think we should really be adding anything to this file.

classes/controllers/FrmFormsController.php
```diff
@@ -238,9 +238,14 @@ public static function duplicate() {
...
-		}
+			$url = admin_url( 'admin.php?page=formidable&frm_action=edit&id=' . absint( $form ) );
+
+			if ( $message ) {
+				$url .= '&message=' . $message;
+			}
 
+			wp_safe_redirect( $url );
+		}
 		self::display_forms_list( $params, __( 'There was a problem creating the new template.', 'formidable' ) );
```
Crabcyborg, line 249: @AbdiTolesa will there be issues with trying to duplicate a form that does not exist with the new position in the code?

Crabcyborg, line 249: @AbdiTolesa will there be issues with trying to duplicate a form that does not exist with the new position in the code?

classes/controllers/FrmFormsController.php
```diff
@@ -238,9 +238,14 @@ public static function duplicate() {
 		$message = $params['template'] ? __( 'Form template was Successfully Created', 'formidable' ) : __( 'Form was Successfully Copied', 'formidable' );
 
 		if ( $form ) {
-			return self::get_edit_vars( $form, array(), $message, true );
-		}
+			$url = admin_url( 'admin.php?page=formidable&frm_action=edit&id=' . absint( $form ) );
+
+			if ( $message ) {
+				$url .= '&message=' . $message;
```
Crabcyborg, line 244: @AbdiTolesa I'm thinking if the message is in the URL that it should probably be added as a key instead like "form_copied" and then the message would be determined server side on load instead.

Crabcyborg: > @Crabcyborg hm, can we probably do that without adding anything to `errors.php` as you mentioned above? I think it is the template used for rendering any messages before the form body.

@AbdiTolesa it's good to follow the code back a bit. If you just search the code for shared/errors.php you can see it's getting called from all over the place, not just this one one page.

Screen Shot 2022-06-09 at 12 24 55 PM

In this case we're only looking at what looks like `views/frm-forms/form.php`.

It looks like you could use the `frm_message_list` filter to add a message from basically anywhere. We just want the code to be close to the form builder code so it's not getting checked for when it isn't necessary either.

AbdiTolesa: @Crabcyborg I think `frm_message_list` is only used for the messages to be displayed in the wrapper with `frm_warning_style` but we actually want it to be inside the `frm_updated_message` wrapper but I could anyway create similar filter for the $message variable but that wouldn't do it as the page is redirected after the filter modifying the message is added from the controller.

Crabcyborg: Thanks @AbdiTolesa thats' a fair point.

Looking a bit higher up in the code I see in `FrmFormsController::get_edit_vars` that `$message` is getting set here if you're editing a form that is a template. We should put our code around here so it's in the same place. Moving all of this into a new function would probably be a good idea.

Screen Shot 2022-06-09 at 4 46 52 PM

AbdiTolesa: @Crabcyborg I think the redirection still would be a problem here too as it should happen only after the form is duplicated successfully. The `$message` variable couldn't be passed down to our template when that happens.

Crabcyborg: @AbdiTolesa yeah, but this is a much better place to check for a parameter than inside of the shared errors template that would effect everywhere.

We already define `$message` here so it's a good place to continue from. I'm picturing something along the lines of a new function like `$message = self::maybe_update_form_builder_message( $form, $message );`. It could handle this template message, and then also support a get parameter. Something like `if ( 'form_duplicated' === FrmAppHelper::simple_get( 'frm_edit_vars_message' ) ) { $message = ... }` so we can avoid having to pass the full message in the URL.

classes/controllers/FrmFormsController.php
Crabcyborg: > @AbdiTolesa will there be issues with trying to duplicate a form that does not exist with the new position in the code?

@AbdiTolesa I just tried this out and it does indeed look very broken. If I visit `/wp-admin/admin.php?page=formidable&frm_action=duplicate&id=0` for example.

The page loaded is not really a functioning page.
Screen Shot 2022-06-14 at 11 56 21 AM

classes/controllers/FrmFormsController.php
```diff
@@ -235,15 +235,24 @@ public static function duplicate() {
 
 		$params  = FrmForm::list_page_params();
 		$form    = FrmForm::duplicate( $params['id'], $params['template'], true );
-		$message = $params['template'] ? __( 'Form template was Successfully Created', 'formidable' ) : __( 'Form was Successfully Copied', 'formidable' );
```
Crabcyborg, line 238: As far as I can tell this `Form template was Successfully Created` text is never getting used at the moment. It looks like it was based off of the page URL (templates used to have a separate page).

As far as I can tell all new templates just use the `Form was successfully copied` message at the moment.

classes/controllers/FrmFormsController.php
```diff
@@ -235,15 +235,24 @@ public static function duplicate() {
...
-		}
+			$url = admin_url( 'admin.php?page=formidable&frm_action=edit&id=' . absint( $form ) );
+			$url .= '&message=' . $message;
 
+			wp_safe_redirect( $url );
+		}
 		self::display_forms_list( $params, __( 'There was a problem creating the new template.', 'formidable' ) );
 	}
 
+	public static function add_message_to_message_list( $show_messages ) {
```
Crabcyborg, line 249: Is this something you added that you don't need anymore? It looks unused now.

classes/controllers/FrmFormsController.php
AbdiTolesa: @Crabcyborg I think the best way to handle this is redirecting back to the form list page with a form duplicate error message in such scenario. I was trying to handle this the way other actions do, showing the error message and form list in the same page (without redirection) but that seems to require many function changes as the duplicate action is hooked to the `admin_init` unlike the other actions and nothing has loaded yet, which breaks the form list calls.

classes/controllers/FrmFormsController.php
```diff
@@ -235,13 +235,25 @@ public static function duplicate() {
...
-		$message = $params['template'] ? __( 'Form template was Successfully Created', 'formidable' ) : __( 'Form was Successfully Copied', 'formidable' );
+		$nonce   = FrmAppHelper::simple_get( '_wpnonce' );
+
+		$frm_settings = FrmAppHelper::get_settings();
+		if ( ! wp_verify_nonce( $nonce ) ) {
+			wp_die( esc_html( $frm_settings->admin_permission ) );
+		}
+
+		$url = admin_url( 'admin.php?page=formidable' );
+		$message = 'form_duplicate_error';
```
Crabcyborg, line 246: @AbdiTolesa what do you think about getting the error to show as an error? Right now seeing a green message for this is a bit weird.

Screen Shot 2022-07-04 at 12 19 59 PM

Crabcyborg:
@AbdiTolesa thanks for updating this! It's looking a lot better.

We need to make sure we're capturing all of our duplicate URLs and including the nonce of them so we're not breaking anything. Right now I've found that it's breaking where it duplicates the form template.

Check out the new form modal. There's a "My Templates" section. You can select any of your forms and turn them into a template here.

Screen Shot 2022-07-04 at 12 22 06 PM

When you click that plus it triggers a duplicate action without a nonce.

As a result it fails to work now.
Screen Shot 2022-07-04 at 12 22 10 PM

classes/controllers/FrmFormsController.php
AbdiTolesa: @Crabcyborg Sorry I just looked into `errors.php` closely and noticed the different types to display message, warning and errors. This is updated accordingly now.

AbdiTolesa:
> @AbdiTolesa thanks for updating this! It's looking a lot better.
> 
> We need to make sure we're capturing all of our duplicate URLs and including the nonce of them so we're not breaking anything. Right now I've found that it's breaking where it duplicates the form template.
> 
> Check out the new form modal. There's a "My Templates" section. You can select any of your forms and turn them into a template here.
> 
> Screen Shot 2022-07-04 at 12 22 06 PM
> 
> When you click that plus it triggers a duplicate action without a nonce.
> 
> As a result it fails to work now. Screen Shot 2022-07-04 at 12 22 10 PM

@Crabcyborg I have added the nonce to templates as well now and the error is gone.

Crabcyborg:
@AbdiTolesa are you sure if there are not any others?

It looks like it comes up a bit in the code.
https://github.com/Strategy11/formidable-forms/search?q=action%3Dduplicate

I see a link in FrmFormsHelper but that file looks untouched in this update. https://github.com/Strategy11/formidable-forms/blob/aa202e1d3fb368e444926480ec0ece46eed0d81c/classes/helpers/FrmFormsHelper.php#L1044

Crabcyborg:
@garretlaxton I think this is ready for testing now.

This update fixes https://github.com/Strategy11/formidable-pro/issues/3161 but it also adds a nonce to all of the links that duplicate a form, which includes things like duplicating a form template when using a custom template from the new form modal.

garretlaxton (approved):
This looks great! The form duplication after refresh is solved and there are nonces added to the duplication links.

classes/controllers/FrmFormsController.php
```diff
@@ -617,7 +628,7 @@ public static function build_template() {
 			}
 
 			$response = array(
-				'redirect' => admin_url( 'admin.php?page=formidable&frm_action=duplicate&id=' . $new_form_id ),
+				'redirect' => admin_url( 'admin.php?page=formidable&frm_action=duplicate&id=' . $new_form_id ) . '&_wpnonce=' . wp_create_nonce(),
```
Crabcyborg, line 631: @AbdiTolesa it looks like on one line you're using `wp_nonce_url` and on another you're adding `&_wpnonce` yourself. Any reason for the difference? Could we make this more consistent? It might make sense to add a new function that just takes in the form id and returns the url with the nonce?

Screen Shot 2022-07-13 at 9 29 46 AM

AbdiTolesa: @Crabcyborg The place I appended nonce myself is because that one is for handling an ajax request and `wp_nonce_url` escapes the url and when that is sent back encoded to the referrer the page redirect breaks.

PR #788, opened 2022-06-13
Fix Switching field IDs could break the serialize data

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3624

### DIFF
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -190,7 +190,15 @@ class FrmForm {
 				continue;
 			}
 
-			$new_val = FrmFieldsHelper::switch_field_ids( $value );
+			if ( 'field_options' === $key ) {
+				// Need to loop through field_options to prevent breaking serialized string when length changed.
+				FrmAppHelper::unserialize_or_decode( $field[ $key ] );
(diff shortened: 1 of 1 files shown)


classes/models/FrmForm.php
```diff
@@ -190,7 +190,15 @@ private static function switch_field_ids_in_field( $field ) {
 				continue;
 			}
 
-			$new_val = FrmFieldsHelper::switch_field_ids( $value );
+			if ( 'field_options' === $key ) {
+				// Need to loop through field_options to prevent breaking serialized string when length changed.
+				FrmAppHelper::unserialize_or_decode( $field[ $key ] );
+				$new_val = FrmFieldsHelper::switch_field_ids( $field[ $key ] );
```
Crabcyborg, line 196: Thanks @truongwp for fixing this! This looks good. I found these lines a bit confusing at first though as everything else uses `$value` but you're accessing `$field[ $key ]` directly, which is the same thing, right?

truongwp: You're right. I updated it.

Crabcyborg (approved):

PR #790, opened 2022-06-14
Set opacity:1 to firefox placeholder style

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/1920249527/101307/

The Read Only Text Color setting is used for placeholders. On Chrome, it uses the color. In Firefox, you need to specify `opacity: 1` or it appears as a more faded out alpha transparency.

Related Stack Overflow question https://stackoverflow.com/questions/19621306/css-placeholder-text-color-on-firefox#answer-19621642

**Before**
Screen Shot 2022-06-14 at 10 38 41 AM

**After**
Screen Shot 2022-06-14 at 10 38 32 AM


stephywells (approved):
Nice find. Looks good.

PR #791, opened 2022-06-15
Add floating labels effect to "inside" label position

truongwp:
Inline documentation is in progress.

Fixes https://github.com/Strategy11/formidable-pro/issues/3617

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1266,6 +1266,7 @@ function frmFrontFormJS() {
 		});
 
 		runOnLoad = function() {
+			console.log( 'onload' );
 			checkDropdownLabel();
 
 			if ( isIE() ) {
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1219,7 +1219,7 @@ function frmFrontFormJS() {
 					firstOpt.text = '';
 				}
 			});
-		}
+		};
 
 		checkPlaceholderIE = function( input ) {
 			if ( input.value ) {
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1178,16 +1178,48 @@ function frmFrontFormJS() {
 	}
 
 	function initFloatingLabels() {
-		var checkFloatLabel, checkPlaceholderIE, selector;
+		var checkFloatLabel, checkDropdownLabel, checkPlaceholderIE, runOnLoad, selector, floatClass;
 
-		selector = '.frm-show-form .frm_inside_container input, .frm-show-form .frm_inside_container select, .frm-show-form .frm_inside_container textarea';
+		selector   = '.frm-show-form .frm_inside_container input, .frm-show-form .frm_inside_container select, .frm-show-form .frm_inside_container textarea';
+		floatClass = 'frm_label_float_top';
 
 		checkFloatLabel = function( event ) {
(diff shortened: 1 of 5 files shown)


js/formidable.js
```diff
@@ -1135,6 +1138,129 @@ function frmFrontFormJS() {
...
+			if ( isIE() ) {
+				document.querySelectorAll( selector ).forEach( function( input ) {
+					checkPlaceholderIE( input );
+				});
+			}
+		};
+
+		runOnLoad();
+
+		jQuery( document ).on( 'frmPageChanged', function( event ) {
```
Crabcyborg, line 1259: Does `documentOn` work here too by any chance?

truongwp: I tried this before but `document.addEventListener()` and `documentOn()` don't work with `jQuery( document ).trigger()`;

Crabcyborg:
@garretlaxton this update is mostly done and could use some solid testing. It updates the "Inside" label position setting so that the label will transition from the input to the label position with an animation.

js/formidable.js
```diff
@@ -1153,7 +1303,7 @@ function frmFrontFormJS() {
 			jQuery( document.getElementById( 'frm_resend_email' ) ).on( 'click', resendEmail );
 
 			jQuery( document ).on( 'change', '.frm-show-form input[name^="item_meta"], .frm-show-form select[name^="item_meta"], .frm-show-form textarea[name^="item_meta"]', frmFrontForm.fieldValueChanged );
-			jQuery( document ).on( 'change keyup', '.frm-show-form .frm_inside_container input, .frm-show-form .frm_inside_container select, .frm-show-form .frm_inside_container textarea', maybeShowLabel );
+			// jQuery( document ).on( 'change keyup', '.frm-show-form .frm_inside_container input, .frm-show-form .frm_inside_container select, .frm-show-form .frm_inside_container textarea', maybeShowLabel );
```
Crabcyborg, line 1306: We probably just want to remove this line if we don't need it anymore. Keeping commented out code usually just confuses things.

tests/fields/test_FrmFieldsController.php
```diff
@@ -18,7 +18,10 @@ public function test_prepare_placeholder() {
 			'required'    => 0,
 		);
 		$placeholder = $this->prepare_placeholder( $field );
-		$this->assertEquals( $name, $placeholder, 'an empty string should be replaced by the label inside of an input.' );
+		// $this->assertEquals( $name, $placeholder, 'an empty string should be replaced by the label inside of an input.' );
```
Crabcyborg, line 21: This line too, I think we can just remove the commented assert.

truongwp:
@garretlaxton Thanks for testing. I fixed those issues. Please check again.

garretlaxton (approved):
This looks great now!

Crabcyborg:
@truongwp I'm still seeing the overlapping issue when I refresh like I mention at the end of https://github.com/Strategy11/formidable-forms/pull/791#issuecomment-1160656751

I played with it for a moment. I have an incomplete update here that tries to correct any inside containers that might have a value on load. https://github.com/Strategy11/formidable-forms/pull/803

I checked Chrome and it looks like it doesn't have this behavior. But this is how it works in Firefox if you enter values into a field, then refresh the page.

PR #796, opened 2022-06-23
New frm_after_destroy_entry_hook

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/1924502130/101690/

There isn't a very practical way to hook into after these get deleted, just before. In this customer's case they want to build their cache after the deleting happens, which doesn't really work if you try to trigger it beforehand.

I opted to still pass the deleted entry so you can check it for specific things like form_id or post_id associated with the entry as well.

Example use.
```
add_action(
	'frm_after_destroy_entry',
	function( $entry_id, $entry ) {
		$target_form_id = 294; // change this.
		if ( $target_form_id !== (int) $entry->form_id ) {
			return;
		}

		// Handle deleted entry for target form.
	},
	10,
	2
);
```

### DIFF
--- a/classes/models/FrmEntry.php
+++ b/classes/models/FrmEntry.php
@@ -251,6 +251,12 @@ class FrmEntry {
 		return $query_results;
 	}
 
+	/**
+	 * Delete an entry.
+	 *
+	 * @param string|int $id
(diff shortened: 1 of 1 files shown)


stephywells (approved):
This looks perfect. Thanks Mike!

PR #802, opened 2022-06-29
Fixes Issue 3637

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3637


Crabcyborg (commented):
@AbdiTolesa I haven't looked over the rest of this but we don't want to commit the translation po/mo files in here. It's showing a huge diff as a result. If you could remove this from the pull request that would be great.

Screen Shot 2022-06-30 at 12 54 20 PM
Screen Shot 2022-06-30 at 12 54 28 PM

classes/controllers/FrmFieldsController.php
```diff
@@ -841,6 +841,18 @@ public static function check_label( $opt ) {
...
 
+	/**
+	 * Sort aria-describedby values.
+	 *
+	 * @param array $aria_describedby_values The list of aria-describedby values.
+	 *
+	 * @return string
+	 */
+	public static function sort_aria_describedby_values( $aria_describedby_values ) {
+		rsort( $aria_describedby_values );
```
Crabcyborg, line 852: I don't think `rsort` will give us the order we're looking for necessarily.

I'm looking more at the customer's ticket. They show their custom values as this, which explains why we're seeing an error with a key as well as an error with an id.

input aria-describedby=frm_error field_ key  frm_desc field  key  autoc

I'm thinking we could try to derive the order from the current value rather than adding a new `frm_aria_describedby_values` filter now. If they have it here as `error, description` we should use that as our order. If it's set here as `description, error` we could do the opposite.

AbdiTolesa: @Crabcyborg I have a few questions yet. What if they just have one custom value, or even no custom value is added here?

Crabcyborg: @AbdiTolesa if there is no custom value we just use a default order. I think it should be error, description by default. If there is any custom value as well, we can just use the order its in in this tag. If the custom attributes don't include error/description, we can just append them, but still prioritize error over description.

Crabcyborg:
@garretlaxton this is a follow-up from the other aria-describedby update from last release.

The customer had a list of feedback items on how to improve it further based off of  https://github.com/Strategy11/formidable-pro/issues/3637

@AbdiTolesa if you turn on submit with AJAX, does this have any issues with the JavaScript adding duplicates or was that all already handled in the JavaScript side? And would it support the custom ordering as expected there?

I also pushed a small commit to drop the old function.

js/formidable.js
```diff
@@ -792,9 +792,14 @@ function frmFrontFormJS() {
 
 				if ( typeof describedBy === 'undefined' ) {
 					describedBy = id;
-				} else if ( describedBy.indexOf( id ) === -1 ) {
-					describedBy = describedBy + ' ' + id;
+				} else if ( describedBy.indexOf( id ) === -1 && describedBy.indexOf( 'frm_error_field_' ) === -1 ) {
+					if ( input.data( 'error-first' ) === 0 ) {
```
Crabcyborg, line 796: It might be nice to avoid jQuery here (the `.data` function) so it'll be easier to move away from jQuery in the future.

garretlaxton:
This looks to work more as described now. 

Sorry, I usually don't look at the PR until I come to QA. It's probably best to ping me in Slack to let me know if there's anything to prioritize.

garretlaxton (approved):

stephywells:
@AbdiTolesa it looks like the unit tests are failing from this PR. Can you please submit a new PR that fixes this?

Thank you!

AbdiTolesa:
@stephywells Thanks for pointing that out. The unit test was not updated after further enhancements were asked by our customer and we updated it based on the new requirements. I will review the unit test shortly.

PR #803, opened 2022-06-30
Check inside container for non empty value on load

Crabcyborg:
**Before**

Screen Shot 2022-06-30 at 1 29 49 PM

**After**

3n0ZphN0Et


truongwp:
Thanks @Crabcyborg. The code looks good. Do you think we should remove the PHP code which makes the label float top since we use JS to check at the page load?

Crabcyborg:
@truongwp I think your other check is probably good to keep as it sets the class right away. In my case, the animation happens right away, which I think is acceptable to catch this edge case, but is nicer that the label is already at the top if it has a value.

If I could detect this with PHP too I would, but it seems to only persist on the client side of the request.

truongwp:
@Crabcyborg Thank you. I'll pull and fix that.

truongwp:
@Crabcyborg I pushed the fix. I think this PR is ready to merge now.

PR #804, opened 2022-07-01
Update upsells for form settings to no longer use a pop up

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3638
Fixes https://github.com/Strategy11/formidable-pro/issues/3641
Fixes https://github.com/Strategy11/formidable-pro/issues/3642

Pretty straight forward. Instead of using the `data-upgrade`, `data-message`, attributes for the modal, they're being used to fill in empty pages. Added new `data-screenshot` attributes for the new images.

I also updated the pop up code so it only uses jQuery for the modal now. And refactored some of the functions so they work outside of the modal.

Right now I'm just borrowing the upgrade call to action from the upgrade modal which still exists on this page for other pro features (like actions).

This update also includes an updated up-sell for reports (screenshot at the bottom). I added a tweak so `'lite-reports'` would count as `'reports'` when checking for the active tab so the tab doesn't appear to be inactive anymore.

Screen Shot 2022-07-01 at 2 14 19 PM

Screen Shot 2022-07-01 at 2 14 14 PM

Screen Shot 2022-07-01 at 2 17 16 PM

Screen Shot 2022-07-01 at 2 17 19 PM

**Updated reports up-sell**

Screen Shot 2022-07-04 at 10 40 50 AM

**And when the `oneclick` install gets triggered.**

Screen Shot 2022-07-04 at 11 28 39 AM

**Global settings**

Screen Shot 2022-07-04 at 3 15 54 PM

Screen Shot 2022-07-04 at 3 15 58 PM


stephywells (approved):
Looks great @Crabcyborg. I optimized the images a little more. 

Thanks so much!

PR #806, opened 2022-07-01
Add upsell for multiple style templates

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3639

Dropping the alpha slider up-sell as I know Steve commented it's a bit funny. I don't know how clear it is or if it's worth trying to point that out as an up-sell. But added more attention to the duplicate button there instead which seems important.

Screen Shot 2022-07-01 at 4 19 38 PM

Screen Shot 2022-07-01 at 4 19 16 PM


stephywells (approved):
The alpha upsell was a little weird. This looks good to me! Thank you for cleaning this up.

PR #807, opened 2022-07-04
Views/reports tabs not faded out when using lite fix

Crabcyborg:
Noticed when using lite only that views/reports don't ever fade out anymore.

Checked git blame and it looks like it happened about 8 months ago when field search was added https://github.com/Strategy11/formidable-forms/commit/aebf5b4ea080ef30b1efadea9a6e550debe1afa8

**Before**
Screen Shot 2022-07-04 at 10 56 05 AM

**After**
Screen Shot 2022-07-04 at 10 55 53 AM


stephywells (approved):
Nice catch! Sorry I missed that.

PR #813, opened 2022-07-13
Repeater-entry-style

stephywells:
This updates the styling of a single entry in a table, based on the design here:
https://www.figma.com/file/K3L40rJTHXwjNeBA3LoWBr/%E2%9C%85-PDF?node-id=604%3A296

Requires https://github.com/Strategy11/formidable-pro/pull/3658

After:
Screen Shot 2022-07-12 at 6 25 37 PM

### DIFF
--- a/classes/models/FrmTableHTMLGenerator.php
+++ b/classes/models/FrmTableHTMLGenerator.php
@@ -54,6 +54,14 @@ class FrmTableHTMLGenerator {
 	protected $td_style = '';
 
 	/**
+	 * Used to add a class in tables. Set in Pro.
+	 *
+	 * @var bool
+	 * @since x.x
+	 */
+	public $is_child = false;
+
+	/**
 	 * FrmTableHTMLGenerator constructor.
 	 *
 	 * @param string $type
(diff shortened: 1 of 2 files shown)


truongwp:
@stephywells Does this PR cover this task https://app.asana.com/0/0/1202597215708055/f?

classes/models/FrmTableHTMLGenerator.php
```diff
@@ -304,12 +310,35 @@ public function generate_two_cell_table_row( $label, $value ) {
...
+		$class = '';
+		if ( $empty ) {
+			// Only add this class on two cell rows.
+			$class .= ' frm-empty-row';
+		}
+		if ( $this->is_child ) {
+			$class .= ' frm-child-row';
+		}
+		if ( $class ) {
+			$class = ' class="' . $class . '"';
```
Crabcyborg, line 340: I don't know if we'd maybe want to trim this to avoid the the funny space in `class=" frm-empty-row"` but that isn't really a big deal.

Crabcyborg (approved):
Looks good!

classes/models/FrmTableHTMLGenerator.php
stephywells: I was thinking the same thing. Incoming.

stephywells:
> https://app.asana.com/0/0/1202597215708055/f

Yep, exactly!

garretlaxton (approved):
The new styling looks great. 

The only feedback I have is if you leave a field inside the repeater blank, the empty row will mess up the zebra stripes in the rows. You can see in the screenshot below the Radio Button row should be white. 
Screenshot 2022-07-13 at 16-14-27 Formidable Entries ‹ Formidable Forms — WordPress

I notice this happened before this update so not sure if you would want to include that here or not.

stephywells:
Thanks @garretlaxton ! That is a bit weird sometimes. That isn't new, so we can address that in another PR. Thanks again!

PR #814, opened 2022-07-13
add aria-label to autocomplete dropdown items.

AbdiTolesa:
As related to this [PR](https://github.com/Strategy11/formidable-views/pull/324), the applications dropdown items id is read instead of the Application name. This pertains to the autocomplete function. This can be fixed by extending the `_renderItem` when the autocomplete is created. [https://api.jqueryui.com/autocomplete/#method-_renderItem](https://api.jqueryui.com/autocomplete/#method-_renderItem)

image


Crabcyborg (approved):
Thanks @AbdiTolesa! Good idea. I tried it out and agree that just hearing it read out "3" might not be very useful 😅.

PR #816, opened 2022-07-14
Add floating labels code to Styles page

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3680

### DIFF
--- a/css/custom_theme.css.php
+++ b/css/custom_theme.css.php
@@ -440,38 +440,38 @@ legend.frm_hidden{
 }
 
 /* Floating labels */
-.with_frm_style .frm_inside_container {
+.with_frm_style .frm_inside_container:not(.frm_not_inside_container) {
 	position: relative;
 	padding-top: 27px;
 	padding-top: calc(0.85 * var(--field-height));
 }
 
-.with_frm_style .frm_inside_container > input,
-.with_frm_style .frm_inside_container > select,
-.with_frm_style .frm_inside_container > textarea {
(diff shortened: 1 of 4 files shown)


css/custom_theme.css.php
```diff
@@ -440,38 +440,38 @@
...
+.with_frm_style .frm_inside_container:not(.frm_not_inside_container) {
 	position: relative;
 	padding-top: 27px;
 	padding-top: calc(0.85 * var(--field-height));
 }
 
-.with_frm_style .frm_inside_container > input,
-.with_frm_style .frm_inside_container > select,
-.with_frm_style .frm_inside_container > textarea {
+.with_frm_style .frm_inside_container:not(.frm_not_inside_container) > input,
```
Crabcyborg, line 449: @truongwp this seems like a not of `:not`... to have to add to the CSS. Does it really need to be an `frm_inside_container` and a `frm_not_inside_container`? Is it possible to just not use `frm_inside_container` for these "not inside" containers?

truongwp: I added `frm_not_inside_container` because I don't want to check if the field supports floating labels in JS when changing Styles settings. That is also a way to disable floating labels on a field. 
If you don't want to add that class and checking field in JS doesn't affect the performance much, I can change this.

Crabcyborg: I think it'd be best to try to keep this in style-sheet specific code as much as we can.

For a style manager fix this is a lot of front end CSS to add to every customer's website.

So if we can avoid more CSS by adding more admin side JavaScript I think it's worth it.

PR #817, opened 2022-07-14
Stop showing autoresponder upsell for action types with no autoresponder support

Crabcyborg:
This becomes pretty apparent one you start trying to add a bunch of outcome quiz actions. The extra up-sell on every one takes up a lot of space.

I got confused by this before, when I went to install an autoresponder from this trigger only to refresh the action and see the settings didn't exist for my action type. I think it could be a misleading up-sell that could get customers frustrated thinking something supports automation when it doesn't.

This removes it from all quiz actions and others that aren't 'email', 'twilio', 'api', 'register', or 'zapier'.

**Before**
Screen Shot 2022-07-14 at 2 30 11 PM

**After**
Screen Shot 2022-07-14 at 2 33 04 PM

### DIFF
--- a/classes/views/frm-form-actions/_action_inside.php
+++ b/classes/views/frm-form-actions/_action_inside.php
@@ -79,7 +79,7 @@ if ( ! FrmAppHelper::pro_is_installed() ) {
 }
 
 // Show Form Action Automation indicator.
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Oh, definitely. Great catch!

PR #818, opened 2022-07-15
Add new static() to FrmAddon::load_hooks()

truongwp:
The conversation for this: https://github.com/Strategy11/formidable-pdfs/pull/22#discussion_r920437309

### DIFF
--- a/classes/models/FrmAddon.php
+++ b/classes/models/FrmAddon.php
@@ -3,6 +3,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 	die( 'You are not allowed to call this page directly.' );
 }
 
+/** @phpstan-consistent-constructor */
 class FrmAddon {
 	public $store_url = 'https://formidableforms.com';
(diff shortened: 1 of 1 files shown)


classes/models/FrmAddon.php
```diff
@@ -37,7 +37,7 @@ public function __construct() {
 
 	public static function load_hooks() {
 		add_filter( 'frm_include_addon_page', '__return_true' );
-		//new static();
+		new static();
```
stephywells, line 40: @truongwp Looks like PHPstan doesn't like this. Let's use this solution:
https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static#add-%40phpstan-consistent-constructor-to-the-class

truongwp: Thanks for finding the solution. I updated it.

stephywells (approved):
This looks great now. Thanks @truongwp !

PR #819, opened 2022-07-15
add function that returns the list of field translatable strings.

AbdiTolesa:
It might be a good idea to create a function that returns the list of translatable strings for a given field pretty much like `FrmForm::translatable_strings` to keep things consistent. We need similar functionality for fields to fix [https://github.com/Strategy11/formidable-wpml/issues/3](https://github.com/Strategy11/formidable-wpml/issues/3)

### DIFF
--- a/classes/models/FrmField.php
+++ b/classes/models/FrmField.php
@@ -1146,4 +1146,23 @@ class FrmField {
 
 		return ! empty( $field_type_obj->is_combo_field );
 	}
+
+	/**
+	 * Get a list of all field settings that should be translated
+	 * on a multilingual site.
+	 *
+	 * @since x.xx
+	 * @param object $field - The field object
+	 */
+	public static function translatable_strings( $field ) {
+		$strings = array(
+			'name',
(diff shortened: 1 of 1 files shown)


Crabcyborg (commented):
@AbdiTolesa so it looks like we do have something like this already.

https://github.com/Strategy11/formidable-forms/blob/c7838f6e970a320241b5e887769d0aae7f46f786/classes/models/fields/FrmFieldType.php#L324

It's in `FrmFieldType` so it can be extended as well by certain field types.

`conf_msg` gets added for the types that support it in Pro. https://github.com/Strategy11/formidable-pro/blob/a44db95fd2d3de65f70aca5004ab7ec3af87e1d4/classes/models/fields/FrmProFieldEmail.php#L58

Toggle fields have their own strings https://github.com/Strategy11/formidable-pro/blob/30a07c8025598b768e745bd51f089b844bee5d41/classes/models/fields/FrmProFieldToggle.php#L49

So I think you just need to use those functions.

PR #820, opened 2022-07-18
Update Zapier icon

stephywells:
Since the Zapier logo changed, we need to update it in plugin too. This also sets the color in the form action to Orange when the Zapier add-on isn't installed.

Before:
Screen Shot 2022-07-18 at 9 57 11 AM

After:
Screen Shot 2022-07-18 at 9 56 52 AM


Crabcyborg (approved):
Looks good!

PR #823, opened 2022-07-19
Fixes Issue #3389

AbdiTolesa:
issue: [#3389](https://github.com/Strategy11/formidable-pro/issues/3389)

Following from [https://github.com/Strategy11/formidable-forms/pull/511](https://github.com/Strategy11/formidable-forms/pull/511), a `Save and Reload` functionality has been added to the popup that appears when activating addons from the settings page. However, that didn't cover the same action from form builder. This PR is to extend the existing code and bring the same functionality to the form builder.

image

image

### DIFF
--- a/classes/controllers/FrmAddonsController.php
+++ b/classes/controllers/FrmAddonsController.php
@@ -1015,9 +1015,12 @@ class FrmAddonsController {
 	 * @return string
 	 */
 	private static function get_activating_page() {
-		if ( false !== strpos( FrmAppHelper::get_server_value( 'HTTP_REFERER' ), 'frm_action=settings' ) ) {
+		$referer = FrmAppHelper::get_server_value( 'HTTP_REFERER' );
+		if ( false !== strpos( $referer, 'frm_action=settings' ) ) {
 			return 'settings';
-		} elseif ( false !== strpos( FrmAppHelper::get_server_value( 'HTTP_REFERER' ), 'frm_action=edit' ) ) {
+		}
+
+		if ( false !== strpos( $referer, 'frm_action=edit' ) ) {
 			return 'form_builder';
 		}
 
--- a/classes/controllers/FrmAddonsController.php
+++ b/classes/controllers/FrmAddonsController.php
@@ -998,21 +998,30 @@ class FrmAddonsController {
 	 * @return array|string
 	 */
 	private static function get_addon_activation_response() {
+		$activating_page = self::get_activating_page();
+
 		$response = array(
-			'message' => __( 'Your plugin has been activated. Would you like to save and reload the page now?', 'formidable' ),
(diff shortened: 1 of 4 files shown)


css/frm_admin.css
```diff
@@ -1838,7 +1838,7 @@ h2.frm-h2 + .howto {
 	font-size: 13px;
 }
 
-#frm_save_and_reload_settings, #frm_save_and_reload_settings + .frm-button-secondary {
+#frm_save_and_reload_settings, #frm_save_and_reload_form_builder, #frm_save_and_reload_settings + .frm-button-secondary, #frm_save_and_reload_form_builder + .frm-button-secondary {
```
Crabcyborg, line 1841: @AbdiTolesa thanks for working on this, it looks like it works well. I think we should just clean it up a bit more.

I think we could probably reduce the code a bit if we just used the same ID for both buttons?

We could just call it `#frm_save_and_reload`. Then in your event listener, since it's nested in the function, you can just refer to the value of `saveAndReload` instead of checking for `button.id`.

classes/controllers/FrmAddonsController.php
```diff
@@ -998,22 +998,30 @@ public static function ajax_activate_addon() {
...
-	 * @return bool
+	 * Return a string that reflects the page from which the addon is being activated on,
+	 * if it is from settings or form builder, otherwise return empty string.
+	 *
+	 * @return string
 	 */
-	private static function activating_from_settings_page() {
-		return false !== strpos( FrmAppHelper::get_server_value( 'HTTP_REFERER' ), 'frm_action=settings' );
+	private static function get_activating_page() {
+		if ( false !== strpos( FrmAppHelper::get_server_value( 'HTTP_REFERER' ), 'frm_action=settings' ) ) {
```
Crabcyborg, line 1018: I think we could make this function a bit easier to follow as well. Instead of calling `FrmAppHelper::get_server_value` it would make sense to just set it to a variable first.

And it's good to not use an else after a return as it wouldn't really reach the following code either way.

It reads a bit simpler if you just check like (this is just pseudo code).

```
If ( referer contains frm_action=settings )
    return settings;

if ( referer contains frm_action=edit )
    return form_builder
```

AbdiTolesa: @Crabcyborg I have used `elseif` actually that it returns an empty string if neither of the conditions are met. Assigning the query string to a variable makes sense indeed.

css/frm_admin.css
AbdiTolesa: Yes this makes sense as there is no need to necessarily have different ids as they are different pages. Thank you!

Crabcyborg (approved):
Thanks @AbdiTolesa!

PR #825, opened 2022-07-20
Fix floating label doesn't work well on Styles page

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3680

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -7540,14 +7540,14 @@ function frmAdminBuildJS() {
 			const firstOpt = input.querySelector( 'option:first-child' );
 
 			if ( shouldFloatTop ) {
-				if ( firstOpt.getAttribute( 'data-label' ) ) {
-					firstOpt.text = firstOpt.getAttribute( 'data-label' );
+				if ( firstOpt.hasAttribute( 'data-label' ) ) {
+					firstOpt.textContent = firstOpt.getAttribute( 'data-label' );
 					firstOpt.removeAttribute( 'data-label' );
 				}
 			} else {
-				if ( firstOpt.text ) {
-					firstOpt.setAttribute( 'data-label', firstOpt.text );
-					firstOpt.text = '';
+				if ( firstOpt.textContent ) {
+					firstOpt.setAttribute( 'data-label', firstOpt.textContent );
+					firstOpt.textContent = '';
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


Crabcyborg:
Thanks @truongwp it looks like it fixes the issue for most fields but I'm still seeing some funny stuff happening with radio buttons, toggles, and sliders.

Screen Shot 2022-07-20 at 9 33 53 AM

truongwp:
@Crabcyborg Thank you. I pushed the fix. Please check it again.

stephywells (changes requested):
Would this be more simple to do in the html? Or is there something that needs to be changed in the sample form HTML so the existing js will work? This seems like it may be tricky to maintain.

truongwp:
@stephywells I created a fix like that here <https://github.com/Strategy11/formidable-forms/pull/816|https://github.com/Strategy11/formidable-forms/pull/816>. You might want to take a look.

js/formidable_admin.js
```diff
@@ -7507,7 +7507,50 @@ function frmAdminBuildJS() {
...
+
+			if ( shouldFloatTop ) {
+				if ( firstOpt.getAttribute( 'data-label' ) ) {
+					firstOpt.text = firstOpt.getAttribute( 'data-label' );
+					firstOpt.removeAttribute( 'data-label' );
+				}
+			} else {
+				if ( firstOpt.text ) {
+					firstOpt.setAttribute( 'data-label', firstOpt.text );
+					firstOpt.text = '';
```
Crabcyborg, line 7550: @truongwp I actually can't find any real documentation on `.text`. It seems to work, but I don't really have a good idea of what the browser support for it actually is. What do you think about using https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent?

js/formidable_admin.js
```diff
@@ -7507,7 +7507,50 @@ function frmAdminBuildJS() {
...
+
+		const shouldFloatTop = input.value || document.activeElement === input;
+
+		container.classList.toggle( 'frm_label_float_top', shouldFloatTop );
+
+		if ( 'SELECT' === input.tagName ) {
+			const firstOpt = input.querySelector( 'option:first-child' );
+
+			if ( shouldFloatTop ) {
+				if ( firstOpt.getAttribute( 'data-label' ) ) {
```
Crabcyborg, line 7543: Since you're not actually getting the attribute value, you could use https://developer.mozilla.org/en-US/docs/Web/API/Element/hasAttribute instead. It's more intentional and likely better for performance. Either that or getting the data-label and checking the variable would work as well.

js/formidable_admin.js
truongwp: Ok. I'll need to update this in the frontend too.

Crabcyborg (approved):
Thanks @truongwp!

PR #828, opened 2022-07-20
Improve JS code when changing label position on Styles page

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3702


Crabcyborg (approved):
Thanks @truongwp! Looks good!

PR #836, opened 2022-07-27
Update checkUniqueOpt

Crabcyborg:
Related Slack conversation https://strategy11.slack.com/archives/C799A2R61/p1658910750598379

> Would there be any way to include a warning in-plugin if identical values are detected, telling users that separate values MUST be unique?

I got it working again, removed the jQuery, added support for separate values, and made the error message more helpful.

Screen Shot 2022-07-27 at 3 30 16 PM

_Branching off of https://github.com/Strategy11/formidable-forms/pull/835 as I forgot to checkout master first. That shouldn't be a real issue as I hope to include both of these in next release._

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -5,12 +5,12 @@ on:
     branches:
       - master
   pull_request:
-    branches:
-      - master
+    types: [ opened, labeled, synchronize ]
 
 name: Inspections
 jobs:
   runPHPCSInspection:
+    if: contains(github.event.pull_request.labels.*.name, 'run analysis')
     name: Run PHPCS inspection
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


garretlaxton:
I only see this warning when I select an option in the builder. Are there any other times this should pop-up?
Screenshot 2022-07-27 at 16-23-46 Formidable Forms ‹ Formidable Forms — WordPress

I also see this pop-up when I select an option and there actually is no duplicates.
Screenshot 2022-07-27 at 16-21-59 Formidable Forms ‹ Formidable Forms — WordPress

Crabcyborg:
That's it. It's intended to trigger when/as you add the duplicate. A visible wanting at all times may make sense too. What are your thoughts @stephywells?

stephywells:
@Crabcyborg when we added this originally, it triggered on blur. Looks like this is already on those fields:
data-frmchange="trim,updateOption"
data-frmchange="trim,updateDefault"

So we could trigger it from inside the updateDefault and updateOption functions.

Crabcyborg:
> data-frmchange="trim,updateOption" data-frmchange="trim,updateDefault"

Thanks @stephywells I just added checkUniqueOpt into that list. It does an explode and checks the loop of strings so adding a new function makes sense rather than trying to add more logic to the others.

stephywells (approved):
Nice! Looks great!

PR #837, opened 2022-07-28
Issue 3679

AbdiTolesa:
Fixes [#3679](https://github.com/Strategy11/formidable-pro/issues/3679)

Based on Mike's research and suggestion in the issue description, I have tried to offload the drag and save icons from initial DOM loading to conditional loading of those icons on interaction with javascript . When a field is selected all its options in the left will get those icons appended. This would greatly increase the page's performance.


classes/views/frm-fields/single-option.php
```diff
@@ -4,7 +4,17 @@
 }
 ?>
 <li id="frm_delete_field_<?php echo esc_attr( $field['id'] . '-' . $opt_key ); ?>_container" data-optkey="<?php echo esc_attr( $opt_key ); ?>" class="frm_single_option <?php echo $opt_key === '000' ? 'frm_hidden frm_option_template' : ''; ?>">
-	<?php FrmAppHelper::icon_by_class( 'frmfont frm_drag_icon frm-drag' ); ?>
+	<?php
+	$show_icons = true;
+	if ( isset( $field['do_not_include_icons'] ) ) {
```
Crabcyborg, line 9: @AbdiTolesa do you see this being used anywhere else? I'm not sure if this is really a flag we need to add or if we can just remove this code from PHP.

js/formidable_admin.js
```diff
@@ -2911,6 +2911,8 @@ function frmAdminBuildJS() {
 			newOption = newOption.replace( new RegExp( '-' + oldKey + '"', 'g' ), '-' + optKey + '"' );
 			newOption = newOption.replace( new RegExp( '\\[' + oldKey + '\\]', 'g' ), '[' + optKey + ']' );
 			newOption = newOption.replace( 'frm_hidden frm_option_template', '' );
+			newOption = jQuery.parseHTML( newOption );
```
Crabcyborg, line 2914: @AbdiTolesa can you come up with a way of doing this without adding jQuery? We're trying to eventually move away from having any jQuery dependencies where possible.

js/formidable_admin.js
```diff
@@ -9923,9 +9925,23 @@ jQuery( document ).ready(
 			html = html.replaceAll( '</li>', '</div>' );
 			ul.outerHTML = html;
 		}
+
+		let saveIcon = '<svg class="frmsvg"><use xlink:href="#frm_save_icon"></use></svg>';
```
Crabcyborg, line 9929: @AbdiTolesa could you use the `svg` function that already exists in this file instead?

https://github.com/Strategy11/formidable-forms/blob/b066b463a84df20f10b01b36d8fd50c5ff58f45d/js/formidable_admin.js#L7314

It should just need to be `const saveIcon = svg({ href: '#frm_save_icon' });`

Crabcyborg (commented):
Thanks for working on this @AbdiTolesa! This looks like a good start, and it looks like it does have a pretty significant performance increase.

classes/views/frm-fields/single-option.php
AbdiTolesa: @Crabcyborg Yes this is used when adding a new select fields for example. We could handle all scenarios at extra js code and a bit of PHP code modification, but that wouldn't have much performance implication and that's why I used the flag to be set in the hook that fires when the builder loads initially. I think it might also require us to create a new filter.

js/formidable_admin.js
AbdiTolesa: @Crabcyborg Understood, I have replaced this with a JS code.

js/formidable_admin.js
AbdiTolesa: @Crabcyborg To be able to use the `svg` function, I think we need to modify it a bit first to allow passing css classes. I have now updated the function to support that.

classes/views/frm-fields/single-option.php
Crabcyborg: @AbdiTolesa thanks for clearing that up. What do you think about using something a bit more clear that it's actually on. An empty string `''` isn't really a clear indication that this is on. Something like `true` should be safer and more clear.

js/formidable_admin.js
```diff
@@ -9923,6 +9929,11 @@ jQuery( document ).ready(
 			html = html.replaceAll( '</li>', '</div>' );
 			ul.outerHTML = html;
 		}
+
+		frmGlobal.icons = {
+			save: frmDom.svg({ href: '#frm_save_icon' }),
```
Crabcyborg, line 9934: @AbdiTolesa I think we should probably avoid adding stuff to `frmGlobal`. That could probably lead to weird tough-to-debug issues down the line if we were to add icons to frmGlobal and not realize this existed, etc, since it would be defining pieces of the object in multiple locations.

If this code isn't inside of the scope of `function frmAdminBuildJS` maybe it should be. We could probably keep `icons` in the scope of these other variables here https://github.com/Strategy11/formidable-forms/blob/0e2e47d6d576beb6840d261cca2c8b2412b843dd/js/formidable_admin.js#L323

Also, right about that, I set `svg` in the function scope https://github.com/Strategy11/formidable-forms/blob/0e2e47d6d576beb6840d261cca2c8b2412b843dd/js/formidable_admin.js#L320. If you move this code you should be able to just call `svg` rather than having to access it from `frmDom.svg`.

js/formidable_admin.js
```diff
@@ -10007,3 +10018,66 @@ function frmImportCsv( formID ) {
 		}
 	});
 }
+
+function addSaveAndDragIconsToOption( fieldId, liObject ) {
+	if ( liObject.newOption ) {
+		const parser = new DOMParser();
+		li = parser.parseFromString( liObject.newOption, 'text/html' ).body.childNodes[0];
```
Crabcyborg, line 10025: @AbdiTolesa it doesn't look like all of these variables get declared? I never see a `let/var` declaration for `li`.

js/formidable_admin.js
```diff
@@ -10007,3 +10018,66 @@ function frmImportCsv( formID ) {
...
+	} else {
+		li = liObject;
+	}
+
+	const icons = li.querySelectorAll( 'svg' );
+	let hasDragIcon = false;
+	let hasSaveIcon = false;
+	icons.forEach( ( svg, key ) => {
+		useTag     = svg.getElementsByTagNameNS( 'http://www.w3.org/2000/svg', 'use' )[0];
```
Crabcyborg, line 10034: Here's another one. I don't see a declaration for `useTag`.

js/formidable_admin.js
```diff
@@ -10007,3 +10018,66 @@ function frmImportCsv( formID ) {
 		}
 	});
 }
+
+function addSaveAndDragIconsToOption( fieldId, liObject ) {
```
Crabcyborg, line 10022: @AbdiTolesa it looks like this function is also in the global scope. Can we move into the functions so it can't be referenced outside? It shouldn't be possible for other people to call the function.

Crabcyborg:
The ESLint workflow is also complaining about some of this.

```
   1906:15  error    There must be a space inside this paren                          space-in-parens
  1906:38  error    There must be a space inside this paren                          space-in-parens
  1906:39  error    Missing semicolon                                                semi
  1910:25  error    Missing semicolon                                                semi
```

classes/views/frm-fields/single-option.php
```diff
@@ -4,7 +4,19 @@
 }
 ?>
 <li id="frm_delete_field_<?php echo esc_attr( $field['id'] . '-' . $opt_key ); ?>_container" data-optkey="<?php echo esc_attr( $opt_key ); ?>" class="frm_single_option <?php echo $opt_key === '000' ? 'frm_hidden frm_option_template' : ''; ?>">
-	<?php FrmAppHelper::icon_by_class( 'frmfont frm_drag_icon frm-drag' ); ?>
+	<?php
+	$show_icons = true;
```
Crabcyborg, line 8: These several lines could be reduced to a simpler one liner. Something like `$show_icons = empty( $field['do_not_include_icons'] );` should cover it.

Crabcyborg:
@garretlaxton I think this is ready enough for testing. The drag/remove icons for options for radios/check boxes/dropdowns will no longer be included on page load and get generated when the field is selected instead. This helps optimize forms with fields that have a lot of options because it only shows the icons for the fields as necessary, instead of loading all of that HTML up front. Let me know if you can find any issues with this optimization!

@AbdiTolesa it looks like this has merge conflicts that will need to be resolved as well.
Screen Shot 2022-08-02 at 1 07 43 PM

garretlaxton:
This definitely performs A LOT better, but I do notice one thing.

If you click a dropdown multiple times, a drag icon is added each time you click the field. Here's a quick video: https://www.loom.com/share/0650d2de5c4b444182028d65b6582836

AbdiTolesa:
> This definitely performs A LOT better, but I do notice one thing.
> 
> If you click a dropdown multiple times, a drag icon is added each time you click the field. Here's a quick video: https://www.loom.com/share/0650d2de5c4b444182028d65b6582836

@garretlaxton Thanks for catching that for us! I have now fixed the issue and you can give it a shot one more time?

garretlaxton:
@AbdiTolesa You're welcome, that looks good now! One last thing I see is when a field is next to another, the drag options are missing for every field but the first field.
Screenshot 2022-08-02 at 15-14-41 Formidable Forms ‹ Formidable Forms — WordPress

AbdiTolesa:
> @AbdiTolesa You're welcome, that looks good now! One last thing I see is when a field is next to another, the drag options are missing for every field but the first field. Screenshot 2022-08-02 at 15-14-41 Formidable Forms ‹ Formidable Forms — WordPress

@garretlaxton Sorry I didn't try that myself and that's why I missed it. That case is handled as well now.

garretlaxton:
@AbdiTolesa Unfortunately, I'm still not seeing the drag icon when they are side by side.
Screenshot 2022-08-02 at 20-23-31 Formidable Forms ‹ Formi QA — WordPress

AbdiTolesa:
> @AbdiTolesa Unfortunately, I'm still not seeing the drag icon when they are side by side. Screenshot 2022-08-02 at 20-23-31 Formidable Forms ‹ Formi QA — WordPress

@garretlaxton I'm wondering why we see different results. Do you see any error in the console by chance? I just created a new form and tried this again but it looks working from my side yet.
image

garretlaxton (approved):
Thanks @AbdiTolesa! This looks great now!

Crabcyborg (approved):
Thanks @AbdiTolesa! I left this for next release just to be on the safe side. I'm excited to get this out soon though! 🎉

AbdiTolesa:
> Thanks @AbdiTolesa! I left this for next release just to be on the safe side. I'm excited to get this out soon though! 🎉

@Crabcyborg Can't wait to see this out as well, ⚐

PR #842, opened 2022-08-02
add missing form strings.

AbdiTolesa:
I'm trying to add the missing translatable form strings, which a form duplication become incomplete without them as their respective translated strings are not copied to the new form.


Crabcyborg (approved):
Thanks @AbdiTolesa!

PR #849, opened 2022-08-04
Check target input id in duplicate option check

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3734


stephywells (approved):
Awesome! This solves the problem for me.

PR #852, opened 2022-08-05
Do not exit early for empty content on denylist check

Crabcyborg:
I mentioned this in Slack. Not sure to it, just removing the content check so it always does the blacklist check on the other fields like IP address.


stephywells (approved):
This looks great! I just double checked the wp_check_comment_disallowed_list() function to make sure this isn't a problem.

PR #853, opened 2022-08-08
Recalculate [auto_id] on entry duplicate.

AbdiTolesa:
Fields with [auto_id] are not currently incrementing when an entry is duplicating, resulting in entries with the same value and that defeats the purpose of having unique values for a given field with the help of the shortcode.

This PR aims to make sure that field values with [auto-id] gets updated when duplicating an entry.

### DIFF
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -672,6 +672,25 @@ class FrmEntriesController {
 	}
 
 	/**
+	 * If a field default value matches [auto_id] patterns, evaluate the shortcode again and assign it to the new field entry meta.
+	 *
+	 * @since x.x
+	 *
+	 * @param array $meta The field entry meta
+	 */
+	public static function autoincrement_on_duplicate( &$meta ) {
+		$field = FrmField::getOne( $meta->field_id );
+		if ( ! preg_match( '/\[auto_id[^\]]*\]/', $field->default_value ) ) {
+			return;
+		}
+
+		$value = $field->default_value;
+
+		FrmProFieldsHelper::replace_non_standard_formidable_shortcodes( array( 'field' => $field ), $value );
+		$meta->meta_value = $value;
+	}
+
+	/**
 	 * @deprecated 4.0
 	 */
 	public static function contextual_help( $help, $screen_id, $screen ) {
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -672,25 +672,6 @@ class FrmEntriesController {
 	}
 
 	/**
-	 * If a field default value matches [auto_id] patterns, evaluate the shortcode again and assign it to the new field entry meta.
-	 *
-	 * @since x.x
-	 *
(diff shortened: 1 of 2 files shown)


classes/controllers/FrmEntriesController.php
```diff
@@ -671,6 +671,25 @@ public static function entry_sidebar( $entry = false ) {
...
+	/**
+	 * If a field default value matches [auto_id] patterns, evaluate the shortcode again and assign it to the new field entry meta.
+	 *
+	 * @since x.x
+	 *
+	 * @param array $meta The field entry meta
+	 */
+	public static function autoincrement_on_duplicate( &$meta ) {
+		$field = FrmField::getOne( $meta->field_id );
+		if ( ! preg_match( '/\[auto_id[^\]]*\]/', $field->default_value ) ) {
```
Crabcyborg, line 683: Thanks for working on this @AbdiTolesa!

I think we can update the lite add on but we shouldn't add direct references to anything specific to Pro. In this case, I don't want `auto_id` to be in the lite add on, but a new hook that we can use in Pro would be better.

I see we have a similar hook for after a form is duplicated here https://github.com/Strategy11/formidable-forms/blob/7befe99471b3edd04365250439b24e5b0d1894f3/classes/models/FrmForm.php#L138

But we're probably better hook into before the loop after `$metas = self::get_entry_meta_info( $old_id );` and adding a new `frm_before_duplicate_entry_values` instead.

classes/controllers/FrmEntriesController.php
```diff
@@ -671,6 +671,25 @@ public static function entry_sidebar( $entry = false ) {
...
+			return;
+		}
+
+		$value = $field->default_value;
+
+		FrmProFieldsHelper::replace_non_standard_formidable_shortcodes( array( 'field' => $field ), $value );
```
Crabcyborg, line 689: Another reason why this shouldn't be in lite is because it's calling Pro functions.

classes/controllers/FrmEntriesController.php
AbdiTolesa: @Crabcyborg I think that might be a better approach but how about adding the filter inside the loop to avoid looping over `$metas` twice?

Crabcyborg: I think one hook is probably less overhead than a possiblu being called hundreds of times.

classes/models/FrmEntryMeta.php
```diff
@@ -166,6 +166,7 @@ public static function update_entry_metas( $entry_id, $values ) {
 
 	public static function duplicate_entry_metas( $old_id, $new_id ) {
 		$metas = self::get_entry_meta_info( $old_id );
+		$metas = apply_filters( 'frm_before_duplicate_entry_values', $metas );
```
Crabcyborg, line 169: @AbdiTolesa could you add comments to the filter? It's nice to have more information on these especially since they're customer-facing.

Crabcyborg (approved):
Thanks @AbdiTolesa!

PR #854, opened 2022-08-09
Fix data-toggleclass doesn't work with content from AJAX

truongwp:
To replicate, just add a checkbox with `data-toggleclass` inside a form action.


Crabcyborg (approved):
Thanks @truongwp! Does this fix any current issues or does it just prevent you from doing something new?

Crabcyborg:
Thanks @truongwp sounds good.

PR #855, opened 2022-08-09
Handle image upload clicks for dynamic content in settings

Crabcyborg:


stephywells (approved):
The code looks good! Do you think this needs any testing?

Crabcyborg:
Thanks @stephywells! That's a good point.

@garretlaxton technically you've already looked over this if you tested the quiz outcomes but it might benefit from a bit more specific testing covering what was updated here in lite:

- Do the wysiwyg editors in quiz outcomes work as expected? Are there any issues when you switch between the Visual and Text tabs? Do new actions work the same as actions you've already opened? Are duplicated actions fine (I didn't test that much).
- Are there any issues with limiting quiz types? You should only be allowed one scored quiz, or unlimited quiz outcomes. Not both.
- Is the image upload all good? Deleting an image, etc all works as expected?

Thanks!

garretlaxton:
1. When I duplicate an action this is as much as I can see when I expand it. If you have trouble replicating this, try updating the form then duplicating an action.
localhost_qa-form_wp-admin_admin php_page=formidable frm_action=settings id=230 t=email_settings

2. If you export a quiz that has an image in the action, the image isn't there when you import on a different site although it shows the delete option. If you click delete, then upload a new image the preview doesn't show.
localhost_qa-form_wp-admin_admin php_page=formidable frm_action=settings id=231

3. For some reason, this one image is acting weird when I delete it and replace it with another image. It shows the old image after uploading the new image until you update the form. It's so weird bc it only happens with this image. Here's a video to better understand what I mean: https://www.loom.com/share/6cb8ddcac8094da1ae6b46e025756cd3
Here's said image: 
BD

Crabcyborg:
Thanks @garretlaxton!

> 1. When I duplicate an action this is as much as I can see when I expand it. If you have trouble replicating this, try updating the form then duplicating an action.

Unfortunately I didn't have much luck replicating this. It seems like it's probably more specific to actions though, and not necessarily related to this update, unless it has to do with TinyMCE initializing.

I'm noticing there's an error in Chrome though that I wasn't seeing in Firefox coming from TinyMCE. I think I’ll leave this out of this release and test TInyMCE more when I work on https://github.com/Strategy11/formidable-pro/issues/3737.
```
tinymce.min.js?ver=49110-20201110:2 Uncaught TypeError: Cannot read properties of undefined (reading 'onpageload')
```

>     2. If you export a quiz that has an image in the action, the image isn't there when you import on a different site although it shows the delete option. If you click delete, then upload a new image the preview doesn't show.

Thanks for catching this. I added an extra check now in Quizzes so it won’t treat it as having a value if the value isn’t mapping to an image attachment (https://github.com/Strategy11/formidable-quizzes/pull/133/commits/de0d7adcc3435fd46cf03f5ef7719338aea6b0e1).

>     3. For some reason, this one image is acting weird when I delete it and replace it with another image. It shows the old image after uploading the new image until you update the form. It's so weird bc it only happens with this image.

Thanks for catching that! It looks like there’s a srcset on the image that wasn’t changing from the original value it loads as so it would use old images. I’ve removed the attribute on update and it appears to have fixed it.

garretlaxton:
@Crabcyborg The last thing I'm seeing is imported entries show an ID for the **Outcome** column instead of the actual outcome on the entries page.

The top entry is one I manually created, and the bottom 2 were imported.
localhost_qa-form_wp-admin_admin php_page=formidable-entries frm-full=1 frm_action=list form=242

garretlaxton (approved):
Good to go here then!

Crabcyborg:
> 1. When I duplicate an action this is as much as I can see when I expand it. If you have trouble replicating this, try updating the form then duplicating an action.

@garretlaxton I was able to replicate this in Chrome eventually. It looks like something is forcing 15px height on the widget-inside element. I added a `height:auto` style on it so it forces that off (https://github.com/Strategy11/formidable-quizzes/pull/133/commits/b607d61f7ac70307eb0bb05489996f6a68ae7a75).

I also pushed a few other updates to account. Now when you duplicate an outcome, it will trigger the auto-numbering which it wasn't doing before. It will also fix issues with the wysiwyg not initializing properly after duplicating because IDs were conflicting between the two wysiwygs. I'm fixing this now by removing the TinyMCE before clone from the original action and then adding it back after the duplicate is finished.

These parts could use some additional testing.

Thank you!

garretlaxton:
This looks great now! When you duplicate an outcome now the name shouldn't duplicate as well correct?
localhost_qa-form_wp-admin_admin php_page=formidable frm_action=settings id=254 t=email_settings

Duplicating my **Nope** outcome is named as **Quiz Outcome 3**. It makes sense, but just want to make sure this is intended now.

Crabcyborg:
> This looks great now! When you duplicate an outcome now the name shouldn't duplicate as well correct?

Thanks @garretlaxton. I just pushed an update to the quiz outcomes to only auto-number it now if it's the default. So If it's "Quiz Outcome 1", "Quiz Outcome 2", etc, it will count up, but otherwise it will continue to use the name of the duplicated action.

garretlaxton:
One thing I see with the outcome names is if you add a new outcome, the count will start based on the number of actions.
Here I have 5 actions and when I add another outcome, the count starts at `5`.
localhost_qa-form_wp-admin_admin php_page=formidable frm_action=settings id=254
If I have 3 actions, the count will start at `3`.
localhost_qa-form_wp-admin_admin php_page=formidable frm_action=settings id=254 t=email_settings (1)


By HTML field description I guess you are referring to the content section? The wysiwyg works good with that still on front-end/back-end.
localhost_qa-form_wp-admin_admin php_page=formidable frm_action=edit id=254

Crabcyborg:
> One thing I see with the outcome names is if you add a new outcome, the count will start based on the number of actions.

@garretlaxton How do you think the counting should work? I figured just `"Quiz Outcome " + # of outcomes` was suitable. If you have 4 other outcomes, technically it's the 5th one. Any other number would seem inaccurate to me. I'm mostly just trying to avoid them all just being called "Quiz Outcome" if someone doesn't rename them all.

Maybe the numbering should only happen if the others are all numbered as well?

> By HTML field description I guess you are referring to the content section? The wysiwyg works good with that still on front-end/back-end.

Yes, exactly that. The code for handling that is shared now, cleaning it up quite a bit. I didn't see any issues, just wanted to get some extra testing on it. Thanks!

PR #860, opened 2022-08-15
Add extra checks for data in actions

Crabcyborg:
Referencing a comment from Garret https://github.com/Strategy11/formidable-quizzes/pull/133#issuecomment-1212400320

He mentioned an issue.
```
PHP Notice: Trying to get property 'post_status' of non-object in wp-content\plugins\formidable\classes\models\FrmFormAction.php on line 425
```

I added a unit test to replicate the issue. I'm not sure how to get it to happen on the front end exactly. I'm just adding a few extra type checks.

The problem is really that there's a line in the code here.

`$old_instance = isset( $all_instances[ $number ] ) ? $all_instances[ $number ] : array();`.

That if `$all_instances[ $number ]` is not set will all back to an array rather than an object. But the status update function is assuming that `$old_instance` is never an array.

The unit test catches this. Though in PHPUnit the error I see instead is "Attempt to read property "post_status" on array".

I also added a check for the `classes` key in the action option. I was seeing `Undefined array key "classes"` as well. The variable tries to set an `array()` default so this is a bug as well.

### DIFF
--- a/classes/models/FrmFormAction.php
+++ b/classes/models/FrmFormAction.php
@@ -106,7 +106,7 @@ class FrmFormAction {
 			$action_options['color'] = 'var(--' . reset( $colors ) . ')';
 		}
 
-		$upgrade_class = $action_options['classes'] === 'frm_show_upgrade';
+		$upgrade_class = isset( $action_options['classes'] ) && $action_options['classes'] === 'frm_show_upgrade';
 		if ( $action_options['group'] === $id_base ) {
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
I'm not sure how I reproduced this myself other than enabling JS validation. I just enabled JS validation and updated the form and saw the notice. I don't see this anymore though!

stephywells (approved):
Looks like a good solution

PR #862, opened 2022-08-16
Fix issues with landing page upgrade tab message not reflecting license

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3755

Not all of the oneclick logic was being carried over. Add on statuses were missing and not getting updated, etc.

I removed all of the ID-specific stuff and made everything work with classes and allowing for multiple instances of things. The upgrade tabs still borrow from the upgrade modal but will use their own elements after the elements are initially cloned now.

More code is being re-used now. And there's a bit less jQuery and a bit more ES6 than before as well.

Screen Shot 2022-08-16 at 12 11 46 PM
Screen Shot 2022-08-16 at 12 32 06 PM

### DIFF
--- a/classes/views/shared/upgrade_overlay.php
+++ b/classes/views/shared/upgrade_overlay.php
@@ -25,12 +25,12 @@ if ( ! defined( 'ABSPATH' ) ) {
 				</h2>
 				<div class="cta-inside">
 
-					<p id="frm-oneclick" class="frm_hidden">
+					<p class="frm-oneclick frm_hidden">
 						<?php esc_html_e( 'That add-on is not installed. Would you like to install it now?', 'formidable' ); ?>
 					</p>
-					<p id="frm-addon-status"></p>
+					<p class="frm-addon-status"></p>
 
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
This looks pretty solid. Only suggestion I would make is when the add-on is installed but not activated we could change it to say active instead of install.
localhost_qa-form_wp-admin_admin php_page=formidable frm_action=settings id=233

Crabcyborg:
Thanks @garretlaxton good point! I pushed an update for that.

Screen Shot 2022-08-17 at 11 50 44 AM

I also noticed that I had left the "Close" button for the pop up on the tab page beside the Save and Reload option which doesn't really make sense 😅. So I removed that too for the tab, but kept it for the pop up.

Screen Shot 2022-08-17 at 11 51 58 AM

garretlaxton (approved):
Shweet! Looking good!
localhost_qa-form_wp-admin_admin php_page=formidable frm_action=settings id=233 (1)

stephywells:
Sorry I wasn't able to look at this sooner! The code looks good. Thanks so much for cleaning this up!

PR #868, opened 2022-08-18
Update post parent id if imported page id parent matches

Crabcyborg:
Fixes an issue mentioned in https://secure.helpscout.net/conversation/1983340821/106865/

> When pages are imported as part of an application they do not remember their parent page when it's brought across in the same application.


garretlaxton (approved):
This looks good! Importing applications with pages that have a parent page now import in the same order.
localhost_qa-form_wp-admin_edit php_post_type=page

stephywells (approved):
Looks great!

PR #870, opened 2022-08-23
Get meta with entry in FrmEntry::destroy function

Crabcyborg:
Fixes https://secure.helpscout.net/conversation/1987233335/107342/

Deleted actions do not trigger conditional logic if the conditions are based off of item meta because the item meta is missing from the object.

### DIFF
--- a/classes/models/FrmEntry.php
+++ b/classes/models/FrmEntry.php
@@ -261,7 +261,7 @@ class FrmEntry {
 		global $wpdb;
 		$id = (int) $id;
 
-		$entry = self::getOne( $id );
+		$entry = self::getOne( $id, true ); // Item meta is required for conditional logic in actions with 'delete' events.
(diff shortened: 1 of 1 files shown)


garretlaxton:
@Crabcyborg I'm not sure what a form action with a delete event is. Any pointers?

Crabcyborg:
@garretlaxton I'm just referring to this "Trigger this action when" "Entry is deleted" setting. Sorry that wasn't more clear!

Screen Shot 2022-08-23 at 5 01 52 PM

garretlaxton (approved):
Thanks Mike!

This looks good. I tested this from all the places you can delete an entry and I received the email when the condition was met.
mail google com_mail_u_0_

stephywells (approved):
Nice, simple fix!

PR #871, opened 2022-08-24
Avoid empty error pop up when aborting jquery

Crabcyborg:
This pop up happens if you try to insert a field and then click save and try to leave the page before the field has inserted. It triggers the AJAX error function as it's aborted. Checking for a status code of 0 is the closest I could find to checking for that.

It's the third answer here but I don't think either of the first two are perfect either. The biggest criticism it has is that it will catch some other types of errors as well.
https://stackoverflow.com/questions/4807572/jquery-ajax-error-handling-to-ignore-aborted#answer-5958734

image

I tested the pop up with an error by failing the nonce check. It still works great. And the summary field re-enables fine. However I'm seeing other issues that can lead to 2 summary fields that I plan to look into as well.

Screen Shot 2022-08-24 at 2 10 43 PM

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1485,8 +1485,8 @@ function frmAdminBuildJS() {
 				}
 				toggleSectionHolder();
 			},
-			error: function( jqXHR, textStatus, errorThrown ) {
-				maybeReenableSummaryBtnAfterAJAX( fieldType, addBtn, fieldButton, errorThrown );
+			error: function( jqXHR, _, errorThrown ) {
+				maybeReenableSummaryBtnAfterAJAX( fieldType, addBtn, fieldButton, errorThrown, jqXHR );
 			}
 		});
 	}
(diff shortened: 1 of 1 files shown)


garretlaxton:
@Crabcyborg To replicate this, does this need to happen fast? I can't seem to insert a field, click update, & leave the page before the field is inserted. Not sure if I'm doing this wrong, or just too slow haha.

Crabcyborg:
@garretlaxton Yeah exactly. It might not be easy to test. I've been using Local recently which is notoriously slow 😅.

Slowing down the server response could help. You could add a sleep() call in the frm_insert_field endpoint or something to slow it down artificially.

garretlaxton (approved):
Thanks! I found a super easy way to slow down speeds in the chrome network tab.
H3qPt
localhost_qa-form_wp-admin_admin php_page=formidable frm_action=edit id=269

I no longer see the popup when this happens with these changes.

stephywells (approved):
This looks good. I mentioned this in Slack just now, but I think the summary code causing the problem here is old and should be removed after all.

PR #873, opened 2022-08-25
add textarea height to popup options top offset.

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3757
This is just a quick fix to make sure the advanced options box doesn't overlap with the textarea content, adding the textarea height to the popup's top offset.

Before:
image

With this update:
image

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -8558,8 +8558,8 @@ Responsive Design
 }
 
 .frm-custom-field-group-layout:hover,
-.frm-break-field-group:hover, .tracking-date {
-	background-color: #f5f5f5;
+.frm-break-field-group:hover {
+	background-color: #f5f5f5f5;
 }
 
 .frm-custom-field-group-layout svg,
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -777,9 +777,13 @@ p.frm_has_shortcodes {
 
 .frm-nav-tabs a {
 	color: var(--dark-grey);
-	padding: 5px 1px;
+	padding: 10px 1px;
 	margin: 0 9px;
 	border-bottom: 2px solid transparent;
+}
+
+.frm_form_settings .frm-nav-tabs a {
+	padding-top: 5px 1px;
 }
 
 .frm-nav-tabs.frm-compact-nav a {
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -652,6 +652,9 @@ p.frm_has_shortcodes {
 	min-height: 200px;
 	overflow-y: scroll;
 	z-index: 100;
+	box-shadow: 0px 12px 24px rgb(72 78 84 / 20%);
+	border-radius: 4px;
+	border: none;
 }
 
 #form_global_settings .columns-2 .frm-right-panel,
@@ -745,7 +748,7 @@ p.frm_has_shortcodes {
 	position: -webkit-sticky;
 	position: sticky;
 	top: 0;
-	background-color: var(--sidebar-color);
+	background-color: #fff;
 	z-index: 98; /* must be < 99 */
 	margin-top: 0;
 	padding-top: 5px;
@@ -769,7 +772,7 @@ p.frm_has_shortcodes {
 
 .frm-nav-tabs a {
 	color: var(--dark-grey);
-	padding: 10px 1px;
+	padding: 5px 1px;
 	margin: 0 9px;
 	border-bottom: 2px solid transparent;
 }
@@ -784,13 +787,14 @@ p.frm_has_shortcodes {
 }
 
 #frm_adv_info .frm-nav-tabs a {
-	color: var(--dark-grey);
+	color: rgba(40, 47, 54, 0.45);
 }
 
 .frm-nav-tabs a:hover,
 .frm-nav-tabs .frm-tabs a {
 	color: var(--primary-color) !important;
 	border-color: var(--primary-color);
+	font-weight: 600;
 }
 
 #frm-bulk-modal .howto,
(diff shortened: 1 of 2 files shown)


css/frm_admin.css
```diff
@@ -652,6 +652,9 @@ p.frm_has_shortcodes {
 	min-height: 200px;
 	overflow-y: scroll;
 	z-index: 100;
+	box-shadow: 0px 12px 24px rgb(72 78 84 / 20%);
```
Crabcyborg, line 655: @AbdiTolesa Is this a valid syntax for color? It looks a bit odd.

We usually use rgba and commas, no slashes. More like `box-shadow: 0px 8px 24px rgba(40, 47, 54, 0.15);`.

css/frm_admin.css
```diff
@@ -774,6 +782,10 @@ p.frm_has_shortcodes {
 	border-bottom: 2px solid transparent;
 }
 
+.frm_form_settings #frm_adv_info .frm-nav-tabs a {
+	padding-top: 5px 1px;
```
Crabcyborg, line 786: @AbdiTolesa Is this supposed to be just `padding` or something? I don't think you can have 2 `padding-top` values.

css/frm_admin.css
AbdiTolesa: @Crabcyborg Yes it is a valid color in different syntax. I have changed to the one you mentioned here just for consistency though.

css/frm_admin.css
AbdiTolesa: @Crabcyborg Oh! It was supposed to be `padding`, thanks for catching that!

Crabcyborg:
Thanks @AbdiTolesa for updating that!

I'm working on trying to change this textarea to a TinyMCE wysiwyg now instead. I have a draft pull request already https://github.com/Strategy11/formidable-forms/pull/885.

I played with trying out this update with my update. I have a separate branch for that right now https://github.com/Strategy11/formidable-forms/compare/tinymce_email_action_plus_issue-3757?expand=1

Unfortunately the new textarea breaks this pop up quite a bit.
Screen Shot 2022-09-05 at 1 24 04 PM

Could you see if we can get this working with TinyMCE now?

Thank you!

AbdiTolesa:
> Thanks @AbdiTolesa for updating that!
> 
> I'm working on trying to change this textarea to a TinyMCE wysiwyg now instead. I have a draft pull request already #885.
> 
> I played with trying out this update with my update. I have a separate branch for that right now https://github.com/Strategy11/formidable-forms/compare/tinymce_email_action_plus_issue-3757?expand=1
> 
> Unfortunately the new textarea breaks this pop up quite a bit. I see it off to the left. Screen Shot 2022-09-05 at 1 24 04 PM
> 
> Could you see if we can get this working with TinyMCE now?
> 
> Thank you!

@Crabcyborg I tried to pull the branch you created from both updates and checked out the email action but it looks to be on position for me. Is your update just for the email actions but not for form messages yet though?

image

Crabcyborg:
> I tried to pull the branch you created from both updates and checked out the email action but it looks to be on position for me.

Thanks @AbdiTolesa. It looks like it works differently in Firefox and in Chrome. But i can replicate this in Chrome too as soon as I switch tabs.

Here's a recording.
DZdd8QqsDb

> Is your update just for the email actions but not for form messages yet though?

Yes, this isn't a universal update for everything so this update might still also apply to textareas too, not just the new wysiwyg.

AbdiTolesa:
> > I tried to pull the branch you created from both updates and checked out the email action but it looks to be on position for me.
> 
> Thanks @AbdiTolesa. It looks like it works differently in Firefox and in Chrome. But i can replicate this in Chrome too as soon as I switch tabs.
> 
> Here's a recording. DZdd8QqsDb [ DZdd8QqsDb ](https://user-images.githubusercontent.com/9134515/188496339-4ceb486f-b3d4-4b88-a430-5137ec66c66a.gif) [ ](https://user-images.githubusercontent.com/9134515/188496339-4ceb486f-b3d4-4b88-a430-5137ec66c66a.gif)
> 
> > Is your update just for the email actions but not for form messages yet though?
> 
> Yes, this isn't a universal update for everything so this update might still also apply to textareas too, not just the new wysiwyg.

@Crabcyborg I was able to replicate this switching between the tabs. Does your update seem to work with master without this issue? I sort of tried to use `frm_admin.css` from the master branch and it looks the issue happens with that as well but I'm still trying to look into the root cause.

Crabcyborg:
@AbdiTolesa Your update doesn't break it, this is an issue on master.

But it needs to work with this new type as well so if you could help figure out a fix that would be great.

AbdiTolesa:
> @AbdiTolesa Your update doesn't break it, this is an issue on master.
> 
> But it needs to work with this new type as well so if you could help figure out a fix that would be great.

@Crabcyborg Surely, I would like to help on that as well.

AbdiTolesa:
> > @AbdiTolesa Your update doesn't break it, this is an issue on master.
> > But it needs to work with this new type as well so if you could help figure out a fix that would be great.
> 
> @Crabcyborg Surely, I would like to help on that as well.

@Crabcyborg This issue seems to be addressed by reimplementing `getIconForInput` and `getInputForIcon` functions using recursion.

Crabcyborg:
Thanks @AbdiTolesa that looks better in Chrome but I'm still seeing it in the same place in Firefox. Do you think you could try testing it in Firefox too and seeing if you can get it working in multiple browsers?

Crabcyborg:
Thanks @AbdiTolesa this looks good!

I'm going to close this pull request and have Garret test both of those updates together in a new pull request https://github.com/Strategy11/formidable-forms/pull/895

Thank you!

PR #875, opened 2022-08-26
Update deprecated code

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3788

That code is deprecated in v3.5, the current version of Elementor is 3.7.2 so I think we don't need to check if new code exists.


Crabcyborg (approved):
Thanks @truongwp!

PR #879, opened 2022-08-30
User tracking rock update

AbdiTolesa:
This update is complementary to the rock update done to our User Tracking plugin. The view part is moved away to the respective plugin. There are also a couple of css update required for the frontend in this update.


classes/views/frm-entries/sidebar-shared.php
```diff
@@ -69,76 +69,3 @@
 		<?php FrmEntriesHelper::maybe_render_captcha_score( $entry->id ); ?>
 	</div>
 </div>
-
-<div class="frm_with_icons">
```
Crabcyborg, line 73: @AbdiTolesa I think we want to keep the sidebar User Information in lite. Maybe we want to hide this when the User Tracking is on, but I don't think we want to get rid of all of this.

css/frm_admin.css
```diff
@@ -3678,10 +3678,28 @@ label input[type="checkbox"], label input[type="radio"] {
 	padding:20px;
 }
 
-.frm-alt-table tr:nth-child(even) {
+.frm-alt-table tr:nth-child(even), .frm-usr-trk-date-row {
 	background-color: var(--sidebar-color);
 }
 
+.frm-usr-trk-date-row {
```
Crabcyborg, line 3685: @AbdiTolesa If all of this is required only for the Tracking add on it would be better if we just loaded a new CSS file from that add on instead of adding stuff we don't need in lite.

classes/views/frm-entries/sidebar-shared.php
AbdiTolesa: @Crabcyborg One of the improvements I thought are worth is making sure that the user tracking info is displayed only if a plugin handling that is turned on as I we discussed in the slack channel last time so that it is now controllable whether to display such data or not. Wouldn't that make sense?

Crabcyborg: @AbdiTolesa All of this information is tracked in lite even if you don't have tracking on though, right? It's just basic information about the entry.

The user tracking is more about session tracking. This is all just information like User, IP, and Browser.

Crabcyborg: Thanks @AbdiTolesa, it looks like that's because it shows the referer data and the user tracking uses referer as well.

Maybe if the user tracking add on is active and up to date, we can skip the referer piece here. But otherwise keep that so we're not removing lite functionality either.

AbdiTolesa: @Crabcyborg The user tracking plugin uses a different key ( 'user_journey' ) to store the pages info. But checking the user tracking add on and its version would be available only with the new lite version to be released yet and this approach wouldn't work for the current lite versions out there because if only the add on is updated, that would lead to displaying the same view both in the sidebar (due to old lite) and in the main area ( by the updated add on ).

css/frm_admin.css
AbdiTolesa: @Crabcyborg Got it, will create a css file for this.

classes/views/frm-entries/sidebar-shared.php
Crabcyborg: @AbdiTolesa It should be pretty simple to confirm the versions are right. Something like `class_exists( 'FrmUsrTrkAppHelper' )` should only be true on versions that support the User Journey. If that passes, we can avoid showing those pieces in lite.

AbdiTolesa: @Crabcyborg In the later case, of course we can hide the add on view if the lite is old so that the data will be displayed in the sidebar but the problem is for new entries created with this updated version of the add on, the old lite doesn't know how to parse and display it as the data is array now.

Crabcyborg: @AbdiTolesa Is there a way to avoid this issue? Maybe it's best if only started using the new format if Lite is already up to date. That way it would only be an issue if you reverted, which isn't as likely.

AbdiTolesa: @Crabcyborg I think the only way to avoid this issue is adding snippets for saving the pages data in the old format if Lite is not updated. I don't see any other way to turn this around.

Crabcyborg: I think just showing a compatibility message would do. If the version of lite it out of date, would it just print it as serialized?

Can we switch to JSON? We're trying to avoid serializing where possible. 

We have this unserialize_or_decode function that helps with backward compatibility.
https://github.com/Strategy11/formidable-forms/blob/54f2e1079dce57d5a39133230b3d954ba08403ea/classes/helpers/FrmAppHelper.php#L2645

Crabcyborg: @AbdiTolesa Yeah, I tried it out and I got it working.

Here's my pull request https://github.com/Strategy11/formidable-user-tracking/pull/8

The `$entry` object gets passed to a `frm_show_entry_sidebar` action before this, so modifying the description isn't too difficult.

AbdiTolesa: @Crabcyborg Thank you for your time on this and my apologies for not getting back to update this before you did. I jumped off to the other issues I'm working on in the middle.

classes/views/frm-entries/sidebar-shared.php
```diff
@@ -69,76 +69,14 @@
...
-	<h3><?php esc_html_e( 'User Information', 'formidable' ); ?></h3>
-	<div class="inside">
-		<?php if ( $entry->user_id ) { ?>
-			<div class="misc-pub-section">
-				<?php
-				FrmAppHelper::icon_by_class( 'frmfont frm_user_icon', array( 'aria-hidden' => 'true' ) );
-
-				printf(
-					/* translators: %1$s: User display name. */
-					esc_html__( 'Created by: %1$s', 'formidable' ),
```
Crabcyborg, line 83: @AbdiTolesa I still think you're getting rid of way too much from here.

Created by, Updated by, IP, Browser are all useful and should be available with just Lite installed.

AbdiTolesa: @Crabcyborg right, user info should always show with lite. Just updated that accordingly, thanks!

Crabcyborg:
@AbdiTolesa Is it fine to just close this pull request? I'm not sure if it's necessary. I just tried out the user tracking add on without it and it looked good to me.

AbdiTolesa:
> @AbdiTolesa Is it fine to just close this pull request? I'm not sure if it's necessary. I just tried out the user tracking add on without it and it looked good to me.

@Crabcyborg I think so as there were some changes made to the plan, this seems not required anymore.

PR #880, opened 2022-08-31
make sure autocomplete option has priority over input shortcode.

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3780

This update avoids having double autocomplete html attributes in a field by making sure that the option set in the field settings override the input shortcode html.

### DIFF
--- a/classes/controllers/FrmFieldsController.php
+++ b/classes/controllers/FrmFieldsController.php
@@ -769,7 +769,7 @@ class FrmFieldsController {
 		}
 
 		foreach ( $field['shortcodes'] as $k => $v ) {
-			if ( 'opt' === $k || ( ! empty( $field['autocomplete'] && $k === 'autocomplete' ) ) ) {
+			if ( 'opt' === $k || ( ! empty( $field['autocomplete'] ) && $k === 'autocomplete' ) ) {
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


classes/controllers/FrmFieldsController.php
```diff
@@ -769,7 +769,7 @@ private static function add_shortcodes_to_html( $field, array &$add_html ) {
 		}
 
 		foreach ( $field['shortcodes'] as $k => $v ) {
-			if ( 'opt' === $k ) {
+			if ( 'opt' === $k || ( ! empty( $field['autocomplete'] ) && $k === 'autocomplete' ) ) {
```
Crabcyborg, line 772: @AbdiTolesa I think this could be added in a more readable place, and possibly doesn't need to be in the loop at all.

Is it safe to modify `$field['shortcodes']`? It might make sense to just `unset( $field['shortcodes']['autocomplete'] )` if `! empty( $field['autocomplete'] )` before the loop?

AbdiTolesa: @Crabcyborg Thanks and I just updated it accordingly.

Crabcyborg (approved):
Thanks @AbdiTolesa!

garretlaxton (approved):
This looks solid! Checked with a few different settings.
input
html

PR #883, opened 2022-09-01
Vertically center background images in formidable dropdowns

Crabcyborg:
This fixes a conflict with Astra.

I noticed dropdowns all look weird like this in Astra. It's easy to fix though by adding a `background-position-y: center` style to a select.

**Before**
Screen Shot 2022-09-01 at 10 29 02 AM

**After**
Screen Shot 2022-09-01 at 10 27 38 AM


stephywells (approved):
I can't think of any issues this would cause. Looks good!

PR #884, opened 2022-09-02
Add form name shortcode to email actions

Crabcyborg:
I went through the feature requests more today and found this one.

https://team.strategy11.com/feature-requests/2021/04/30/form-name-shortcode-for-emails/?modal=1

> I need to include the form name in the email message admin receives.
> I can't get [form_name] to work.

It has 10 upvotes and it is a pretty small task so I gave it a go.

This update adds support for the `[form_name]` shortcode in email subject and in email message.

**Before**
Screen Shot 2022-09-02 at 3 38 18 PM
Screen Shot 2022-09-02 at 3 38 15 PM

**After**
Screen Shot 2022-09-02 at 3 43 17 PM
Screen Shot 2022-09-02 at 3 43 16 PM

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1442,6 +1442,7 @@ class FrmFormsController {
 			$entry_shortcodes['default-message'] = __( 'Default Msg', 'formidable' );
 			$entry_shortcodes['default-html']    = __( 'Default HTML', 'formidable' );
 			$entry_shortcodes['default-plain']   = __( 'Default Plain', 'formidable' );
+			$entry_shortcodes['form_name']       = __( 'Form Name', 'formidable' );
 		}
 
 		/**
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1530,7 +1530,7 @@ class FrmFormsController {
 	 * @param stdClass|string|int $form
 	 * @return string
 	 */
-	private static function replace_form_name_shortcodes( $string, $form ) {
+	public static function replace_form_name_shortcodes( $string, $form ) {
 		if ( false === strpos( $string, '[form_name]' ) ) {
 			return $string;
 		}
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1497,11 +1497,19 @@ class FrmFormsController {
 		wp_die();
 	}
 
+	/**
+	 * @param string                    $content
+	 * @param stdClass|string|int       $form
+	 * @param stdClass|string|int|false $entry
+	 * @return string
+	 */
 	public static function filter_content( $content, $form, $entry = false ) {
 		self::get_entry_by_param( $entry );
 		if ( ! $entry ) {
 			return $content;
 		}
+
+		$content = self::replace_form_name_shortcodes( $content, $form );
 
 		if ( is_object( $form ) ) {
 			$form = $form->id;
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
This looks good! I tested with multiple forms and duplicate actions.
 Before:
mail google com_mail_u_0_ (2)

After:
mail google com_mail_u_0_ (1)

garretlaxton:
Actually, one question... Would it be worth supporting this in the From field?
formi grunge us eu org_wp-admin_admin php_page=formidable frm_action=settings id=161 t=email_settings
mail google com_mail_u_0_ (3)

Crabcyborg:
Thanks Garret! Maybe, I don't see a harm in it. What do you think @stephywells?

stephywells:
@Crabcyborg I don't see any reason not to include it there. There have been times I've wanted to be able to use it in the after html box too and in the success message. It could be nice as a general helper in the views too.

Crabcyborg:
> I don't see any reason not to include it there.

Thanks @stephywells I've pushed an update to include it there as well!

> There have been times I've wanted to be able to use it in the after html box too and in the success message. It could be nice as a general helper in the views too.

Noted. I'll see about adding this in a few more places!

garretlaxton (approved):
Took a quick look over this, looks good!

stephywells (commented):
This looks good! Is this included in the UI at all so people know it's available?

Crabcyborg:
> This looks good! Is this included in the UI at all so people know it's available?

Thanks @stephywells I did that in views and didn't think about it here. I pushed an update for that now.

Screen Shot 2022-09-07 at 9 48 46 AM

Since this pop up is used in a few places I also added support for this update in the success message so the shortcode appearing in the pop up wouldn't be an issue!

classes/helpers/FrmFormsHelper.php
```diff
@@ -250,14 +250,37 @@ public static function get_invalid_error_message( $args ) {
...
+	 *     @type string   $message
+	 *     @type stdClass $form
+	 *     @type int      $entry_id
+	 *     @type string   $class
+	 * }
+	 * @return string
+	 */
 	public static function get_success_message( $atts ) {
-		$message = apply_filters( 'frm_content', $atts['message'], $atts['form'], $atts['entry_id'] );
+		$message = self::replace_form_name_shortcodes( $atts['message'], $atts['form'] );
```
stephywells, line 263: It would be great if 'frm_content' would filter out the form name instead of focusing on specific spots. Can we do this inside of FrmFormsController::filter_content()? That might cover the View filtering as well.

Crabcyborg: @stephywells I was considering that but I didn't like how many extra calls it would require to get the form object. It was my first choice, but I think it would be slower as most use cases already have access to the form.

See this line https://github.com/Strategy11/formidable-forms/blob/81090d2c9c493570143f4210385cbcda22e57f57/classes/controllers/FrmFormsController.php#L1506

`$form` is a mixed variable type so I'd have to get a `FrmForm::getOne` call into the middle of `FrmFormsController::filter_content` which gets a lot of traffic.

stephywells: Yeah, that doesn't sound very optimal... What if we check for form_name right before that line? That way, it could all be in one place without extra overhead. Also, we can check if '[form_name]' exists before getting the form if we don't already have the object.

Crabcyborg: Thanks @stephywells I pushed a commit that does it that way now instead. It looks like it covered all of those other cases, and likely several others as well this way, so it cleaned up the other code for sure.

Crabcyborg:
@garretlaxton This could use some additional testing as the scope has changed a bit.

Now this should apply to most Formidable content that gets filtered including the success message which also includes quiz outcomes.

I still need to revisit the updates for views but that likely mostly works in master already now.

classes/helpers/FrmFormsHelper.php
stephywells: Thank you! That covers it for the PDF file name update I was working on too!

classes/controllers/FrmFormsController.php
```diff
@@ -1512,6 +1521,31 @@ public static function filter_content( $content, $form, $entry = false ) {
...
+	public static function replace_form_name_shortcodes( $string, $form ) {
+		if ( false === strpos( $string, '[form_name]' ) ) {
+			return $string;
+		}
+
+		if ( ! is_object( $form ) ) {
+			$form = FrmForm::getOne( $form );
+		}
+
+		return str_replace( '[form_name]', $form->name, $string );
```
stephywells, line 1542: Can we add a check here and replace with an empty string if ! $form?

Crabcyborg: Sounds good! I added an extra check.

stephywells: Perfect!

stephywells (approved):

garretlaxton (approved):
This looks good still!

Success message:
success message

Outcomes:
outcomes

I also saw this in the advanced UI tab now, but couldn't screenshot it without it closing on me haha.

PR #886, opened 2022-09-06
Check if array keys exist in init_user_info function

Crabcyborg:
I noticed these two warnings when I viewed an entry that I created from a Post. Because it doesn't get submitted normally, the description is just "Copied from Post" and these array keys don't exist.

Screen Shot 2022-09-06 at 11 25 39 AM


stephywells (approved):
Nice find!

PR #887, opened 2022-09-06
Treat single field in group as selected group for multiselect

Crabcyborg:
This is just a small enhancement for side-by-side fields.

The way the multi-select works when you're holding the shift key, it would only consider the field group as selected if the actual field group was. Since a single field fills in the whole group it isn't very easy to click on the field's group without clicking on the field.

This means that when you shift+click it would just work like the `Cmd`/`Ctrl` key and just select another field group without using the shift rule of selecting all groups in between as well.

**Before**
ROIAU5cVdr

**After**
yTDLS2BO2e

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -3831,28 +3831,27 @@ function frmAdminBuildJS() {
 	}
 
 	function fieldGroupClick( e ) {
-		var hoverTarget, ctrlOrCmdKeyIsDown, shiftKeyIsDown, groupIsActive, $selectedFieldGroups, numberOfSelectedGroups, selectedField, $firstGroup, $range;
-
 		if ( 'ul' !== e.originalEvent.target.nodeName.toLowerCase() ) {
 			// only continue if the group itself was clicked / ignore when a field is clicked.
 			return;
 		}
 
-		hoverTarget = document.querySelector( '.frm-field-group-hover-target' );
-		if ( hoverTarget === null ) {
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
Woah, I never knew about this. The new changes look good though!
localhost_qa-form_wp-admin_admin php_page=formidable frm_action=edit id=284

Crabcyborg:
Thanks @garretlaxton! There's an open issue about letting people know more about this feature (https://github.com/Strategy11/formidable-pro/issues/3138). I plan to look into this soon!

stephywells (approved):
That's a cool feature! The education would be cool for this too.

I've still been having trouble with getting consistent results with moving fields inside a section. If you're looking for something to work on right now, I'd love a revisit to the dragging and dropping in rows.
https://github.com/Strategy11/formidable-pro/issues/3118
https://github.com/Strategy11/formidable-pro/issues/3175

stephywells:
@Crabcyborg I'm fine with whatever you think is best. How much of a time commitment do you think that would be?

Crabcyborg:
@stephywells It might not be too much. I think I'll need to give it a go before I really know what weird hurdles there may be.

But I think a lot of my struggles have been because of jQuery's auto-sorting not doing what makes sense, so having it out of the way might really make things easier if anything.

I'm guessing it might be about a day or two if it goes well.

stephywells:
Awesome!! Then let's try it!

PR #888, opened 2022-09-07
New function FrmAppHelper::get_plugins that adds missing file

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3802

This exact issue is fixed all over the place. I added a new helper function that handles this check to reduce the copied/pasted checks we have all over.


stephywells (approved):
This is SO much better!

PR #889, opened 2022-09-07
Issue 199

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-registration/issues/199

This update just calls a function that initializes autocomplete on selected elements.

### DIFF
--- a/js/formidable_admin_global.js
+++ b/js/formidable_admin_global.js
@@ -1,4 +1,4 @@
-/*global jQuery:false, frmGlobal, tb_remove, ajaxurl, adminpage */
+/*global jQuery:false, frmGlobal, tb_remove, ajaxurl, adminpage, frmDom */
 
 /* exported frm_install_now, frmSelectSubnav, frmCreatePostEntry */
 
--- a/js/formidable_admin_global.js
+++ b/js/formidable_admin_global.js
@@ -27,6 +27,8 @@ jQuery( document ).ready( function() {
 			li.classList.add( 'frm-submenu-highlight' );
 		}
 	}
+
+	frmDom.autocomplete.initSelectionAutocomplete();
 });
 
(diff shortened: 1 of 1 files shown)


js/formidable_admin_global.js
```diff
@@ -27,6 +27,8 @@ jQuery( document ).ready( function() {
 			li.classList.add( 'frm-submenu-highlight' );
 		}
 	}
+
+	frmDom.autocomplete.initSelectionAutocomplete();
```
Crabcyborg, line 31: @AbdiTolesa This definitely isn't the right place to add this. We want to initialize this only where we need it.

Right now it happens in lite on a line like this:
https://github.com/Strategy11/formidable-forms/blob/52edb689f48728af36956db3a257dc4a506c605c/js/formidable_admin.js#L9218

So it only happens on the else case, which should mostly only fall on the "New form selection page" (the index page when click Formidable, the main list of forms).

We can also reduce what gets called and only call `frmDom.autocomplete.initAutocomplete( 'page', inside );` as we don't need the user autocomplete in this case, only the page autocomplete.

Ideally we could get this line into the registration plugin as it isn't really needed in Lite. Since it's just one line maybe we could use https://developer.wordpress.org/reference/functions/wp_add_inline_script/?

AbdiTolesa: @Crabcyborg Thank you for showing me a better way to do that! I have pushed an update so that we can initialize the autocomplete in the add on itself but using  jQuery's ready event as I'm not sure if we can use `DOMContentLoaded` event since it is not supported in IE older than 9.

Crabcyborg: Thanks @AbdiTolesa. As this is only on the admin-side, it's fine if it doesn't support IE at all.

I'm going to close this pull request now as it isn't required.

PR #891, opened 2022-09-08
Add a static installed bool for PHPUnit

Crabcyborg:
- Checks if the install functions have been called yet, and avoid calling them a second time.
- Adds a second check for file path in case the file isn't in the formidable folder but in a formidable/form_id folder instead.
- Only try to create users for the unit tests that require users to pass.

### DIFF
--- a/tests/base/FrmUnitTest.php
+++ b/tests/base/FrmUnitTest.php
@@ -85,16 +85,17 @@ class FrmUnitTest extends WP_UnitTestCase {
 			define( 'WP_IMPORTING', false );
 		}
 
-		if ( ! self::$installed ) {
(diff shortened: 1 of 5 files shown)


stephywells:
Wow! That's way faster!!

Do we still want to take out either 7.2 or 7.3?

Crabcyborg:
> Do we still want to take out either 7.2 or 7.3?

I did that in Pro but I left it here in Lite. Do we have limited minutes in Lite? I wasn't sure if Public repositories might be unlimited or something.

stephywells:
Oh, yep! I didn't realize this was lite. That's no problem then. Thank you for fixing this!!

stephywells (approved):

PR #892, opened 2022-09-09
Replace sortable with draggable

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3175

The drag/drop code has been rewritten to work better with sections and field groups.

This also fixes a bug with deleted groups. When sections were deleted, the children fields weren't also being deleted when using the delete group option.

And it fixes a couple of items from https://github.com/Strategy11/formidable-pro/issues/3118 including:
- "I added a summary field and the grouping that happened is odd. It should wrap the page break and the summary separately."

It also appears to resolve this issue I can appear to replicate on master when I drop a field into a row really quickly. It won't sync my layout but will set the layout class value to a double class like `frm6 frm6`, looking like this as a result:
Screen Shot 2022-09-15 at 4 26 49 PM

**Recording**
QRx9245Dl8

**Completed**
- [x] Do we want class `this.classList.add( 'drop-me' );`?
- [x] A new dragged in field is not draggable.
- [x] Dragging does not work yet between rows.
- [x] Don’t lose any logic from updateFieldAfterMovingBetweenSections
    - [x] And insertNewFieldByDragging
- [x] Do I still need to maintain a $previousFieldContainer variable?
- [x] Only if the previous container had other sibling fields, remove the previous layout class.
- [x] Never place anything after the edit_field_type_end_divider.
- [x] Call allowDrop to make sure that some field types can’t be dragged into a section (see comment don't allow some field types inside section in sortable change function. It uses classes no-drop-placeholder and sortable-placeholder
- [x] Dragging a group into/out of a section should assign the whole group.
- [x] There are issues right now with dragging field groups.
- [x] There's code that tries to update a collapsed page when you drag a field. This doesn't appear to actually happen.
- [x] Page break+summary doesn’t init properly.
- [x] Reduce the negative margin on sections.
- [x] Trying to place a field beside itself back into it's original field group is weird when it in the only item in the group. The placeholder follows the edge of the field as it drags around.
- [x] Break into rows didn’t break into new groups properly
- [x] I can't move around a field within the group it was just in.
- [x] I don’t think sections properly delete in bulk-delete. The child fields seem to get left behind.
- [x] Right now it's calling frm_update_field_after_move too often (after after time a field is dragged). Only call it when section id changes.
- [x] It breaks when you load a long form with AJAX. The loaded items do not become draggable or droppable.
- [x] The field being dragged has an inconsistent z-index.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1497,9 +1497,11 @@ function frmAdminBuildJS() {
 
 				if ( ! $siblings.length ) {
 					makeDroppable( replaceWith.get( 0 ).querySelector( 'ul.frm_sorting' ) );
+					makeDraggable( replaceWith.get( 0 ).querySelector( 'li.form-field' ), '.frm-move' );
+				} else {
+					makeDraggable( replaceWith.get( 0 ), '.frm-move' );
 				}
 
-				makeDraggable( replaceWith.get( 0 ).querySelector( 'li' ), '.frm-move' );
 			},
 			error: handleInsertFieldError
 		});
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1010,7 +1010,13 @@ function frmAdminBuildJS() {
 			if ( insertAtIndex === $children.length ) {
 				const $lastChild = jQuery( $children.get( insertAtIndex - 1 ) );
 				top = $lastChild.offset().top + $lastChild.outerHeight();
-				$list.append( placeholder ); // TODO do not put it after the end divider.
+				$list.append( placeholder );
+
+				// Make sure nothing gets inserted after the end divider.
+				const $endDivider = $list.children( '.edit_field_type_end_divider' );
+				if ( $endDivider.length ) {
+					$list.append( $endDivider );
+				}
 			} else {
 				top = jQuery( $children.get( insertAtIndex ) ).offset().top;
 				jQuery( $children.get( insertAtIndex ) ).before( placeholder );
@@ -1655,7 +1661,6 @@ function frmAdminBuildJS() {
 		return allowMoveFieldToGroup( draggable, droppable );
 	}
 
-	// TODO I cannot drag a field group anywhere.
 	function allowMoveFieldGroup( fieldGroup, droppable ) {
 		if ( droppable.classList.contains( 'start_divider' ) && null === fieldGroup.querySelector( '.start_divider' ) ) {
 			// Allow a field group with no section inside of a section.
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -820,6 +820,11 @@ function frmAdminBuildJS() {
 			list => {
 				makeDroppable( list );
 				Array.from( list.children ).forEach( child => makeDraggable( child, '.frm-move' ) );
+
+				const $sectionTitle = jQuery( list ).children( '[data-type="divider"]' ).children( '.divider_section_only' );
+				if ( $sectionTitle.length ) {
+					makeDroppable( $sectionTitle );
+				}
 			}
 		);
 		setupFieldOptionSorting( jQuery( '#frm_builder_page' ) );
@@ -836,7 +841,7 @@ function frmAdminBuildJS() {
 	}
 
 	function onDragOverDroppable( event, ui ) {
-		const droppable = event.target;
+		const droppable = getDroppableForOnDragOver( event.target );
 		const draggable = ui.draggable[0];
 
 		if ( ! allowDrop( draggable, droppable ) ) {
@@ -848,6 +853,20 @@ function frmAdminBuildJS() {
 		document.querySelectorAll( '.frm-over-droppable' ).forEach( droppable => droppable.classList.remove( 'frm-over-droppable' ) );
 		droppable.classList.add( 'frm-over-droppable' );
 		jQuery( droppable ).parents( 'ul.frm_sorting' ).addClass( 'frm-over-droppable' );
+	}
+
+	/**
+	 * Maybe change the droppable.
+	 * Section titles are made droppable, but are not a list, so we need to change the droppable to the section's list instead.
+	 *
+	 * @param {Element} droppable 
+	 * @returns {Element}
+	 */
+	function getDroppableForOnDragOver( droppable ) {
+		if ( droppable.classList.contains( 'divider_section_only' ) ) {
+			droppable = jQuery( droppable ).nextAll( '.start_divider.frm_sorting' ).get( 0 );
+		}
+		return droppable;
 	}
 
 	function onDraggableLeavesDroppable( event ) {
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1129,11 +1129,22 @@ function frmAdminBuildJS() {
 	}
 
 	function getFieldsInRow( $row ) {
-		return $row.children( 'li.form-field' ).not( '.edit_field_type_end_divider' ).filter(
-			function() {
-				return 'none' !== this.style.display;
+		let $fields = jQuery();
+		Array.from( $row.get( 0 ).children ).forEach(
+			child => {
+				if ( 'none' === child.style.display ) {
+					return;
+				}
+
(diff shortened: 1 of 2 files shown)


Crabcyborg:
@garretlaxton This isn't 100% finished yet but it's at a point where I could use your help testing it.

It's fairly likely there are still errors. I'd like your help tracking those down.

Some items that I know about are in the description.

Thanks!

garretlaxton:
I noticed a few things you have pointed out already, but here's some new things I see.
- [ ] Before you save the form, if you try to duplicate a group or section field it duplicates out of style.
https://www.loom.com/share/4130c9f69a214f2bb51e9fcf017fbdcc
- [ ] After you save the form, when you duplicate the group it puts it in the same group.
localhost_form-ex_wp-admin_admin php_page=formidable frm_action=edit id=27
- [ ] When groups are duplicated they duplicate in alphabetical order. Below, Section A & Section B should be next to each other as they were before duplicating.
localhost_form-ex_wp-admin_admin php_page=formidable frm_action=edit id=27 (1)
- [ ] After duplicating a section with a field, it still shows the `.frm_no_section_fields` border until you save or edit the section.
localhost_form-ex_wp-admin_admin php_page=formidable frm_action=edit id=27 (2)

Crabcyborg:
Thanks @garretlaxton I've logged a couple of these issues so far as new issues. I think some of these could be handled later and were already issues prior to this update.

https://github.com/Strategy11/formidable-pro/issues/3816
https://github.com/Strategy11/formidable-pro/issues/3818

I haven't addressed the other items you mentioned yet either but I'll look into those soon.

I've pushed a few more updates though:

- There will be fewer network calls to update section IDs as it will only make a request when the sections actually change.
- I added a debounce so it syncs faster without hitting performance issues.
- When you delete a group with a section, the children fields of the section were not deleting, just the section itself.

garretlaxton:
How could I test the performance issue?

The first 2 look great! The fields sticking was a bit annoying,so glad that's solved!

Crabcyborg:
> How could I test the performance issue?

Sorry about the delay answering this @garretlaxton. I guess you could just test it with a lot of fields and sections, and if it's still fast, it's working. Testing that is less important than just making sure it all works and that it doesn't feel slow.

I've pushed a few more updates to this. I think this update should be good to go now.

Let me know if you can find any other issues! 🚀

garretlaxton:
I'm not sure if this is intended or related to this update, but sections are collapsed by default on the front-end when collapsible is selected.
image

I also notice section label positions are a bit messed up on the front-end as well.
- The right label still shows on the left and the section area is very small.
image
- The none and hidden but leave space options hide the section completely.
image
- If you have collapsible selected and select the none or hidden but leave space option the whole section disappears including the field in the section

stephywells:
This is so much smoother now! Nice job @Crabcyborg !

I'm seeing a couple minor things.
1. When I hover a field over a section, it feels like it should go inside it on drop. Instead, it goes on the right where the drop line is showing.
Screen Shot 2022-10-03 at 12 19 49 PM

2. The lines aren't quite lined up here:
Screen Shot 2022-10-03 at 12 19 19 PM

@semelyak do you have any input on this? I think there might be related designs somewhere, but I'm not sure where to find them.

js/formidable_admin.js
```diff
@@ -1388,23 +1466,24 @@ function frmAdminBuildJS() {
...
+			getFieldsInRow( jQuery( currentItem.get( 0 ).firstChild ) ).each(
+				function() {
+					updateFieldAfterMovingBetweenSections( jQuery( this ) );
+				}
+			);
+			return;
 		}
 
-		currentItem[0].addEventListener( 'click', function() {
-			maybeAddSaveAndDragIcons( this.dataset.fid );
```
Crabcyborg, line 1403: @AbdiTolesa I removed this code. As far as I can tell, we didn't need this here. Just in the other place it was added. Can you confirm that?

Thanks!

Crabcyborg:
>     1. When I hover a field over a section, it feels like it should go inside it on drop. Instead, it goes on the right where the drop line is showing.

Thanks @stephywells! I made the section title droppable now, and then swap out the droppable target so anything that goes over the section titles will try to enter the section's list. I also added a pointer-events: none style on the section header while the field is being dragged because I don't think the big blue outline helps either there.

> Screen Shot 2022-10-03 at 12 19 49 PM

>     2. The lines aren't quite lined up here:

Thanks! I changed the padding in the section to fix that. It should look more centred now.

Screen Shot 2022-10-04 at 11 52 09 AM

stephywells:
@Crabcyborg those are some weird issues. Let's log those and work on them in a separate PR. It seems like it would only make sense for sections to either show the title or not show it. The left, right, and inline stuff should probably all go.

Steve took a look at the changes here, and he would really like the field to be small again when dragging to make it easier to see where things will go. Can we add some css to the field being dragged so it looks more like adding a new field? We can use css to truncate the label, and set a width and bg color on it. It would be nice if we can show the field icon like we do when first dragging in a field. Is this easily to change?

Crabcyborg:
@stephywells What about when you're dragging a field group?

That's likely the weirdest looking use case (a large group), but in that case there is no "field type" icon unless we come up with one for groups (or use the row layout icon or something), or if we just grabbed one from one of the fields in the group.

js/formidable_admin.js
AbdiTolesa: @Crabcyborg I just pulled this update and tested it and found no issues so far after moving a field around sections and fields. Couldn't remember if that was needed really.

stephywells:
I've got a couple small styling requests.

1. The drop zones are showing inside the box on the left, but outside the box on the right. Can the left one be outside the box too? 
Screen Shot 2022-10-12 at 9 49 32 AM

2. When moving a field, can we add a color overlay or something on it, so it's clear what is being moved? Like this:
Screen Shot 2022-10-12 at 9 59 09 AM

Crabcyborg:
> 1. The drop zones are showing inside the box on the left, but outside the box on the right. Can the left one be outside the box too?

@stephywells Good idea. I added a 4 pixel offset there. I'm already offsetting  in-between the fields in the group so it was pretty easy to change.

Screen Shot 2022-10-12 at 4 08 35 PM

>  2. When moving a field, can we add a color overlay or something on it, so it's clear what is being moved? Like this:

I'm not sure I totally follow the request but I made a few more adjustments.

I added a `box-shadow` to the draggable, and I set the item that you are trading to `opacity: 0.5` so it fades out, so it's more clear what you're actually dragging.

I also added another `pointer-events: none` to the field while dragging because I don't think it makes sense to have them show their hover effects while dragging - I noticed that from your screenshot and didn't like it. I've already done this for section headers but I guess I missed fields.

Screen Shot 2022-10-12 at 4 09 56 PM

stephywells:
1. Awesome!

So with #2, instead of showing the radio field, it would have a color block on top of it. Here it is in action. While dragging, the element being dragged changes in the background:
https://user-images.githubusercontent.com/1116876/195432955-4383157d-b952-4d92-b993-224a5d24c142.mp4

Crabcyborg:
@stephywells Thanks, I didn't know we were talking about how how Gutenberg handles it.

How does this look?

ccjT4Fd51x

stephywells:
Yes! Exactly!! You came up with a way better solution for that than I did.

garretlaxton:
I'm noticing section fields can't be moved side by side after being placed in a form. I'm also having issues placing a field in a section when they are side by side. I am using the PR that was merged, not sure if I should be using this instead though.
https://www.loom.com/share/422cbb622ba542a8a0da2f01dc3276a1

Crabcyborg:
I think that's old code Garret. This look like caching to me.

The dragging item should look like the new field buttons now. It still looks like a section in your video which make some think it's not up to date.

garretlaxton:
Thanks for that. I was using the other PR that was merged so that was my issue. Switching to this worked great for me.
Screenshot (21)

The only issue I see is if you grab a new field, you can't put it back incase you grab the wrong field. You'll just have to delete it.

PR #898, opened 2022-09-15
Truncate super long entry names

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3819

This is the first thing I tried to fix this. The item name has a length limit of 255, the example string was too long. Worked like a charm 🥳


garretlaxton (approved):
This looks good! I tested with JS validation, AJAX submit, and also imported the entries as well as duplicated it.

stephywells (approved):
Perfect!

PR #900, opened 2022-09-21
Avoid fatal error when setting up a field with string field options

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2016033128/110605/

image

I'm not too sure how they have a string `field_options`. But I figure it's best to be on the safe side and avoid calling `array_merge` on a string.


stephywells (approved):
Interesting. Not sure how that would have happened.

PR #911, opened 2022-09-26
add a new filter to modify referrer info

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-user-tracking/issues/9
This update adds a new filter that allows users to modify the referrer info of a user tracking.


Crabcyborg (approved):
Thanks @AbdiTolesa this looks good! I'll include it in this release so we don't need to wait if we want to release user tracking sooner! Assuming this is all we'll need at least 🤞.

PR #912, opened 2022-09-27
Avoid posts page when setting a random page for preview

Crabcyborg:
Fixes https://secure.helpscout.net/conversation/2021199211/111203/

The preview query function sets a random page. In this customer's case it was setting their "Posts Page" or "Blog Page" which meant that it was showing the 10 recent posts, and we were overwriting all 10 of them resulting in 10 forms.


stephywells (commented):
Sorry for being late on this! It looks great though!

PR #917, opened 2022-10-03
Reduce space between fields using Inside position

truongwp:
This fixes https://github.com/Strategy11/formidable-pro/issues/3844

Thanks @stephywells for the CSS code.

### DIFF
--- a/css/custom_theme.css.php
+++ b/css/custom_theme.css.php
@@ -443,8 +443,8 @@ legend.frm_hidden{
 /* Floating labels */
 .with_frm_style .frm_inside_container {
 	position: relative;
-	padding-top: 27px;
-	padding-top: calc(0.85 * var(--field-height));
+	padding-top: 16px;
+	padding-top: calc(0.5 * var(--field-height));
 }
 
 .with_frm_style .frm_inside_container > input,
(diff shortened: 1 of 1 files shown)


stephywells (changes requested):
Thanks @truongwp for being so fast!

I just found a design from Yaro on this. Let's do this one where the label doesn't go all the way above the field.
https://www.figma.com/file/Y8rOu6zzVHBFvE0YZwhxbU/Visual-reviews?node-id=15%3A591

Let's also change the label in the settings and styler from "Placeholder leave the space" to "Floating"

truongwp:
@stephywells Which design will we implement? 1 or 2?

Screen Shot 2022-10-04 at 10 38 17

stephywells:
@truongwp sorry that wasn't very clear. I was thinking the first one so we can keep the field heights the same as they are now.

truongwp:
@stephywells Since we decided to work on that design later. Can this PR be merged?

Crabcyborg:
@stephywells Should I merge this or should we hold off on this update?

stephywells:
@Crabcyborg @truongwp I just made one more small update on this so the label would be aligned with the field. It looked weird for fields with a description or when fields had a top label too.

This looks good to go! Thanks @truongwp for your help and patience on this!

stephywells (approved):

PR #919, opened 2022-10-04
Set the global post object when querying for random page to avoid PHP…

Crabcyborg:
… warnings when loading theme comments

Fixes https://github.com/Strategy11/formidable-pro/issues/3004

Before working on https://github.com/Strategy11/formidable-forms/pull/912 I didn't realize we actually queried for a page on in-theme preview. Since we're already doing this, I think we should also set the global `$post` object.

This way we don't trigger the `PHP Warning:  Attempt to read property "comment_count" on null in wp-includes/general-template.php on line 3159` warning that happens when it tries to display comments.

I tried replicating the other console errors from the issue, and they no longer appear to happen, at least not for me with my current environment.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -318,7 +318,6 @@ class FrmFormsController {
 		}
 
 		$random_page = get_posts( $page_query );
-
 		if ( ! $random_page ) {
 			return;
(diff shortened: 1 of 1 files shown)


stephywells (approved):
This looks prefect. Thank you!

PR #920, opened 2022-10-05
Remove title and description options from embed examples

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3852

Since these are settings now, it make sense not to have the options set by default in the embed examples since they override the setting. It seemed to confuse a customer, and I think we can avoid future support tickets by having it rely on the options in settings by default.

Screen Shot 2022-10-05 at 10 31 11 AM


stephywells (approved):
Great idea!

PR #922, opened 2022-10-05
New draggable helper function

Crabcyborg:
With this update the dragged item no longer "clones". It now always uses the "frmbutton" look for new fields. For a field group, it says "Field Group" instead.

@stephywells @srwells Is this what you had in mind? Ignore the code for now. It's still in a prototype stage.

I'm also cleaning up some styles a bit.

For now I'm using the row layout icon for field groups. Trying to get a field count is more difficult than it should be at the moment.

Screen Shot 2022-10-05 at 2 35 33 PM

tI4rTw6gB8

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -7175,7 +7175,7 @@ input[disabled],
 	height: 25px !important;
 	width: 140px;
 	text-align: center;
-	display: block !important;
+	display: inline !important;
 	background-color: var(--lightest-grey) !important;
 	border-radius: 35px !important;
 	padding: 5px 20px !important;
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -3378,7 +3378,7 @@ li.ui-state-default.selected > .frm_inner_field_container > label {
 }
 
 .frm-dragging * {
-	cursor: move !important;
+	cursor: grabbing !important;
 }
 
 .frm-dragging .divider_section_only {
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -7183,7 +7183,6 @@ input[disabled],
 }
 
 #frm-show-fields .frmbutton.ui-sortable-helper i,
-.frmbutton.ui-sortable-helper span,
 .frmbutton.ui-draggable-dragging i,
 .frmbutton.ui-draggable-dragging .frmsvg {
 	color: #fff !important;
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -3828,7 +3828,7 @@ li.sortable-placeholder {
 	border: none;
 	margin: 0;
 	visibility: visible !important;
-	box-shadow: 2px -10px 0 1px var(--primary-color);
+	box-shadow: 2px -10px 0 2px var(--primary-color);
 }
 
 #frm_form_editor_container > ul > .frm_single_option.ui-sortable-placeholder,
@@ -5455,7 +5455,7 @@ a.frm_action_icon:hover {
 }
 
 .frm-move {
-	cursor: move;
+	cursor: grab;
 }
 
 span.howto {
@@ -7173,11 +7173,12 @@ input[disabled],
 	flex-direction: row;
 	justify-content: center;
 	align-items: center;
-	background-color: var(--lightest-grey) !important;
+	background-color: var(--sidebar-color) !important;
 	border-radius: 35px !important;
 	padding: 5px 10px !important;
 	color: var(--dark-grey) !important;
 	box-sizing: border-box;
+	border: 1px solid var(--border-color);
 }
 
 .frmbutton.ui-draggable-dragging span {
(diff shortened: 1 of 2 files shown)


stephywells:
Yes! It is possible to have the draggable thing sitting right below the cursor?

srwells:
Yeah the field placeholder needs to be at the same location as the cursor. This helps especially when dragging a field horizontally into an existing row.

Crabcyborg:
Thanks @stephywells and @srwells. The position was a bit funny to me too, that's where jQuery was choosing to place it. After a bit more searching around I found a `cursorAt` setting that supports this positioning. Right now I set it to 0/0, so it appears directly to the right of the cursor.

Would we want it to more centred or offset relative to the cursor? I feel like the place I have it now it's not really in the way, but it's always close and easy to view.

jYehJh1u9t

srwells:
I like that a lot better…regarding the position, it is difficult to say without seeing or trying it myself…I am wondering if centered and just above the cursor would be better…also, what if the cursor style changed to cursor:grab or cursor:grabbing would feel more appropriate for a drag and drop action.

Steve Wells
COO at Strategy11 LLC

Our Brands:
Formidable Forms | Business Directory Plugin | AWP Classifieds

srwells:
After looking around a bit and playing with drag and drop interfaces a bit more. I definitely think any draggable element (fields in the left sidebar, and the "Move" icon of an added field) should use the cursor:grab CSS on Hover and cursor:grabbing when moving. 

https://user-images.githubusercontent.com/1116894/194386906-b1396eb8-10b8-4b81-a68d-a0d2a9cc60e2.mp4

Crabcyborg:
@srwells We use a move icon instead of that "grab" icon, so I don't know if it makes sense to use grab/grabbing as much in our case.

Screen Shot 2022-10-06 at 3 54 01 PM

I took out some code about cursor that likely didn't work as-is anymore, forgot to add remove it. I updated it to use `move` at the moment though. Thanks for the reminder.

stephywells:
I'm not sure if you're waiting on review for this right now or not.... Just in case you are, I'm seeing the draggable mostly cut off when I start dragging. This might be what you're talking about with the offsetting though.
https://user-images.githubusercontent.com/1116876/195130727-761d5396-6808-4b8b-8395-63b3c92b6d0e.mp4

Crabcyborg:
Sorry @stephywells I haven't really circled back to this.

I don't think getting it to move above the mouse is very easy, but having it horizontally offset works great. Right now I have all of the draggable buttons set to 140 width, so I'm just hard coding a 70 left offset. I added a comment so that doesn't get missed if the draggable size was changed.

I also set the the draggable item to be inline now. The `display: block` styling was causing a radio button to break into two lines because the text was too long. Setting it to inline makes the icon smaller/more to the left so the label has more space.

Here's a recording with the new offset.
hdZksL7gxe

srwells:
I am not sure if we were talking about the same thing regarding the cursor. In your recent screencast, it is still using the cursor:pointer when dragging a field. Can we switch that to use cursor:grab on hover over the move icon, and cursor:grabbing while moving/dragging the field?
Screenshot 2022-10-11 at 11 38 18 AM

Crabcyborg:
@srwells My screen recording software seems to fake the mouse pointer. It doesn't do that on my end when I actually drag in my browser.

srwells:
Ah, okay…good to know. Thanks for clarifying.

Steve Wells
COO at Strategy11 LLC

Our Brands:
Formidable Forms | Business Directory Plugin | AWP Classifieds

stephywells:
@Crabcyborg are you seeing this too?
Screen Shot 2022-10-11 at 7 34 36 PM

Crabcyborg:
> Are you seeing this too?

@stephywells Yeah, I am. It looks like it happens when it's up against the right side. Thanks for catching that.

Just switching it from block to inline didn't seem to really fix everything. I've updated it a bit more to use some flex styling now, and border-box, and less padding.

Screen Shot 2022-10-12 at 9 58 51 AM
Screen Shot 2022-10-12 at 9 59 06 AM
Screen Shot 2022-10-12 at 9 59 19 AM
Screen Shot 2022-10-12 at 9 59 29 AM
Screen Shot 2022-10-12 at 9 59 41 AM

stephywells:
This looks great now @Crabcyborg !

I just added a border on the dragging element, but it only shows when moving fields and not when adding a new one. Do you know why that would happen?

Crabcyborg:
> Do you know why that would happen?

Yes @stephywells I was able to inspect it and Firefox has a warning about it (`--border-color` isn't set).

Screen Shot 2022-10-12 at 1 10 05 PM

I updated it to just use `--grey-border` which is set.

Crabcyborg:
I'm going to merge this into the other update so it's all together again.

@garretlaxton If you find any issues you can comment on them in https://github.com/Strategy11/formidable-forms/pull/892.

Thanks!

PR #923, opened 2022-10-05
Run do_shortcode on post content in simple_form_render

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3853

@stephywells Do you see an issue just solving it this way? Or can you think of a better solution to this?

### DIFF
--- a/classes/controllers/FrmSimpleBlocksController.php
+++ b/classes/controllers/FrmSimpleBlocksController.php
@@ -140,12 +140,28 @@ class FrmSimpleBlocksController {
 			return '';
 		}
 
+		self::maybe_run_shortcodes_on_page_global();
+
 		$params       = array_filter( $attributes );
 		$params['id'] = $params['formId'];
 		unset( $params['formId'] );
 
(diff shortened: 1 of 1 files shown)


stephywells:
@Crabcyborg I wonder if it would be worth checking for the frm-set-get shortcode before we run this. It's possible this could cause conflicts from plugins that use include_once or something like that. Maybe we could add a hook with a list of shortcodes we should run, and then process only the shortcodes we want?

Ideally, we don't use that shortcode at all. We use [formidable id=1 dailycaloric="My Value"] instead these days. But that's not supported in the block either.

Crabcyborg:
Thanks @stephywells!

I've pushed a couple updates. Now all of the logic is in Pro (as `[frm-set-get]` is a Pro feature) and it will only process the `frm-set-get`/`frm_set_get` shortcodes. Here I've just left a new action so I can hook into this from Pro. https://github.com/Strategy11/formidable-pro/pull/3854

stephywells (approved):
This looks like a great option.

PR #924, opened 2022-10-06
Check for simplexml_load_string function, show form install error in …

Crabcyborg:
…modal

Related ticket https://secure.helpscout.net/conversation/2029422497/112116/

It turns out that it tries to show the error on "form.elements.show_response.value" `show_response` doesn't exist in this context. It looks like it exists in FrmSolution, though it's in a hidden input so I don't understand what it's intention there is anyway.

I'm creating a modal on modal now instead. And I added a new error message to check that `simplexml_load_string` actually exists.

Otherwise they hit an unexpected fatal error:
```
PHP Fatal error:  Uncaught Error: Call to undefined function simplexml_load_string() in /var/www/html/wp-content/plugins/formidable/classes/controllers/FrmXMLController.php:50
Stack trace:
```

Screen Shot 2022-10-06 at 9 13 09 AM
Screen Shot 2022-10-06 at 9 13 32 AM

### DIFF
--- a/classes/controllers/FrmXMLController.php
+++ b/classes/controllers/FrmXMLController.php
@@ -35,10 +35,19 @@ class FrmXMLController {
 	 * Use the template link to install the XML template
 	 *
 	 * @since 3.06
+	 * @return void
 	 */
 	public static function install_template() {
 		FrmAppHelper::permission_check( 'frm_edit_forms' );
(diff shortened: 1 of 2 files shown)


stephywells (approved):
I'm not sure why that was there either. This fix looks great!

PR #927, opened 2022-10-07
replace newline with <br>

AbdiTolesa:
This update just uses `wpautop` to replace double line breaks with paragraphs for HTML emails.

### DIFF
--- a/classes/controllers/FrmFormActionsController.php
+++ b/classes/controllers/FrmFormActionsController.php
@@ -101,7 +101,7 @@ class FrmFormActionsController {
 	 */
 	public static function update_email_message( $message, $atts ) {
 		if ( ! $atts['plain_text'] ) {
-			$message = wpautop( $message );
+			$message = wpautop( $message, false );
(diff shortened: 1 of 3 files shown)


Crabcyborg:
@AbdiTolesa I see you haven't requested review yet but I have a few comments already.

I think we want to avoid this for the "plain text" option. Does that work already with `"\n"`?

Screen Shot 2022-10-07 at 3 30 37 PM

Second thing, I think we may want to use https://developer.wordpress.org/reference/functions/wpautop/ instead of this `str_replace`. I think it should do the same thing, and possibly some additional things as well.

garretlaxton:
Before:
image

After:
image

Should the first line have the `<p>` tag as well in the `[default_message]` shortcode? Other than that, this looks good.

Crabcyborg:
Thanks for testing @garretlaxton! I think that's fine. The default shortcode makes a table, which I don't think really needs to change sense in a `wpautop` context.

PR #928, opened 2022-10-10
Issue 3841

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3841 and Fixes https://github.com/Strategy11/formidable-pro/issues/3850

Its a bit tricky to find an ideal solution for this because sending an html to tinymce editor parses it and create DOM element (table in default-html case) and the browser auto fix changes the html which in unintended side effect. This update avoids that though and is backward compatible.

### DIFF
--- a/classes/controllers/FrmFormActionsController.php
+++ b/classes/controllers/FrmFormActionsController.php
@@ -95,7 +95,7 @@ class FrmFormActionsController {
 	/**
 	 * Updates email message
 	 *
-	 * @param $message
+	 * @param string $message
 	 *
 	 * @return string
 	 */
--- a/classes/controllers/FrmFormActionsController.php
+++ b/classes/controllers/FrmFormActionsController.php
@@ -93,6 +93,18 @@ class FrmFormActionsController {
 	}
 
 	/**
+	 * Updates email message
+	 *
+	 * @param $message
+	 *
+	 * @return string
+	 */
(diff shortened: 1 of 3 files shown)


garretlaxton:
I'm still seeing the shortcodes come through in the email after this. Any ideas why?
image

Crabcyborg:
@garretlaxton I was for a moment when I tried but it started working eventually. It may just be a persistent cache issue.

garretlaxton:
@Crabcyborg Unfortunately, I still couldn't get this to work.

Crabcyborg:
> Unfortunately, I still couldn't get this to work.

Thanks for trying @garretlaxton. I'm not sure really. The more I test this, I can only sometimes get it to not work, at the start, and then it seems to work pretty well for me. I tried another browser too but it seemed pretty consistent in Chrome and Firefox.

I uploaded a beta to the QA site to test it out. Does it not work for you in this form either?

https://qa.formidableforms.com/mike2/wp-admin/admin.php?page=formidable&frm_action=settings&id=195

garretlaxton:
I'm guessing I was using this wrong. I typed in the shortcodes rather than going to the advanced tab and adding it from there. When I do that it works fine.

Crabcyborg:
Thanks @garretlaxton. Sorry I didn't consider mentioning that. That's how it works though, yes.

PR #929, opened 2022-10-11
Add hook for user tracking

stephywells:
This prevents errors in the sidebar with the user tracking plugin installed

@AbdiTolesa is there already something for this that I missed?

### DIFF
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -666,6 +666,13 @@ class FrmEntriesController {
 			if ( isset( $data['browser'] ) ) {
 				$browser = FrmEntriesHelper::get_browser( $data['browser'] );
 			}
+			/**
+			 * Add or remove information in the entry sidebar.
+			 *
+			 * @since 5.5.2
+			 * @param array $data
+			 */
+			$data = apply_filters( 'frm_sidebar_data', $data, compact( 'entry' ) );
 		}
 
 		include( FrmAppHelper::plugin_path() . '/classes/views/frm-entries/sidebar-shared.php' );


AbdiTolesa:
> This prevents errors in the sidebar with the user tracking plugin installed
> 
> @AbdiTolesa is there already something for this that I missed?

@stephywells We are already taking care of that using `frm_show_entry_sidebar` ( https://github.com/Strategy11/formidable-forms/blob/master/classes/views/frm-entries/show.php#L77 ) to unset `user_journey` key from the entry description. The reason you're seeing the error in the sidebar is because `FrmUsrTrkAppController::lite_is_up_to_date` function assumes lite is upto date if it is >= 5.5.1 but that version is already out and we now have to update that to the next lite version to be released with the add on.

Crabcyborg (approved):
Thanks @stephywells this looks good.

I was wondering about this too. We have a version check in user tracking that isn't really tied to any update in Lite anymore.

I think it was originally this commit https://github.com/Strategy11/formidable-forms/pull/879/commits/50f2ce57da0ebf21663a5c74c854921b9805c3db

But after a few updates it wasn't part of https://github.com/Strategy11/formidable-forms/pull/879 anymore and I just ended up closing it because the branch didn't seem to be doing anything.

So I think something got mixed up when @AbdiTolesa was trying to figure out which parts go into which add on.

The only lite pull request that has been merged so far is this one that got merged into last release but it seems to intended for the opposite purpose https://github.com/Strategy11/formidable-forms/pull/911/files

stephywells:
> @stephywells We are already taking care of that using `frm_show_entry_sidebar` 
@AbdiTolesa this wasn't working for me. I just removed that code since it's on a do_action. This can only add code, and doesn't return any changes. Please test all your changes and additions before submitting them for review.

> The only lite pull request that has been merged so far is this one that got merged into last release but it seems to intended for the opposite purpose https://github.com/Strategy11/formidable-forms/pull/911/files
@Crabcyborg that one was intended to add the info inside the email notifications and the frm-show-entry shortcode. I reopened that issue just now too.

AbdiTolesa:
> > @stephywells We are already taking care of that using `frm_show_entry_sidebar`
> 
> @AbdiTolesa this wasn't working for me. I just removed that code since it's on a do_action. This can only add code, and doesn't return any changes. Please test all your changes and additions before submitting them for review.
> 
> > The only lite pull request that has been merged so far is this one that got merged into last release but it seems to intended for the opposite purpose https://github.com/Strategy11/formidable-forms/pull/911/files
> 
> @Crabcyborg that one was intended to add the info inside the email notifications and the frm-show-entry shortcode. I reopened that issue just now too.

@stephywells I meant this one below, we just grabbed $entry to unset the keys we don't need there:
`	public static function maybe_remove_user_journey_from_entry_description( $entry ) {
		if ( ! is_array( $entry->description ) ) {
			return;
		}

		$lite_is_up_to_date = self::lite_is_up_to_date();

		if ( ! $lite_is_up_to_date ) {
			unset( $entry->description['user_journey'] );
			unset( $entry->description['referrer'] );
		}
	}
`

stephywells:
@AbdiTolesa did you test that? It wasn't working for me.

AbdiTolesa:
> @AbdiTolesa did you test that? It wasn't working for me.

Yes @stephywells the reason it wasn't working for you was the lite version checking logic needed an update as v5.5.1 is already out and we need to increase the minimum lite version to the future release that we are going to do next, which was not supposed to have the code that generates the error.

classes/controllers/FrmEntriesController.php
```diff
@@ -666,6 +666,13 @@ public static function entry_sidebar( $entry = false ) {
 			if ( isset( $data['browser'] ) ) {
 				$browser = FrmEntriesHelper::get_browser( $data['browser'] );
 			}
+			/**
+			 * Add or remove information in the entry sidebar.
+			 *
+			 * @since 5.5.2
+			 * @param array $data
+			 */
+			$data = apply_filters( 'frm_sidebar_data', $data, compact( 'entry' ) );
```
AbdiTolesa, line 675: @stephywells A bit late to comment on this but wouldn't error still be thrown if the user updated their add on and not lite probably?

stephywells: @AbdiTolesa Yes, you're right that there could still be an error in the sidebar in that case. But since we are releasing this ahead of the add-on update, I don't expect it to be an issue. The bigger conflicts we need to watch out for happen when Lite is auto updated and add-ons and Pro are not.

PR #930, opened 2022-10-11
New filter "frm_rich_text_emails"

Crabcyborg:
Jonathan mentioned to me that he's seen a few tickets so far that has requested to option to turn off the new rich text emails.

It isn't completely clear why, but I know some people actually have 99+ email actions, and loading TinyMCE for all of those would probably be a bad idea, so I can understand some cases when it makes sense.

Example snippets:

```
add_filter( 'frm_rich_text_emails', '__return_false' );
add_filter( 'frm_rich_text_emails', 'disable_rich_text_email_for_specific_action', 10, 2 );
add_filter( 'frm_rich_text_emails', 'disable_rich_text_email_for_specific_form', 10, 2);

function disable_rich_text_email_for_specific_action( $rich_text_emails, $args ) {
	$target_action_id = 2381;
	if ( $target_action_id === $args['form_action']->ID ) {
		$rich_text_emails = false;
	}
	return $rich_text_emails;
}

function disable_rich_text_email_for_specific_form( $rich_text_emails, $args ) {
	$target_form_id = 338;
	if ( $target_form_id === (int) $args['form']->id ) {
		$rich_text_emails = false;
	}
	return $rich_text_emails;
}
```

### DIFF
--- a/classes/views/frm-form-actions/_email_settings.php
+++ b/classes/views/frm-form-actions/_email_settings.php
@@ -77,16 +77,33 @@ if ( ! defined( 'ABSPATH' ) ) {
 		<?php esc_html_e( 'Message', 'formidable' ); ?>
 	</label>
 	<?php
-	$editor_args = array(
-		'textarea_name' => $this->get_field_name( 'email_message' ),
-		'textarea_rows' => 6,
-		'editor_class'  => 'frm_not_email_message',
-	);
-	wp_editor(
-		$form_action->post_content['email_message'],
-		$this->get_field_id( 'email_message' ),
-		$editor_args
-	);
+	/**
+	 * @since 5.5.2
+	 *
+	 * @param bool  $rich_text_emails
+	 * @param array $args {
+	 *     @type stdClass $form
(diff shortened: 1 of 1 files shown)


stephywells:
@Crabcyborg It looks like the rich text is still on when the "send emails in plain text" box is checked. Can we fix that? I think that would solve a lot of this issue.

garretlaxton:
If a message has formatted text and then you use a filter to disable the the editor, should the message keep the format?
image

Crabcyborg:
> If a message has formatted text and then you use a filter to disable the the editor, should the message keep the format? 

@garretlaxton Yeah, whatever string is saved should be the same as before in the textarea. The only difference should be the rich text editing.

Crabcyborg:
> It looks like the rich text is still on when the "send emails in plain text" box is checked. Can we fix that?

@stephywells I just pushed an update here so the initial `$rich_text_emails` variable is based on the `plain_text` setting. I didn't try having it dynamically switch over (it would be fairly complicated to switch back and forth), but having it work on load should work well enough I think.

PR #934, opened 2022-10-12
Exclude license types from application categories

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3857

I'm just including a check against `FrmFormsHelper::ignore_template_categories` which is used for this with other templates already.

**Before**
Screen Shot 2022-10-12 at 1 32 42 PM

**After**
Screen Shot 2022-10-12 at 1 32 47 PM


stephywells (approved):
Great plan! Thank you!

PR #938, opened 2022-10-14
Add JS validation for password strength

truongwp:
I decided not to fix the issue I mentioned here https://github.com/Strategy11/formidable-pro/issues/3823#issuecomment-1271677233. Because it can be solved easily by setting the Password field required.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -2799,6 +2799,8 @@ class FrmAppHelper {
 			'include_alert_role' => self::should_include_alert_role_on_field_errors(),
 		);
 
+		self::maybe_add_password_checks_data();
+
 		$data = $wp_scripts->get_data( 'formidable', 'data' );
 		if ( empty( $data ) ) {
 			wp_localize_script( 'formidable', 'frm_js', $script_strings );
@@ -2864,6 +2866,22 @@ class FrmAppHelper {
 				wp_localize_script( 'formidable_admin', 'frm_admin_js', $admin_script_strings );
 			}
 		}
+	}
+
+	private static function maybe_add_password_checks_data() {
+		if ( ! class_exists( 'FrmProFieldPassword', false ) ) {
+			return;
+		}
+		$field          = new stdClass();
+		$field->name    = 'password';
+		$field->type    = 'password';
+		$password_field = new FrmProFieldPassword( $field, 'password' );
+
+		wp_localize_script(
+			'formidable',
+			'frm_password_checks',
+			$password_field->password_checks()
+		);
 	}
 
 	/**
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -2799,8 +2799,6 @@ class FrmAppHelper {
 			'include_alert_role' => self::should_include_alert_role_on_field_errors(),
 		);
 
-		self::maybe_add_password_checks_data();
-
 		$data = $wp_scripts->get_data( 'formidable', 'data' );
 		if ( empty( $data ) ) {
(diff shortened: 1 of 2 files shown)


js/formidable.js
```diff
@@ -410,12 +402,34 @@ function frmFrontFormJS() {
...
-		return errors;
 	}
 
 	function checkPasswordField( field, errors ) {
 		confirmField( field, errors );
-		return errors;
+		checkPasswordStrength( field, errors );
+	}
+
+	function checkPasswordStrength( field, errors ) {
```
Crabcyborg, line 412: Thanks for working on this @truongwp!

What do you think about trying to get the majority of this code into Pro? Maybe it's possible to trigger an event here and then hook into it in Pro?

Since Passwords are a pro feature it would be nice if we didn't have it in lite to begin with. It can be a lot harder to move the code later for backward compatibility and we end up having to put things in both places.

js/formidable.js
```diff
@@ -410,12 +402,34 @@ function frmFrontFormJS() {
...
+		}
+
+		if ( -1 === field.className.indexOf( 'frm_strong_pass' ) || 0 === field.id.indexOf( 'field_conf_' ) ) {
+			return;
+		}
+
+		checks = window.frm_password_checks;
+
+		for ( check in checks ) {
+			regex = checks[ check ].regex.slice( 1, checks[ check ].regex.length - 1 )
```
Crabcyborg, line 426: This line is also getting caught in the linting. There's no semi-colon here.

truongwp:
@Crabcyborg Thanks for checking. I moved it.

js/formidable.js
```diff
@@ -48,6 +48,20 @@ function frmFrontFormJS() {
...
+	 * Triggers custom JS event.
+	 *
+	 * @since 5.x
+	 *
+	 * @param {HTMLElement} el        The HTML element.
+	 * @param {String}      eventName Event name.
+	 * @param {mixed}       data      The passed data.
+	 */
+	function triggerCustom( el, eventName, data ) {
+		var event = new CustomEvent( eventName, { detail: data });
```
Crabcyborg, line 61: Thanks @truongwp! This is looking better.

`CustomEvent` is a feature with questionable IE support.

> While a <code>window.CustomEvent</code> object exists, it cannot be called as a constructor. Instead of <code>new CustomEvent(...)</code>, you must use <code>e = document.createEvent('CustomEvent')</code> and then <code>e.initCustomEvent(...)</code>

I've come up with a solution for this before already. Right here https://github.com/Strategy11/formidable-chat/blob/ca4653f54446479ffed84432d830c9d58e34b8e4/js/chat.js#L1473

My trigger custom event code is a bit larger,

```
function triggerEvent( element, eventType ) {
		var event;

		if ( typeof window.CustomEvent === 'function' ) {
			event = new CustomEvent( eventType );
		} else if ( document.createEvent ) {
			event = document.createEvent( 'HTMLEvents' );
			event.initEvent( eventType, false, true );
		} else {
			return;
		}

		element.dispatchEvent( event );
	}
```

classes/helpers/FrmAppHelper.php
```diff
@@ -2866,6 +2868,27 @@ public static function localize_script( $location ) {
 		}
 	}
 
+	/**
+	 * Adds password checks data to JS.
+	 *
+	 * @since 5.x
+	 */
+	private static function maybe_add_password_checks_data() {
```
Crabcyborg, line 2876: @truongwp Is it possible to get this function into Pro as well?

js/formidable.js
truongwp: Thank you. I worked with this code before but forgot it.

PR #940, opened 2022-10-14
Drag and drop enhancements

Crabcyborg:
- It was possible to drag a new field into a field group with a page break or a hidden field. I'm preventing this the other way (no adding a new page break or a hidden field to a group), but missed this check on the other end.
- Steve commented about issues with dragging. It looks like cursorAt doesn't really sync with scroll events. I've fixed this with a `transform: translateY( offset )` style that calculates based on the container's scroll position.
- I swapped the cursor setting to the JavaScript and removed it from CSS. I didn't realize `cursor` was a draggable setting until reading more into the documentation today.

**Before**
dHfHigkxYa

**After**
JcwZad88G0

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -3390,10 +3390,6 @@ li.ui-state-default.selected > .frm_inner_field_container > label {
 	max-width: calc(100% - 100px);
 }
 
-.frm-dragging * {
-	cursor: grabbing !important;
-}
-
 .frm-drag-fade {
 	background-color: var(--lightest-grey) !important;
 	border-radius: 4px;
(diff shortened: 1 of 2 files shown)


garretlaxton:
I'm not sure if I'm misunderstanding the first part, but a hidden field shouldn't be allowed in a group field correct?

Crabcyborg:
Yeah you shouldn't be able to inline a hidden or page break field with another field as they're not really fields that make sense in the same row with something else.

garretlaxton:
The page breaks look good, but I can still move a hidden field into a group.

Once you place a hidden field in a form, you can then move it into a group.

Crabcyborg:
Thanks @garretlaxton, there's a lot of cases to think about here, missed that one too. That should be prevented now as well.

garretlaxton (approved):
Sweet, this looks good now!

stephywells (approved):
This looks good! Thank you!

PR #941, opened 2022-10-17
Fix text input advanced option button style

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3860
It looks like text inputs with advanced option are injected into the DOM when form actions are loaded but textarea fields are already loaded in different places on page loads. It might make sense to leave the css style for textarea SVGs as is but fix these for text fields when form actions are loaded.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -7776,6 +7776,7 @@ function frmAdminBuildJS() {
 		event.target.closest( '.frm_form_action_settings' ).querySelectorAll( '.frmsvg.frm-show-box' ).forEach( ( svg ) => {
 			if ( svg.nextElementSibling.type === 'text' ) {
 				svg.style.bottom = '-3px';
+				svg.style.marginRight = '0';
 			}
 		});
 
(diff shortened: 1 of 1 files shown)


Crabcyborg:
@AbdiTolesa I think we also want to keep the old right positioning. It's a lot more offset inside of the field with the previous update too.

**Before last update**
Screen Shot 2022-10-21 at 9 46 54 AM

**After**
Screen Shot 2022-10-21 at 9 47 20 AM

AbdiTolesa:
> @AbdiTolesa I think we also want to keep the old right positioning. It's a lot more offset inside of the field with the previous update too.
> 
> **Before last update** Screen Shot 2022-10-21 at 9 46 54 AM
> 
> **After** Screen Shot 2022-10-21 at 9 47 20 AM

@Crabcyborg This is fixed now, thanks!

garretlaxton:
Here's what I'm seeing.
image

Crabcyborg (approved):
Thanks for making a look @garretlaxton.

The icons look a lot better with this update to me!

Thanks @AbdiTolesa!

PR #942, opened 2022-10-17
override datepicker pseudo elements styles using svg as background image

AbdiTolesa:


css/custom_theme.css.php
```diff
@@ -364,6 +364,30 @@
 	padding:25px;
 }
 
+.ui-datepicker .ui-datepicker-prev:before{
```
Crabcyborg, line 367: @AbdiTolesa The datepicker is only a Pro feature so we should keep this in Pro.

To make it more Formidable-specific it would be nice if it used the `FrmProFieldsHelper::get_datepicker_class` class as well, so it would use `ui-datepicker` for older versions, but `frm-datepicker` for versions that support our new datepicker class with fewer conflicts.

It should probably go into `pro_fields.css.php`

AbdiTolesa: @Crabcyborg I gave it a shot with the datepicker class but it doesn't seem to work with it as the calendar fragment seem to get appended to the body tag directly.

Crabcyborg: @AbdiTolesa What is your styling rule like that isn't working. It shouldn't be an issue as the class is on the datepicker element.

You can see "frm-datepicker" is my last class on my `ui-datepicker` here.

Screen Shot 2022-10-19 at 12 14 38 PM

So `.ui-datepicker` should be replaceable with the dynamic class.

Crabcyborg:
I'm closing this as it's all in Pro now. New pull request https://github.com/Strategy11/formidable-pro/pull/3877

PR #943, opened 2022-10-17
Make honeypot field untabbable after first tab press

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3856

I looked at the original suggestion, and the suggestion of adding a `-1` tabindex to the input is the best solution I can think of to handle this issue.

I figure a bot might be more likely to check for something like `tabindex="-1"` so I put it in an additional event listener first so it only happens if it detects a tab press.

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1098,7 +1098,7 @@ function frmFrontFormJS() {
 	}
 
 	function maybeMakeHoneypotFieldsUntabbable() {
-		document.addEventListener( 'keyup', handleKeyUp );
+		document.addEventListener( 'keydown', handleKeyUp );
 
 		function handleKeyUp( event ) {
 			var code;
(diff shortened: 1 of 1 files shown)


garretlaxton:
This looks good! But I do notice that if you only have one field and click on the field, then click the tab key, this will still happen.

Crabcyborg:
> But I do notice that if you only have one field and click on the field, then click the tab key, this will still happen.

Thanks @garretlaxton, that slipped my mind for a moment which event type to use here before I set the attribute. I switched it from `keyup` to `keydown` now so it should happen before it tabs to the field, instead of after.

garretlaxton (approved):
Sweet, this looks great now!

stephywells (approved):
This looks like a good way of doing it. Does that mean we can remove aria-hidden="true"?

Crabcyborg:
> Does that mean we can remove aria-hidden="true"?

@stephywells As I think about that, I'm not sure if it would cause issues. When I test VoiceOver it will sometimes read all of the contents of the page out to me before I even do anything. If we remove that `aria-hidden` I believe you'll get told about the element still if you don't happen to hit tab yet.

stephywells:
Okay, I don't think that's a problem. It's hard to know what the bots are looking for.

PR #944, opened 2022-10-17
Remove ie8 polyfills

Crabcyborg:
This update drops polyfills for `Array.prototype.indexOf` and and for `Object.keys`.

I kept the `String.prototype.trim` and `Array.prototype.filter` as those are actually not supported in later versions of IE as well.

I renamed them a bit and added more comments to make that a bit more clear as these aren't just for IE8, but for IE in general.


garretlaxton (approved):
This looks good still!

stephywells (approved):
This looks great! Thanks Mike!

PR #945, opened 2022-10-18
Only append to array, only iterate an array during csv export

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2022152412/111312

The XML file is included in one of Kristine's notes there. I'm not including it here because this is public.

This update works to remove a fatal error and a warning that I get when exporting this customer's entries as CSV.

I spent some time trying to debug the data coming through. As far as I can tell the CSV exports fine with these changes.

I ended up just adding type checks so the code won't break. I haven't really figured out the root of this issue. But I don't know if it's worth the effort either.

> PHP Fatal error:  Uncaught Error: [] operator not supported for strings in .../plugins/formidable/classes/helpers/FrmCSVExportHelper.php:481

> [18-Oct-2022 19:26:44 UTC] PHP Warning:  foreach() argument must be of type array|object, bool given in ...plugins/formidable/classes/helpers/FrmCSVExportHelper.php on line 525

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1449,7 +1449,7 @@ class FrmAppHelper {
 
 		// Do not treat the view listing as full screen when no form id is being filtered.
 		global $pagenow;
-		return 'edit.php' !== $pagenow || FrmAppHelper::simple_get( 'form' );
+		return 'edit.php' !== $pagenow || self::simple_get( 'form' );
 	}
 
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
This looks good, I no longer see the error in the exported CSV file.
image

stephywells (approved):
This looks good. It's probably not necessary to find the corrupted data. It might have been imported originally.

PR #947, opened 2022-10-19
Do not hide update button on styles2 page

Crabcyborg:
Looking at our full screen code I noticed "styles2" being referenced in a few places. I eventually figured out that's there's a "Forms" link from the Appearance menu item that uses this.

I noticed that there's still an old WP 5.7+ bug here on this page because the CSS was only targeting the one page. https://github.com/Strategy11/formidable-forms/pull/361

It seems like no one may really use this page if this hasn't come up in support after all this time. Do we actually think people are visiting this page? Do we track that? It may make sense to remove the secondary URL.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -8518,7 +8518,8 @@ Responsive Design
 	}
 }
 
-.frm-admin-page-styles #frm-publishing #save_menu_header {
+.frm-admin-page-styles #frm-publishing #save_menu_header,
+.frm-admin-page-styles2 #frm-publishing #save_menu_header {
 	display: inline-block;
 }
 


stephywells (approved):
Thanks for the fix!

We don't have any tracking on this page. That would be nice data to have, I'm just not really sure how we'd get it... Any ideas?

Crabcyborg:
> Any ideas?

Right, I guess we don't really have any page visit specific information in our `FrmUsage` yet. Maybe we could start? Maybe something simple, even just to track if this specific page is being tracked.

Mostly page information isn't very important to store, but I could see it coming in handy in other cases too. Maybe specific journey-type milestones. Like "Went from the welcome screen and ..." "clicked the video first", "created a new blank form", etc.

stephywells:
That's a good idea. We can think more on this for a future update. It would also be nice to be able to trigger inbox notices based on those activities or if they are missing. (i.e. If didn't create a form after 1 day, trigger an inbox notice or something...)

PR #948, opened 2022-10-19
Stop treating views index with no filtered form ID as full screen

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-views/issues/347

I refactored it a bit as well. Having this check for `self::is_admin_page( 'formidable-styles' ) || self::is_admin_page( 'formidable-styles2' )` everywhere felt problematic so I moved that into a new function that describes that's really going on because I wasn't even aware of this secondary URL in the theme menu. And most others must not be either if the update button was missing all of this time (see https://github.com/Strategy11/formidable-forms/pull/947 😅).

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -499,7 +499,7 @@ class FrmAppController {
 			'formidable_embed',
 		);
 
-		if ( FrmAppHelper::is_on_style_editor() ) {
+		if ( FrmAppHelper::is_style_editor_page() ) {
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
This looks good! The X at the top is now removed only from the main views page.

stephywells (approved):
Thank you for fixing this!!

And nice find on the style page issue. It's possible it has come up in support, but it's tricky to search for it. Or maybe it's just pointless... Really hard to say without any data.

PR #949, opened 2022-10-19
turn on $br param for wpautop function

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3847
This PR fixes the issue of line breaks not being ouput in html emails because the email body is passed through `wpautop` function but the second parameter was set to false.


garretlaxton (approved):
This is looking good to me!
image

PR #950, opened 2022-10-21
Hide screen options on entry duplicate action

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3880

But I also logged a new issue https://github.com/Strategy11/formidable-pro/issues/3886. I don't think this page should exist at this URL endpoint in the first place as it's problematic and causes issues like this one.

That's a larger fix though so I figure I'd still commit this fix first for now.


stephywells (approved):
Thank you! I've been seeing the duplicate entry issue today too. It doesn't seem to check the url nonce.

PR #951, opened 2022-10-21
Fix shortcode typo in strpos check when importing views and pages

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3866

I looked at a few other things before I realized this was just a shortcode mistake. I always mix this shortcode order up because it doesn't start with frm like everything else 😅.


garretlaxton (approved):
The view is now attached to the draft and other page types when importing an application.
image

Crabcyborg:
Thanks for testing @garretlaxton!

PR #952, opened 2022-10-21
Trigger event instead of continuing

Crabcyborg:
This is required for Stripe link to work. I need to hook into before the form clears so my Stripe elements don't get destroyed when the form submits.

This hooks in only on success so the error handling is still handled normally in lite. If an entry is created successfully and the `frm_trigger_event_on_submit` class name is set on the form, it will just call that and exit early.


stephywells (approved):
This looks great!

PR #954, opened 2022-10-26
Change how the placeholder ul is built when duplicating field groups …

Crabcyborg:
…as the old way no longer works

Fixes https://github.com/Strategy11/formidable-pro/issues/3894

With this update a duplicated field group should be in a new row.

Screen Shot 2022-10-26 at 11 34 35 AM

With out, since the last drag-drop enhancement update the field groups merge into one big field group instead.

Screen Shot 2022-10-26 at 11 35 03 AM

This update also includes some refactoring. It removes some jQuery stuff, and uses more ES6 JavaScript than before.


garretlaxton (approved):
This works great now!
image

PR #957, opened 2022-10-28
Issue 3889

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/3889
This update avoids passing the whole html content through autop for html emails and just extracts the body content (if exists), wpply `wpautop` to it and then put it back into the original content.

Before:
```
<p><html><head></p>
<style>h3{text-decoration:underline;}.label{font-size:14px;font-weight:bold;padding-bottom:5px;}.value{font-size:17px;padding:0 0 0 5px;}.question{font-size:17px;padding:0 0 10px 0;}table{margin-bottom:60px;}tr,td{border:none;padding:6px;}.blue{background:#f0f8ff;}.grey{background:#f8f8f8;}</style>
<p></head><body><br />
LINE 1<br />
LINE 2<br />
</body></html></p>
```

After:
```
<html><head><style>h3{text-decoration:underline;}.label{font-size:14px;font-weight:bold;padding-bottom:5px;}.value{font-size:17px;padding:0 0 0 5px;}.question{font-size:17px;padding:0 0 10px 0;}table{margin-bottom:60px;}tr,td{border:none;padding:6px;}.blue{background:#f0f8ff;}.grey{background:#f8f8f8;}</style></head><body><p>LINE 1<br />
			LINE 2</p>
</body></html>

```

### DIFF
--- a/classes/models/FrmEmail.php
+++ b/classes/models/FrmEmail.php
@@ -398,12 +398,14 @@ class FrmEmail {
 			$this->message = wp_specialchars_decode( strip_tags( $this->message ), ENT_QUOTES );
 		} else {
 			$message = $this->message;
-			if ( strpos( $message, '<body>' ) !== false ) {
+			$result  = preg_match( '/<body[^>]*>([\s\S]*?)<\/body>/', $message, $match );
+
+			if ( ! empty( $match[1] ) ) {
 				$result        = preg_match( '/<body[^>]*>([\s\S]*?)<\/body>/', $message, $match );
 				$message       = preg_replace( '/(<body[^>]*>)([\s\S]+?)(<\/body>)/', '${1}' . wpautop( $match[1] ) . '${3}', $message );
 				$this->message = $message; // HTML emails should use autop.
 			} else {
-				$this->message = wpautop( $this->message );
+				$this->message = wpautop( $message );
 			}
 		}
 
--- a/classes/models/FrmEmail.php
+++ b/classes/models/FrmEmail.php
@@ -398,20 +398,15 @@ class FrmEmail {
 			$this->message = wp_specialchars_decode( strip_tags( $this->message ), ENT_QUOTES );
 		} else {
 			$message = $this->message;
-			if ( strpos( $message, '<html' ) !== false ) {
-				$message = preg_replace( '|\s*(<html[^>]*>)\s*|', '$1', $message );
-				$message = preg_replace( '|\s*</html>\s*|', '</html>', $message );
-			}
-			if ( strpos( $message, '<head' ) !== false ) {
-				$message = preg_replace( '|\s*(<head[^>]*>)\s*|', '$1', $message );
-				$message = preg_replace( '|\s*</head>\s*|', '</head>', $message );
-			}
-			if ( strpos( $message, '<body' ) !== false ) {
-				$message = preg_replace( '|\s*(<body[^>]*>)\s*|', '$1', $message );
-				$message = preg_replace( '|</body>\s*|', '</body>', $message );
-			}
 
-			$this->message = wpautop( $message ); // HTML emails should use autop.
+			if ( strpos( $message, '<body>' ) !== false ) {
+				$result        = preg_match( '/<body[^>]*>([\s\S]*?)<\/body>/', $message, $match );
+				$body_inside   = wpautop( $match[1] );
+				$message       = preg_replace( '/(<body[^>]*>)([\s\S]+?)(<\/body>)/', '${1}' . $body_inside . '${3}', $message );
+				$this->message = $message; // HTML emails should use autop.
+			} else {
+				$this->message = wpautop( $this->message ); // HTML emails should use autop.
+			}
 		}
 
 		$this->message = apply_filters( 'frm_email_message', $this->message, $this->package_atts() );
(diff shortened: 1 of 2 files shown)


classes/models/FrmEmail.php
```diff
@@ -397,7 +397,16 @@ private function set_message() {
 		if ( $this->is_plain_text ) {
 			$this->message = wp_specialchars_decode( strip_tags( $this->message ), ENT_QUOTES );
 		} else {
-			$this->message = wpautop( $this->message ); // HTML emails should use autop.
+			$message = $this->message;
+			$result  = preg_match( '/<body[^>]*>([\s\S]*?)<\/body>/', $message, $match );
```
stephywells, line 401: Let's follow the WP rout here so we aren't getting the values twice.

if ( strpos( $this->message, '<body' ) !== false ) {

classes/models/FrmEmail.php
```diff
@@ -397,7 +397,16 @@ private function set_message() {
...
+
+			if ( ! empty( $match[1] ) ) {
+				$result        = preg_match( '/<body[^>]*>([\s\S]*?)<\/body>/', $message, $match );
+				$message       = preg_replace( '/(<body[^>]*>)([\s\S]+?)(<\/body>)/', '${1}' . wpautop( $match[1] ) . '${3}', $message );
```
stephywells, line 405: This is extra duplication as well. Let's only run the regex once. I think this should work:
$this->message = str_replace( $match[1], wpautop( $match[1] ), $this->message );

stephywells (changes requested):
Thanks for trying that @AbdiTolesa !

In addition to my comments below, can you please move this section of code into a new function?

classes/models/FrmEmail.php
AbdiTolesa: @stephywells I think that could be risky because the html might include `<body` but not necessarily match `/<body[^>]*>([\s\S]*?)<\/body>/` which can cause removal of anything after `<body`, just in case the email html is malformed. I have updated the PR except this condition check and could do it as you say if my point doesn't make much sense as well.

classes/models/FrmEmail.php
```diff
@@ -394,13 +394,29 @@ private function set_message() {
...
+
+		$this->message = apply_filters( 'frm_email_message', $this->message, $this->package_atts() );
+	}
+
+	/**
+	 * Decode the message if it is plain text or runs it through autop for html messages.
+	 *
+	 * @param string $message
+	 */
+	private function decode_or_autop( $message ) {
```
stephywells, line 407: The fewest parameters possible is best for clean code. In this case, we don't need to pass any parameters. We can use this->message instead of passing it. 

Also, I don't think decode tells us what is happening. But this would be more clean if the new function only does one thing like function add_autop(). The plain text part can stay where it was until we do a bigger refactor here.

stephywells (changes requested):
@AbdiTolesa i think it would be a big help if you would read this book:
https://thixalongmy.haugiang.gov.vn/media/1175/clean_code.pdf

I learned a lot from it, and it made a huge improvement to the way I write code.

classes/models/FrmEmail.php
AbdiTolesa: Thanks for sharing me a book that would help me cleaner code, @stephywells! I was thinking the function name could get very long to make it more sensible based on what it does but I think the name you suggested best describe it if I should remove remove the code for plain text indeed. Thank you!

classes/models/FrmEmail.php
```diff
@@ -395,14 +395,27 @@ private function set_message() {
 		$this->message = do_shortcode( $this->message );
 
 		if ( $this->is_plain_text ) {
-			$this->message = wp_specialchars_decode( strip_tags( $this->message ), ENT_QUOTES );
+			$this->message = wp_specialchars_decode( strip_tags( $message ), ENT_QUOTES );
```
stephywells, line 398: @AbdiTolesa this breaks plan text emails. Please review your own PRs line by line, and make fixes from the automated checks before requesting reviews. This takes a lot of extra time when you aren't checking your own work in detail.

stephywells (changes requested):

classes/models/FrmEmail.php
AbdiTolesa: I know it might sound using an excuse but the reality is that I was just trying to review this PR on my mobile and unintentionally tapped the button to request for review but I couldn't undo that although I see the failing workflows and it was too late for me and I was also a bit tired to check that ( almost 1:00 am). I understand the frustration and I would make sure such silly mistake wont happen again. I have fixed the issue and writing a unit test to cover that now.

Thank you in advance for understanding this!

garretlaxton:
Should the CSS be stripped when this is sent as plain text? This is what I currently get.
image

Other than that, this looks good!

Crabcyborg:
Thanks @garretlaxton! I believe that's how it worked before since we're not really changing that. If they want it to be plain text it wouldn't make too much sense if they also had the content filled in with all of those styles.

Crabcyborg:
@AbdiTolesa The unit tests are breaking at the moment. Can you get those fixed?

Screen Shot 2022-11-03 at 9 11 12 AM

garretlaxton (approved):

Crabcyborg (approved):

PR #958, opened 2022-10-29
Hcaptcha implementation

AbdiTolesa:
This update integrates hCAPTCHA into our plugin, since the api looks pretty much similar to reCAPTCHA, most of the update just does the counter part for hCAPTCHA.

https://github.com/Strategy11/formidable-pro/pull/3903 also complements with this.


classes/controllers/FrmSettingsController.php
```diff
@@ -180,11 +180,11 @@ public static function message_settings() {
 	/**
 	 * @since 4.0
 	 */
-	public static function recaptcha_settings() {
+	public static function captcha_settings() {
```
Crabcyborg, line 183: We'll need to deprecate the old function name.

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -64,21 +64,36 @@ protected function extra_field_opts() {
 	 * @return string
 	 */
 	protected function before_replace_html_shortcodes( $args, $html ) {
-		return str_replace( ' for="field_[key]"', ' for="g-recaptcha-response"', $html );
+		$frm_settings = FrmAppHelper::get_settings();
+		if ( $frm_settings->active_captcha === 'recaptcha' ) {
+			$replaced_for = str_replace( ' for="field_[key]"', ' for="g-recaptcha-response"', $html );
```
Crabcyborg, line 69: This could use less repeating. The code is the same in both except for the value of `$replace` so it makes snse to just set that to a variable and re-use the rest.

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -64,21 +64,36 @@ protected function extra_field_opts() {
...
+		} else {
+			$replaced_for = str_replace( ' for="field_[key]"', ' for="h-captcha-response"', $html );
+		}
+		return $replaced_for;
 	}
 
 	public function front_field_input( $args, $shortcode_atts ) {
 		$frm_settings = FrmAppHelper::get_settings();
-		if ( empty( $frm_settings->pubkey ) ) {
+		if ( ( $frm_settings->active_captcha === 'recaptcha' && empty( $frm_settings->pubkey ) ) || ( $frm_settings->active_captcha === 'hcaptcha' && empty( $frm_settings->hcaptcha_pubkey ) ) ) {
```
Crabcyborg, line 78: This is basically the same code as in another spot. I think it would benefit to move this into a function. Something like `active_captcha_key_is_set`.

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -219,7 +252,11 @@ public function validate( $args ) {
 	 */
 	private function should_show_captcha() {
 		$frm_settings = FrmAppHelper::get_settings();
-		return ! empty( $frm_settings->pubkey );
+		if ( $frm_settings->active_captcha === 'recaptcha' ) {
```
Crabcyborg, line 255: It looks like there is already a setting for this. This should be getting used in more places. It just needs to be made public.

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -219,7 +252,11 @@ public function validate( $args ) {
...
+			return ! empty( $frm_settings->pubkey );
+		} else {
```
Crabcyborg, line 257: There really isn't any need for this else. It's just extra code.

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -233,14 +270,97 @@ protected function should_validate() {
...
+	 * Updates error message displayed based on the captcha activated.
+	 *
+	 * @since x.x
+	 *
+	 * @param array $errors
+	 * @param array $params
+	 *
+	 * @return array
+	 */
+	public static function update_captcha_field_error_message( $errors, $params ) {
```
Crabcyborg, line 329: What is the need for this setting? This is already happening in `validate_against_api`. I don't think we need this function.

classes/views/frm-fields/back-end/settings.php
```diff
@@ -21,6 +21,14 @@
 	</h3>
 
 	<div class="frm_grid_container frm-collapse-me">
+		<?php
+		/**
+		 * @since x.x
+		 *
+		 * @param array $field
+		 */
+		do_action( 'frm_before_field_settings', $field );
```
Crabcyborg, line 30: I don't think we need to add a new action here. Can we avoid it?

classes/views/frm-settings/captcha/captcha.php
```diff
@@ -0,0 +1,107 @@
...
+		<div class="captcha_option <?php echo esc_attr( $recaptcha_is_active ? 'active' : '' ); ?>">
+			<input type="radio" name="frm_active_captcha" id="recaptcha" value="recaptcha" <?php checked( $frm_settings->active_captcha, 'recaptcha' ); ?>>
+			<label for="recaptcha">
+				<?php
+				FrmAppHelper::icon_by_class( 'frmfont frm_recaptcha' );
+				?>
+			</label>
+		</div>
+		<div class="captcha_option <?php echo esc_attr( $recaptcha_is_active ? '' : 'active' ); ?>">
+			<input type="radio" name="frm_active_captcha" id="hcaptcha" value="hcaptcha" <?php checked( $frm_settings->active_captcha, 'hcaptcha' ); ?>>
```
Crabcyborg, line 27: It's good to include the `/` at the end of `<input />` tags.

classes/views/frm-settings/captcha/captcha_keys.php
```diff
@@ -0,0 +1,48 @@
...
+	die( 'You are not allowed to call this page directly.' );
+}
+?>
+
+<p class="howto">
+	<?php
+	if ( $captcha === 'recaptcha' ) {
+		printf(
+			/* translators: %1$s: Start link HTML, %2$s: End link HTML */
+			esc_html__( 'reCAPTCHA requires a Site and Private API key. Sign up for a %1$sfree reCAPTCHA key%2$s.', 'formidable' ),
```
Crabcyborg, line 12: These are more-or-less the same message. We could make the captcha type into a variable as well as well and reduce this.

classes/views/frm-settings/recaptcha.php
```diff
@@ -1,82 +0,0 @@
-<?php
```
Crabcyborg, line 1: We'll need to deprecate this file.

css/frm_admin.css
```diff
@@ -4429,6 +4429,86 @@ label.frm-example-icon {
...
+	border: none;
+	border-width: 0 !important;
+	display: none;
+}
+
+.captcha_option.active label:before {
+	content: '';
+	background-color: none;
+	position: absolute;
+    left: -10px;
```
Crabcyborg, line 4493: The indentation with these styles looks pretty inconsistent.

garretlaxton:
I have a page with 2 forms embedded and after completing the first form second form captcha disappears.
image
image

This happens with Google's captcha as well even though I have `Allow multiple reCAPTCHAs to be used on a single page` selected. Is that what this option is for or am I misunderstanding that?

classes/models/fields/FrmFieldCaptcha.php
AbdiTolesa: @Crabcyborg The error message is pretty inconsistent in old forms as it was set to 'The reCAPTCHA was not...' but it should rather be 'The CAPTCHA was not...' this time because the current captcha could really be hCAPTCHA and hCAPTCHA appears instead of reCAPTCHA in the form. It needed a minor update though as the new invalid message is common to both fields.

classes/views/frm-fields/back-end/settings.php
AbdiTolesa: @Crabcyborg Sure we can remove it, I just thought it could must make the field settings more customizable in case similar requirements came in the future as well.

classes/helpers/FrmFieldsHelper.php
```diff
@@ -159,13 +159,11 @@ private static function fill_cleared_strings( $field, array &$field_array ) {
...
 
-		if ( '' == $field_array['invalid'] ) {
-			if ( 'captcha' === $field->type ) {
-				$field_array['invalid'] = $frm_settings->re_msg;
-			} else {
-				/* translators: %s: Field name */
-				$field_array['invalid'] = sprintf( __( '%s is invalid', 'formidable' ), $field_array['name'] );
-			}
+		if ( 'captcha' === $field->type ) {
+			$field_array['invalid'] = $frm_settings->re_msg;
```
Crabcyborg, line 163: I'm seeing some weird issue and I think it's related to this code. I'm thinking it may be in the wrong place.

The invalid message should be using the setting.

Screen Shot 2022-11-11 at 2 17 11 PM

When I'm on this branch it's always displaying `The reCAPTCHA was not entered correctly` in that space instead of my saved custom value.

classes/models/fields/FrmFieldCaptcha.php
Crabcyborg: Screen Shot 2022-11-11 at 2 24 37 PM

We need to have the error message in one place, which is already here. It should be using the invalid message setting.

classes/views/frm-fields/back-end/field-captcha.php
```diff
@@ -3,13 +3,18 @@
...
-	?>
-<div class="howto frm_no_captcha_text"><?php
-	/* translators: %1$s: Link HTML, %2$s: End link */
-	printf( esc_html__( 'Your captcha will not appear on your form until you %1$sset up%2$s the Site and Secret Keys', 'formidable' ), '<a href="?page=formidable-settings" target="_blank">', '</a>' );
-?></div>
-<?php } ?>
-<img src="<?php echo esc_url( FrmAppHelper::plugin_url() . '/images/recaptcha.png' ); ?>" class="recaptcha_placeholder" alt="reCaptcha"/>
+$frm_settings      = FrmAppHelper::get_settings();
+$active_captcha    = $frm_settings->active_captcha;
+$captcha_not_setup = $active_captcha === 'recaptcha' && empty( $frm_settings->pubkey ) || $active_captcha === 'hcaptcha' && empty( $frm_settings->hcaptcha_pubkey );
```
Crabcyborg, line 8: I'm seeing this exact same `$captcha_not_setup = $active_captcha === 'recaptcha' && empty( $frm_settings->pubkey ) || $active_captcha === 'hcaptcha' && empty( $frm_settings->hcaptcha_pubkey );` line in two places. This should be in a function and shared in both places.

classes/views/frm-settings/captcha/captcha_keys.php
```diff
@@ -0,0 +1,48 @@
...
+		'</a>'
+	);
+	?>
+</p>
+<?php
+$prefix = $captcha === 'recaptcha' ? '' : 'hcaptcha_';
+?>
+<p class="frm6 frm_form_field">
+	<label class="frm_help" for="frm_<?php echo esc_attr( $prefix ); ?>pubkey" title="
+	<?php
```
Crabcyborg, line 31: The `title` attribute should really be a variable or something. It isn't great that it opens a tag and then line breaks and then starts PHP. It's hard to ready and easy to not notice one of the quotes and break it, etc. I generally try to avoid HTML tags that break into multiple lines. If it does it can make sense to use `FrmAppHelper::array_to_html_params` to clean it up in a lot of cases.

classes/views/frm-fields/back-end/settings.php
```diff
@@ -21,6 +21,20 @@
...
 
 	<div class="frm_grid_container frm-collapse-me">
+		<?php
+		$frm_settings      = FrmAppHelper::get_settings();
+		$active_captcha    = $frm_settings->active_captcha;
+		$captcha_not_setup = $active_captcha === 'recaptcha' && empty( $frm_settings->pubkey ) || $active_captcha === 'hcaptcha' && empty( $frm_settings->hcaptcha_pubkey );
+		if ( $field['type'] === 'captcha' && $captcha_not_setup ) {
+			echo '<div class="frm_builder_captcha frm_warning_style">';
+			FrmAppHelper::icon_by_class( 'frm_icon_font frm_alert_icon' );
+			echo '<div><b>' . esc_html__( 'Setup a captcha', 'formidable' ) . '</b>';
```
Crabcyborg, line 31: This is in a view file so it really should be HTML that isn't echoed from PHP. Can we make sure that all of the HTML is not in PHP strings here?

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -233,14 +285,69 @@ protected function should_validate() {
...
+
+		return $errors;
+	}
+
+	/**
+	 * Updates field name in page builder to the currently activated captcha if it is set to the default.
+	 *
+	 * @since x.x
+	 *
+	 * @param array $values
```
Crabcyborg, line 341: It's good to always include the `@return` comment too whenever possible.

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -233,14 +285,69 @@ protected function should_validate() {
 	}
 
 	protected function send_api_check( $frm_settings ) {
+		if ( $frm_settings->active_captcha === 'recaptcha' ) {
```
Crabcyborg, line 288: I think we could pull more of the type checking out of this function into a new one.

It would be nice if could have a standardized object for this, like a class, that holds the secret, token field, and endpoint so we can just get ` $captcha_settings` object and access it the same for both types.

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -87,16 +111,20 @@ public function front_field_input( $args, $shortcode_atts ) {
...
+			$api_js_url = $this->hcaptcha_api_url();
+		}
 
-		wp_register_script( 'recaptcha-api', $api_js_url, array( 'formidable' ), '3', true );
-		wp_enqueue_script( 'recaptcha-api' );
+		wp_register_script( 'captcha-api', $api_js_url, array( 'formidable' ), '3', true );
+		wp_enqueue_script( 'captcha-api' );
 	}
 
-	protected function api_url() {
```
Crabcyborg, line 96: I'd still keep `api_url` rather than deprecate it. It could just handle the type checking and calling the two new functions underneath.

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -64,21 +64,45 @@ protected function extra_field_opts() {
...
+
+	/**
+	 * Returns true if a the active captcha is configured properly.
+	 *
+	 * @param object $frm_settings
+	 *
+	 * @return bool
+	 */
+	private static function active_captcha_key_is_set( $frm_settings ) {
+		return ( $frm_settings->active_captcha === 'recaptcha' && empty( $frm_settings->pubkey ) ) || ( $frm_settings->active_captcha === 'hcaptcha' && empty( $frm_settings->hcaptcha_pubkey ) );
```
Crabcyborg, line 82: This line is quite large. It helps for readability to break it up into pieces, and it's safer too.

In your example you're checking too much. If `$frm_settings->active_captcha === 'recaptcha'` it should return false right away when `$frm_settings->pubkey` is empty but here it isn't. Instead it checks the active captcha type again for hcaptcha.

AbdiTolesa: @crabcyborg I actually have replaced this with` FrmFieldCaptcha::should_show_captcha` as it does pretty much the same a better way but I realized that my last commits were not actually pushed for some reason. I just synced my local with remote now.

classes/models/fields/FrmFieldCaptcha.php
AbdiTolesa: @Crabcyborg We can do that but we still need to check the active captcha even though we move the code to a new class, right? I mean I couldn't see how that simplifies the code if so, or may be I didn't get your point well?

classes/views/frm-fields/back-end/settings.php
AbdiTolesa: This is done, thanks!

classes/views/frm-settings/captcha/captcha_keys.php
AbdiTolesa: Thank you for the feedback and letting me know a better way to do this in the future!

classes/views/frm-fields/back-end/field-captcha.php
AbdiTolesa: @Crabcyborg I just made `FrmFieldCaptcha::should_show_captcha` public and static and reuse it in different places, thank you!!

classes/helpers/FrmFieldsHelper.php
AbdiTolesa: Right, I think there is no need to touch this function that much.

classes/models/FrmField.php
```diff
@@ -9,6 +9,16 @@ class FrmField {
 	public static $transient_size = 200;
 
 	public static function field_selection() {
+		$frm_settings   = FrmAppHelper::get_settings();
+		$active_captcha = $frm_settings->active_captcha;
+		$captcha_is_set = ( $active_captcha === 'recaptcha' && ! empty( $frm_settings->pubkey ) ) || ( $active_captcha === 'hcaptcha' && ! empty( $frm_settings->hcaptcha_pubkey ) );
```
Crabcyborg, line 14: This should probably also use `should_show_captcha` rather than being duplicate code everywhere.

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -205,7 +249,7 @@ public function validate( $args ) {
 		}
 
 		// phpcs:ignore WordPress.Security.NonceVerification.Missing
-		if ( ! isset( $_POST['g-recaptcha-response'] ) ) {
+		if ( ! isset( $_POST['g-recaptcha-response'] ) || ! isset( $_POST['h-captcha-response'] ) ) {
```
Crabcyborg, line 252: This line confuses me. There should really only be a `g-recaptcha-response` or a `h-captcha-response` response, right? This is checking if both of them are not set but if one is set, this would still be false.

classes/models/fields/FrmFieldCaptcha.php
Crabcyborg: I just want to keep functions small and reusable.

I think we could basically do something like `$captcha_settings = self::get_active_captcha_settings()` that returns a `stdClass` with secret, token_field and endpoint.

It still needs the check somewhere but it shouldn't be inside of `send_api_check` as it's really intended to send the api check, not check the active settings. I'm just trying to reduce scope.

AbdiTolesa: That's clear now, I just refactored it accordingly.

classes/models/FrmField.php
AbdiTolesa: Definitely, don't know how it escaped my eyes.

classes/models/fields/FrmFieldCaptcha.php
AbdiTolesa: This was wrong indeed, I have corrected it now, thank you!

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -160,21 +199,26 @@ protected function validate_against_api( $args ) {
...
 			}
 		}
 
 		if ( isset( $response['success'] ) && ! $response['success'] ) {
 			// What happens when the CAPTCHA was entered incorrectly
-			$invalid_message                 = FrmField::get_option( $this->field, 'invalid' );
-			$errors[ 'field' . $args['id'] ] = ( $invalid_message == '' ? $frm_settings->re_msg : $invalid_message );
+			$invalid_message = FrmField::get_option( $this->field, 'invalid' );
+			if ( $invalid_message === __( 'The reCAPTCHA was not entered correctly', 'formidable' ) ) {
+				$invalid_message = __( 'The CAPTCHA was not entered correctly', 'formidable' );
```
Crabcyborg, line 219: We could use `re_msg` here to avoid duplicate translations, right?

Crabcyborg: Or even just set it to `''` since it seems the code under this sets it to `re_msg` anyway.

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -205,7 +249,7 @@ public function validate( $args ) {
 		}
 
 		// phpcs:ignore WordPress.Security.NonceVerification.Missing
-		if ( ! isset( $_POST['g-recaptcha-response'] ) ) {
+		if ( ! ( isset( $_POST['g-recaptcha-response'] ) || isset( $_POST['h-captcha-response'] ) ) ) {
```
Crabcyborg, line 252: Instead of wrapping this in more brackets I think it's ultimately more readable as `if ( ! isset( $_POST['g-recaptcha-response'] ) && ! isset( $_POST['h-captcha-response'] ) ) {`.

classes/models/fields/FrmFieldCaptchaSettings.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
Crabcyborg, line 1: I'm thinking this makes more sense in `models` instead of `models/fields`. If you check out the `models/fields` in Pro you can see how it's really just used for the things that extend `FrmFieldType`. Everything else just goes into the models folder.

Crabcyborg:
@garretlaxton I'm planning to merge and release this hopefully this week.

Could you test it once more? Thank you!

Crabcyborg (changes requested):
@AbdiTolesa The new SVGs for reCAPTCHA/hCAPTCHA are a bit bigger than they need to be. Could you run those SVGs through https://jakearchibald.github.io/svgomg/ and use the optimized SVG data instead?

Thank you!

I also pushed a small update to optimize the images. I generally try to run them through https://imagecompressor.com/ first so they don't take up more space than necessary.

garretlaxton:
I'm seeing this warning in the form builder when there's a captcha in the form.
`PHP Warning:  Undefined variable $image_name in localhost/formi/wp-content/plugins/formidable/classes/views/frm-fields/back-end/field-captcha.php on line 6`

AbdiTolesa:
> I'm seeing this warning in the form builder when there's a captcha in the form. `PHP Warning: Undefined variable $image_name in localhost/formi/wp-content/plugins/formidable/classes/views/frm-fields/back-end/field-captcha.php on line 6`

This is fixed now.

AbdiTolesa:
> 

Good to know that, @Crabcyborg. Thank you for optimizing the SVGs!

Crabcyborg:
> Thank you for optimizing the SVGs!

@AbdiTolesa I didn't. I only optimized the images. I asked if you could update the SVGs, if you don't mind. I don't have the original Figma design to copy from so I left the SVGs.

AbdiTolesa:
> > Thank you for optimizing the SVGs!
> 
> @AbdiTolesa I didn't. I only optimized the images. I asked if you could update the SVGs, if you don't mind. I don't have the original Figma design to copy from so I left the SVGs.

Oh, sorry for getting confused, I would like to do that definitely, @Crabcyborg

garretlaxton (commented):
This is looking good to me. Has anything changed with the way reCAPTCHA should work?

AbdiTolesa:
> This is looking good to me. Has anything changed with the way reCAPTCHA should work?

Not really, the last updates were just style changes, anything else you observed?

garretlaxton (approved):
Nope, this is functioning fine for me.

classes/views/frm-settings/captcha/captcha.php
```diff
@@ -0,0 +1,109 @@
...
+		<label for="frm_re_multi">
+			<input type="checkbox" name="frm_re_multi" id="frm_re_multi"
+			value="1" <?php checked( $frm_settings->re_multi, 1 ); ?> />
+			<?php esc_html_e( 'Allow multiple reCAPTCHAs to be used on a single page', 'formidable' ); ?>
+		</label>
+	</p>
+</div>
+
+<div id="hcaptcha_settings" class="frm_grid_container <?php echo esc_attr( $recaptcha_is_active ? 'frm_hidden' : '' ); ?>">
+	<h3>hCaptcha Settings</h3>
```
Crabcyborg, line 104: @AbdiTolesa This string here isn't translated. Everything should go through `esc_html_e`.

classes/views/frm-settings/captcha/captcha.php
```diff
@@ -0,0 +1,109 @@
...
+	</div>
+</div>
+
+<div class="alert frm_hidden">
+	<?php FrmAppHelper::icon_by_class( 'frm_icon_font frm_tooltip_icon' ); ?>
+	<span><?php esc_html_e( 'Changing the captcha type here will replace it in all any forms where it is used.', 'formidable' ); ?></span>
+</div>
+
+<div id="recaptcha_settings" class="frm_grid_container <?php echo esc_attr( $recaptcha_is_active ? '' : 'frm_hidden' ); ?>">
+	<h3>reCaptcha Settings</h3>
```
Crabcyborg, line 45: Same as my other comment but `reCAPTCHA` settings also isn't translated here.

PR #959, opened 2022-10-31
add extra class specific to entry page

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/3896
I think adding extra class that doesn't load on form builder page makes sense to avoid styles that affects form builder page.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -3345,13 +3345,13 @@ input[type="checkbox"] {
 	margin-right: 5px;
 }
 
-.with_frm_style .frm_radio input[type=radio] {
+.frm_single_entry_page .with_frm_style .frm_radio input[type=radio] {
 	margin: 5px 0;
 	width: 18px;
 	position: relative; /* override Bootstrap */
 }
 
-.with_frm_style .frm_radio input[type=radio]:before {
+.frm_single_entry_page .with_frm_style .frm_radio input[type=radio]:before {
 	position: relative !important;
 	left: calc(50% - 6px);
 	top: calc(50% - 6px);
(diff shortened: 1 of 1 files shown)


AbdiTolesa:
> @AbdiTolesa what about the width change that was made in the last update? We probably want to only change it on this page as well, right?
> 
> Screen Shot 2022-11-01 at 11 51 41 AM

oh yes, I have seen that effect  but not sure why I forgot to add the class there as well, @Crabcyborg. It is updated to inlcude that now, thank you!

Crabcyborg:
@AbdiTolesa I still don't think this looks right.

Prior to that last update radios and checkboxes had the same size set. I think we should put that back as well:

```
.with_frm_style .frm_radio input[type=radio],
.with_frm_style .frm_checkbox input[type=checkbox] {
	width: 25px !important;
}
```

AbdiTolesa:
> @AbdiTolesa I still don't think this looks right.
> 
> Prior to that last update radios and checkboxes had the same size set. I think we should put that back as well:
> 
> ```
> .with_frm_style .frm_radio input[type=radio],
> .with_frm_style .frm_checkbox input[type=checkbox] {
> 	width: 25px !important;
> }
> ```

@Crabcyborg I think that was done on purpose to prevent the radio buttons from looking stretched horizontally. D

> @AbdiTolesa I still don't think this looks right.
> 
> Prior to that last update radios and checkboxes had the same size set. I think we should put that back as well:
> 
> ```
> .with_frm_style .frm_radio input[type=radio],
> .with_frm_style .frm_checkbox input[type=checkbox] {
> 	width: 25px !important;
> }
> ```

@Crabcyborg Yes I think that was forgotten in the old PR, that is covered with this now.

PR #970, opened 2022-11-04
Make infoModal() global

truongwp:
This is used for the ACF addon.


Crabcyborg (approved):
Thanks @truongwp this looks nice and simple!

PR #979, opened 2022-11-11
Filter cached form objects

Crabcyborg:
The way this works right now makes `frm_form_object` fairly unreliable.

It sets the result to a cache, then filters it.

Screen Shot 2022-11-11 at 11 08 47 AM

Prior to this update it would then pull the result from cache and then just return it unfiltered. Meaning that any cached results don't really work the same as uncached results.

Screen Shot 2022-11-11 at 11 09 14 AM


stephywells (approved):
Great find!

PR #980, opened 2022-11-11
Do not call triggerCustomEvent on duplicate entry error

Crabcyborg:
Related fix https://github.com/Strategy11/formidable-stripe/pull/181

So it turns out the reason I'm seeing the error I was getting after reloading was because I was submitting a duplicate entry. I didn't realize that instead of passing errors that duplicate entry errors just appear in the content. I'm treating that error case as a success case right now when it really isn't.

I'm not sold this is the final solution though. It may make sense to have the duplicate error come through as an AJAX error? I'm going to look into that next.

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -620,7 +620,7 @@ function frmFrontFormJS() {
 			} else if ( response.content !== '' ) {
 				// the form or success message was returned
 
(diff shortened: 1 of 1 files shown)


Crabcyborg:
Fixed with https://github.com/Strategy11/formidable-pro/pull/3925 instead.

PR #982, opened 2022-11-15
change let and const to var

AbdiTolesa:
This is a quick fix to make sure errors are not thrown from our script in IE,


Crabcyborg (approved):

PR #984, opened 2022-11-17
Avoid a fatal error if hidden columns are a string

Crabcyborg:
I was looking at the Support for a second in WordPress and noticed this issue https://wordpress.org/support/topic/frmentriescontroller-php/.

I haven't seen it come up in support but every time I see a fatal error with the stack trace I try to look at the code for a better idea.

This is simple to replicate with a basic snippet but I'm not sure how it comes up naturally. Possibly the option is a string, or another plugin is filtering the value and returning a string when it shouldn't.

```
add_filter(
	'get_user_option_manageformidable_page_formidable-entriescolumnshidden',
	function() {
		return '';
	},
	9
);
```

I added a unit test that also covers this case. And I repeated the fix for the forms list which can have the same issue.

### DIFF
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -306,7 +306,18 @@ class FrmEntriesController {
 		return apply_filters( 'frm_field_column_is_sortable', $is_sortable, $field );
 	}
 
+	/**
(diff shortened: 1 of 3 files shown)


stephywells (approved):
Awesome! This looks like a great fix to avoid conflicts.

PR #987, opened 2022-11-22
Fix in_section value is unreliable

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3024


Crabcyborg:
Thanks for looking into this @truongwp! I took a deeper look into it. Your update would cause it to update after every time a field moves that isn't in a section, which would be too often.

The issue is really that the `previousSectionId` value wasn't accurate so it wasn't matching the `!==` check. They were both 0.

I have two new related pull requests I'd like you to review since you're familiar with this issue now.

https://github.com/Strategy11/formidable-forms/pull/994 and https://github.com/Strategy11/formidable-pro/pull/3948

One update fixes how I get the value of `previousSectionId` and it also passes a `previousFormId` value now to the action so that the Pro update can delete the field transient to avoid the field moved outside of a repeater from appearing in both the repeater and outside of the repeater.

PR #989, opened 2022-11-24
Run JS code in script tags after adding form action

truongwp:
Fixes https://github.com/Strategy11/formidable-acf/issues/37

### DIFF
--- a/js/admin/dom.js
+++ b/js/admin/dom.js
@@ -385,6 +385,30 @@
 				callback( event );
 			};
 			element.addEventListener( 'click', listener );
+		},
+		/**
+		 * Sets innerHTML with JS code inside <script> tags are executed.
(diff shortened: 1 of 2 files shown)


Crabcyborg:
@garretlaxton Could you help test this? I haven't tried it yet myself. I know there's separate code for opening a form action that already exists, so I'd make sure that that works as well.

And this applies to all new form actions so we'll want to make sure that none of the others end up with any weird side effect if you could try playing with several types of actions, not just post actions.

Thank you!

garretlaxton:
I tested the Register user, hubspot, and quiz outcome actions so far. I notice the following error in my console when I go to formidable settings and try to switch to a new tab. It also doesn't switch tabs and just stays on the general settings page.
image

PR #991, opened 2022-11-25
Update pencil icon

Crabcyborg:
I'm trying to work toward some better icon consistency as I work on the visual styler. The thicket pencil doesn't look great in the dropdown, but I think it may be worth avoiding adding another icon for the same thing when we can just replace our current one. It's a bit smaller and has less detail. I looked at the code and identified it's used in a few places. I think it looks okay.

I'm not sure if we want to just change all icons at once but I figure it's easier to do this one at a time since each one could be used in multiple places. Do we have updated icons for all of these in one of the designs?

It's also used when you edit a summary, but the SVG is separate code so it doesn't get updated with this update. If we want to change it there as well, it isn't difficult to update In Pro too.

Or maybe we want to leave the old icon alone for now?

**Before**

Screen Shot 2022-11-25 at 12 00 05 PM

**In the Edit Entry sidebar link.**

Screen Shot 2022-11-25 at 11 49 13 AM

**In the Other option setting in the form builder.**

Screen Shot 2022-11-25 at 11 44 30 AM


stephywells (approved):
This looks good to me. Thanks for making the update!

PR #993, opened 2022-11-28
use message from url if not set already

AbdiTolesa:
This PR goes along https://github.com/Strategy11/formidable-pro/pull/3935 to fix https://github.com/Strategy11/formidable-pro/issues/3886. We want to show the message an entry was duplicated successfully.

### DIFF
--- a/classes/views/frm-entries/errors.php
+++ b/classes/views/frm-entries/errors.php
@@ -8,8 +8,16 @@ if ( isset( $include_extra_container ) ) { ?>
 	<?php
 }
 
-if ( empty( $message ) && FrmAppHelper::simple_get( 'message' ) !== '' ) {
-	$message = FrmAppHelper::simple_get( 'message' );
+if ( empty( $message ) && FrmAppHelper::simple_get( 'frm_message' ) ) {
+
+	/**
+	 * Updates message in an entry edit page.
+	 *
+	 * @since 5.x
+	 *
+	 * @param string $message
+	 */
+	$message = apply_filters( 'frm_message', '' );
 }
 
 if ( isset( $message ) && $message != '' ) {
--- a/classes/views/frm-entries/errors.php
+++ b/classes/views/frm-entries/errors.php
@@ -8,20 +8,6 @@ if ( isset( $include_extra_container ) ) { ?>
 	<?php
 }
 
-$message_code = FrmAppHelper::simple_get( 'frm_entries_updated_message' );
-if ( empty( $message ) && $message_code ) {
-
-	/**
-	 * Updates message in an entry edit page.
-	 *
-	 * @since 5.x
-	 *
-	 * @param string $message
-	 * @param string $frm_entries_updated_message
-	 */
-	$message = apply_filters( 'frm_entries_updated_message', '', $message_code );
-}
-
 if ( isset( $message ) && $message != '' ) {
 	if ( FrmAppHelper::is_admin() ) {
 		?>
--- a/classes/views/frm-entries/errors.php
+++ b/classes/views/frm-entries/errors.php
@@ -8,7 +8,8 @@ if ( isset( $include_extra_container ) ) { ?>
 	<?php
 }
 
-if ( empty( $message ) && FrmAppHelper::simple_get( 'frm_message' ) ) {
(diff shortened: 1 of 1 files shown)


classes/views/frm-entries/errors.php
```diff
@@ -7,6 +7,11 @@
 <div class="<?php echo esc_attr( $include_extra_container ); ?>" id="frm_form_<?php echo esc_attr( $form->id ); ?>_container">
 	<?php
 }
+
+if ( empty( $message ) && FrmAppHelper::simple_get( 'message' ) !== '' ) {
+	$message = FrmAppHelper::simple_get( 'message' );
```
Crabcyborg, line 12: I think we could probably come up with a less exploitable way of passing the message. I don't like the idea of having this here where anyone could force any message in the URL. Possibly an error code so it would only allow a single value.

Or we may want a filter somewhere so we could set the message from another add on.

AbdiTolesa: @Crabcyborg I think the idea of using a hook might not work as the page is redirected already. I want to try setting an error code to prevent more than one value of the message param, is there probably an example I can refer to in our code that does something similar?

Crabcyborg: @AbdiTolesa I'm picturing something like this switch/case in Stripe that checks for expected error codes and shows messages for them.

https://github.com/Strategy11/formidable-stripe/blob/9609d98ecfd8b069b03070c4448bf8ad71cb7efe/models/FrmStrpAuth.php#L121

The error message shouldn't be in lite though. We could filter the message if there is a key set, and handle the translation in Pro.

AbdiTolesa: @Crabcyborg I have updated this to check for a message code in the url and add a filter hook if the key is set, which I hooked a function that decodes the message from pro. I'm not quite sure if that is how you want it implemented. Let me know if I'm a bit off the track yet.

classes/views/frm-entries/errors.php
```diff
@@ -7,6 +7,19 @@
 <div class="<?php echo esc_attr( $include_extra_container ); ?>" id="frm_form_<?php echo esc_attr( $form->id ); ?>_container">
 	<?php
 }
+
+if ( empty( $message ) && FrmAppHelper::simple_get( 'frm_message' ) ) {
```
Crabcyborg, line 11: @AbdiTolesa Maybe we should always call the filter. I think we could do something like `$message_code = FrmAppHelper::simple_get( 'frm_message' );` and then pass that into a second parameter into the new filter so it can be used in Pro without having to call `FrmAppHelper::simple_get( 'frm_message' )` a second time.

I also don't love the name of the `frm_message` filter. I think it could probably be more descriptive. Something more like `frm_entries_updated_message` seems more fitting.

AbdiTolesa: @Crabcyborg Thank you for your time reviewing it and suggesting how to make it cleaner! I have updated it accordingly.

classes/views/frm-entries/errors.php
```diff
@@ -7,6 +7,21 @@
 <div class="<?php echo esc_attr( $include_extra_container ); ?>" id="frm_form_<?php echo esc_attr( $form->id ); ?>_container">
 	<?php
 }
+
+$message_code = FrmAppHelper::simple_get( 'frm_entries_updated_message' );
+if ( empty( $message ) && $message_code ) {
+
+	/**
+	 * Updates message in an entry edit page.
```
Crabcyborg, line 15: Thanks for updating this @AbdiTolesa! This looks pretty close now. But I still think we'd probably benefit to make it better.

What do you think about moving this out of the view entirely so it's all code that renders this view. This isn't really code that makes sense in a view as it's all logic, no HTML.

And that way we can break this apart more and make it more readable too.

AbdiTolesa: I have moved that part to pro @Crabcyborg so I think we can delete this branch now.

Crabcyborg: Oh perfect, thanks @AbdiTolesa that's great that we can keep it all in one plugin. I'll go ahead and close this.

PR #995, opened 2022-12-01
Fix plugin connection always falling

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3947

@Crabcyborg Somehow, the unit test fails. The code before `$wp_roles->add_cap()` runs correctly (capabilities should be added to role `administrator`). But that method seems not to run. I printed some strings inside that method, but nothing showed in the terminal.

### DIFF
--- a/classes/models/FrmAddon.php
+++ b/classes/models/FrmAddon.php
@@ -251,6 +251,10 @@ class FrmAddon {
 	protected function update_pro_capabilities() {
 		global $wp_roles;
 
+		if ( ! function_exists( 'get_editable_roles' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/user.php';
+		}
+
 		$caps     = FrmAppHelper::frm_capabilities( 'pro_only' );
 		$roles    = get_editable_roles();
 		$settings = new FrmSettings();
--- a/tests/misc/test_FrmAddon.php
+++ b/tests/misc/test_FrmAddon.php
@@ -83,6 +83,17 @@ class test_FrmAddon extends FrmUnitTest {
 
 		}
 	}
+
+	public function test_update_pro_capabilities() {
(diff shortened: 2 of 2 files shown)


truongwp:
@Crabcyborg Yes. But I couldn't find out why it's failing. So I created this PR to seek help. I tried printing data before this line https://github.com/Strategy11/formidable-forms/pull/995/files#diff-19643824179376f80b6730a6f29427bbd5d4bc3d445c40cb89c7b8f8edc781c9R271. The data is correct. But nothing shows when I print data inside the `add_cap()` method.

Crabcyborg:
Thanks @truongwp I think I looked at this too quickly and missed your description. I'll take a look!

Crabcyborg:
@truongwp Could you take a look at my update to the unit test? I couldn't get it to fail locally without removing the caps first. Then it seems to be an issue with the `role_objects` being cached. Calling `new WP_Roles()` resolves that for me.

Do you think that caching would be an issue after the function is called? Would it make more sense to include that `new WP_Roles()` code, or something similar, in the update function itself?

Crabcyborg:
I'm merging this. Regardless of my question above, I think this stands alone and fixes the main issue. I'm still curious if it makes sense to update the update function though as well.

Thanks for your help @truongwp!

truongwp:
@Crabcyborg Thanks for working on this. I see 2 `get_role()` lines. Is that intentional?

For your concern, I think we should ask Garret to test.

Crabcyborg:
@truongwp Yeah, these objects hold onto old inaccurate data everywhere so I need to get a fresh object or it still fails the capability check.

I think we're fine with this. It's all just short term memoization so it would resolve immediately. I don't think you need those permissions in the same request where you're activating a license.

PR #996, opened 2022-12-01
Add JS hooks after adding and opening form action

truongwp:


js/formidable_admin.js
```diff
@@ -6430,6 +6439,15 @@ function frmAdminBuildJS() {
 			if ( widgetTop ) {
 				jQuery( widgetTop ).trigger( 'frm-action-loaded' );
```
Crabcyborg, line 6440: @truongwp Is it not possible to hook into `jQuery( b ).trigger( 'frm-action-loaded' );`? It's called in both of these places. Adding a second way to hook into here might complicate things if it isn't necessary.

But if you need this, I don't see a big problem with it either.

truongwp: @Crabcyborg I just avoid using jQuery. And if using `jQuery.on()`, we need to query the `.widget-top`.

Crabcyborg (approved):

PR #998, opened 2022-12-02
Add more backend validation for frm_install and doJsonFetch, remove h…

Crabcyborg:
…idden id from frm_install_message

Related ticket https://secure.helpscout.net/conversation/2085567282/123503/

This updates adds a nonce and permission check for the `FrmAppController::ajax_install` function. The function doesn't take any user input. At most it could be abused to constantly try triggering the install process that gets used when this spinner is added to the page:

Testing this is quite easy as you can just add a custom snippet to trigger the code: `add_action( 'admin_notices', 'FrmAppController::install_js_fallback' );`.

Screen Shot 2022-12-02 at 10 26 17 AM

This update also fixes that spinner which wasn't previously visible because it had 2 IDs, one called "hidden" which just seems to be a bug. I looked at the git blame but this line hasn't changed in 8 years 😂.

The report also mentions the application data GET endpoint, though it later mentions that its read only and that it could be ignored. I added a nonce into `doJsonFetch` though and added the check for applications data, just to have the improved coverage. Everywhere I call this in Pro is through `doJsonFetch` so it should just work without any other changes.

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -380,7 +380,10 @@ class FrmAppController {
 	 */
 	public static function install_js_fallback() {
 		FrmAppHelper::load_admin_wide_js();
-		echo '<div id="hidden frm_install_message"></div><script type="text/javascript">jQuery(document).ready(function(){frm_install_now();});</script>';
+		?>
+			<div id="frm_install_message"></div>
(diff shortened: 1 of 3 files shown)


stephywells (approved):
This looks good to me! Thank you Mike!

PR #1002, opened 2022-12-05
Update stripe icon

Crabcyborg:
Following up from https://github.com/Strategy11/formidable-stripe/pull/199

This update replaces the current `frm_stripe_icon` with the smaller square one.

The `frm_stripe_icon` doesn't really appear to be used anywhere so I don't think replacing it should be a problem.

**New icon**
Screen Shot 2022-12-05 at 4 55 34 PM


stephywells:
@Crabcyborg Thank you! That screenshot doesn't look quite right. The icon is cut off a bit on the bottom, and isn't vertically aligned. Maybe this one would work better?
https://images.ctfassets.net/fzn2n1nzq965/HTTOloNPhisV9P4hlMPNA/cacf1bb88b9fc492dfad34378d844280/Stripe_icon_-_square.svg?q=80&w=1082

Crabcyborg:
Thanks @stephywells! The 40px border radius looks pretty different, like a circle, when I try applying it to the SVG. I tried to get it close to what it looks like in Figma, which is closer to around 5px I think.

We could probably fix it in the SVG but I think it's simpler with just a CSS `border-radius` rule.

Screen Shot 2022-12-06 at 3 37 59 PM

stephywells:
Good idea! That looks great!

stephywells (approved):

PR #1005, opened 2022-12-07
try to get style id from database even if name is numeric

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/2242
This update fixes the issue of style not getting applied to the form it should on by trying the same action after the forms are imported plus getting around the cache issue in `FrmXMLHelper::update_custom_style_setting_on_import` when `FrmDb::get_var` is called clearing the cache saved with the key generated from the database query.

### DIFF
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -746,10 +746,10 @@ class FrmXMLHelper {
 			);
 			$select   = 'ID';
 
-			$style = $wpdb->get_row( $wpdb->prepare( "SELECT ID FROM $table WHERE post_type=%s AND post_name=%s", 'frm_styles', $form['options']['custom_style'] ) );
+			$style_id = $wpdb->get_var( $wpdb->prepare( "SELECT ID FROM $table WHERE post_type=%s AND post_name=%s", 'frm_styles', $form['options']['custom_style'] ) );
 
-			if ( $style ) {
-				$form['options']['custom_style'] = $style->ID;
+			if ( $style_id ) {
+				$form['options']['custom_style'] = $style_id;
 			} else {
 				// save the old style to maybe update after styles import
 				$form['options']['old_style'] = $form['options']['custom_style'];
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -882,8 +882,10 @@ class FrmXMLHelper {
 		}
 		unset( $posts_with_shortcodes, $view_ids );
 
-		// clear imported forms style cache to make sure the new styles are applied to the forms
-		self::clear_forms_style_caches( $imported['forms'] );
+		if ( ! empty( $imported['forms'] ) ) {
+			// clear imported forms style cache to make sure the new styles are applied to the forms
+			self::clear_forms_style_caches( $imported['forms'] );
+		}
 
 		self::maybe_update_stylesheet( $imported );
 
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -881,24 +881,29 @@ class FrmXMLHelper {
 		}
 		unset( $posts_with_shortcodes, $view_ids );
 
+		$where = array(
+			'id' => $imported['forms'],
+			'options LIKE' => '"old_style"',
+		);
+		$forms = FrmDb::get_results( 'frm_forms', $where );
+
+		foreach ( $forms as $form ) {
+			$form_options = unserialize( $form->options );
+
+			$where = array(
+				'post_name' => $form_options['old_style'],
+				'post_type' => 'frm_styles',
+			);
+
+			$select = 'ID';
+
+			$cache_key = FrmDb::generate_cache_key( $where, array( 'limit' => 1 ), $select, 'var' );
+			FrmDb::delete_cache_and_transient( $cache_key, 'post' );
+		}
+
 		self::maybe_update_stylesheet( $imported );
 
 		flush_rewrite_rules();
-
-		foreach ( $imported['forms'] as $form_id ) {
-			$form  = FrmForm::getOne( $form_id );
-			if ( ! isset( $form->options['old_style'] ) ) {
-				continue;
-			}
-			$where = array(
-				'post_name' => $form->options['old_style'],
-				'post_type' => 'frm_styles',
-			);
-			$select    = 'ID';
-			$cache_key = FrmDb::generate_cache_key( $where, array( 'limit' => 1 ), $select, 'var' );
-			FrmDb::delete_cache_and_transient( $cache_key, 'post' );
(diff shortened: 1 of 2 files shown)


classes/helpers/FrmXMLHelper.php
```diff
@@ -745,17 +745,11 @@ private static function update_custom_style_setting_on_import( &$form ) {
 				'post_type' => 'frm_styles',
 			);
 			$select   = 'ID';
-			$style_id = FrmDb::get_var( $table, $where, $select );
```
Crabcyborg, line 742: @AbdiTolesa it seems like the issue here may be caching related. It looks like the original query really isn't any different than the one using `$wpdb`.

`FrmDb::get_var` uses a `FrmDb::check_cache` function. We may just need to clear that cache value after the style imports?

AbdiTolesa: @Crabcyborg I was able to do achieve that but clearing the cache required me generating a cache key and that function (`FrmDb::generate_cache_key`) need to be changed to public from private and I have done that but not confident if that is the way to go. Does that sound safe for you?

Crabcyborg: @AbdiTolesa I think that could possibly be far too often if we had it there in the code.

Rather than having it happen before every single imported form (which could be thousands of times for a single style), I think we'd want to try to trigger it after we import style. Despite its name, `import_xml_views` is really the function that imports all posts in the XML, which includes styles and form actions, as well as actual posts and pages.

Can we clear the cache there instead so it happens once per style rather than once per form?

AbdiTolesa: I have moved the patch down to `import_xml_view` and I checked that only happens once for the import I'm using. Is that probably as you were imagining now?

Crabcyborg: @AbdiTolesa Not quite. Is it possible to do this still when it's importing the form, but with the cache clearing after it imports the style?

This loop looks really bad for performance. I imagine this could be slower than what you had before.

AbdiTolesa: @Crabcyborg I have tried doing that while it is importing the form to avoid repeating the loop but that didn't work out and that's why I looped over the forms again after all imports are finished. We can keep the cache clearing where it is now ( in `import_xml_view` ) but move the loop back to where it was (after all imports are done) possibly.

Crabcyborg: Thanks @AbdiTolesa.

Maybe we can just try to reduce this more. I want to avoid `FrmForm::getOne( $form_id );` in a loop.

Instead we could try to use `FrmDb::get_results` and pass the whole array of `$imported['forms']` ids into a single query, with a second check if `old_style` exists. So we'd only pull data once, and only for the forms with an old style key.

Also, this gets called from `maybe_update_stylesheet` so I think we need to make sure it doesn't update anything twice. If it gets updated there, it means we don't need to update here again here, right? Or maybe we want to remove some of that previous code instead?

AbdiTolesa: @Crabcyborg I found out that clearing the cache before `maybe_update_stylesheet` is called is enough at the end! I reduced the db trip to one based on your suggestion. We could actually avoid that if we do `FrmDb::cache_delete_group( 'post' )` but that might be worse as it could lead to extra trips to the database on the go.

classes/helpers/FrmXMLHelper.php
```diff
@@ -882,6 +882,26 @@ public static function import_xml_views( $views, $imported ) {
...
 		unset( $posts_with_shortcodes, $view_ids );
 
+		$where = array(
+			'id' => $imported['forms'],
+			'options LIKE' => '"old_style"',
+		);
+		$forms = FrmDb::get_results( 'frm_forms', $where );
+
+		foreach ( $forms as $form ) {
+			$form_options = unserialize( $form->options );
```
Crabcyborg, line 892: We have a `FrmAppHelper::unserialize_or_decode` function that we usually use instead of `unserialize`. And it would be best to check that the result is actually an error to be on the safe side.

Otherwise I think this looks good!

classes/helpers/FrmXMLHelper.php
```diff
@@ -882,6 +882,26 @@ public static function import_xml_views( $views, $imported ) {
...
+
+			$where = array(
+				'post_name' => $form_options['old_style'],
+				'post_type' => 'frm_styles',
```
Crabcyborg, line 896: One other thing. Let's avoid hard coding this here. Let's use` FrmStylesController::$post_type` instead.

AbdiTolesa: Understood, will do accordingly going forward.

classes/helpers/FrmXMLHelper.php
AbdiTolesa: Great, all done!

garretlaxton (approved):
This looks good! Imported styles are now assigned to the proper form.

classes/helpers/FrmXMLHelper.php
```diff
@@ -882,6 +882,28 @@ public static function import_xml_views( $views, $imported ) {
 		}
 		unset( $posts_with_shortcodes, $view_ids );
 
+		$where = array(
```
Crabcyborg, line 885: @AbdiTolesa One more thing before we merge this.

`import_xml_views` is a huge function. And this new chunk is pretty large too. Can you move this into a new function, with a description about the issue you're fixing here? This update isn't very readable yet.

AbdiTolesa: @Crabcyborg  Surething, I just did that.

classes/helpers/FrmXMLHelper.php
```diff
@@ -882,13 +882,45 @@ public static function import_xml_views( $views, $imported ) {
 		}
 		unset( $posts_with_shortcodes, $view_ids );
 
+		// clear imported forms style cache to make sure the new styles are applied to the forms
+		self::clear_forms_style_caches( $imported['forms'] );
```
Crabcyborg, line 886: @AbdiTolesa I think this may be problematic if you don't import any forms but you import a post.

Here's a basic XML example of an application with only a page.

[application-with-a-page.xml.zip](https://github.com/Strategy11/formidable-forms/files/10315629/application-with-a-page.xml.zip)

It ends up querying for `'id' => array()` which yields all forms with an old style. Instead it should skip all of this code in that case as it isn't necessary when no forms are imported.

AbdiTolesa: Understood, I had a misunderstanding on that part when empty array is provided, thank you for helping me clear that, I have added a check before the function call now.

Crabcyborg: Thanks @AbdiTolesa!

Crabcyborg (approved):
🚀

PR #1006, opened 2022-12-08
Issue 3914

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3914

I think we can hook into `frm_submit_button_class` and `frm_back_button_class` to add extra classes if twenty twenty-one theme is active.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1347,7 +1347,7 @@ class FrmFormsController {
 	}
 
 	/**
-	 * Updates classes used in submit buttons.
+	 * Updates classes used in submit and prev buttons to avoid conflict with twenty twenty-one theme.
 	 *
 	 * @param array $classes
 	 *
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1346,6 +1346,14 @@ class FrmFormsController {
 		}
 	}
 
+	public static function update_submit_button_classes( $classes ) {
+		if ( wp_get_theme()->get('Name') === 'Twenty Twenty-One' ) {
+			$classes[] = 'has-text-color has-background';
+		}
+	
+		return $classes;
+	}
+
 	public static function mb_tags_box( $form_id, $class = '' ) {
 		$fields = FrmField::get_all_for_form( $form_id, '', 'include' );
 
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1354,7 +1354,7 @@ class FrmFormsController {
 	 * @return array
 	 */
 	public static function update_classes( $classes ) {
-		if ( wp_get_theme()->get( 'Name' ) === __( 'Twenty Twenty-One', 'formidable' ) ) {
+		if ( function_exists( 'twenty_twenty_one_setup' ) ) {
 			$classes[] = 'has-text-color has-background';
 		}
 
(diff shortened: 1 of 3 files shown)


classes/controllers/FrmFormsController.php
```diff
@@ -1346,6 +1346,21 @@ private static function clean_submit_html( &$values ) {
...
 
+	/**
+	 * Updates classes used in submit and prev buttons to avoid conflict with twenty twenty-one theme.
+	 *
+	 * @param array $classes
+	 *
+	 * @return array
+	 */
+	public static function update_classes( $classes ) {
+		if ( wp_get_theme()->get( 'Name' ) === __( 'Twenty Twenty-One', 'formidable' ) ) {
```
Crabcyborg, line 1357: I don't believe theme names are translated. If it was, it wouldn't really be in the `formidable` text domain either way. I think removing the `__` call here probably makes sense.

classes/controllers/FrmFormsController.php
```diff
@@ -1346,6 +1346,21 @@ private static function clean_submit_html( &$values ) {
...
 	}
 
+	/**
+	 * Updates classes used in submit and prev buttons to avoid conflict with twenty twenty-one theme.
+	 *
+	 * @param array $classes
+	 *
+	 * @return array
+	 */
+	public static function update_classes( $classes ) {
```
Crabcyborg, line 1356: This `update_classes` function name isn't very descriptive. This is specific for updating button classes, which should be mentioned in the function name.

classes/controllers/FrmHooksController.php
```diff
@@ -69,6 +69,8 @@ public static function load_hooks() {
 		add_filter( 'frm_replace_content_shortcodes', 'FrmFormsController::replace_content_shortcodes', 20, 3 );
 		add_action( 'admin_bar_init', 'FrmFormsController::admin_bar_css' );
 		add_action( 'wp_footer', 'FrmFormsController::footer_js', 1, 0 );
+		add_filter( 'frm_submit_button_class', 'FrmFormsController::update_classes' );
```
Crabcyborg, line 72: Can we possibly add these filters later?

It would help cut back on the number of filters when a form isn't loaded if we only add these when we know we're loading a form.

classes/controllers/FrmFormsController.php
Crabcyborg: It also may be useful to check for the name of a parent theme is one is applicable. A child theme of Twenty Twenty-One would likely have the same problems.

Crabcyborg: I was looking around for other options on how to handle this better.

I think our safest bet is to check if `function_exists( 'twenty_twenty_one_setup' )` instead of checking the name.

classes/controllers/FrmHooksController.php
AbdiTolesa: @Crabcyborg Thank you for suggesting that and I see `FrmHooksController::load_form_hooks` is there and useful for the same purpose. Let me know if you have a better place than that.

classes/controllers/FrmFormsController.php
AbdiTolesa: updated accordingly!

classes/controllers/FrmFormsController.php
AbdiTolesa: just renamed the function name for making it more readable

garretlaxton (approved):
This looks good! I tested the default submit button, repeater buttons, and page break buttons.
image

Crabcyborg (approved):

PR #1008, opened 2022-12-12
Fix redirect setTimeout runs too fast

truongwp:
I found this issue when working on the On Submit action. I'm not sure how to replicate this with the current version.

`setTimeout(window.location='xxx', 8000);` will execute window.location assignment immediately.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -2194,7 +2194,7 @@ class FrmFormsController {
 		} else {
 			add_filter( 'frm_use_wpautop', '__return_true' );
 			echo FrmAppHelper::maybe_kses( $redirect_msg ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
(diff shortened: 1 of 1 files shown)


Crabcyborg:

I can't replicate it either.

It looks like it depends on your `output_buffering` setting in `php.ini`. In my case, `headers_sent()` appears to always return false. https://stackoverflow.com/questions/17588563/headers-sent-return-false-but-headers-were-sent

truongwp:
@Crabcyborg You can replicate it by adding a condition to force using that code. I did that when working on the On Submit action.

Crabcyborg:
Thanks @truongwp it seems that when I hit that code I get the message twice.

Screen Shot 2022-12-14 at 10 49 16 AM

Any idea what may be causing that? I tried adding a `wp_die` call after it echos the first time and it only shows the message once, but it also doesn't load any of the styles then so it doesn't look very good either. It looks like `redirect_after_submit` is being called twice.

I tried disabling all of my plugins and it still seems to happen with just lite active.

Crabcyborg:
Thanks @truongwp that sounds good to me. I'll go ahead and close this for now.

PR #1013, opened 2022-12-16
Add upgrade tip for ACF addon

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3982

@stephywells Please adjust the `link` param.

### DIFF
--- a/classes/helpers/FrmTipsHelper.php
+++ b/classes/helpers/FrmTipsHelper.php
@@ -216,6 +216,14 @@ class FrmTipsHelper {
 				'tip'  => __( 'Send an SMS message when a form is submitted.', 'formidable' ),
 				'call' => __( 'Get the Twilio integration.', 'formidable' ),
 			),
+			array(
+				'link' => array(
+					'content' => 'acf-integration',
+					'param'   => 'acf-integration',
+				),
+				'tip'  => __( 'Map your form with ACF.', 'formidable' ),
+				'call' => __( 'Upgrade to Pro.', 'formidable' ),
+			),
 		);
 
 		return $tips;
--- a/classes/helpers/FrmTipsHelper.php
+++ b/classes/helpers/FrmTipsHelper.php
@@ -218,11 +218,11 @@ class FrmTipsHelper {
 			),
 			array(
 				'link' => array(
-					'content' => 'acf-integration',
-					'param'   => 'acf-integration',
+					'content' => 'acf-tip',
+					'param'   => 'acf-tip',
 				),
-				'tip'  => __( 'Map your form with ACF.', 'formidable' ),
-				'call' => __( 'Upgrade to Pro.', 'formidable' ),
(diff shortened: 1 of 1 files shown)


classes/helpers/FrmTipsHelper.php
```diff
@@ -216,6 +216,14 @@ public static function get_form_action_tip() {
 				'tip'  => __( 'Send an SMS message when a form is submitted.', 'formidable' ),
 				'call' => __( 'Get the Twilio integration.', 'formidable' ),
 			),
+			array(
+				'link' => array(
+					'content' => 'acf-integration',
+					'param'   => 'acf-integration',
```
stephywells, line 222: Let's use 'acf-tip' here for consistency.

classes/helpers/FrmTipsHelper.php
```diff
@@ -216,6 +216,14 @@ public static function get_form_action_tip() {
...
+				),
+				'tip'  => __( 'Map your form with ACF.', 'formidable' ),
+				'call' => __( 'Upgrade to Pro.', 'formidable' ),
```
stephywells, line 225: Let's change this up a bit since ACF isn't included in all plans.
tip: 'Fill Advanced Custom Fields from a form.'
call: 'Add ACF Integration'

stephywells (changes requested):
Thanks for adding this Truong. I should have had the text ready for you first. Sorry about that!

truongwp:
@stephywells I updated them.

stephywells (approved):
Thank you! Looks good.

garretlaxton:
For some reason I don't see the updates for this. Does this require a certain license? I don't have the ACF add-on or ACF installed.
image

stephywells:
@garretlaxton this PR adds a pro tip only in the lite version.

There's a message inside the Create post form action if you try to turn on ACF fields in any Pro version without our ACF add-on installed. (This one is already released)

Are you seeing both of these?

garretlaxton:
@stephywells I do not see them. Could it be because I had them installed previously?

truongwp:
@garretlaxton The tip shows randomly. Maybe you need to refresh the page a couple of times to see it.

garretlaxton:
Thanks @truongwp! I do see the option now at the bottom of the form actions after a few refreshes.
image

I just don't see the `New` tag in the Create Post action and the install button doesn't look like the one I see in figma.
image

Crabcyborg (approved):

I don't know why I asked for QA on this 😅. I should've looked at the code first.

@garretlaxton That pop up is actually part of another update. This isn't related to the post action, which isn't supported in Lite. This is a tip in lite. You need to have Pro deactivated to see it, so it can't be in the post action itself.

Screen Shot 2023-01-09 at 1 49 31 PM

As for the content of that pop up, it may make sense to make that nicer. I think ours just uses the standard default upgrade pop up at the moment.

PR #1004, opened 2022-12-07
Style tab

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3934
Fixes https://github.com/Strategy11/formidable-pro/issues/3714
Related pull request for pro features https://github.com/Strategy11/formidable-pro/pull/3970

**Release notes**
- The `frm_style_settings_top` hook is no longer called.
- Star rating fields are no longer blue in the builder or entries list.
- Custom CSS is no longer merged on import.
- I added some `id` and `for` attributes to the edit page for better accessibility.
- The style name setting is now at the bottom of the General section in the edit page.
- Color pickers were not visibly resetting after a reset action.
- Repeater icon dropdowns were not visibly appearing as reset after resetting a style in older versions.
- This update adds a new `frm_is_admin` filter so it's easier to control the value of `FrmAppHelper::is_admin` from another add on.
- The Styling & Buttons category is now just "Buttons" now, with a different icon.
- The change styling endpoint for the preview has an added permission check.

PR #1014, opened 2022-12-19
On submit action

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3958


classes/controllers/FrmFormsController.php
```diff
@@ -2144,6 +2176,85 @@ public static function run_success_action( $args ) {
...
+		$redirect_action = null;
+		foreach ( $args['conf_method'] as $action ) {
+			if ( 'redirect' === FrmOnSubmitHelper::get_action_type( $action ) ) {
+				$redirect_action = $action;
+				continue;
+			}
+
+			self::run_single_on_submit_action( $args, $action );
+
+			unset( $action, $new_args );
```
Crabcyborg, line 2231: It looks like `$new_args` doesn't ever exist here to begin with.

classes/helpers/FrmOnSubmitHelper.php
```diff
@@ -0,0 +1,183 @@
...
+		?>
+		<div class="frm_form_field frm_has_shortcodes frm_has_textarea">
+			<label for="<?php echo esc_attr( $id_attr ); ?>">
+				<?php esc_html_e( 'Message on submit', 'formidable' ); ?>
+			</label>
+			<textarea
+				name="<?php echo esc_attr( $args['action_control']->get_field_name( 'success_msg' ) ); ?>"
+				id="<?php echo esc_attr( $id_attr ); ?>"
+				rows="4"
+			><?php echo esc_textarea( $args['form_action']->post_content['success_msg'] ); ?></textarea>
```
Crabcyborg, line 46: We have a specific `FrmAppHelper::esc_textarea` that may make sense to use here. It works a bit differently, and may help avoid some issues.

classes/models/FrmOnSubmitAction.php
```diff
@@ -0,0 +1,49 @@
...
+
+class FrmOnSubmitAction extends FrmFormAction {
+
+	public static $slug = 'on_submit';
+
+	public function __construct() {
+		$action_ops = array(
+			'classes'   => 'frm_icon_font frm_location_arrow',
+			'active'    => true,
+			'event'     => FrmAppHelper::pro_is_connected() ? array( 'create', 'update' ) : array( 'create' ),
```
Crabcyborg, line 21: Would it make sense to use the `frm_on_submit_control_settings` filter to set this in Pro instead?

classes/views/frm-form-actions/on_submit_settings.php
```diff
@@ -0,0 +1,87 @@
...
+$type_args                   = $args;
+$type_args['form_action']    = $form_action;
+$type_args['action_control'] = $this;
+foreach ( $types as $type_name => $type ) {
+	$css_class = 'frm_on_submit_' . esc_attr( $type_name ) . '_settings';
+	if ( $success_action !== $type_name ) {
+		$css_class .= ' frm_hidden';
+	}
+
+	echo '<div class="' . esc_attr( $css_class ) . '" data-sub-settings data-type="' . esc_attr( $type_name ) . '">';
```
Crabcyborg, line 78: Can we avoid echoing HTML in a string? This is a view file so it feels especially funny that it isn't just HTML.

js/formidable_admin.js
```diff
@@ -9523,6 +9508,31 @@ function frmAdminBuildJS() {
...
+			}
+
+			const actionEl = event.target.closest( '.frm_form_action_settings' );
+			actionEl.querySelectorAll( '[data-sub-settings]:not(.frm-fade)' ).forEach( el => {
+				fadeOut( el, () => el.classList.add( 'frm_hidden' ) );
+			});
+
+			setTimeout( () => {
+				const activeEl = actionEl.querySelector( '[data-sub-settings][data-type="' + event.target.value + '"]' );
+				console.log( activeEl  );
```
Crabcyborg, line 9524: There's a `console.log` here.

Crabcyborg:
I'm seeing a couple workflow issues as well.

**From Psalm**
> ParamNameMismatch: Argument 1 of FrmOnSubmitAction::form has wrong name $form_action, expecting $instance as defined by FrmFormAction::form

As for the pro toggle, I've been working toward moving that into Lite but it's currently part of my style tab branch (https://github.com/Strategy11/formidable-forms/pull/1004).

Mostly these new files
https://github.com/Strategy11/formidable-forms/blob/e3602259ad3aa71e45ed1c398bc2bc5c1da01920/classes/helpers/FrmHtmlHelper.php
https://github.com/Strategy11/formidable-forms/blob/e3602259ad3aa71e45ed1c398bc2bc5c1da01920/classes/views/shared/toggle.php

And this new CSS in https://github.com/Strategy11/formidable-forms/blob/e3602259ad3aa71e45ed1c398bc2bc5c1da01920/css/frm_admin.css

```
/*
These are the styles for the "Enable Formidable styling" toggle in the visual styler.
It uses the same style rules as the font end toggle field in Pro.
To use this element, you can call the FrmHtmlHelper::toggle function.
*/
.frm_switch_opt {
	padding: 0 8px 0 0;
	white-space: normal;
	display: inline;
	vertical-align: middle;
	font-weight: normal;
}

.frm_on_label {
	padding:0 0 0 8px;
}

.frm_on_label,
.frm_off_label {
	color: #444444;
}

.frm_switch {
	position: relative;
	display: inline-block;
	width: 40px;
	height: 25px;
	vertical-align: middle;
}

.frm_switch_block input {
	display:none !important;
}

.frm_slider {
	position: absolute;
	cursor: pointer;
	top: 0;
	left: 0;
	right: 0;
	bottom: 0;
	background-color: #eaeaea;
	transition: .4s;
	border-radius: 30px;
}

.frm_slider:before {
	border-radius: 50%;
	position: absolute;
	content: "";
	height: 23px;
	width: 23px;
	left: 1px;
	bottom: 1px;
	background-color: white;
	transition: .4s;
	box-shadow:0 2px 6px rgba(41, 58, 82, 0.31);
}

input:checked + .frm_switch .frm_slider {
	background-color: #579AF6;
}

input:focus + .frm_switch .frm_slider {
	box-shadow: 0 0 1px #3177c7;
}

input:checked + .frm_switch .frm_slider:before {
	transform: translateX(15px);
}

.frm_rtl .frm_switch_opt {
	padding: 0 8px;
}

.frm_rtl .frm_slider:before {
	left: 16px;
}

.frm_rtl input:checked + .frm_switch .frm_slider:before {
	transform: none!important;
	left: 1px;
}
/* End toggle styling */
```

That's most of the required code at least. There's some JavaScript as well we'd need to make it accessible so the `aria-checked` value syncs and so that the space key can toggle it. I pulled out some functionality to make it simpler as some of it was only necessary for the front end toggle and should be avoided.

Depending on what order we go in with these updates, it may be easier to just let the style tab updates merge and then use some code from there. The solution right now isn't bad but I do think it makes sense to consistently use the same UI elements in Lite and Pro rather than trying to support multiple.

garretlaxton:
Should this be automatically added to each form? That's currently what I see.

Also, if the submit button is disabled for the form should the `On Submit` action still be available to select?

truongwp:
@Crabcyborg Thanks for checking. I'm beginning to fix them. I spent almost time on the unit tests.

@garretlaxton When you update the plugin, a new On Submit action will be created automatically and old settings are removed. When you create a new form, no On Submit action will be created.

You're able to create multiple On Submit actions.

@stephywells Garret asked "if the submit button is disabled for the form should the On Submit action still be available to select?". What do you think?

garretlaxton:
>When you create a new form, no On Submit action will be created.

@truongwp Should the same apply to imported forms? I imported a form from a different site without this update.

truongwp:
@garretlaxton On submit action is not created after importing until you view the form settings page. It converts old settings to On submit action automatically when you view that page.

stephywells:
> @stephywells Garret asked "if the submit button is disabled for the form should the On Submit action still be available to select?". What do you think?

I think it's fine to have available. In most cases, the submit button would be disabled conditionally, right?

truongwp:
> In most cases, the submit button would be disabled conditionally, right?

@stephywells I didn't understand this question. Can you describe it more?

garretlaxton:
@truongwp Pretty sure she means when the submit button is disabled, there is no submit button on the form anyway.

stephywells:
Sorry about that. I was trying to say I don't think there's a setting to remove the submit button. I think our settings are just to conditionally hide it. So I don't think there is a need to hide the On Submit action. We don't hide form actions for things like MailChimp if there's no email address in the form. 

I think it's fine to show all the available form actions, unless there's a reason we want to restrict them.

truongwp:
@Crabcyborg Can you please take a look at the failing unit test? The `$actions` in the test is `array()`, because `FrmFormAction::get_all()` return `array()`.

This is the `$query` I print_r() before the `check_cache()` call https://github.com/Strategy11/formidable-forms/blob/master/classes/models/FrmFormAction.php#L568:

```
Array
(
    [post_type] => frm_form_actions
    [post_status] => any
    [numberposts] => 99
    [orderby] => title
    [order] => ASC
    [menu_order] => 15
    [suppress_filters] =>
)
```

This is the `$message_action` I print_r() after creating it in the test https://github.com/Strategy11/formidable-forms/pull/1014/files#diff-be81915f505ea7d61afc1c514e7f02f01d9b724cdea75cab5748c18eb9893f7aR178:

```
WP_Post Object
(
    [ID] => 24
    [post_author] => 0
    [post_date] => 2023-01-06 19:21:10
    [post_date_gmt] => 2023-01-06 19:21:10
    [post_content] => {"event":["create"],"success_action":"message","success_msg":"Success message!"}
    [post_title] => Post title 0000038
    [post_excerpt] => on_submit
    [post_status] => publish
    [comment_status] => closed
    [ping_status] => closed
    [post_password] =>
    [post_name] => post-title-0000038
    [to_ping] =>
    [pinged] =>
    [post_modified] => 2023-01-06 19:21:10
    [post_modified_gmt] => 2023-01-06 19:21:10
    [post_content_filtered] =>
    [post_parent] => 0
    [guid] => http://example.org/?frm_form_actions=post-title-0000038
    [menu_order] => 15
    [post_type] => frm_form_actions
    [post_mime_type] =>
    [comment_count] => 0
    [filter] => raw
)
```

I saw that at least that post object should match the query. But the query returns empty. I tested with a direct `get_posts()` call. It just returns an array of posts if I unset the `suppress_filters` and `menu_order` from the `$query`.

garretlaxton:
When I import a form with only lite installed I get 2 On Submit actions in my form. It should only use the redirect to Google.
image

Also the redirect seems to take a little bit before it actually redirects. If AJAX is enabled it's instant.

truongwp:
@garretlaxton Can you share the import file?

In your case, both show message and redirect actions will run. So it will show the message first, then redirect after about 8 seconds.

garretlaxton:
@truongwp here's my import file.
[email-form.zip](https://github.com/Strategy11/formidable-forms/files/10362903/email-form.zip)

truongwp:
@garretlaxton Look like you trashed the form, then imported that form again. The old action wasn't deleted when the form is trashed. Then the new action is created when the form is import.

@Crabcyborg This looks like our limitation. Do you have any ideas on this?

Crabcyborg:
> The old action wasn't deleted when the form is trashed. Then the new action is created when the form is import.

@truongwp That sounds like something we could check for and try to improve upon.

If we import a form and it was deleted, we should still try to match the pre-existing actions with the actions from the export file.

I think we could log that as a separate issue though. It probably isn't a hard requirement for this update.

Crabcyborg:
> Can you please take a look at the failing unit test? The `$actions` in the test is `array()`, because `FrmFormAction::get_all()` return `array()`.

@truongwp Are you sure? I spent some time trying to look into this, playing with a few ways of clearing the WP caches.

And I found that when I commented out this line the tests started passing https://github.com/Strategy11/formidable-forms/blob/f238708d1eca6c87979375114a8296010d69b6f9/classes/helpers/FrmOnSubmitHelper.php#L158

As far as I can tell, the caching is fine outside of this memoization here.

Crabcyborg:
I didn't confirm if this is related to this branch or not but I saw an error in my console when I was testing this.

Screen Shot 2023-01-09 at 3 14 51 PM

Crabcyborg:
@truongwp What do you think about using rich text editors for the success message now like we do with email actions and quiz outcomes? With it in an action it's very similar to how those are done so it should be fairly straight-forward to get working.

Screen Shot 2023-01-09 at 3 17 28 PM
Screen Shot 2023-01-09 at 3 17 21 PM

classes/controllers/FrmFormsController.php
```diff
@@ -2026,7 +2028,23 @@ public static function get_form_contents( $form, $title, $description, $atts ) {
 				$pass_args['reset']       = true;
 				$pass_args['conf_method'] = self::get_confirmation_method( compact( 'form', 'entry_id' ) );
 
-				self::run_success_action( $pass_args );
+				// If conf_method is an array, run On Submit actions.
+				if ( is_array( $pass_args['conf_method'] ) ) {
+					if ( ! $pass_args['conf_method'] ) {
+						// Use default message.
+						self::populate_on_submit_data( $form->options );
```
Crabcyborg, line 2035: @truongwp this function is quite large. What do you think about moving this chunk into a new function?

classes/controllers/FrmFormsController.php
```diff
@@ -2135,6 +2169,8 @@ public static function run_success_action( $args ) {
 		$opt = ( ! isset( $args['action'] ) || $args['action'] === 'create' ) ? 'success' : 'edit';
 
 		$args['success_opt'] = $opt;
+		$args['ajax']        = ! empty( $GLOBALS['frm_vars']['ajax'] );
```
Crabcyborg, line 2172: We usually use the `global $frm_vars` syntax. It may be nice to stay consistent with that so it's easy to find references to `$frm_vars` throughout our code. When I search for ` $GLOBALS['frm_vars']` I think I see it used only once.

classes/controllers/FrmFormsController.php
```diff
@@ -2582,6 +2705,192 @@ function() {
...
+
+			default:
+				$form_options[ $opt . 'msg' ] = ! empty( $action->post_content['success_msg'] ) ? $action->post_content['success_msg'] : FrmOnSubmitHelper::get_default_msg();
+				$form_options['show_form']   = ! empty( $action->post_content['show_form'] );
+		}
+	}
+
+	/**
+	 * Maybe migrate submit settings to On Submit action.
+	 *
```
Crabcyborg, line 2785: It might be nice to add some additional comments explaining why this exists for someone who may come to this without knowing the history of form actions. It would be useful to know before looking at the code that it used to be a form setting and was moved to a form action.

classes/views/frm-forms/settings-advanced.php
```diff
@@ -156,20 +115,11 @@
...
-		<td class="frm_has_shortcodes frm_has_textarea">
-			<label for="success_msg"><?php esc_html_e( 'On Submit', 'formidable' ); ?></label>
-			<textarea id="success_msg" name="options[success_msg]" cols="50" rows="2" class="frm_long_input"><?php echo FrmAppHelper::esc_textarea( $values['success_msg'] ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></textarea>
-		</td>
-	</tr>
-	<?php do_action( 'frm_add_form_msg_options', $values ); ?>
-</table>
+<?php if ( has_action( 'frm_add_form_msg_options' ) ) : ?>
+	<table class="form-table frm-fields">
+		<?php do_action( 'frm_add_form_msg_options', $values ); ?>
```
Crabcyborg, line 120: I think a comment here explaining why this is here and why it is still here would help make this code easier to understand in the future.

classes/controllers/FrmFormsController.php
```diff
@@ -2144,6 +2180,85 @@ public static function run_success_action( $args ) {
...
+		}
+	}
+
+	public static function run_single_on_submit_action( $args, $action ) {
+		$new_args = self::get_run_success_action_args( $args, $action );
+		self::run_success_action( $new_args );
+	}
+
+	private static function get_run_success_action_args( $args, $action ) {
+		$new_args = $args;
```
Crabcyborg, line 2251: It isn't very clear why `$new_args` is getting created from `$args` instead of just continuing to use `$args`. Maybe a comment would help, or maybe we should just update `$args` here instead?

classes/controllers/FrmFormsController.php
```diff
@@ -2144,6 +2180,85 @@ public static function run_success_action( $args ) {
...
+		}
+
+		if ( $redirect_action ) {
+			// Show script to delay the redirection.
+			$args['force_delay_redirect'] = true;
+			self::run_single_on_submit_action( $args, $redirect_action );
+		}
+	}
+
+	public static function run_single_on_submit_action( $args, $action ) {
```
Crabcyborg, line 2245: It would be nice to fill in the missing type comments everywhere.

classes/helpers/FrmOnSubmitHelper.php
```diff
@@ -0,0 +1,184 @@
...
+		</div>
+
+		<?php $id_attr = $args['action_control']->get_field_id( 'redirect_msg' ); ?>
+		<div class="frm_form_field">
+			<label for="<?php echo esc_attr( $id_attr ); ?>"><?php esc_html_e( 'Redirect message (used when multiple On Submit actions run)', 'formidable' ); ?></label>
+			<textarea
+				name="<?php echo esc_attr( $args['action_control']->get_field_name( 'redirect_msg' ) ); ?>"
+				id="<?php echo esc_attr( $id_attr ); ?>"
+				rows="4"
+			><?php echo esc_textarea( $args['form_action']->post_content['redirect_msg'] ); ?></textarea>
```
Crabcyborg, line 114: I think we want to use `FrmAppHelper::esc_textarea` here.

classes/helpers/FrmOnSubmitHelper.php
```diff
@@ -0,0 +1,184 @@
...
+	 *     @type object        $form           Form data.
+	 *     @type string        $action_key     Action key.
+	 *     @type array         $values         Contains `fields` (form fields) and `id` (form ID).
+	 * }
+	 */
+	public static function show_page_settings( $args ) {
+		$name_attr = $args['action_control']->get_field_name( 'success_page_id' );
+		?>
+		<div class="frm_form_field">
+			<label for="<?php echo esc_attr( $name_attr ); ?>"><?php esc_html_e( 'Select a page', 'formidable' ); ?></label>
```
Crabcyborg, line 136: I don't think this `for` attribute works as expected. It should use the inputs ID, but it also looks like this input doesn't currently have an ID set.

Crabcyborg:
@truongwp I'm seeing an odd visual difference here between on submit actions any other actions.

You can see here in an email action that the line separates the header and body, and it's a straight line along the left.
Screen Shot 2023-01-09 at 3 56 38 PM

I'm seeing extra border radius styling here with the on submit action.
Screen Shot 2023-01-09 at 3 56 40 PM

js/formidable_admin.js
```diff
@@ -9523,6 +9508,30 @@ function frmAdminBuildJS() {
...
+			if ( ! event.target.checked ) {
+				return;
+			}
+
+			const actionEl = event.target.closest( '.frm_form_action_settings' );
+			actionEl.querySelectorAll( '[data-sub-settings]:not(.frm-fade)' ).forEach( el => {
+				fadeOut( el, () => el.classList.add( 'frm_hidden' ) );
+			});
+
+			setTimeout( () => {
```
Crabcyborg, line 9522: I don't know how I feel about the fading out and then immediately turning visible that happens when you switch between types. It feels inconsistent with other settings and the transition feels a bit jarring.

Was this planned on ahead of time? Is it like this anywhere else in any of our plugins?

tests/forms/test_FrmFormsController.php
```diff
@@ -150,12 +150,79 @@ public function test_front_head() {
...
+		$this->assertEquals( wp_list_pluck( $actions, 'ID' ), array( $page_action->ID, $redirect_action_2->ID ) );
+	}
+
 	/**
 	 * Test redirect after create
 	 *
 	 * @covers FrmFormsController::redirect_after_submit
 	 */
 	public function test_redirect_after_create() {
+		$this->markTestSkipped(
```
Crabcyborg, line 222: @truongwp I don't think we need to use `markTestSkipped` here.

Are you planning to write new tests to cover these cases? I think the same test assertions should still work, just with action settings instead.

classes/views/frm-form-actions/on_submit_settings.php
```diff
@@ -0,0 +1,88 @@
...
+		$css_class .= ' frm_hidden';
+	}
+	?>
+	<div class="<?php echo esc_attr( $css_class ); ?>" data-sub-settings data-type="<?php echo esc_attr( $type_name ); ?>">
+		<?php
+		if ( is_callable( $type['sub_settings'] ) ) {
+			call_user_func( $type['sub_settings'], $type_args );
+		}
+		?>
+	</div><!-- End .frm_on_submit_<?php echo esc_attr( $type_name ); ?>_settings -->
```
Crabcyborg, line 84: Is it necessary to have this HTML comment?

classes/controllers/FrmHooksController.php
```diff
@@ -124,6 +124,7 @@ public static function load_admin_hooks() {
 			add_action( 'frm_add_form_style_tab_options', 'FrmFormsController::add_form_style_tab_options' );
 		}
 		add_action( 'frm_after_duplicate_form', 'FrmFormActionsController::duplicate_form_actions', 20, 3 );
+		add_filter( 'frm_skip_form_action', 'FrmFormActionsController::skip_actions', 10, 2 );
```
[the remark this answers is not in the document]
truongwp: Thanks. It's in the wrong place. I removed the add_filter and use the check directly.

classes/controllers/FrmFormsController.php
```diff
@@ -2582,6 +2705,192 @@ function() {
...
+		$base_action['post_title']   = FrmOnSubmitAction::get_name();
+		$base_action['menu_order']   = $form_id;
+		$base_action['post_status']  = 'publish';
+		$base_action['post_content'] = array(
+			'event' => array( 'create' ),
+		);
+
+		$action_data = self::get_on_submit_action_data_from_form_options( $form->options );
+
+		if ( FrmAppHelper::pro_is_connected() && intval( $form->editable ) ) {
```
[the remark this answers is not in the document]
truongwp: It might be a `'0'` string if the `$form` object is not prepared correctly. So I used `intval()` here.

classes/controllers/FrmFormsController.php
truongwp: I planned to add a filter for the result of that method. So I kept the old `$args`.

classes/views/frm-form-actions/on_submit_settings.php
truongwp: It's helpful in case there are HTML errors that make the open and close tags not match. But I'm ok to remove it.

js/formidable_admin.js
truongwp: I didn't check other places in our plugins. I found the `fadeOut()` function, but didn't find the `fadeIn()`. So I did that. Do you want to add the fade-in effect, or just remove the fade-out?

truongwp:
> @truongwp I'm seeing an odd visual difference here between on submit actions any other actions.
> 
> You can see here in an email action that the line separates the header and body, and it's a straight line along the left. Screen Shot 2023-01-09 at 3 56 38 PM
> 
> I'm seeing extra border radius styling here with the on submit action. Screen Shot 2023-01-09 at 3 56 40 PM
> 
> It's pretty small so it's a bit hard to see. Here's a more zoomed in screenshot. Screen Shot 2023-01-09 at 3 59 13 PM

I just see it during opening and closing a form action. Maybe there is an error that happens in yours?

truongwp:
> > @truongwp I'm seeing an odd visual difference here between on submit actions any other actions.
> > You can see here in an email action that the line separates the header and body, and it's a straight line along the left. Screen Shot 2023-01-09 at 3 56 38 PM
> > I'm seeing extra border radius styling here with the on submit action. Screen Shot 2023-01-09 at 3 56 40 PM
> > It's pretty small so it's a bit hard to see. Here's a more zoomed in screenshot. Screen Shot 2023-01-09 at 3 59 13 PM
> 
> I just see it during opening and closing a form action. Maybe there is an error that happens in yours?

I can not replicate this. There might be a js error that makes the CSS class isn't added correctly. Can you check again on your side?

classes/helpers/FrmOnSubmitHelper.php
truongwp: Our pages dropdown use `field_name` as dropdown id: https://github.com/Strategy11/formidable-forms/blob/41688e410352e2e57511040eee4699e678a159ab/classes/helpers/FrmAppHelper.php#L1353

The autocomplete doesn't have id attribute.

Do you want to update those helper methods to add the id attribute?

truongwp:
> @truongwp I'm seeing an odd visual difference here between on submit actions any other actions.
> 
> You can see here in an email action that the line separates the header and body, and it's a straight line along the left. Screen Shot 2023-01-09 at 3 56 38 PM
> 
> I'm seeing extra border radius styling here with the on submit action. Screen Shot 2023-01-09 at 3 56 40 PM
> 
> It's pretty small so it's a bit hard to see. Here's a more zoomed in screenshot. Screen Shot 2023-01-09 at 3 59 13 PM



> get_run_success_action_args

I got this issue when try using WordPress editor:

Screen Shot 2023-01-10 at 20 19 08

Screen Shot 2023-01-10 at 20 19 28

I just copied and modified this code:

```
$editor_args = array(
	'textarea_name' => $this->get_field_name( 'email_message' ),
	'textarea_rows' => 6,
	'editor_class'  => 'frm_not_email_message',
);
wp_editor(
	$form_action->post_content['email_message'],
	$this->get_field_id( 'email_message' ),
	$editor_args
);
```

Do I need to add anything in JS?

truongwp:
> > Can you please take a look at the failing unit test? The `$actions` in the test is `array()`, because `FrmFormAction::get_all()` return `array()`.
> 
> @truongwp Are you sure? I spent some time trying to look into this, playing with a few ways of clearing the WP caches.
> 
> And I found that when I commented out this line the tests started passing
> 
> https://github.com/Strategy11/formidable-forms/blob/f238708d1eca6c87979375114a8296010d69b6f9/classes/helpers/FrmOnSubmitHelper.php#L158
> 
> As far as I can tell, the caching is fine outside of this memoization here.

This is a great spot. I guess the method is called once before the actions are created. Then in the second call, it uses the previous result. I'm checking this further.

Crabcyborg:
> Do I need to add anything in JS?

Yeah. We re-initialize it when it loads, which is both when you open one that exists, and when you create one by duplicating.

https://github.com/Strategy11/formidable-forms/blob/6c7b67dfdf565174a1ec51d0899df47ba9a33f44/js/formidable_admin.js#L7847

I think it needs that `addFocusEvents` option to support the shortcode pop up.

`onEmailActionLoaded` is just for handling this one thing, so it could be refactored to support emails and on submit actions instead.

Crabcyborg:
Here is an accessibility issue I'm noticing. You can't tab to the type options. It just skips them and goes to the message input instead.

ZvZb3CLbh9

Crabcyborg (commented):
@truongwp I merged master into this branch. I was seeing a JavaScript error after duplicating an on submit action, but it was because of a bug unrelated to this branch that has been fixed already.

That may have had to do with the rounded corners I was seeing. I'm not seeing that while I'm testing right now. Thanks for confirming that!

classes/controllers/FrmFormsController.php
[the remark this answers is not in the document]
Crabcyborg: That should still be fine to just check `if ( $form->editable )` without it.

https://www.php.net/manual/en/language.types.boolean.php

> When converting to bool, the following values are considered false:
> 
>     the [boolean](https://www.php.net/manual/en/language.types.boolean.php) false itself
>     the [integer](https://www.php.net/manual/en/language.types.integer.php) 0 (zero)
>     the [float](https://www.php.net/manual/en/language.types.float.php)s 0.0 and -0.0 (zero)
>     the empty [string](https://www.php.net/manual/en/language.types.string.php) "", and the [string](https://www.php.net/manual/en/language.types.string.php) "0"
>     an [array](https://www.php.net/manual/en/language.types.array.php) with zero elements
>     the unit type [NULL](https://www.php.net/manual/en/language.types.null.php) (including unset variables)
>     Internal objects that overload their casting behaviour to bool. For example: [SimpleXML](https://www.php.net/manual/en/ref.simplexml.php) objects created from empty elements without attributes.

It's how it's checked everywhere else. I think it should be fine.

https://github.com/Strategy11/formidable-pro/blob/7209d2f15360adb16c22aa62b0f0fa42f351f1ac/classes/controllers/FrmProFormsController.php#L1143

https://github.com/Strategy11/formidable-pro/blob/20eb30510edc36d1e39a17c599252eea3423ea61/classes/helpers/FrmProEntriesHelper.php#L24

https://github.com/Strategy11/formidable-pro/blob/dd41fda1339293b387c15469334748e73af73f65/classes/models/FrmProEntry.php#L47

classes/helpers/FrmOnSubmitHelper.php
Crabcyborg: It probably isn't too urgent. It doesn't seem like our autocompletes are very accessible anywhere. It would make sense to update that all at once, which could be done in another branch.

truongwp: Ok. I'll note this in an issue.

classes/controllers/FrmFormsController.php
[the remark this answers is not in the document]
truongwp: Thank you so much, Mike. That's great information.

classes/controllers/FrmFormsController.php
```diff
@@ -2144,6 +2177,141 @@ public static function run_success_action( $args ) {
...
+				self::run_success_action( $args );
+			} elseif ( 1 === count( $args['conf_method'] ) ) {
+				self::populate_on_submit_data( $args['form']->options, reset( $args['conf_method'] ) );
+				$args['conf_method'] = $args['form']->options['success_action'];
+				self::run_success_action( $args );
+			} else {
+				self::run_multi_on_submit_actions( $args );
+			}
+		} else {
+			self::run_success_action( $args );
```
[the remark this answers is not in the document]
truongwp: Yeah. After moving to a separate method, we can do that to make the code cleaner.

classes/controllers/FrmFormsController.php
```diff
@@ -2144,6 +2177,141 @@ public static function run_success_action( $args ) {
...
+	 * @since 5.x.x
+	 *
+	 * @param array $args See {@see FrmFormsController::run_success_action()}.
+	 */
+	public static function run_multi_on_submit_actions( $args ) {
+		$redirect_action = null;
+		foreach ( $args['conf_method'] as $action ) {
+			if ( 'redirect' === FrmOnSubmitHelper::get_action_type( $action ) ) {
+				$redirect_action = $action;
+				continue;
```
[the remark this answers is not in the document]
truongwp: I added a comment there.

classes/models/FrmForm.php
```diff
@@ -768,6 +769,7 @@ public static function getAll( $where = array(), $order_by = '', $limit = '' ) {
 
 		if ( $results ) {
 			foreach ( $results as $result ) {
+				FrmFormsController::update_get_form_success_action( $result );
```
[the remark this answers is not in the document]
truongwp: The `$result` is cached. So I think it just hits once per page load.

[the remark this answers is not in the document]
Crabcyborg: Right. This is probably fine then. Thanks.

js/formidable_admin.js
Crabcyborg: I've asked about this in check-in this morning. I also tagged Razvan for his input. I'm not totally sure which direction we'd want to go in with this.

truongwp:
> Here is an accessibility issue I'm noticing. You can't tab to the type options. It just skips them and goes to the message input instead.
> 
> ZvZb3CLbh9 [ ZvZb3CLbh9 ](https://user-images.githubusercontent.com/9134515/211639002-65c738f3-0ba8-4efe-aa03-9c734c960881.gif) [ ](https://user-images.githubusercontent.com/9134515/211639002-65c738f3-0ba8-4efe-aa03-9c734c960881.gif)

@Crabcyborg Now you can use tab, then arrow keys to select that option. But I think when using tab to focus on that option, we should use a different border or add outline to make it different, so user will know that it's focused on.

js/formidable_admin.js
truongwp: Ok. I removed it.

PR #1015, opened 2022-12-22
Adding a new filter that allows updating new entry values

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3975

Just adding a new filter that allows updating new entry value for a given field and extract reusable code into a new function and call it twice.

### DIFF
--- a/classes/helpers/FrmEntriesHelper.php
+++ b/classes/helpers/FrmEntriesHelper.php
@@ -103,6 +103,23 @@ class FrmEntriesHelper {
 			$new_value = str_replace( '"', '&quot;', $new_value );
 		}
 
+		$field_obj = self::get_field_id_and_object( $field )[1];
+
+		$args = array(
+			'allow_array' => false,
+			'field'       => $field_obj,
+		);
+
+		/**
+		 * Allows updating posted entry value for a field.
+		 *
+		 * @since x.x
+		 *
+		 * @param string|int $new_value
+		 * @param array $args
+		 */
+		$new_value = apply_filters( 'frm_new_entry_value', $new_value, $args );
+
 		return $new_value;
 	}
 
@@ -339,7 +356,7 @@ class FrmEntriesHelper {
 		$_POST['item_meta'][ $args['parent_field_id'] ][ $args['key_pointer'] ][ $field->id ] = $value; // phpcs:ignore WordPress.Security.NonceVerification.Missing
 	}
 
-	public static function get_posted_value( $field, &$value, $args ) {
+	private static function get_field_id_and_object( $field ) {
 		if ( is_array( $field ) ) {
 			$field_id  = $field['id'];
 			$field_obj = FrmFieldFactory::get_field_object( $field['id'] );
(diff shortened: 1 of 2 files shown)


classes/helpers/FrmEntriesHelper.php
```diff
@@ -349,11 +366,21 @@ public static function get_posted_value( $field, &$value, $args ) {
...
+	public static function get_posted_value( $field, &$value, $args ) {
+		if ( ! ( is_array( $field ) || is_object( $field ) || is_numeric( $field ) ) ) {
 			$value = self::get_posted_meta( $field, $args );
 			FrmAppHelper::sanitize_value( 'sanitize_text_field', $value );
 			return;
 		}
+		$field_id_and_object = self::get_field_id_and_object( $field );
+
+		$field_id  = $field_id_and_object[0];
+		$field_obj = $field_id_and_object[1];
```
Crabcyborg, line 383: @AbdiTolesa Why are you passing around `$field_obj` if you never use it?

AbdiTolesa: That has no purpose assigning it to another variable indeed, just removed that line now.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -349,11 +366,20 @@ public static function get_posted_value( $field, &$value, $args ) {
...
+		return array( $field_id, $field_obj );
+	}
+
+	public static function get_posted_value( $field, &$value, $args ) {
+		if ( ! ( is_array( $field ) || is_object( $field ) || is_numeric( $field ) ) ) {
 			$value = self::get_posted_meta( $field, $args );
 			FrmAppHelper::sanitize_value( 'sanitize_text_field', $value );
 			return;
 		}
+		$field_id_and_object = self::get_field_id_and_object( $field );
```
Crabcyborg, line 380: Could this be refactored in a way that we're not calling `get_field_id_and_object` twice? Ideally we'd get it once and use it in both cases.

AbdiTolesa: I gave this a try but since `get_posted_value` is a public function, it seems it a bit complex to refactor it further.

AbdiTolesa:
@Crabcyborg Psalm analysis is failing and I tried to look into it but not too sure what should the return type for `FrmShowForm::update` should be as it is a public function. What do you suggest?

Crabcyborg:
Thanks @AbdiTolesa those errors appear to be new. Likely from a new version of Psalm.

I'm seeing them in other pull requests too. It's not related to this update.

I'm also not entirely familiar with the widget. I think we could probably just add the exception for that file for now as I don't think it's too important to fix.

AbdiTolesa:
> Thanks @AbdiTolesa those errors appear to be new. Likely from a new version of Psalm.
> 
> I'm seeing them in other pull requests too. It's not related to this update.
> 
> I'm also not entirely familiar with the widget. I think we could probably just add the exception for that file for now as I don't think it's too important to fix.

Got it, this is done now.

psalm.xml
```diff
@@ -13,6 +13,7 @@
 		<ignoreFiles>
 			<directory name="vendor" />
 			<directory name="classes/views/" />
+			<file name="classes/widgets/FrmShowForm.php" />
```
Crabcyborg, line 16: This isn't exactly what I meant. It's a lot to fully ignore the whole file.

Psalm is really good at ignoring a specific rule for a specific file. You can just use this:

```
<MethodSignatureMustProvideReturnType>
			<errorLevel type="suppress">
				<file name="classes/widgets/FrmShowForm.php" />
    </errorLevel>
</MethodSignatureMustProvideReturnType>
```

Crabcyborg:
I'm closing this pull request.

I left a comment on the other pull request https://github.com/Strategy11/formidable-pro/pull/3989#issuecomment-1406625345

 I don't think we need the new filter to solve this issue.

PR #1016, opened 2022-12-23
Fix JS error when duplicating a form action

truongwp:
This can be replicated by duplicating a form action that doesn't have a WYSIWYG editor.


Crabcyborg (approved):
Thanks for catching that @truongwp! This looks good!

PR #1017, opened 2022-12-26
Logic styling

AbdiTolesa:
This update would ideally be kept in pro had there been admin css file there. It is adding to https://github.com/Strategy11/formidable-pro/pull/3976/files that pertains to making the required updates to match the new design (https://www.figma.com/file/ILeT8x5Ks4cxEUzRix1ma8/Form-Actions?node-id=801%3A2184&t=cR37l3v540PtR2tO-0).

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -5304,10 +5304,6 @@ table td, .form-table tr td {
 	vertical-align: bottom;
 }
 
-.frm_action_id {
-	padding-top: 10px;
-}
-
 .frm_logic_field_opts {
 	max-width: 35%;
 }
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -4778,14 +4778,16 @@ h2.frm-h2,
 }
 
 .frm_settings_form h3,
-.frm_form_settings h3, .frm_form_settings p.frm_add_logic_link {
+.frm_form_settings h3,
+.frm_form_settings span.frm_add_logic_link {
 	font-size: 16px;
-	font-weight: 600;
 	border-top: 1px solid var(--grey-border);
-	margin: 20px 0 10px 0;
-	padding-top: 20px;
+	padding-top: 10px;
+	margin: 1em 0;
+	font-weight: normal;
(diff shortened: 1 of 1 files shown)


Crabcyborg:
@AbdiTolesa I don't think it would hurt to add any pro-specific styles into Pro with a new file for this form settings page.

Anything specific to Pro that we can avoid adding into the Lite plugin is ideal.

stephywells:
This is adding/removing too much styling that makes the settings look worse. Here is before:
Screen Shot 2023-01-03 at 10 02 11 AM
Screen Shot 2023-01-03 at 9 59 29 AM

Here is after:
Screen Shot 2023-01-03 at 10 04 33 AM
Screen Shot 2023-01-03 at 10 02 47 AM

@AbdiTolesa for all PRs, can you please include these before and after screenshots? This will save time in reviewing.

stephywells (approved):

AbdiTolesa:
> I went ahead and updated this. I also switched the p back to a span to avoid extra HTML changes. This should be good to go now.

@stephywells I should have made the style updates more specific not to mess with the other h3 styles but the spacing and font styles look a bit different than the design with this update yet. That's why I added the extra margin and padding you have removed.

Current:
image

Design:
image

AbdiTolesa:
> This is adding/removing too much styling that makes the settings look worse. Here is before: Screen Shot 2023-01-03 at 10 02 11 AM Screen Shot 2023-01-03 at 9 59 29 AM
> 
> Here is after: Screen Shot 2023-01-03 at 10 04 33 AM Screen Shot 2023-01-03 at 10 02 47 AM
> 
> @AbdiTolesa for all PRs, can you please include these before and after screenshots? This will save time in reviewing.

@stephywells Ok, I will be careful to include the before and after screenshots going forward! I have tried to explain why I added each style update below, but not sure if that makes sense.

stephywells:
Thanks for clarifying that. The problem is the changes seemed further from the design to me. For this update, we were just switching from the button to the heading. This isn't intended to make any extra styling updates.

PR #1018, opened 2022-12-27
Report date filter styles

AbdiTolesa:
I'm adding styles for the report filter here as we don't have a pro admin css file but I'm not sure if it should ideally be in lite even yet. I can move it away to any place you would recommend if we should @Crabcyborg.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -5684,7 +5684,6 @@ span.howto {
     border: none;
     color: var(--primary-color);
     font-size: 14px;
-    line-height: 21px;
     font-weight: 600;
 	cursor: pointer;
 }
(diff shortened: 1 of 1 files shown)


Crabcyborg:
@AbdiTolesa I think long term we're best off making new CSS files for individual pages. These huge files with shared CSS get out of hand, and it can be impossible to remove CSS from lite after we add it, so it's best to avoid it from the beginning.

A new `reports.css` file in Pro would make a lot of sense. In the future we'll likely want to remove Reports from Pro, into a new add on, so having everything for Reports separate will help make that easier to do later as well.

Crabcyborg:
I'll go ahead and close this as I don't think this pull request will be needed.

PR #1019, opened 2022-12-28
Replace Rich text fields placeholder setting field with RTE

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3987
This update changes the html field used to set the default value for rich text editor to the same field.
Before:
image

After:
image


classes/views/frm-fields/back-end/settings.php
```diff
@@ -179,7 +179,14 @@
 						}
 						$field_obj->default_value_to_string( $default_value );
 
-						if ( $display['type'] === 'textarea' || $display['type'] === 'rte' ) {
+						if ( $display['type'] === 'rte' ) {
+							$e_args  = array(
+								'textarea_name' => 'default_value_' . absint( $field['id'] ),
```
Crabcyborg, line 184: It looks like there's already a `$default_name` variable with this set that you could use instead of coding it a second time.

classes/views/frm-fields/back-end/settings.php
```diff
@@ -179,7 +179,14 @@
...
+								'textarea_rows' => 8,
```
Crabcyborg, line 185: 8 rows may be a bit large. It looks like textareas are only set to 3 rows?

classes/views/frm-fields/back-end/settings.php
```diff
@@ -179,7 +179,14 @@
 						}
 						$field_obj->default_value_to_string( $default_value );
 
-						if ( $display['type'] === 'textarea' || $display['type'] === 'rte' ) {
+						if ( $display['type'] === 'rte' ) {
+							$e_args  = array(
```
Crabcyborg, line 183: @AbdiTolesa Since rich text fields are a Pro feature, do you think we could try to get more of this update into the Pro plugin?

garretlaxton:
This looks good! The default value for the RTE fields show the editor now.
image

Just curious, is there a way to choose which tab is the default when viewing the form? Ex: Visual or Text? It currently defaults to text, but if a default value is set I think it would be more ideal for the Visual tab to be default. 

It might be a good idea to have an option to set the Visual or Text as default in the field options. But that can be an enhancement idea for later.

garretlaxton (approved):

classes/views/frm-fields/back-end/settings.php
AbdiTolesa: @Crabcyborg I think we can check if pro installed and show rich text or opt to the default textarea otherwise, in the same file here. If we have to move more of the update to Pro literally, I imagine it would be calling a function that just calls `wp_editor` as we could prepare the function arguments here in lite. Do you think the first approach is good for now as we don't have much of extra code for this update?

Crabcyborg: It becomes important for the change log even. It's weird to put an update in the lite add on for a rich text editor.

If we ended up changing `textarea_rows` I have to put that in the change log for the lite add on but it's for a pro feature so it should be an update in pro.

I think we could add functionality to the `field_obj` here and overwrite the function in `FrmProFieldRte`.

Something like
```
if ( ! $field_obj->maybe_echo_placeholder_setting() ) {
    ... // Handle the current code as a fallback.
}
```

`FrmFieldType::maybe_echo_placeholder_setting` could return false so it handles it in the view.

And then `FrmProFieldRte::maybe_echo_placeholder_setting` could return true and handle the echo in Pro.

AbdiTolesa: @Crabcyborg Not sure but is the following approach probably better as it doesn't require updating `FrmFieldType`?
```
if ( $display['type'] === 'rte' ) {
	if ( method_exists( $field_obj, 'maybe_echo_placeholder_setting' )  ) {
		$field_obj->maybe_echo_placeholder_setting()
	} else {
		// fallback to show textarea
	}
}
```

Crabcyborg: I don't love it. I think it's best that a function overwrites a function that exists in the parent.

Crabcyborg: Now that I think about this a bit more, ai think we could pull out the maybe from the function name and just handle all of the placeholder settings from the model files.

FrmFieldType would handle the default case, we'd add a function for textareas, then this new function for Rte fields.

AbdiTolesa: Sounds even better, I have updated my code like that.

AbdiTolesa:
> This looks good! The default value for the RTE fields show the editor now. image
> 
> Just curious, is there a way to choose which tab is the default when viewing the form? Ex: Visual or Text? It currently defaults to text, but if a default value is set I think it would be more ideal for the Visual tab to be default.
> 
> It might be a good idea to have an option to set the Visual or Text as default in the field options. But that can be an enhancement idea for later.

It defaults to the visual tab for me actually.

classes/models/fields/FrmFieldType.php
```diff
@@ -397,6 +397,16 @@ public function show_field_options( $args ) {
...
 	}
 
+	/**
+	 * @param array  $field
+	 * @param object $field_obj
+	 * @param string $default_name
+	 * @param mixed  $default_value
+	 */
+	public function show_placeholder_setting( $field, $field_obj, $default_name, $default_value ) {
+		echo '<input type="text" name="' . esc_attr( $default_name ) . '" value="' . esc_attr( $default_value ) . '" id="frm_default_value_' . esc_attr( $field['id'] ) . '" class="default-value-field" data-changeme="field_' . esc_attr( $field['field_key'] ) . '" data-changeatt="value" data-sep="' . esc_attr( $field_obj->displayed_field_type( $field ) ? ',' : '' ) . '" />';
```
Crabcyborg, line 407: It would be nice to get these into view files still. We usually don't want to echo HTML as a string.

Crabcyborg:
> choose which tab is the default when viewing the form
> It defaults to the visual tab for me actually.

@garretlaxton @AbdiTolesa As far as I understand, there's something that controls this value between sessions. If I select Visual, it generally becomes my active tab in future iterations. And the same if I select Text. I'm not sure how it's tracked, somewhere in TinyMCE or WordPress code, but we generally don't control this setting. It's automatic.

css/frm_admin.css
```diff
@@ -6594,6 +6594,16 @@ i.frm-show-inline-modal,
 	font-size: 18px;
 }
 
+.frm_rte_placeholder{
```
Crabcyborg, line 6597: This CSS could go into Pro. There's a `builder.css` file there for this page.

classes/helpers/FrmAppHelper.php
```diff
@@ -974,6 +974,7 @@ public static function kses_icon( $icon ) {
 	 */
 	public static function add_allowed_icon_tags( $allowed_html ) {
 		$allowed_html['svg']['data-open'] = true;
+		$allowed_html['svg']['data-html_id'] = true;
```
Crabcyborg, line 977: I'd rather not need to add more allowed attributes for our `kses`.

Can we just add the html ID on the `.frm_rte_placeholder` element?

classes/views/frm-fields/back-end/placeholder-setting.php
```diff
@@ -0,0 +1,16 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+
+if ( $field['type'] !== 'textarea' ) {
```
Crabcyborg, line 6: This isn't very object orientated. I think we should still have two views. This just makes testing more complicated and add unnecessary if statements to a view.

js/formidable_admin.js
```diff
@@ -5796,6 +5797,16 @@ function frmAdminBuildJS() {
 			if ( ! inputTrigger ) {
 				input = getInputForIcon( icon );
 			}
+			if ( input === null ) {
+				iconWrapper = icon.closest( '.frm-rte-smart-values-wrapper' );
+				if ( iconWrapper ) {
+					inputID = iconWrapper.getAttribute( 'data-html_id' );
```
Crabcyborg, line 5803: Since this is an admin script we can use `const`. Sine `inputID` and `iconWrapper` are only used at this scope it makes more sense to use `const` in this case.

classes/views/frm-fields/back-end/settings.php
```diff
@@ -161,13 +161,16 @@
...
 						<?php
 						$special_default = ( isset( $field['post_field'] ) && $field['post_field'] === 'post_category' ) || $field['type'] === 'data';
-						FrmAppHelper::icon_by_class(
-							'frm_icon_font frm_more_horiz_solid_icon frm-show-inline-modal',
-							array(
-								'data-open' => $special_default ? 'frm-tax-box-' . $field['id'] : 'frm-smart-values-box',
-								'title'     => esc_attr__( 'Toggle Options', 'formidable' ),
-							)
-						);
+						if ( $field['type'] !== 'rte' ) {
```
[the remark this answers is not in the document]
AbdiTolesa: I have now added removed the code copied from lite and added the icon with js.

js/formidable_admin.js
```diff
@@ -5796,6 +5796,16 @@ function frmAdminBuildJS() {
 			if ( ! inputTrigger ) {
 				input = getInputForIcon( icon );
 			}
+			if ( input === null ) {
+				const iconWrapper = icon.closest( '.frm-rte-smart-values-wrapper' );
```
Crabcyborg, line 5800: @AbdiTolesa Are we able to get these into the Pro `builder.js` file instead?

AbdiTolesa: @Crabcyborg Yes that was indeed possible but a bit tricky. No js update left in lite now!

classes/views/frm-fields/back-end/settings.php
```diff
@@ -153,21 +153,31 @@
...
 				</span>
 				<?php } ?>
-
-				<p class="frm-has-modal default-value-section-<?php echo esc_attr( $field['id'] . ( isset( $default_value_types['default_value']['current'] ) ? '' : ' frm_hidden' ) ); ?>" id="default-value-for-<?php echo esc_attr( $field['id'] ); ?>">
+				<?php
+				$html_id      = 'frm_default_value_' . absint( $field['id'] );
+				$field_is_rte = $field['type'] === 'rte';
+				?>
+				<p
+				class="frm-has-modal <?php echo esc_attr( $field_is_rte ? 'frm_rte_default' : '' ); ?> default-value-section-<?php echo esc_attr( $field['id'] . ( isset( $default_value_types['default_value']['current'] ) ? '' : ' frm_hidden' ) ); ?>"
```
Crabcyborg, line 161: @AbdiTolesa one of my goals here is to avoid anything in Lite that is explicitly about `rte` fields. Ideally nothing would reference them here, except for some function descriptions where it would be useful information.

I see you're using this to tell them apart.

We already have a `.frm-single-settings.frm-type-rte` parent class we can use to identify that the default value input is part of an RTE or not. I don't think we should add more classes than that specific for field type.

Can we give this class a class like `frm-default-value-input-wrapper` or something instead and then target it with  `.frm-single-settings.frm-type-rte .frm-default-value-input-wrapper`?

AbdiTolesa: @Crabcyborg Thanks for the feedback. I have updated to a more generic class name and also avoided use of `$field_is_rte` variable but I'm still adding a check for the field type when adding the data attributes.

classes/views/frm-fields/back-end/settings.php
```diff
@@ -153,21 +153,28 @@
 					<?php } ?>
 				</span>
 				<?php } ?>
-
-				<p class="frm-has-modal default-value-section-<?php echo esc_attr( $field['id'] . ( isset( $default_value_types['default_value']['current'] ) ? '' : ' frm_hidden' ) ); ?>" id="default-value-for-<?php echo esc_attr( $field['id'] ); ?>">
+				<?php $html_id = 'frm_default_value_' . absint( $field['id'] ); ?>
+				<p
+				class="frm-has-modal frm-default-value-wrapper default-value-section-<?php echo esc_attr( $field['id'] . ( isset( $default_value_types['default_value']['current'] ) ? '' : ' frm_hidden' ) ); ?>"
+				<?php echo $field['type'] === 'rte' ? 'data-modal-trigger-title="' . esc_attr__( 'Toggle Options', 'formidable' ) . '"' : ''; ?>
```
Crabcyborg, line 159: @AbdiTolesa Maybe we can add another function to `FrmFieldType` that handles allows us to echo additional attributes so we can manage this all in Pro.

classes/views/frm-fields/back-end/settings.php
```diff
@@ -153,21 +153,28 @@
...
 						<?php
 						$special_default = ( isset( $field['post_field'] ) && $field['post_field'] === 'post_category' ) || $field['type'] === 'data';
-						FrmAppHelper::icon_by_class(
-							'frm_icon_font frm_more_horiz_solid_icon frm-show-inline-modal',
-							array(
-								'data-open' => $special_default ? 'frm-tax-box-' . $field['id'] : 'frm-smart-values-box',
-								'title'     => esc_attr__( 'Toggle Options', 'formidable' ),
-							)
-						);
+						if ( $field['type'] !== 'rte' ) {
```
Crabcyborg, line 168: Since this is also still here, maybe the new function should handle rendering the whole `<p>` tag here. Just bear in mind that we're still trying to re-use code.

Crabcyborg (approved):
🥳

PR #1020, opened 2022-12-30
individual action type limit should be accounted for in the total act…

AbdiTolesa:
…ions limit


Crabcyborg:
I'm going to close this as I don't think we want to go in this direction.

PR #1021, opened 2022-12-30
exclude repeaters from exportable forms list

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3988

### DIFF
--- a/classes/controllers/FrmXMLController.php
+++ b/classes/controllers/FrmXMLController.php
@@ -282,8 +282,8 @@ class FrmXMLController {
 
 	public static function form( $errors = array(), $message = '' ) {
 		$where = array(
-			'status' => array( null, '', 'published' ),
-			'parent_form_id' => 0,
+			'status'         => array( null, '', 'published' ),
+			'parent_form_id' => array( null, 0 ),
 		);
 		$forms = FrmForm::getAll( $where, 'name' );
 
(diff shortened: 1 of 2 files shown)


classes/views/xml/import_form.php
```diff
@@ -123,7 +123,12 @@
 						</tr>
 					</thead>
 					<tbody>
-						<?php foreach ( $forms as $form ) { ?>
+						<?php
+						foreach ( $forms as $form ) {
```
Crabcyborg, line 127: @AbdiTolesa What do you think about only getting the `$forms` from the database without a parent form ID in the first place? It could help reduce the amount of memory required and speed up the database call.

I don't believe we need repeater forms anywhere here. `$forms` is passed to a `frm_csv_opts` filter but when we use that filter we're hiding repeater forms there already as well.

Screen Shot 2022-12-30 at 10 03 40 AM

AbdiTolesa: Done, thanks!

classes/controllers/FrmXMLController.php
```diff
@@ -283,6 +283,7 @@ public static function route() {
 	public static function form( $errors = array(), $message = '' ) {
 		$where = array(
 			'status' => array( null, '', 'published' ),
+			'parent_form_id' => 0,
```
Crabcyborg, line 286: We may also want to check for `NULL` as `parent_form_id` is a nullable column.

AbdiTolesa: Done, good to know about that.

garretlaxton (approved):
This looks good! The repeater field no longer shows in the export section. I also tested import/export which works fine with repeaters still.

Crabcyborg (approved):

PR #1026, opened 2023-01-12
Add a new filter frm_should_sanitize_field_opt_string

Crabcyborg:
Related pull request https://github.com/Strategy11/formidable-pro/pull/4010
Required to fix the issue I mention in https://github.com/Strategy11/formidable-pro/pull/4007

### DIFF
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -382,15 +382,38 @@ class FrmForm {
 		return $values;
 	}
 
+	/**
+	 * @param string $opt
+	 * @param mixed  $value
+	 * @return void
+	 */
 	private static function sanitize_field_opt( $opt, &$value ) {
-		if ( is_string( $value ) ) {
-			if ( $opt === 'calc' ) {
-				$value = self::sanitize_calc( $value );
-			} else {
-				$value = FrmAppHelper::kses( $value, 'all' );
-			}
-			$value = trim( $value );
+		if ( ! is_string( $value ) ) {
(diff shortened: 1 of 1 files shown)


classes/models/FrmForm.php
```diff
@@ -382,15 +382,38 @@ public static function update_fields( $id, $values ) {
...
+		/**
+		 * Allow the option to turn off sanitization for a field. This way a custom rule can be used instead.
+		 * Make sure to add custom sanitization using the frm_update_field_options filter as the data will no longer be sanitized.
+		 *
+		 * @since x.x
+		 *
+		 * @param bool   $should_sanitize
+		 * @param string $opt
+		 */
+		$should_sanitize = apply_filters( 'frm_should_sanitize_field_opt_string', true, $opt );
```
stephywells, line 404: Actually... I just looked at the other PR. I wonder if it should be more like this, so we can handle the sanitizing.
```
$sanitized_value = apply_filters( 'frm_sanitized_field_opt_string', $value, $opt );
if ( $value !== $sanitized_value ) {
return $sanitized_value;
}
```

Crabcyborg: @stephywells The problem with that is that it will always sanitize it here still if it doesn't get sanitized.

Like a valid value of `' '` wouldn't change, so it would still get trimmed.

stephywells: Thanks. I didn't think that one through very well. This looks great!

PR #1028, opened 2023-01-13
Remove margin bottom from combo inputs container fields only on builder page

Crabcyborg:
I'm hitting this conflict when I try to preview the styler. In a normal preview this is useful.

@truongwp We only need this rule on the builder page, right?

Screen Shot 2023-01-13 at 1 23 07 PM

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -1400,7 +1400,7 @@ form .frm_primary_label input {
 	margin-bottom: 20px;
 }
 
-.with_frm_style form .frm_combo_inputs_container > .form-field {
+#frm_builder_page .with_frm_style form .frm_combo_inputs_container > .form-field {
 	margin-bottom: 0;
 }
 


truongwp:
@Crabcyborg Where do you see that issue? As I see on the Builder page, the Address field doesn't use the `frm_combo_inputs_container` class.

That CSS rule is just for the Builder page. I didn't find other places in the backend containing the Address field.

Crabcyborg:
@truongwp This comes up for me when I try to display an address field on an admin page using the front end rendering (like when the form shortcode function is called).

That class gets used in the front end views, which are used in that case.

This happens now in the visual styler so it becomes fairly obvious. You can see that here https://qa.formidableforms.com/beta/wp-admin/themes.php?page=formidable-styles&form=27

But I can also reproduce this in Visual Views.
Screen Shot 2023-01-18 at 1 04 54 PM

I think the builder page is the only place we want this.

truongwp (approved):
Ok. That CSS should be used when the backend template of the Address field shows, which I think it's just in the form builder.

Crabcyborg:
Thanks for taking a look and confirming @truongwp!

PR #1033, opened 2023-01-19
On submit action

stephywells:
Reopen merged PR https://github.com/Strategy11/formidable-forms/pull/1014


garretlaxton:
When I change the Delay Before Redirect time, after saving it goes back to the default 5 seconds. The page also redirects instantly.

truongwp:
> When I change the Delay Before Redirect time, after saving it goes back to the default 5 seconds. The page also redirects instantly.

The redirection will be delayed when it runs with the Show Message or Show Page action. If there is just one Redirect action runs, it will redirect instantly.

garretlaxton:
When I use the Redirect to URL option I don't see the message before I'm redirected.
image

I also don't see the delay option on the Show Message and Show Page options.
image

stephywells:
> > When I change the Delay Before Redirect time, after saving it goes back to the default 5 seconds. The page also redirects instantly.
> 
> The redirection will be delayed when it runs with the Show Message or Show Page action. If there is just one Redirect action runs, it will redirect instantly.

I fixed this yesterday. The input was missing "name" so it wasn't saving. It was also not being used in several cases.

stephywells:
> I also don't see the delay option on the Show Message and Show Page options.

Since these aren't options to redirect, I thought it didn't make sense to include redirect settings there. @garretlaxton what do you think?

garretlaxton:
@stephywells Yes, that makes sense. I'm just confused about what Truong said "The redirection will be delayed when it runs with the Show Message or Show Page action."

stephywells:
@garretlaxton Here's how it should work:
- If ajax submit is turned on and there are only redirect actions, the redirect message and delay settings should be hidden and it should work the same way as it does in the current version.
- If ajax submit is off or there are other types of success actions, the delay setting on redirect should show
- Any time the redirect message and delay setting show and have a value, they should be used when the form is submitted
- If the redirect message and delay are empty, it should redirect without a delay.

@truongwp is this how it is set up to work?

truongwp:
@stephywells It works differently. There are 2 cases that the Redirect Message and Delay Before Redirect are used:

1. When form is submitted, there is just one Redirect action matches the conditional logics, AJAX submit is off, and some output is printed on the page before we redirect that causes error if we redirect instantly.
2. When form is submitted, there are multiple On Submit actions match the conditional logics, that contain both Redirect and Message, Redirect and Page, or all of three. In this case, we show messages, page content, and the Redirect message, then redirect after the amount of delay time.

garretlaxton:
Thanks for the extra info! I notice the default has `0` as the delay, but when you save the page it will get stuck in a loading loop since the value must be greater than 1.
image

truongwp:
@garretlaxton Yes. I updated the default value of the Delay Before Redirect setting. It's also fixed in my last commit. Can you get the latest code, set that setting to `0` then check again?

classes/controllers/FrmFormsController.php
```diff
@@ -2288,7 +2288,6 @@ private static function run_on_submit_actions( $args ) {
 		} elseif ( 1 === count( $args['conf_method'] ) ) {
 			self::populate_on_submit_data( $args['form']->options, reset( $args['conf_method'] ) );
 			$args['conf_method'] = $args['form']->options['success_action'];
-			$args['time_to_read'] = isset( $args['form']->options['time_to_read'] ) ? $args['form']->options['time_to_read'] : 0;
```
stephywells: @truongwp Does it work for you without this line to add a delay when there is one redirect action with ajax turned off?

garretlaxton:
I just updated and the redirect message is missing after I saved the page.
image

classes/models/FrmOnSubmitAction.php
```diff
@@ -44,7 +44,7 @@ public function get_defaults() {
 			'success_url'     => '',
 			'success_page_id' => '',
 			'redirect_msg'    => FrmOnSubmitHelper::get_default_redirect_msg(),
-			'time_to_read'    => 0, // In seconds.
+			'time_to_read'    => 8, // In seconds. If this is 0, it always redirects instantly.
```
stephywells, line 47: @truongwp we still want to default to be 0. Can you tell me where you found the 8 second delay? I'm not seeing that in the current version.

truongwp: It's here https://github.com/Strategy11/formidable-forms/blob/master/classes/controllers/FrmFormsController.php#L2212

classes/controllers/FrmFormsController.php
truongwp: It works for me. You can try printing something before the checks in `redirect_after_submit()`, it will make `headers_sent()` be `true`.

classes/models/FrmOnSubmitAction.php
truongwp: Please check this comment https://github.com/Strategy11/formidable-forms/pull/1033#issuecomment-1405472226. Maybe you're thinking differently about when it delays redirecting.

classes/controllers/FrmFormsController.php
stephywells: @truongwp does it work for you to set the delay to 2 seconds? It's always using 8 seconds for me now since this value needs to be passed. Can you please test this with a different delay time?

classes/models/FrmOnSubmitAction.php
stephywells: Thanks for pointing that out. So the problem with this is that a redirect action will always have an 8 second delay, even if ajax submit is turned on. Before, that delay was only used as a fallback. This means this is now the default instead of the fallback.

Can you switch this back to 0 as the default setting so we don't have this problem?

classes/controllers/FrmFormsController.php
truongwp: You're right. It's always using 8 seconds when just one Redirect action runs. Let me update the code.

classes/models/FrmOnSubmitAction.php
truongwp: I tested this case. The `time_to_read` won't be populated if there is just one Redirect action runs. So it will redirect instantly, even AJAX submit is on or off.

classes/controllers/FrmFormsController.php
truongwp: @stephywells I pushed the fix for this issue.

classes/models/FrmOnSubmitAction.php
stephywells: @truongwp we want the delay to happen if ajax submit is off. We need to allow time to read the redirect message even if only on action.

classes/controllers/FrmFormsController.php
stephywells: @truongwp instead of adding the force_delay_redirect, can you please return to using a single variable as I had it before? I don't think this is working the way we want it to because the default still must be 0.

classes/models/FrmOnSubmitAction.php
truongwp: > without ajax turned on, they should have to option to show a message and delay the redirect.

Is this a new thing from the current version?

stephywells: Yes. The redirect message is also new in this version.

stephywells: Actually... This option should show with ajax turned on too. It would be confusing if it comes and goes.

But it's super important to default to a 0 second delay and and empty redirect message.

stephywells: Do we have a way to check if it was migrated? In most cases, the current behavior is no delay so we need to be careful not to change that.

truongwp: Yes, we have a way to check.

Where will we get the delay time when there are multiple actions run, when time_to_read is 0 since I want to redirect instantly if just one redirect runs?

truongwp:
@garretlaxton We decided to remove the Redirect Message and Delay Before Redirect settings from this version.

garretlaxton:
I have 2 on submit actions in my form. One is off and the other is on, when I submit the form both are triggered so my `On Submit Redirect` action still triggers while it's marked as off.
image

truongwp:
@garretlaxton Thank you. I fixed it. Please test again.

garretlaxton:
The last thing I see is when Show Page Content is selected, after I submit a form the page is blank.

truongwp:
@garretlaxton I can't replicate that issue. Can you double-check if the page content is empty? If it's not, can you replicate it on the QA site?

garretlaxton:
@truongwp The page does have content. Can you upload this to the QA so I can try it there?

truongwp:
@garretlaxton I uploaded this version to the QA site. Please activate the `5.5.5-on-submit` version.

stephywells:
@garretlaxton Would you mind going through this list and check off everything that you know works? I'm not sure how much I need to check:
- [ ] Set up editing with **redirect** in the current version. Switch to this branch only for lite and test submitting and editing a form **without making any changes** to the settings.
- [ ] Repeat with a custom edit message instead of redirect.
- [ ] Save the form settings and test again.
- [ ] Repeat all of the above but update Pro too.
- [ ] Test one success message action with **ajax submit off**
- [ ] Switch the action to redirect and test again
- [ ] Test one success action with **ajax submit on**
- [ ] Switch the action to redirect and test again
- [ ] Test with multiple actions without logic with and without ajax submit
- [ ] Test with multiple actions with logic with and without ajax submit
- [ ] Test editing an entry with the above setups (ajax, multiple, single...)

Anything else I'm missing? This is super important that this works right and there are a lot of moving parts.

truongwp:
@garretlaxton It works for me: 


https://user-images.githubusercontent.com/19748318/215786404-fe092cb4-f698-4c5b-886b-4bf61dade1de.mov

garretlaxton:
This is what I see when I submit the form. Is this working for you @stephywells?
image

truongwp:
@garretlaxton I saw it when Previewing In theme. Let me check.

garretlaxton:
When editing an entry from the front-end it always shows `Your responses were successfully submitted. Thank you!` whether show message or redirect is selected. This happens with both only Lite and the Pro on submit version.

truongwp:
That issue happens with the code on the master branch too, and just in the 
Preview.

> When editing an entry from the front-end it always shows `Your responses were successfully submitted. Thank you!` whether show message or redirect is selected. This happens with both only Lite and the Pro on submit version.

Did you check "Update" in the Trigger this action setting?

garretlaxton:
@truongwp Whoops! I did not. 

So without the on submit pro, should I still see the `Your responses were successfully submitted. Thank you!` message since the trigger options aren't there?

truongwp:
@garretlaxton Without the On Submit version of Pro, you should see what you set in the Form Settings > Form Permissions > Allow front-end editing of entries.

garretlaxton:
@truongwp Thanks for clearing that up for me! If I try and change the On Update option to something other than Show Message, it seems to always go back to Show Message.

truongwp:
@garretlaxton I fixed the empty page content issue. It happens only when you preview In theme.

> Thanks for clearing that up for me! If I try and change the On Update option to something other than Show Message, it seems to always go back to Show Message.

I will check this.

truongwp:
@garretlaxton I fixed the issue with old Pro version.

garretlaxton:
I don't have the Pro on submit version installed, but I see the trigger action in the On Submit action now. 
image

Also, if I have On Update set to show page content, when I update an entry it shows a blank page like you previously fixed here: https://github.com/Strategy11/formidable-forms/pull/1033/commits/f2cfef9ff3ee2442b20194b4f34d51c57217544a
image

truongwp:
> I don't have the Pro on submit version installed, but I see the trigger action in the On Submit action now.

Steph added it as a fallback to prevent "update" On Submit action becomes "create" On Submit action. For now, it shows when Pro is installed, not matter what version.

> I have On Update set to show page content, when I update an entry it shows a blank page.

@garretlaxton Did you get this issue with the new or old Pro?

truongwp:
Oh. I see it's the old Pro. Let me check it.

truongwp:
@garretlaxton It was an issue with the current version of Pro. I fixed it in the On Submit action PR in Pro: https://github.com/Strategy11/formidable-pro/pull/3995/commits/c6fe252b1fd311ba7af8788a531e1cd6c420d472#diff-0b783bc2474a31b50b78b8e63872466196b161f94bd6c7a01f4cd72205e8f060R1102

Can you check again to confirm that that is an issue with current version of Lite and Pro?

garretlaxton:
@truongwp Yes, I took a look and it only happens on older versions of Pro.

truongwp:
@garretlaxton I meant it happens with old Lite too. It's a bug. And this PR fixes it.

garretlaxton:
@truongwp Sorry for the confusion. Yes, you're right.

truongwp:
@stephywells Can you take a look at my last 2 commits? When migrating, it get all On Submit actions of the form, compare the `event` array and the `success_method`. If there is at least one existing action matches the action we want to create, it stop creating.

I also added a check to not migrating "update" On Submit action if Pro is old.

To test these things, I modified the form options serialized data in the DB, so it might be hard to ask help from Garret.

classes/controllers/FrmFormsController.php
```diff
@@ -2597,6 +2838,251 @@ function() {
...
 	}
 
+	/**
+	 * Gets the first On Submit action to update the success action data in form object.
+	 *
+	 * @since 6.0
+	 *
+	 * @param object $form
+	 */
+	public static function update_get_form_success_action( &$form ) {
```
Crabcyborg, line 2848: `$form` is an object so it probably doesn't need to be passed by reference, right? It should just work.

truongwp: Thank you. This method is removed since we decided to save success settings when saving form actions instead of populating the data when getting form.

truongwp:
@garretlaxton Can you test the migration again, then test this https://github.com/Strategy11/formidable-pro/issues/4049?

garretlaxton:
I got the following warning when using this with the current version of pro (5.5.7).
```
PHP Warning:  Undefined array key "success_action" in C:\wamp64\www\qa-form\wp-content\plugins\formidable\classes\controllers\FrmFormsController.php on line 2939
PHP Stack trace:
PHP   1. {main}() C:\wamp64\www\qa-form\wp-admin\admin.php:0
PHP   2. do_action($hook_name = 'toplevel_page_formidable') C:\wamp64\www\qa-form\wp-admin\admin.php:259
PHP   3. WP_Hook->do_action($args = [0 => '']) C:\wamp64\www\qa-form\wp-includes\plugin.php:517
PHP   4. WP_Hook->apply_filters($value = '', $args = [0 => '']) C:\wamp64\www\qa-form\wp-includes\class-wp-hook.php:332
PHP   5. FrmFormsController::route('') C:\wamp64\www\qa-form\wp-includes\class-wp-hook.php:308
PHP   6. FrmFormsController::settings($id = [], $message = *uninitialized*) C:\wamp64\www\qa-form\wp-content\plugins\formidable\classes\controllers\FrmFormsController.php:1729
PHP   7. FrmFormsController::maybe_migrate_submit_settings_to_action($form_id = 122) C:\wamp64\www\qa-form\wp-content\plugins\formidable\classes\controllers\FrmFormsController.php:143
PHP   8. FrmFormsController::get_on_submit_action_data_from_form_options($form_options = ['akismet' => '', 'honeypot' => 'basic', 'antispam' => '1', 'ajax_submit' => '1', 'logged_in_role' => [0 => ''], 'single_entry_type' => 'user', 'cookie_expiration' => '8000', 'editable_role' => [0 => ''], 'draft_msg' => 'Your draft has been saved.', 'protect_files' => '0', 'noindex_files' => '0', 'open_status' => '', 'open_date' => '2023-01-06 19:31', 'close_date' => '', 'max_entries' => '', 'closed_msg' => '<p>This form is currently closed for submissions.</p>', 'custom_style' => '1', 'transition' => '', 'submit_value' => 'Submit', 'edit_value' => 'Update', 'draft_label' => 'Save Draft', 'submit_align' => '', 'submit_conditions' => ['show_hide' => 'show', 'any_all' => 'all'], 'start_over_label' => 'Start Over', 'form_class' => '', 'before_html' => '<legend class=\\"frm_screen_reader\\">[form_name]</legend>\r\n[if form_name]<h3 class=\\"frm_form_title\\">[form_name]</h3>[/if form_name]\r\n[if form_description]<div class=\\"frm_description\\">[form_description]</div>[/if form_description]', 'after_html' => '', 'submit_html' => '<div class=\\"frm_submit\\">\r\n[if back_button]<button type=\\"submit\\" name=\\"frm_prev_page\\" formnovalidate=\\"formnovalidate\\" class=\\"frm_prev_page\\" [back_hook]>[back_label]</button>[/if back_button]\r\n<button class=\\"frm_button_submit\\" type=\\"submit\\"  [button_action]>[button_label]</button>\r\n[if save_draft]<a href=\\"#\\" tabindex=\\"0\\" class=\\"frm_save_draft\\" [draft_hook]>[draft_label]</a>[/if save_draft]\r\n[if start_over]<a href=\\"#\\" tabindex=\\"0\\" class=\\"frm_start_over\', 'show_form' => 0, 'no_save' => 0, 'ajax_load' => 0, 'js_validate' => 0, 'show_title' => 0, 'show_description' => 0, 'save_draft' => 0, 'open_editable_role' => '-1', 'copy' => 0, 'single_entry' => 0, 'success_page_id' => '', 'success_url' => '', 'prev_value' => 'Previous', 'rootline' => '', 'pagination_position' => '', 'rootline_titles_on' => 0, 'rootline_titles' => [], 'rootline_lines_off' => 0, 'rootline_numbers_off' => 0], $event = *uninitialized*) C:\wamp64\www\qa-form\wp-content\plugins\formidable\classes\controllers\FrmFormsController.php:2897
```

truongwp:
@garretlaxton Can you test with a new form, or save the form in the `master` version?

garretlaxton:
Thanks @truongwp. I updated the form and that resolved it. The on submit issue with the styler is resolved for me.

I am seeing the blank page again when I have Show Page Content selected.
image

garretlaxton:
It's happening when the form is on a page and I use a different page other than the sample page.

I'm also seeing 2 on submit actions added when I create a new form.

truongwp:
@garretlaxton I fixed the 2 actions issue. I'm checking the empty page issue. Look like it's not related to this PR.

garretlaxton:
Thanks, this looks good! I am using the current release of the Pro version if that's any help.

truongwp:
I am able to replicate it with both new versions. 

This is strange. The code runs correctly. I can log the page content. Maybe it redirects twice, or the form shortcode doesn't work correctly.

I need to sleep now. I will check this later.

classes/views/frm-form-actions/form_action.php
```diff
@@ -7,9 +7,16 @@
 
 $form_action = apply_filters( 'frm_form_action_settings', $form_action, $form_action->post_excerpt );
 $form_action = apply_filters( 'frm_form_' . $form_action->post_excerpt . '_action_settings', $form_action );
-
+$data_attrs  = 'data-actionkey="' . esc_attr( $action_key ) . '"';
```
stephywells, line 10: Can we switch this to an array, and use FrmAppHelper::array_to_html_params to echo it? Then we can skip the phpcs:ignore too.

stephywells (commented):
@truongwp I have a few requests. I haven't tested this in detail, but this is from a look at most of the changed code in this PR.

I'm not sure if this is stable enough to have finalized today. What are your thoughts on this?

truongwp:
@garretlaxton That issue is fixed.

tests/forms/test_FrmFormsController.php
```diff
@@ -150,35 +150,123 @@ public function test_front_head() {
...
+		wp_cache_delete( $form_id, 'frm_form' );
+
+		$form = $this->factory->form->get_object_by_id( $form_id );
 
 		$entry_key = 'submit-redirect';
 		$response = $this->post_new_entry( $form, $entry_key );
 
-		if ( headers_sent() ) {
-			// since headers are sent by phpunit, we will get the js redirect
-			$this->assertNotFalse( strpos( $response, "window.location='http://example.com'" ) );
```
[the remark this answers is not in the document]
truongwp: I'm not sure why we have a custom handler for `redirect` in https://github.com/Strategy11/formidable-forms/pull/1033/commits/b53e0424c2d8a364d1f68b22f7cd4051df32dde9#diff-1b880a2d11e84622570383ea0503da990c04c1ccbffe4f5f81ab21d68fa00280R2157. I thought it's unnecessary since we used form actions.

I added this test back and added On Submit handler there.

tests/forms/test_FrmFormsController.php
```diff
@@ -205,8 +293,23 @@ public function run_message_after_create( $show_form = 0 ) {
 				),
 			)
 		);
+
+		// Test default action.
 		$this->assertEquals( $form->options['success_action'], 'message' );
 
+		$this->create_on_submit_action(
```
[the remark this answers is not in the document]
truongwp: The default On Submit action is created in `FrmFormsController::build_new_form()` method: https://github.com/Strategy11/formidable-forms/pull/1033/files#diff-1b880a2d11e84622570383ea0503da990c04c1ccbffe4f5f81ab21d68fa00280R661. While in the tests, we just run `FrmForm::create()` method: https://github.com/Strategy11/formidable-forms/blob/master/tests/base/frm_factory.php#L22. So in the tests, default On Submit action isn't created.

truongwp:
@stephywells I fixed them. I tested all cases once again and they worked correctly. I think this is stable enough.

tests/forms/test_FrmFormsController.php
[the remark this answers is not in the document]
stephywells: Thanks for explaining that!

stephywells:
@truongwp have you tested this with import and export?

truongwp:
@stephywells Yes. I tested with import and export.

garretlaxton (approved):
This is looking good now!

classes/controllers/FrmFormsController.php
```diff
@@ -2584,6 +2843,137 @@ function() {
...
+	/**
+	 * Populates the On Submit data to form options.
+	 *
+	 * @since 6.0
+	 *
+	 * @param array  $form_options Form options.
+	 * @param object $action       Optional. The On Submit action object.
+	 * @param string $event        Form event. Default is `create`.
+	 */
+	public static function populate_on_submit_data( &$form_options, $action = null, $event = 'create' ) {
```
Crabcyborg, line 2855: @truongwp Would this function make more sense in the `OnSubmitHelper`? I'd like to avoid adding too much extra to the big controller files.

classes/helpers/FrmTipsHelper.php
```diff
@@ -13,20 +13,28 @@ public static function pro_tip( $callback, $html = '' ) {
...
+		if ( isset( $tip['link'] ) && ! isset( $tip['link']['medium'] ) ) {
 			$tip['link']['medium'] = 'tip';
 		}
 
-		$link = FrmAppHelper::admin_upgrade_link( $tip['link'], $tip['page'] );
+		if ( 'p' === $html ) {
+			echo '<p class="frmcenter frm-mt-0">';
+		}
+
+		$link = empty( $tip['link'] ) ? $tip['page'] : FrmAppHelper::admin_upgrade_link( $tip['link'], $tip['page'] );
```
Crabcyborg, line 35: @truongwp Is it possible to just use `FrmAppHelper::admin_upgrade_link` outside of this function in the `page` param? I'm not sure if we need to add more options to this function?

js/formidable_admin.js
```diff
@@ -9438,6 +9424,71 @@ function frmAdminBuildJS() {
...
+	 * Does the same as jQuery( document ).on( 'event', 'selector', handler ).
+	 *
+	 * @since 5.4.2
+	 *
+	 * @param {String}         event    Event name.
+	 * @param {String}         selector Selector.
+	 * @param {Function}       handler  Handler.
+	 * @param {Boolean|Object} options  Options to be added to `addEventListener()` method. Default is `false`.
+	 */
+	function documentOn( event, selector, handler, options ) {
```
Crabcyborg, line 9437: @truongwp What do you think about moving this to `dom.js`, maybe as part of the `util` object.

I had copied this over to https://github.com/Strategy11/formidable-forms/blob/c1a35c93abb20ab58e931d3dfc249d8a8387930a/js/admin/style.js#L1393 because we only needed styler when I was working on it.

Some admin pages I try to avoid this big file altogether. But `dom.js` is intended for global use without anything specific to one page, so this makes sense there.

js/formidable_admin.js
```diff
@@ -9438,6 +9424,71 @@ function frmAdminBuildJS() {
...
+		if ( 'undefined' === typeof options ) {
+			options = false;
+		}
+
+		document.addEventListener( event, function( e ) {
+			var target;
```
Crabcyborg, line 9443: This also isn't a big deal but I'm trying to move away from using `var` on our admin scripts as we work on updating pieces of the code. So let's change this to `let`.

Crabcyborg:
@truongwp It looks like when you create a new action, the autocomplete doesn't initialize right away so I can't pick a page.

Crabcyborg:
It seems the autocomplete is a bit funny for the page autocomplete as well. I'm not sure why some of it is gray and some isn't, but I can see this in multiple browsers. I tried pulling the latest styler changes in here too, and it's still an issue, so I don't think this has been fixed in another branch.

Screen Shot 2023-02-09 at 3 00 11 PM

truongwp:
@Crabcyborg I removed the gray background. It's from `jquery-ui.css`.

classes/helpers/FrmTipsHelper.php
truongwp: @Crabcyborg I don't understand this comment. Can you describe it more details?

js/formidable_admin.js
truongwp: I moved it and updated style.js too. Please take a look at it.

Crabcyborg:
@truongwp Thanks for updating this! A new action initializes properly now for a new action, but I'm still seeing an issue if I use the duplicate trigger.

Screen Shot 2023-02-10 at 3 26 35 PM

js/formidable_admin.js
```diff
@@ -9465,6 +9452,71 @@ function frmAdminBuildJS() {
...
+
+			// loop parent nodes from the target to the delegation node.
+			for ( target = e.target; target && target != this; target = target.parentNode ) {
+				if ( target.matches( selector ) ) {
```
Crabcyborg, line 9475: There's a support ticket today that I think could be related to this line. https://secure.helpscout.net/conversation/2152523030/150610

But I think in this case it's specific to the version of `documentOn` that exists in our front end JS in lite,
`frm.min.js:289 Uncaught TypeError: target.matches is not a function error when using a repeater field.
Already tried to disable all other scripts`

I think `target.parentNode` eventually becomes null and then it tries to use `.matches` on null. Can we add a check first that `.matches` is actually a function maybe?

js/formidable_admin.js
Crabcyborg: I'm not sure if this commit didn't get pushed or something. It looks the same as before to me.

classes/helpers/FrmTipsHelper.php
Crabcyborg: If we just did `'page' => `FrmAppHelper::admin_upgrade_link( $var1, $var2 )` outside of this, we wouldn't need a new option for basically the same thing.

I'm jut not sure if it's necessary to make this function more complex. This new option isn't really documented at all and it's a bit confusing why it's needed or what sets it apart from page from looking at the function.

I guess one difference is the `target="_blank"` too so a new option may be necessary. I just think this could be done in a more readable way.

js/formidable_admin.js
truongwp: Oh sorry. I didn't pushed it.

classes/helpers/FrmTipsHelper.php
truongwp: I'm not sure about this. I think we need to ask Steph's opinion. If we change that, we'll need to update everywhere we use it, right?

Crabcyborg: I think an Extra option makes sense. Page and link are really the same name for the same thing. You basically have to read the function to know what page and link do.

I want it to be more descriptive and easier to follow.

Maybe we call the new option upgrade_link?

js/formidable_admin.js
truongwp: I added checks for target.matches in both backend and frontend JS.

classes/helpers/FrmTipsHelper.php
truongwp: I added docs comment for this method. @stephywells Since you added this method, can you take a look at it?

PR #1034, opened 2023-01-19
6.0 styles

stephywells:
Reopen https://github.com/Strategy11/formidable-forms/pull/1032

Add Inter font
Update top nav to use flex
Start adding more generic, self-explanatory css vars
Switch to caret icon everywhere
Add .frm-button-tertiary
Add .frm-button-sm


Crabcyborg:
@stephywells I notice that in the branch branch the floating button font weight is really heavy now.

Is this an intentional change?

**The design**
Screen Shot 2023-01-23 at 11 19 59 AM

**How it looks on my branches**
Screen Shot 2023-01-23 at 11 19 54 AM

**How the beta looks**
Screen Shot 2023-01-23 at 11 22 24 AM

css/frm_admin.css
```diff
@@ -7772,13 +7851,32 @@ button.frm_choose_image_box,
...
+	color: var(--grey-800);
+	font-size: var(--text-md);
 	margin: 0;
 	background-color: #fff;
-	line-height: 1;
+	line-height: 24px;
+}
+
+.frm_wrap textarea:focus,
+.frm_wrap input[type="tel"]:focus,
```
Crabcyborg, line 7864: Is there a specific reason this is for all of these types and not just something like `input:focus`? What type are we avoiding here? Maybe a comment would help for context?

stephywells: Yeah. We have to use the types to override other styles in WP. We could probably us an important instead though. I'll be working on this more today. I suppose this PR should be marked as a draft.

stephywells:
> @stephywells I notice that in the beta branch the floating button font weight is really heavy now.
> 
> Is this an intentional change?

The `button` element has font weight now so it'll be consistent. It looks like this will need to be overridden in the .frm_floating_style_button class instead.

stephywells:
Squashing this into a new PR. I messed up and merged a branch into this one.

PR #1035, opened 2023-01-23
remove repeaters from exportable forms only for xml export

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4024

This update uses js to show or hide repeaters in the form export page based on the type of export selected. When CSV is selected, both forms and repeaters would be shown and the repeaters are hidden when XML is selected.

image

image

### DIFF
--- a/classes/views/xml/import_form.php
+++ b/classes/views/xml/import_form.php
@@ -124,7 +124,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 					</thead>
 					<tbody>
 						<?php foreach ( $forms as $form ) { ?>
-							<tr class="frm-row <?php echo ! empty( $form->parent_form_id ) ? esc_attr( 'is-repeater' ) : ''; ?>">
+							<tr class="frm-row <?php echo ! empty( $form->parent_form_id ) ? esc_attr( 'frm-is-repeater' ) : ''; ?>">
 								<td>
 									<input type="checkbox" name="frm_export_forms[]" value="<?php echo esc_attr( $form->id ); ?>" id="export_form_<?php echo esc_attr( $form->id ); ?>" />
 								</td>
--- a/classes/views/xml/import_form.php
+++ b/classes/views/xml/import_form.php
@@ -124,7 +124,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 					</thead>
 					<tbody>
 						<?php foreach ( $forms as $form ) { ?>
-							<tr class="frm-row">
+							<tr class="frm-row <?php echo ! empty( $form->parent_form_id ) ? esc_attr( 'is-repeater' ) : ''; ?>">
 								<td>
 									<input type="checkbox" name="frm_export_forms[]" value="<?php echo esc_attr( $form->id ); ?>" id="export_form_<?php echo esc_attr( $form->id ); ?>" />
 								</td>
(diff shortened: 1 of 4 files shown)


classes/views/xml/import_form.php
```diff
@@ -124,7 +124,7 @@
 					</thead>
 					<tbody>
 						<?php foreach ( $forms as $form ) { ?>
-							<tr class="frm-row">
+							<tr class="frm-row <?php echo ! empty( $form->parent_form_id ) ? esc_attr( 'is-repeater' ) : ''; ?>">
```
Crabcyborg, line 127: Can we use a `frm-` prefix to avoid any possible conflicts?

js/formidable_admin.js
```diff
@@ -9053,13 +9072,20 @@ function frmAdminBuildJS() {
 			addons.add( 'frm-limited-actions' );
 		}
 
+		const exportOption = document.querySelector( 'select[name="format"]' ).value;
+
 		for ( i = 0; i < items.length; i++ ) {
 			var innerText = items[i].innerText.toLowerCase();
+			const itemCanBeShown = ! ( exportOption === 'xml' && items[i].classList.contains( 'is-repeater' ) );
```
Crabcyborg, line 9079: Is it possible to just call `showOrHideRepeaters` on load as well or avoid having the same logic twice?

AbdiTolesa: I'm adding this check here so that repeaters are not shown dynamically on search if `xml` is selected as export type. Calling `showOrHideRepeaters` on load wouldn't prevent that from happening.

js/formidable_admin.js
```diff
@@ -8188,6 +8197,29 @@ function frmAdminBuildJS() {
...
+			return;
+		}
+
+		if ( exportOption === 'csv' ) {
+			repeaters.forEach( form => {
+				form.classList.remove( 'frm_hidden' );
+			});
+		} else {
+			repeaters.forEach( form => {
+				form.classList.add( 'frm_hidden' );
```
Crabcyborg, line 8216: There's a `classList.toggle` function we could use here. We could reduce this if/else. Something more like:

```
const showRepeater = exportOption === 'csv';
repeaters.forEach( form => {
		form.classList.toggle( 'frm_hidden', showRepeater );
});
```

garretlaxton (approved):
Looks good. Repeaters are only hidden from the XML option.

js/formidable_admin.js
```diff
@@ -10367,6 +10404,10 @@ function frmAdminBuildJS() {
 			jQuery( '#frm_export_xml input, #frm_export_xml select' ).on( 'change', removeExportError );
 			jQuery( 'input[name="frm_import_file"]' ).on( 'change', checkCSVExtension );
 			jQuery( 'select[name="format"]' ).on( 'change', checkExportTypes ).trigger( 'change' );
+			document.querySelector( 'select[name="format"]' ).addEventListener( 'change', event => {
```
Crabcyborg, line 10407: This place still doesn't check if the format input actually exists.

But I think we should just use `checkExportTypes` considering there's already an event listener on this input. It seems weird to have two listeners for the same input defined back to back,

AbdiTolesa: @Crabcyborg I originally placed it in that function but moved it out thinking it makes maintaining the code easier. I should indeed need to add a check for the element first indeed, or just place it inside that function otherwise.

Crabcyborg: I think we can improve maintainability by calling two functions from inside of a single event listener. It probably shouldn't actually be inside of `checkExportTypes` but it makes sense to just listen for the change once and call both.

AbdiTolesa: Makes sense, I just updated this accordingly.

PR #1036, opened 2023-01-24
Extra double quote in recaptcha field fix

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2134228507/148173/

There was an extra double quote left from the hCaptcha update. This update removes the weird `"=""` you can see in this example.

Screen Shot 2023-01-24 at 9 28 50 AM


AbdiTolesa (approved):
Thanks for the quick fix!

PR #1038, opened 2023-01-25
Version/6.0b

stephywells:
New combo branch since I corrupted the other


js/formidable_admin.js
```diff
@@ -7831,110 +7811,16 @@ function frmAdminBuildJS() {
...
-			}
-		});
-
 		const settings = event.target.closest( '.frm_form_action_settings' );
-		if ( settings && settings.classList.contains( 'frm_single_email_settings' ) ) {
-			onEmailActionLoaded( settings );
-		}
-	}
-
-	function onEmailActionLoaded( settings ) {
```
Crabcyborg, line 7847: @truongwp It may be safer to keep the type check and just target both email and submit actions.

Outcome quizzes initialize a wysiwyg as well, with the initializing all happening in the quizzes add on. It's pretty much identical to this, but it makes it more flexible if we wanted to treat it differently in another type of action.

I believe this change would cause it to always initialize the quiz outcome wysiwyg twice, which would probably be fine, but it could lead to future conflicts and it does more work than it should be.

truongwp: Ok. I updated it.

Crabcyborg:
I think the dropdown for the bulk actions in the Form Entries page is off.

Screen Shot 2023-02-08 at 2 07 22 PM

There's much more space between the text and the arrow in the forms list.
Screen Shot 2023-02-08 at 2 08 06 PM

stephywells:
> @stephywells I think the dropdown for the bulk actions in the Form Entries page is off.

@Crabcyborg I'm not seeing this. I did make some updates for that in the past day or two. Do you see it here too?
https://qa.formidableforms.com/beta/wp-admin/admin.php?page=formidable-entries&frm_action=list&form=17

Crabcyborg:
@stephywells I don't see it on https://qa.formidableforms.com/beta/wp-admin/admin.php?page=formidable-entries&frm_action=list&form=17 but I see it on https://qa.formidableforms.com/beta/wp-admin/admin.php?page=formidable-entries

When there isn't a form in the query parameters, it looks different.

stephywells:
> When there isn't a form in the query parameters, it looks different.

Thanks! Got it!

Crabcyborg:
I just merged master into this and it's just a line break different from master now.

Screen Shot 2023-02-14 at 1 08 15 PM

So I'm just going to close this 🚀.

PR #1037, opened 2023-01-25
6.0 styles (#1034)

stephywells:
* a few layout fixes for small screens
* set secondary button to white instead of transparent
* update a few icons 
* more consistent padding/margins
* icon vertical alignment
* Make verify on destructive actions red
* Move some buttons into header 
* More consistent search box
* Use small buttons in table filtering
* Use css grids for form template list
* Apply more css to modals and Views page with frm-white-body class
* Style all buttons on frm-white-body
* Psalm fixes
* use compatible return type

PR #1039, opened 2023-01-25
Styler design v2

Crabcyborg:
This includes https://github.com/Strategy11/formidable-forms/pull/1037 and https://github.com/Strategy11/formidable-forms/pull/1004 as well as several new design changes for the styler.

This requires https://github.com/Strategy11/formidable-pro/pull/4040 for all of the Pro support.


classes/helpers/FrmStylesCardHelper.php
```diff
@@ -65,17 +65,13 @@ public function __construct( $active_style, $default_style, $form_id, $enabled )
...
-			$params['class'] .= ' frm-default-style-card';
-		}
+		$is_default_style = $style->ID === $this->default_style->ID;
+		$is_active_style  = $style->ID === $this->active_style->ID;
+		$is_locked        = $this->locked;
+		$params           = $this->get_params_for_style_card( $style );
+
 		if ( $is_active_style ) {
-			$params['class'] .= ' frm-active-style-card';
+			$params['class'] .= ' frm-active-style-card frm-currently-set-style-card';
```
stephywells, line 74: Can the active style apply to the applied style as well? I keep clicking it but it feels like nothing is happening.

Also, would it be hard to switch to edit mode if the current style is already selected when I click it? This would save a click to not need to open the dropdown menu.

classes/views/styles/_style-preview-container.php
```diff
@@ -26,11 +27,11 @@
 			?>
 		</div>
 		<?php if ( 'edit' !== $view ) { ?>
-			<a href="<?php echo esc_url( admin_url( 'admin.php?page=formidable-styles&frm_action=edit&form=' . $form->id ) ); ?>" id="frm_edit_style" class="frm_floating_style_button" tabindex="0" role="button">
+			<a href="<?php echo esc_url( admin_url( 'admin.php?page=formidable-styles&frm_action=edit&form=' . $form->id ) ); ?>" id="frm_edit_style" class="frm_floating_style_button button frm-button-secondary frm-button-sm" tabindex="0" role="button">
```
stephywells, line 30: Can we keep the sample form selected if we are switching to the edit screen? I wanted to see the sample form in edit mode, but clicked sample first. Then when I clicked edit it went back to the custom form.

classes/helpers/FrmStylesCardHelper.php
Crabcyborg: @stephywells This is how Razvan had designed it to work. If it's selected it doesn't look selected intentionally.

It confused me at first though too - It took some conversation to know I was implementing it the way he wanted. And Steve has mentioned it as well in his QA. So changing this probably makes sense.

Crabcyborg: > Also, would it be hard to switch to edit mode if the current style is already selected when I click it? This would save a click to not need to open the dropdown menu.

That sounds a bit jarring. Maybe if we had a pop up that prompted to go to the edit page, but I don't think there's anything intuitive about having only the applied style link to the edit page. It would probably confuse people. I think I'd just assume I clicked something by mistake.

Crabcyborg:
@stephywells I noticed the vertical alignment on the "My templates" icon pushes it way down now so it looks too close to the template underneath.

My Templates

classes/helpers/FrmStylesCardHelper.php
stephywells: That's probably true. We can pass on that one.

classes/views/styles/_style-preview-container.php
Crabcyborg: I just pushed a commit that passes along a `sample=1` parameter in the edit links dynamically on click events if the sample form is active.

Should we do the same on the back button on the edit page? Anywhere else?

Crabcyborg:
@stephywells It looks like several icons were removed in https://github.com/Strategy11/formidable-forms/pull/1039/commits/e771afb8fb3a18ffa01d057f35e66d4580b16c27

I'm not really sure if this is just a work in progress? What icons do we want now if the new ones are being removed?
Screen Shot 2023-01-31 at 2 52 42 PM

stephywells:
> **Edit -** Nevermind! I see there are updates in Pro for this. It looks good once I pulled those.

I was seeing a lot of inconsistent icons. I know this is an ongoing issue, but I'd like to keep them the same style. If we decide we can switch everything out to a thin line, we'll need to do it all at once.

classes/views/styles/_style-preview-container.php
stephywells: Awesome! That sounds like it would cover it. Thank you!

Crabcyborg:
@stephywells I see that you updated the forms list so the style column is now just an icon.

One funny thing I'm seeing is a big gap from the landing page icon. It seems that the space is hard coded into the landing page add on with two `&nbsp;` characters. It doesn't seem easy to ignore those characters with CSS, but we could just remove them even with something like `$val = str_replace( '&nbsp;', '', $val );` after the `frm_form_list_actions` filter.

What do you think?

Screen Shot 2023-02-01 at 9 53 25 AM

css/frm_admin.css
```diff
@@ -849,8 +989,15 @@ p.frm_has_shortcodes {
 	padding-top: 0 !important;
 }
 
+.frm-flex-justify,
+.frm-right-panel h3 {
+	display: flex;
+	justify-content: space-between;
```
Crabcyborg, line 995: @stephywells This new style rule seems to be pushing the style categories way to the right.

Screen Shot 2023-02-01 at 10 14 02 AM

garretlaxton:
I notice the color doesn't fully fit the box. I can see the transparent part on the right a tiny bit. This is on the beta site so not sure if that's been fixed or not.
image

Crabcyborg:
Thanks @garretlaxton it's always been that way, in Pro. I think it has to do with showing the opacity of the color or something. It happens with the alpha color picker library. I should try updating it, maybe that will change it.

garretlaxton:
Thanks, I've never noticed it. I also didn't see it in any screenshots posted in Slack so that's why I mentioned it.

css/frm_admin.css
stephywells: This is fixed now.

stephywells:
That color picker issue is fixed now too. I actually fixed it before I saw your comment Garret. Thank you!

Crabcyborg:
@stephywells I'm not entirely sure why this is happening but I'm noticing that the color pickers are getting cut off.

Screen Shot 2023-02-02 at 4 29 38 PM

It seems that removing `overflow: hidden` from the `.accordion-section-content` parent fixes it, but that's how the styles are in the master branch and it works fine there so I'm not really sure what changes to result in this.

Crabcyborg:
@stephywells Is this an intentional style change? I noticed the multiselect dropdowns have gray backgrounds now.

As far as I can tell it's because they're button elements and have a `background-color: buttonface` style by default. So it doesn't look intentional as far as I can tell.
Screen Shot 2023-02-06 at 12 01 14 PM

Crabcyborg:
@stephywells The conditional logic in a field seems off to me. There isn't an accordion arrow until the first time you've clicked and opened it. It doesn't look very clickable either, as just a section header with no body.

Is this final right now? I think it could use a more obvious call to action here.

X6SZXYb4YI

Crabcyborg:
@stephywells It looks like the CSS layout class options are a bit funny now.

**Before**
Screen Shot 2023-02-06 at 3 33 00 PM

**After**
Screen Shot 2023-02-06 at 3 32 21 PM

garretlaxton:
I'm not sure how updated the QA is, but I don't see the dropdown arrows at the moment.
image

Crabcyborg:
@garretlaxton I just pushed more recent updates. It looks different now. The backend dropdown had no arrow, but I added an !important to make sure it does.

The front end one appears to be back as well but I didn't change anything myself for that. Something else I merged in likely helped, or some setting was changed.

@stephywells When I was looking at the dropdown I noticed we're styling the background image position but that seems to make it look worse

We're setting `background-position-y: center;` here https://github.com/Strategy11/formidable-forms/blob/a585c5615d651136e38720c6fa6a84396462311b/css/custom_theme.css.php#L343

That causes the background position to be in the wrong place in the Twenty Twenty One theme:
Screen Shot 2023-02-10 at 4 16 28 PM

There's a style rule coming from Twenty Twenty One `background-position: right var(--form--spacing-unit) top 60%;` that expects the top position to be 60%. If I remove our center rule it looks center.
Screen Shot 2023-02-10 at 4 16 14 PM

Crabcyborg:
@garretlaxton I think it was a setting thing. If you set the default style to important it will never have a background image because we force it to none here https://github.com/Strategy11/formidable-forms/blob/a585c5615d651136e38720c6fa6a84396462311b/css/custom_theme.css.php#L239

stephywells:
> When I was looking at the dropdown I noticed we're styling the background image position but that seems to make it look worse
> 
> We're setting `background-position-y: center;` here
> 
> https://github.com/Strategy11/formidable-forms/blob/a585c5615d651136e38720c6fa6a84396462311b/css/custom_theme.css.php#L343

@Crabcyborg This PR should only affect form ons the backend. That background position was added here for Astra:
https://github.com/Strategy11/formidable-forms/pull/883

@garretlaxton would you mind creating logging a new issue for that conflict?

Crabcyborg:
> That background position was added here for Astra: #883

Oh, thanks @stephywells! Forgot I was the one who added that rule.

js/admin/dom.js
```diff
@@ -642,6 +687,33 @@
...
+	function success( content ) {
+		const container           = document.getElementById( 'wpbody' );
+		const notice              = div({
+			className: 'notice notice-info frm-review-notice frm_updated_message',
+			child: div({
+				className: 'frm-satisfied',
+				child: 'string' === typeof content ? document.createTextNode( content ) : content
+			})
+		});
+		notice.style.borderRadius = '4px';
```
stephywells, line 706: @Crabcyborg can we move this styling into the css file to make it easier to keep track of and use consistently?

classes/models/FrmSettings.php
```diff
@@ -174,14 +184,33 @@ public function fill_with_defaults( $params = array() ) {
 				$this->{$setting} = $default;
 			}
 
-			if ( $filter_html && in_array( $setting, $filter_keys, true ) ) {
-				$this->{$setting} = FrmAppHelper::kses( $this->{$setting}, 'all' );
```
stephywells, line 178: @Crabcyborg did the filtered html check get removed intentionally?

js/admin/dom.js
Crabcyborg: For sure! https://github.com/Strategy11/formidable-forms/pull/1063

PR #1040, opened 2023-01-26
Specify required license in applications modal

Crabcyborg:
We talked about this in check-in this morning for the style templates modal. I figured I'd do the same here with Applications too.

Screen Shot 2023-01-26 at 4 34 47 PM
Screen Shot 2023-01-26 at 4 34 42 PM


stephywells (approved):
Awesome!

PR #1043, opened 2023-01-30
Stop setting body to overflow-y: initial

Crabcyborg:
I think this may be the cause of https://github.com/Strategy11/formidable-pro/issues/3918 but I'm not 100% positive it's totally resolved. I think it may at least be related. When I noticed this behaviour in the styler, I linked to this issue as it seems similar.

With the visual styler I noticed that every time I closed a modal that the body became scrollable and I'd end up with a double scroll bar.

I had added this `overflowY: initial` rule when a modal closes previously to fix an issue with all modals scrolling to the top of the page which gets weird when you want to use the embed pop up anywhere where you've scrolled (like in the list of actions or in the list of forms).

I'm going with a less conflicting method now. Instead of changing the `overflowY` on the element directly, I added a new `frm-body-with-open-modal` class that I toggle on and off, which sets a `overflowY` rule only when it's open.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -1930,6 +1930,10 @@ h2.frm-h2 + .howto {
 	border-top: 1px solid var(--sidebar-hover);
 }
 
+body.frm-body-with-open-modal {
+	overflow-y: hidden;
+}
+
 #frm_new_form_modal .postbox > div.inside {
 	max-height: 490px;
 	overflow-y: auto;
--- a/js/admin/dom.js
+++ b/js/admin/dom.js
@@ -525,6 +525,8 @@
 	}
 
 	function makeModalIntoADialogAndOpen( modal, { width } = {}) {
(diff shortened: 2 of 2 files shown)


stephywells:
@Crabcyborg do the changes here solve this for you?
https://github.com/Strategy11/formidable-forms/pull/1039/commits/a9186a69c7c3462f3caeefab2925f83a1b5de586

stephywells (approved):
I removed the page container class from most pages in 6.0 to help with this. It does look like it's related to https://github.com/Strategy11/formidable-pro/issues/3918

If I have enough time before we need 6.0 finalized, I'd still like to get that into it as well.

Crabcyborg:
Thanks @stephywells, that update sounds good too.

I can still replicate my issue with setting the overflow-y to `initial` though, as I think that's the root of the specific issue I'm trying to fix here.

I think this update is an improvement to my previous fix so I'm going to merge this.

PR #1044, opened 2023-01-30
fallback to textarea for rte default value field if pro is <= 5.5.6

AbdiTolesa:
This is an extension from the previous update added to make sure paragraph is used instead of text field for RTE field default value setting when lite is updated but pro is not.

Before this update:
image

After:
image

### DIFF
--- a/classes/models/fields/FrmFieldType.php
+++ b/classes/models/fields/FrmFieldType.php
@@ -446,14 +446,16 @@ DEFAULT_HTML;
 	 * @return void
 	 */
 	public function show_default_value_field( $field, $default_name, $default_value ) {
-		$pro_version              = class_exists( 'FrmProDb' ) ? FrmProDb::$plug_version : '0';
-		$min_pro_version_not_met  = version_compare( $pro_version, '5.5.6', '<=' );
-
-		if ( $field['type'] === 'rte' && $min_pro_version_not_met ) {
-			include FrmAppHelper::plugin_path() . '/classes/views/frm-fields/back-end/textarea-default-value-field.php';
-		} else {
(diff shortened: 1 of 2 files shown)


classes/models/fields/FrmFieldType.php
```diff
@@ -446,7 +446,14 @@ public function display_smart_values_modal_trigger_icon( $field ) {
 	 * @return void
 	 */
 	public function show_default_value_field( $field, $default_name, $default_value ) {
-		include FrmAppHelper::plugin_path() . '/classes/views/frm-fields/back-end/default-value-field.php';
+		$pro_version              = class_exists( 'FrmProDb' ) ? FrmProDb::$plug_version : '0';
```
Crabcyborg, line 449: Can we use `FrmAppHelper::meets_min_pro_version` here?

AbdiTolesa: @Crabcyborg Good to know that we have that. The next version would be `5.5.7` right?

AbdiTolesa:
@Crabcyborg I noticed that two `@since x.x` annotations have been merged into main from the previous update.

Crabcyborg:
Thanks @AbdiTolesa sometimes I'll miss them for a couple of releases. They get updated eventually. It would be great to automate it but I haven't gotten to that yet.

Crabcyborg:
@AbdiTolesa It looks like the previous update is causing a phpcs issue.

```
FILE: ...asses/views/frm-fields/back-end/textarea-default-value-field.php
----------------------------------------------------------------------
----------------------------------------------------------------------
 9 | ERROR | [x] Tabs must be used to indent lines; spaces are not
   |       |     allowed
   |       |     (Generic.WhiteSpace.DisallowSpaceIndent.SpacesUsed)
```

Could you also fix that?

PR #1045, opened 2023-01-31
Add new admin toggle element

stephywells:
Moved from https://github.com/Strategy11/formidable-forms/commit/f5269f9ada5a22dc8dcd8876aa5d26dee3057886

Used in place of the toggle in the form actions
Added an option to remove the div on the toggle

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -480,7 +480,8 @@ class FrmAppController {
 		wp_register_script( 'formidable_embed', $plugin_url . '/js/admin/embed.js', array( 'formidable_dom', 'jquery-ui-autocomplete' ), $version, true );
 		self::register_popper1();
 		wp_register_script( 'bootstrap_tooltip', $plugin_url . '/js/bootstrap.min.js', array( 'jquery', 'popper' ), '4.6.1', true );
-		wp_register_script( 'formidable_style_settings', $plugin_url . '/js/admin/style-settings.js', array(), $version, true );
+		wp_register_script( 'formidable_settings', $plugin_url . '/js/admin/settings.js', array( 'formidable-admin' ), $version, true );
+
 
 		$page = FrmAppHelper::simple_get( 'page', 'sanitize_title' );
 
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -480,7 +480,7 @@ class FrmAppController {
 		wp_register_script( 'formidable_embed', $plugin_url . '/js/admin/embed.js', array( 'formidable_dom', 'jquery-ui-autocomplete' ), $version, true );
 		self::register_popper1();
 		wp_register_script( 'bootstrap_tooltip', $plugin_url . '/js/bootstrap.min.js', array( 'jquery', 'popper' ), '4.6.1', true );
-		wp_register_script( 'formidable_settings', $plugin_url . '/js/admin/settings.js', array( 'formidable-admin' ), $version, true );
+		wp_register_script( 'formidable_settings', $plugin_url . '/js/admin/settings.js', array(), $version, true );
 
 
 		$page = FrmAppHelper::simple_get( 'page', 'sanitize_title' );
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -480,6 +480,7 @@ class FrmAppController {
 		wp_register_script( 'formidable_embed', $plugin_url . '/js/admin/embed.js', array( 'formidable_dom', 'jquery-ui-autocomplete' ), $version, true );
 		self::register_popper1();
 		wp_register_script( 'bootstrap_tooltip', $plugin_url . '/js/bootstrap.min.js', array( 'jquery', 'popper' ), '4.6.1', true );
+		wp_register_script( 'formidable_style_settings', $plugin_url . '/js/admin/style-settings.js', array(), $version, true );
 
 		$page = FrmAppHelper::simple_get( 'page', 'sanitize_title' );
 
--- a/classes/helpers/FrmHtmlHelper.php
+++ b/classes/helpers/FrmHtmlHelper.php
@@ -10,7 +10,6 @@ class FrmHtmlHelper {
 
 	/**
 	 * Create a toggle and either echo or return the string.
-	 * This isn't completely compatible with Pro. The input_html option has been omitted as it isn't secure.
 	 * This is intended for use on admin pages only. The CSS is included in frm_admin.css.
 	 *
 	 * @since x.x
--- a/classes/helpers/FrmHtmlHelper.php
+++ b/classes/helpers/FrmHtmlHelper.php
@@ -24,6 +24,7 @@ class FrmHtmlHelper {
 	 * @return string|void
 	 */
 	public static function toggle( $id, $name, $args ) {
+		wp_enqueue_script( 'formidable_style_settings' );
 		return FrmAppHelper::clip(
 			function() use ( $id, $name, $args ) {
 				require FrmAppHelper::plugin_path() . '/classes/views/shared/toggle.php';
--- a/classes/helpers/FrmHtmlHelper.php
+++ b/classes/helpers/FrmHtmlHelper.php
@@ -24,7 +24,7 @@ class FrmHtmlHelper {
 	 * @return string|void
 	 */
 	public static function toggle( $id, $name, $args ) {
-		wp_enqueue_script( 'formidable_style_settings' );
+		wp_enqueue_script( 'formidable_settings' );
 		return FrmAppHelper::clip(
 			function() use ( $id, $name, $args ) { // @phpstan-ignore-line
 				require FrmAppHelper::plugin_path() . '/classes/views/shared/toggle.php';
--- a/classes/helpers/FrmHtmlHelper.php
+++ b/classes/helpers/FrmHtmlHelper.php
@@ -1,0 +1,34 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+
+/**
+ * @since x.x
+ */
+class FrmHtmlHelper {
+
+	/**
+	 * Create a toggle and either echo or return the string.
+	 * This isn't completely compatible with Pro. The input_html option has been omitted as it isn't secure.
+	 * This is intended for use on admin pages only. The CSS is included in frm_admin.css.
+	 *
+	 * @since x.x
+	 *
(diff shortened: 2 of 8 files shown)


Crabcyborg:
Thanks @stephywells! This looks mostly good to me.

The one thing I see missing here is the accessibility scripts. Unfortunately this is pretty inconsistent right now.

Originally we only had this toggle in Pro, so a lot of this stuff all exists in Pro. That includes this JS file for handling the toggle on Pro settings pages (which covered all toggle cases before 6.0 - the ACF form action, the attachment toggles for email actions, the conversational/landing page toggles) https://github.com/Strategy11/formidable-pro/blob/master/js/admin/settings.js

Right now all that file does is handle all of the toggle accessibility.

If you're using Pro, that script loads on this page, so this toggle you're updating in the email action will have the accessibility, but not if you're only using lite, which is an issue.

In the styler I added this `makeToggleAccessible` function, which isn't very reusable in its current state as I just kept it simple and kept it in the styler JS and made it work only for a single toggle https://github.com/Strategy11/formidable-forms/blob/0023e15c178d89728ccc81ac6445af9890c05daa/js/admin/style.js#L95

stephywells:
Thanks for pointing that out @Crabcyborg . I copied that file over since it doesn't look like it will cause a problem if it runs twice temporarily. Do you see any issues with that?

This PR also adds more styling with the labels don't show to make it more clear whether the toggle is on or off.
Screen Shot 2023-02-02 at 9 49 05 AM

classes/views/shared/toggle.php
```diff
@@ -0,0 +1,69 @@
...
+?>
+<?php
+$div_class    = isset( $args['div_class'] ) ? $args['div_class'] : false;
+$show_labels  = isset( $args['show_labels'] ) ? $args['show_labels'] : false;
+$off_label    = isset( $args['off_label'] ) ? $args['off_label'] : '';
+$on_label     = isset( $args['on_label'] ) ? $args['on_label'] : 1;
+$value        = isset( $args['value'] ) ? $args['value'] : $on_label;
+$checked      = isset( $args['checked'] ) && ( true === $args['checked'] || false !== strpos( $args['checked'], 'checked="checked"' ) );
+$aria_checked = $checked ? 'true' : 'false';
+$input_html   = isset( $args['input_html'] ) ? $args['input_html'] : array();
```
Crabcyborg, line 24: @stephywells I think `$args['input_html']` might normally be a string.

It looks like it gets defined in `FrmFieldType::get_field_input_html_hook` which ends with:

```
$input_html = ob_get_contents();
ob_end_clean();
return $input_html;
```

You can see the original code for it in Pro is `echo $input_html; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>`

classes/helpers/FrmHtmlHelper.php
```diff
@@ -0,0 +1,35 @@
...
+}
+
+/**
+ * @since x.x
+ */
+class FrmHtmlHelper {
+
+	/**
+	 * Create a toggle and either echo or return the string.
+	 * This isn't completely compatible with Pro. The input_html option has been omitted as it isn't secure.
```
Crabcyborg, line 13: This comment is now a bit outdated if we're trying to support input_html again. It was only required for the front end field in Pro so I had initially left it out.

classes/controllers/FrmAppController.php
```diff
@@ -480,6 +480,7 @@ public static function admin_js() {
 		wp_register_script( 'formidable_embed', $plugin_url . '/js/admin/embed.js', array( 'formidable_dom', 'jquery-ui-autocomplete' ), $version, true );
 		self::register_popper1();
 		wp_register_script( 'bootstrap_tooltip', $plugin_url . '/js/bootstrap.min.js', array( 'jquery', 'popper' ), '4.6.1', true );
+		wp_register_script( 'formidable_style_settings', $plugin_url . '/js/admin/style-settings.js', array(), $version, true );
```
Crabcyborg, line 483: I think this URL is off. This is pulling a style JS that only exists by this name in Pro. I think this shouldn't have the `style-` in it.

classes/views/shared/toggle.php
stephywells: Yep! Thanks for catching this. I didn't realize it was already there since I copied this from the style branch.

classes/controllers/FrmAppController.php
stephywells: Doh! Sure is. Fixed.

classes/helpers/FrmHtmlHelper.php
stephywells: Give me a sec. I'm looking at using a string only as a fallback.

classes/views/shared/toggle.php
stephywells: This works for an array or string now and adds a deprecated message if a string is used.

classes/controllers/FrmAppController.php
```diff
@@ -480,6 +480,8 @@ public static function admin_js() {
 		wp_register_script( 'formidable_embed', $plugin_url . '/js/admin/embed.js', array( 'formidable_dom', 'jquery-ui-autocomplete' ), $version, true );
 		self::register_popper1();
 		wp_register_script( 'bootstrap_tooltip', $plugin_url . '/js/bootstrap.min.js', array( 'jquery', 'popper' ), '4.6.1', true );
+		wp_register_script( 'formidable_settings', $plugin_url . '/js/admin/settings.js', array( 'formidable-admin' ), $version, true );
```
Crabcyborg, line 483: It looks like `formidable_admin` is the name of the script and `formidable-admin` is the name of the style 😅.

I think we want the underscore here for the script dependency.

classes/views/shared/toggle.php
```diff
@@ -0,0 +1,75 @@
...
+		<input type="checkbox" name="<?php echo esc_attr( $name ); ?>[]" id="<?php echo esc_attr( $id ); ?>" value="<?php echo esc_attr( $value ); ?>"
+			<?php checked( $checked, true ); ?>
+			<?php if ( $show_labels && $off_label ) { ?>
+				data-off="<?php echo esc_attr( $off_label ); ?>"
+			<?php } ?>
+			<?php
+			if ( is_array( $input_html ) ) {
+				FrmAppHelper::array_to_html_params( $input_html, true );
+			} else {
+				_deprecated_argument( '$args[input_html]', '6.0', 'Use an array for "input_html"' );
```
Crabcyborg, line 59: If this is getting deprecated, do we need to update `FrmProToggle::front_field_input` too?

classes/controllers/FrmAppController.php
stephywells: Oh man! Clearly I need a lunch break. That was the style name instead of a script name.

But I removed the dependency since it doesn't actually need it.

classes/views/shared/toggle.php
stephywells: Hmm... I didn't realize this would be used on the front-end too. I just made more changes so we don't mess with the front-end toggles, or the toggles in the form previews.

classes/views/shared/toggle.php
```diff
@@ -0,0 +1,75 @@
...
+}
+
+if ( $use_container ) {
+	?>
+<div <?php FrmAppHelper::array_to_html_params( $div_params, true ); ?>>
+	<?php
+	$div_params = array();
+}
+?>
+	<label class="frm_toggle_block" <?php FrmAppHelper::array_to_html_params( $div_params, true ); ?>>
```
Crabcyborg, line 45: Do we actually want all of the div params on the label too? Maybe a new name that isn't div_params makes sense if it's not being used only a div anymore.

classes/views/shared/toggle.php
Crabcyborg: Cool, then we probably don't need to support the string then. This line can probably go.

stephywells: I think this should be ready. Now the admin and front-end toggles are clearly separated, and the admin toggles use different classes so we don't have styling conflicts with form previews and builder.

The CSS is merged where possible, but we'll need to replace all the frm_toggle_long occurrences before we can remove the extra css.

Crabcyborg:
I think this is looking good! It should at least be ready enough to pull into the submit action too. I doubt we'll need many more changes after this.

🚀

classes/views/shared/toggle.php
stephywells: Yeah, this is intentional. If we don't include the div, the classes are used on the label instead. There are spots where we have toggles inside a span or p, so we can't include a div.

stephywells:
Thanks for all your help! I thought this would be quick since I was just extracting parts from the styler branch. Sorry about the silly mistakes!

classes/views/shared/toggle.php
stephywells: I was thinking about that, and removed it then added it back. My concern is that we'll switch the toggle() calls to admin_toggle() or the lite toggle() function, but not realize a string input_html isn't doing anything. I was thinking we might be a bit safer with a graceful fallback. But on the other hand, since this could be a potential security problem, maybe it's worth it?

Which way would you prefer?

Crabcyborg: Hm, what do you think about using https://developer.wordpress.org/reference/functions/_doing_it_wrong/ instead? Since it's not really a deprecation in this case.

garretlaxton (approved):
This looks good! As Mike mentioned, there are a bit of toggle switches in Pro that don't match this design yet. Other than that, this looks good for Lite.

classes/views/shared/toggle.php
stephywells: This is updated. I also found an error after saving the inbox settings since the name was adding '[]'. That's fixed now too. I think this is ready to merge.

Crabcyborg:
Just noting here that I merged this into https://github.com/Strategy11/formidable-forms/pull/1039 so it's available there as well now.

And the new toggle styling is up on the beta site. 🚀

stephywells:
Awesome!

I switched from using frm_switch to frm_toggle so we can tell the difference between a setting and a form. We were already using frm_toggle in the inbox settings.

This admin branch is updated to handle the js loading for admin toggles using the toggle() function. It loads the js if toggle() is loaded from a Formidable admin page. We'll want to deprecate that officially at some point, but I don't want messages showing all over the place until we update all the add-ons that are using it.

The key/id toggle labels are updated now too. I think we'll want to revisit this with a design since it's inconsistent with toggles in some places and "IDs | Keys" links in other places.

Crabcyborg (approved):
I think this is good to merge now 🚀

Thanks @stephywells!

PR #1049, opened 2023-02-03
Fix broken auto scroll in form builder with scroll bars

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3917

This update fixes the broken autoscroll when dragging fields around a form with many fields that is scrollable. The form now auto scrolls up and down as you grab a field and move the cursor around the top/bottom inside the builder.

BEFORE:
https://user-images.githubusercontent.com/41271840/216592817-68b9ef47-69f4-440e-a8ef-49514b8530ee.mp4

AFTER:
https://user-images.githubusercontent.com/41271840/216592213-471b72b7-adfb-4cbf-9b88-63a9db862d1c.mp4

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -971,12 +971,11 @@ function frmAdminBuildJS() {
 
 	function handleDrag( event, ui ) {
 		var container = jQuery( '#post-body-content' );
-		var startSort = 0;
 
 		container.scrollTop( function( i, v ) {
 			var moved, h, relativePos, y;
 
-			moved = event.clientY - startSort;
+			moved = event.clientY;
 			h = this.offsetHeight;
 			relativePos = event.clientY - this.offsetTop;
 			y = relativePos - h / 2;
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -9771,7 +9771,7 @@ function frmAdminBuildJS() {
 			debouncedSyncAfterDragAndDrop = debounce( syncAfterDragAndDrop, 10 );
 			postBodyContent = document.getElementById( 'post-body-content' );
 			$postBodyContent = jQuery( postBodyContent );
-	
+
 			if ( jQuery( '.frm_field_loading' ).length ) {
 				loadFieldId = jQuery( '.frm_field_loading' ).first().attr( 'id' );
 				loadFields( loadFieldId );
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -332,7 +332,8 @@ function frmAdminBuildJS() {
 		thisFormId = 0,
 		autoId = 0,
 		optionMap = {},
-		lastNewActionIdReturned = 0;
+		lastNewActionIdReturned = 0,
+		$postBodyContent;
 
 	const { __ } = wp.i18n;
 	let debouncedSyncAfterDragAndDrop, postBodyContent;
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -969,6 +970,7 @@ function frmAdminBuildJS() {
 	}
 
 	function handleDrag( event, ui ) {
+		maybeScrollBuilder( event );
```
Crabcyborg, line 973: @AbdiTolesa I think putting this in every `handleDrag` call might be a bit of a performance concern. The drag function can get called a lot so calling scrollTop this often with a jQuery find every drag event seems heavy.

Do you think we can check something first to reduce the number of calls, and possibly get some elements ahead of time and re-use them rather than processing everything in every function call?

AbdiTolesa: @Crabcyborg I was concerned about that as well and I looked into this again but it looks to me there isn't much of the calculations that we can reduce. I just moved out the jQuery selector to an upper lexical scope so that it is called once and its result is reused on every drag event trigger.

Crabcyborg: Thanks @AbdiTolesa I tried profiling it a bit and it seems to work quite fast so we should be fine.

I tweaked it a bit more. In my performance it's now slightly faster.

I changed the first variable to a `_` to reflect that it isn't used at all. I also changed it to an array function, using `postBodyContent` instead of `this` so we can avoid `this`.

I was thinking we could try to maybe pull the function declaration out of the function, with a new `handleScrollTop` function possibly but getting that to work seems difficult with also needing to pass the `event` object. Trying to wrap a smaller function that called an external function didn't seem to have any impact on performance. According to https://stackoverflow.com/questions/80802/does-use-of-anonymous-functions-affect-performance it would likely be much faster if we could only create one function, instead of one every drag event. But performance seems fine as it is so I think we're okay.

js/formidable_admin.js
```diff
@@ -1001,6 +1005,27 @@ function frmAdminBuildJS() {
...
+
+				if ( relativePos > ( h - 50 ) && moved > 5 ) {
+					// Scrolling down.
+					return v + y * 0.1;
+				}
+
+				if ( relativePos < 50 && moved < -5 ) {
+					// Scrolling up.
+					return v - Math.abs( y * 0.1 );
+				}
```
Crabcyborg, line 1024: I couldn't really find any documentation about using a function as the parameter for `scrollTop`. The documentation seems to just reference passing a number.

What I don't love about this function still is that it sometimes returns void and other times it returns a number. I think we should at least add a comment at the end here to state why it's returning void if it need to. Maybe we could return something but I'm not really sure what. Do we return `v` maybe?

AbdiTolesa: @Crabcyborg Thank you for your time on this! I just tried returning `v` at the end and it seems to work rather than returning void.

js/formidable_admin.js
AbdiTolesa: Good to know that, thanks for sharing this!

Crabcyborg:
@AbdiTolesa Just one other thing before we merge this. It feels like I need to drag my field too high up for it to start scrolling up. Like over the header. Can you try to have it scroll when it's less high up?

Crabcyborg:
@AbdiTolesa Nevermind, one more thing. I think I caught a bug. It looks like if I drag a field group, it doesn't scroll, and it disappears off the page.

IamcLmPCaL

AbdiTolesa:
> @AbdiTolesa Nevermind, one more thing. I think I caught a bug. It looks like if I drag a field group, it doesn't scroll, and it disappears off the page.
> 
> IamcLmPCaL [ IamcLmPCaL ](https://user-images.githubusercontent.com/9134515/217832429-cd099e64-e5fe-4b1a-9a47-7413793977af.gif) [ ](https://user-images.githubusercontent.com/9134515/217832429-cd099e64-e5fe-4b1a-9a47-7413793977af.gif)

@Crabcyborg Both done! Thanks for catching that!

Crabcyborg (approved):
Thanks @AbdiTolesa! 🚀

PR #1050, opened 2023-02-03
remove default limit on the number of actions triggered

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3991

It looks that there is a default limit (99) defined for the number of actions triggered in `FrmFormAction::prepare_get_action` which prevents those extra actions to be triggered even if `frm_${action}_action_options` filter is used to extend the number of actions added in a given form.

I imagine two alternative solutions.

First, it looks we actually can use the action's `action_options['limit']` value instead to account for the new limit of actions as in this PR.

If we don't want to do that, I think we need to provide `$args['limit']` value everywhere `FrmFormAction::get_action_for_form` is called so that we override the default limit mentioned above.

### DIFF
--- a/classes/models/FrmFormAction.php
+++ b/classes/models/FrmFormAction.php
@@ -519,6 +519,7 @@ class FrmFormAction {
 			);
 		}
 		$defaults = array(
+			'limit'       => 99,
 			'post_status' => $default_status,
 		);
 		$args     = wp_parse_args( $args, $defaults );
@@ -548,8 +549,13 @@ class FrmFormAction {
 	}
 
 	public function get_all( $form_id = false, $atts = array() ) {
+		if ( is_array( $atts ) && ! isset( $atts['limit'] ) ) {
+			$atts['limit'] = $this->action_options['limit'];
+		}
+
 		self::prepare_get_action( $atts, 'any' );
-		$limit = $this->action_options['limit'];
+
+		$limit = $atts['limit'];
 
 		if ( $form_id ) {
 			$this->form_id = $form_id;
--- a/classes/models/FrmFormAction.php
+++ b/classes/models/FrmFormAction.php
@@ -519,7 +519,6 @@ class FrmFormAction {
 			);
 		}
 		$defaults = array(
-			'limit'       => 99,
 			'post_status' => $default_status,
 		);
 		$args     = wp_parse_args( $args, $defaults );
@@ -550,7 +549,7 @@ class FrmFormAction {
 
 	public function get_all( $form_id = false, $atts = array() ) {
 		self::prepare_get_action( $atts, 'any' );
-		$limit = $atts['limit'];
+		$limit = $this->action_options['limit'];
 
 		if ( $form_id ) {
 			$this->form_id = $form_id;


classes/models/FrmFormAction.php
```diff
@@ -519,7 +519,6 @@ protected static function prepare_get_action( &$args, $default_status = 'publish
 			);
 		}
 		$defaults = array(
-			'limit'       => 99,
```
Crabcyborg, line 522: I don't know if we should remove the default. I think this could cause too big of a change, and may result in query issues. I'm sure the default 99 limit exists for a reason.

This should be something we can override when passing parameters. In quiz outcomes, we call `$actions = FrmFormAction::get_action_for_form( $form_id, self::$action_name, $args );` in `FrmQuizzesFormActionHelper::get_quiz_action_from_form`. Passing the higher limit through `$args` there would possibly make more sense.

AbdiTolesa: @Crabcyborg I would go with the second option I proposed then. But this could probably be required for all actions calling `FrmFormAction::get_action_for_form` without the limit argument as I tried to mention above.

Crabcyborg: Yeah, having multiple limits does seem weird.

I just don't think we should drop support for this `$atts['limit']` either though. We have some code like `$frm_style->get_all( 'post_date', 'ASC', 1 );` where it intentionally only gets a single action.

I think we could consider using `$this->action_options` but only if `$atts['limit']` isn't also set. We'd want to use the option as a default, not the final value.

AbdiTolesa: Thanks @Crabcyborg! I added a check for `$atts['limit']` and set it with `$this->action_options` if not already.

Crabcyborg:
@stephywells I'd like your opinion on this solution. Does it make sense to change the default of `99` in the `action_control->get_all()` function to use the `limit` from the action options?

We're still using the limit set when calling `get_all` if one is specified, but the 99 default seems to be causing issues and it isn't possible to filter against (as calls to get_all are happening internally in our code). The default limit for actions is also usually 99 too.

Because these values are different, we're seeing issues with extending the limit beyond 99 in some cases (like in Quiz outcomes https://github.com/Strategy11/formidable-pro/issues/3991).

This isn't really code I'd touched before so I'd just like your confirmation that there isn't anything wrong with this. I think it's good to merge though.

stephywells:
> Does it make sense to change the default of `99` in the `action_control->get_all()` function to use the `limit` from the action options?

Yes! I think that would be no problem. The limit is used to disable adding a new form action, but we've only used a limit of 1 for this. I don't think there are any other limits we have set.

garretlaxton:
This didn't solve the issue for me. Once I get to 99, any additional outcomes added will remove outcomes starting at 99.

Crabcyborg:
@garretlaxton Is this even with the snippets on for the filters?

I think it requires both of these:

```
add_filter( 'frm_quiz_outcome_action_options', 'increase_quiz_outcome_limit' );
function increase_quiz_outcome_limit( $options ) {
    $options['limit'] = 200;
}

add_filter(
        'frm_form_action_limit',
	function( $limit ) {
		return 200;
	}
);
```

garretlaxton:
@Crabcyborg When I use that filter it seems to limit the outcomes to only one.
image

I was using the filter mentioned in the issue.
```
add_filter('frm_quiz_outcome_action_options', 'increase_quiz_outcome_limit'); function increase_quiz_outcome_limit( $options ) { $options['limit'] = 200; return $options; }
```

Crabcyborg (approved):
Thanks @AbdiTolesa!

PR #1051, opened 2023-02-06
Fix buttons moving on save

stephywells:
Prevent movement when the form builder update button is clicked. This uses 'aria-busy' instead of changing the button label.

This also adds the loading spinner when the styler page is saved.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -5882,8 +5882,8 @@ function frmAdminBuildJS() {
 
 	function submitBuild() {
 		/*jshint validthis:true */
-		var $thisEle = jQuery( this );
-		var p = $thisEle.html();
+		var $thisEle = this;
+		var p = this.innerHTML;
 
 		preFormSave( this );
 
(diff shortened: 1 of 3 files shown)


js/formidable_admin.js
```diff
@@ -5938,31 +5937,26 @@ function frmAdminBuildJS() {
...
+		b.setAttribute( 'aria-busy', 'true' );
 	}
 
-	function afterFormSave( $button, buttonVal ) {
-		$button.removeClass( 'frm_loading_form' ).removeClass( 'frm_loading_button' );
-		$button.html( frm_admin_js.saved );
+	function afterFormSave( button ) {
+		button.classList.remove( 'frm_loading_form' );
+		button.classList.remove( 'frm_loading_button' );
+		button.innterHTML = frm_admin_js.saved;
```
Crabcyborg, line 5951: `button.innterHTML` has an extra t. Also, is it save to just textContent here so it can't be used to inject a script? It looks like saved should just be a string.

stephywells: Oops! This is removed now. That shouldn't have been there anyway. The success message covers this, so we don't need the buttons to change. It looks weird with the button sizes changing.

Crabcyborg (approved):
Looks good! 🚀

PR #1053, opened 2023-02-08
Add more sanitizing to styler colors

Crabcyborg:
This came up in support on Monday.

Someone had a `(rgba(...` style with that extra leading `(` on it. This caused the rest of the styles to break.

We already have some sanitizing in place to make sure colours are formatted as expected. But the checks weren't removing any excess brackets so I've updated this to make sure that it is always consistent.

I tried cleaning it up a bit more in a few other ways too. I noticed the same issue happened with an extra brace after the `rgb` like `rgb((` so I added some trimming to remove that as well.

And I added some other sanitizing. The r/g/b values now get passed through `absint` and the alpha goes through `floatval` so if any non-numeric values are used here, they'll get removed.


stephywells (approved):
This looks great! Thanks for all the unit test cases!

garretlaxton (approved):
This works good. Any extra parentheses will now be removed.

AbdiTolesa (approved):
@Crabcyborg This looks good to me as well!

PR #1054, opened 2023-02-08
Fix close button on WP auth modal on styler pages

Crabcyborg:
I noticed this was my session expired.

In the styler branch I wasn't getting these styles because I removed the `wp-core-ui` class.
Screen Shot 2023-02-08 at 11 07 36 AM

This update just inserts the wp-core-ui class on the auth wrapper so the styles get applied properly.

**Before**
Screen Shot 2023-02-08 at 10 59 10 AM

**After**
Screen Shot 2023-02-08 at 10 56 27 AM

### DIFF
--- a/js/admin/style.js
+++ b/js/admin/style.js
@@ -28,6 +28,7 @@
 
 	initCommonEventListeners();
 	initPreview();
+	fixWpAuthModal();
 
 	/**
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Good catch!

PR #1055, opened 2023-02-08
Hide the WP admin menu and toolbar on full screen pages

stephywells:
This should make scrolling eaiser

Fixes https://github.com/Strategy11/formidable-pro/issues/3918
(I think)

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -450,6 +450,8 @@ class FrmAppController {
 		}
 
 		if ( FrmAppHelper::is_admin_page( 'formidable' ) ) {
+			self::disable_admin_menus();
+
 			$action = FrmAppHelper::get_param( 'frm_action' );
 
 			if ( in_array( $action, array( 'add_new', 'list_templates' ), true ) ) {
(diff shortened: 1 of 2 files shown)


classes/controllers/FrmAppController.php
```diff
@@ -461,6 +463,27 @@ public static function admin_init() {
...
+			return;
+		}
+
+		wp_deregister_script( 'admin-bar' );
+		wp_deregister_style( 'admin-bar' );
+		remove_action( 'admin_footer', 'wp_admin_bar_render', 1000 );
+		add_action(
+			'admin_head',
+			function() {
+				echo '<style>html.wp-toolbar{padding-top:0}</style>';
```
Crabcyborg, line 482: @stephywells I don't think this code is reached in the styler.

`disable_admin_menus` is only called `if ( FrmAppHelper::is_admin_page( 'formidable' ) ) {`. So this isn't applying everywhere full screen at the moment.

Screen Shot 2023-02-08 at 12 39 45 PM

Crabcyborg:
It also looks like this update causes visual views to load as a white screen. I don't see any errors though.

Screen Shot 2023-02-08 at 12 44 00 PM

classes/controllers/FrmAppController.php
stephywells: This is fixed now.

Crabcyborg (approved):
🚀

garretlaxton (approved):
This looks good for me.

PR #1056, opened 2023-02-08
Add new tag to application templates

Crabcyborg:
We talked about this in the meeting today. It's a pretty easy add. None of our templates are new enough to show it right now, but I used a period of 300 days to get a screenshot.

I went with 30 days. We use 10 for form templates. I don't know if we want to be more consistent. 10 seems pretty short. Since we don't add as many applications, a longer duration may make sense.

Screen Shot 2023-02-08 at 2 42 12 PM

### DIFF
--- a/classes/models/FrmApplicationTemplate.php
+++ b/classes/models/FrmApplicationTemplate.php
@@ -161,6 +161,8 @@ class FrmApplicationTemplate {
 			$application['link']       = $application['upgradeUrl'];
 		}
 
+		$application['isNew'] = $this->is_new();
+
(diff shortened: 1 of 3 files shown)


stephywells (approved):
Nice! That was fast. 30 days sounds perfect.

10 days does seem like too short for the form templates. 30 days would probably be best for those too.

Crabcyborg:
I just updated this to make form templates 30 days too. It may make sense to re-use this in more places. I'll probably do that when I try to add this tag for style templates too.

PR #1057, opened 2023-02-08
Stop hard coding a 1 for forms restored

Crabcyborg:
I just noticed if you delete some forms, and then "Undo" that it would always show a 1. I had 13 forms restore, but the count said 1.

Screen Shot 2023-02-08 at 3 12 51 PM

It's just hard coded. I'm using `$count` now and it works as expected.

Screen Shot 2023-02-08 at 3 21 14 PM

I also added a check for 0 so we don't bother showing "0 forms restored" on refresh.


stephywells (approved):
Haha! Good one!

PR #1059, opened 2023-02-08
Add is new to api data, add "NEW" to style templates

Crabcyborg:
I'm just hooking into the shared API code now. It will set a new `is_new` key after it gets a response. If there's a `released` date, it will set the boolean. This way it's consistent and easy to use for other new things.

Screen Shot 2023-02-08 at 4 40 08 PM
Screen Shot 2023-02-08 at 4 39 41 PM
Screen Shot 2023-02-08 at 4 38 07 PM


classes/models/FrmFormApi.php
```diff
@@ -91,12 +91,20 @@ public function get_api_info() {
...
+
+			if ( isset( $addon['categories'] ) ) {
+				$cats = array_intersect( $this->skip_categories(), $addon['categories'] );
+				if ( ! empty( $cats ) ) {
+					unset( $addons[ $k ] );
+					continue;
+				}
+			}
+
+			if ( array_key_exists( 'released', $addon ) ) {
```
stephywells, line 106: This is a great improvement! It would be cool if this were a separate is_new() function to make it a bit easier to override if we want to.

Crabcyborg: Oh, great idea! I just added a new protected is_new function that handles the date comparison.

stephywells: Nice! I thought it would be a quick add to update this for add-ons too. But then I had to override the number of days for each class since add-ons use the base. Would you mind taking a look over my changes?

stephywells (approved):

classes/models/FrmFormApi.php
Crabcyborg: @stephywells Those updates look good! I'll merge this into the styler branch now! 🚀

PR #1065, opened 2023-02-14
Fix PHP notice with success_action

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/4068

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -2339,7 +2339,7 @@ class FrmFormsController {
 	}
 
 	/**
-	 * Gets run_success_action() args from the On Sumbit action.
+	 * Gets run_success_action() args from the On Submit action.
 	 *
 	 * @since 6.0
 	 *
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Thank you @truongwp !

classes/controllers/FrmFormsController.php
```diff
@@ -2929,10 +2929,10 @@ private static function maybe_migrate_submit_settings_to_action( $form_id ) {
 	private static function get_on_submit_action_data_from_form_options( $form_options, $event = 'create' ) {
 		$opt  = 'update' === $event ? 'edit_' : 'success_';
 		$data = array(
-			'success_action' => $form_options[ $opt . 'action' ],
+			'success_action' => isset( $form_options[ $opt . 'action' ] ) ? $form_options[ $opt . 'action' ] : FrmOnSubmitHelper::get_default_action_type(),
 		);
 
-		switch ( $form_options[ $opt . 'action' ] ) {
+		switch ( $data['success_action'] ) {
```
stephywells, line 2935: Actually.. This line isn't quite right. This lost the $opt

stephywells (changes requested):

classes/controllers/FrmFormsController.php
truongwp: I use `$data`, not `$form_options`.

stephywells: What happens if the event is 'edit' instead of 'success'?

truongwp: In the On Submit action, we just use `success_action`, not matter it's an `update` or `create` event.

truongwp: We just need to check in the `$form_options`: https://github.com/Strategy11/formidable-forms/pull/1065/files#diff-1b880a2d11e84622570383ea0503da990c04c1ccbffe4f5f81ab21d68fa00280R2932

stephywells: Oh! I see it now. It's all good.

stephywells (approved):

PR #1069, opened 2023-02-15
Avoid changing the label position container classes if a field does not use the default

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4072

The old code for this was fine because it was for the sample form where this setting doesn't exist. I hadn't considered this field setting.

This update adds a class to any field with the default set, and only those get changed now.

### DIFF
--- a/classes/helpers/FrmStylesPreviewHelper.php
+++ b/classes/helpers/FrmStylesPreviewHelper.php
@@ -21,11 +21,23 @@ class FrmStylesPreviewHelper {
 	private $form_includes_captcha = false;
 
 	/**
+	 * Track the field ids with the default label position.
+	 * A field with a default label position will have a frm-default-label-position class in the styler preview.
+	 * Only fields with this class will change when you update the position setting in the styler.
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This looks good!
image

Should this apply to the form builder though?
image

Crabcyborg:
@garretlaxton That's an interesting question. I think that might just make the field more difficult to update without a label.

That page isn't as true of a preview so it's probably fine leaving it as is for now.

stephywells (commented):
@Crabcyborg Nice! I didn't test this to make sure... Does this cover the preview form when not editing the style too?

stephywells (approved):

Crabcyborg:
> Does this cover the preview form when not editing the style too?

Yes, it does. The HTML for the styler preview is identical for both the edit/list views, and the `setPosClass` function here is used for both as well. It just fakes the position input on the list page based on the label position setting of the selected style card.

PR #1071, opened 2023-02-16
Fix PHP 8.2 notices

truongwp:
Related issue: https://github.com/Strategy11/formidable-pro/issues/4042

### DIFF
--- a/classes/views/frm-forms/settings-advanced.php
+++ b/classes/views/frm-forms/settings-advanced.php
@@ -87,7 +87,7 @@ if ( time() < strtotime( '2023-03-07' ) ) {
 </p>
 
 <?php
-is_callable( 'self::render_spam_settings' ) && self::render_spam_settings( $values );
+is_callable( 'FrmFormsController::render_spam_settings' ) && FrmFormsController::render_spam_settings( $values );
 FrmTipsHelper::pro_tip( 'get_form_settings_tip', 'p' );
 ?>
 
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1323,7 +1323,7 @@ class FrmFormsController {
 	/**
 	 * @param array $values
 	 */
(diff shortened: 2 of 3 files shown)


Crabcyborg (approved):
Thanks @truongwp!

If a deprecation message happens for using `FrmFormsController` with `self` inside of a view, I think Psalm can catch those automatically. We just have that rule turned off right now.

https://github.com/Strategy11/formidable-forms/blob/5cead1d842e4fdee2506308c599521e47a034e0a/psalm.xml#L61

It may be the `InaccessibleMethod` error? Or maybe `InvalidScope`?

classes/views/frm-forms/settings-advanced.php
```diff
@@ -87,7 +87,7 @@
 </p>
 
 <?php
-is_callable( 'self::render_spam_settings' ) && self::render_spam_settings( $values );
+is_callable( 'FrmFormsController::render_spam_settings' ) && FrmFormsController::render_spam_settings( $values );
```
Crabcyborg, line 90: I approved this a bit too quickly.

This function is private. Now PHPStan doesn't like it:

```
Error: Call to private static method render_spam_settings() of class FrmFormsController.
 ------ --------------------------------------------------------------- 
  Line   classes/views/frm-forms/settings-advanced.php                  
 ------ --------------------------------------------------------------- 
  90     Call to private static method render_spam_settings() of class  
         FrmFormsController.                                            
 ------ --------------------------------------------------------------- 
```

Maybe we should just make it public.

truongwp: Oh I missed that. Thank you.

truongwp:
@Crabcyborg I tried enabling both of those rules in https://github.com/Strategy11/formidable-forms/pull/1076. It's related to the fix in this PR. This PHP 8.2 notice is for `self` in a callable.

PR #1072, opened 2023-02-16
Remove extra comma in mb adv info view

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2157723026/151310/

There was an extra comma here. This only breaks in versions of PHP 7.2 and lower.

I spent some time trying to add a rule for this with phpcs, psalm, or phpstan, but it doesn't really appear to be easy to check for with any of these. Maybe with the PHPCompatibilityWP in phpcs but I don't think it's worth pulling that in just for a comma check.

But in that process, I removed some Psalm exceptions from view files. It was previously ignoring the files entirely. Now it's just ignoring specific rules.

### DIFF
--- a/classes/views/shared/mb_adv_info.php
+++ b/classes/views/shared/mb_adv_info.php
@@ -101,7 +101,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 					'value'       => 'id',
 					'show_labels' => true,
 					'echo'        => true,
-				),
+				)
 			);
 			?>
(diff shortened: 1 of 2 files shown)


NathanaelJonesIreland:
Reported again here: https://secure.helpscout.net/conversation/2160526133/151763/

Crabcyborg:
Also reported in https://secure.helpscout.net/conversation/2161397521/151940/

PR #1073, opened 2023-02-16
Update lite so Pro can adjust entry search queries to include repeaters

AbdiTolesa:
Related issue https://github.com/Strategy11/formidable-pro/issues/3843
Required for https://github.com/Strategy11/formidable-pro/pull/4061

This update doesn't add a new functionality or fix but refactors the existing functions so that the pro update calls those new functions with the new update that is aimed at fixing the issue. Just to make existing code reusable in pro.


omidakhavans:
@AbdiTolesa Yes and I have made the reviews on [Pro](https://github.com/Strategy11/formidable-pro/pull/4061)

truongwp (approved):
This looks good! 

Please add `@since` tag for these new methods since they can be extended in child classes.

Thanks @AbdiTolesa!

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -78,17 +98,54 @@ public function prepare_items() {
...
+		}
+
+		return apply_filters( 'frm_entries_list_query', $s_query, compact( 'form_id' ) );
+	}
+
+	/**
+	 * @since x.x
+	 *
+	 * @param int|string $form_id
+	 * @return int|string
```
Crabcyborg, line 145: Return types should not change between parent and child classes.

In Pro, this is an array.

The name `get_form_ids` implies an array.

I think we should just return an array all the time, but still avoid passing the array to `$s_query` so it doesn't change the query for a single form.

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -142,10 +143,10 @@ protected function get_search_query( $form_id, &$join_form_in_query ) {
 	 * @since x.x
 	 *
 	 * @param int|string $form_id
-	 * @return int|string
+	 * @return array
 	 */
 	protected function get_form_ids( $form_id ) {
-		return $form_id;
+		return array( $form_id );
```
Crabcyborg, line 149: @AbdiTolesa One last thing.

It would be nice if we guaranteed the array was all integers.

If we could return `array( (int) $form_id )`, we can do strict checks reliably.

Crabcyborg: I think Psalm and PHPStan can detect `int[]` and `array<int>` type definition formats.

AbdiTolesa:
@Crabcyborg PHPCS is failing but the lines of error doesn't even exist in my branch and I really have no clue how to fix it.

Crabcyborg:
Thanks @AbdiTolesa. That's my fault.

I made a small PR to fix that https://github.com/Strategy11/formidable-forms/pull/1360

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -19,37 +19,68 @@ class FrmEntriesListHelper extends FrmListHelper {
...
+		$this->total_items = $total_items;
+		$this->prepare_pagination();
+	}
+
+	/**
+	 * Prepares pagination.
+	 *
+	 * @since x.x
+	 */
+	public function prepare_pagination() {
```
Crabcyborg, line 40: I think we can keep this as `protected` rather than `public`.

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -19,37 +19,68 @@ class FrmEntriesListHelper extends FrmListHelper {
...
+	 * @since x.x
+	 *
+	 * @param int|string $form_id
+	 * @param int        $per_page
+	 * @param array      $s_query
+	 * @param bool       $join_form_in_query
+	 *
+	 * @return array
+	 */
+	protected function get_entry_items( $form_id, $per_page, &$s_query, &$join_form_in_query ) {
```
Crabcyborg, line 71: @AbdiTolesa Instead of passing `$per_page` as a param, could we just check the global or add a new `per_page` property to this class and access `$this->per_page` directly?

Crabcyborg: Instead of `$form_id` we could just access `$this->params['form']`, right?

AbdiTolesa: Sorry I overlooked that part! I have updated to make sure we don't pass extra function args than needed.

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -19,37 +19,67 @@ class FrmEntriesListHelper extends FrmListHelper {
 	public function prepare_items() {
 		global $per_page;
```
Crabcyborg, line 20: `$per_page` isn't used here anymore.

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -19,37 +19,67 @@ class FrmEntriesListHelper extends FrmListHelper {
...
-			);
-			$join_form_in_query = true;
-		}
+		$join_form_in_query = false;
 
-		$s = self::get_param(
+		$this->items = $this->get_entry_items( $s_query, $join_form_in_query );
+		$total_items = FrmEntry::getRecordCount( $s_query );
+
+		$this->total_items = $total_items;
```
Crabcyborg, line 31: I think we could save some more code in Pro if we added a `$this->set_total_items( $s_query );` function here.

Crabcyborg (approved):

I think we're good to go with this update!

🚀

PR #1075, opened 2023-02-17
Fix overflowing rte box on smaller spaces

stephywells:
Before:
Screen Shot 2023-02-17 at 4 33 25 PM
Screen Shot 2023-02-17 at 4 33 13 PM

After:
Screen Shot 2023-02-17 at 4 32 36 PM
Screen Shot 2023-02-17 at 4 32 54 PM

The message still overlaps, but that text is in Pro.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -4056,6 +4056,8 @@ label input[type="checkbox"], label input[type="radio"] {
 
 .frm_rte {
 	background: url(../images/rte.png) no-repeat;
+	background-size: clamp( 400px, 100%, 654px) auto;
+
 }
 
 .frm_rte .howto {
@@ -4111,9 +4113,10 @@ label input[type="checkbox"], label input[type="radio"] {
 
 .frm_sorting .frm_rte textarea {
 	width: 653px;
+	max-width: 100%;
 	background: #fff;
 	margin: 1px 0 0;
-	border: 1px solid #dfdfdf;
+	border: 1px solid var(--grey-300);
 	border-top: none;
 	-moz-border-radius: 0;
 	-webkit-border-radius: 0;


garretlaxton:
This looks good, but if I switch to a mobile view it isn't compatible. Not sure if it should be or not but figured I'd mention it.
image

stephywells:
Thank for checking! We just finished up some designs to show a message on small screens, and adjust the menu bar to make it work better on more screen sizes. But for this pr, we don't need to cover it since it'll be a bigger job.

garretlaxton (approved):
Gotcha, thanks!

garretlaxton:
I took a look over this again and if you have 2 rte fields next to each other it gets cut off still.
image

Crabcyborg:
@stephywells What do you think about trying to just use TinyMCE instead of an image?

It's more responsive, and then we can drop the message entirely.
Screen Shot 2023-02-21 at 11 34 04 AM

We're already loading the entire form now in the styler preview. I just tried adding 35 rich text fields and they perform fine, so I don't think there's a very big performance concern in having too many rich text fields on the page.

stephywells:
@Crabcyborg Is that an option without message up the builder? I suppose this was originally set as an image because of the problem with multiple RTE on the page. That looks like a much better solution!

I'm trying to make my way back out of the FF dev. Is there someone who could take over this one?

Crabcyborg:
> I'm trying to make my way back out of the FF dev. Is there someone who could take over this one?

Yeah, no problem! Since Abdi is likely out today I can just take it on. It should pretty straight-forward at this point as we've gotten TinyMCE working just about everywhere else already.

Crabcyborg:
I'll close this as this is being replaced with https://github.com/Strategy11/formidable-pro/pull/4085 instead.

Maybe we want to still merge it though? This styling would get used if you don't update Pro right away.

Once the Pro release has been out for long enough we could probably remove this RTE image from lite as well. For now it's probably better to keep it for backward compatibility though.

stephywells:
That sounds like a good plan!

PR #1086, opened 2023-02-23
add right margin to fields box

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4092
Just a small fix to make sure the email message text area is easily expandable.
BEFORE:
image

NOW:
image


garretlaxton (approved):
Works great! Tested in multiple browsers and sizes.

stephywells:
I updated this so we don't mess with the placement I spend time fixing in 6.0.

PR #1089, opened 2023-02-24
Do not set message in else condition fix

Crabcyborg:


truongwp:
The purpose of that else is if no actions match, it uses the default message. I'm adding a check to not set `$method` to `message` if the form isn't migrated.

truongwp:
I created #1091 to fix this but keep fallback to the default message.

PR #1091, opened 2023-02-25
Fix confirmation action is forced to message

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/4101

My last fix added `$method = 'message'` to show the default message if no actions match. But it breaks the form that hasn't been migrated. I added a check if the form hasn't been migrated there.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -140,7 +140,7 @@ class FrmFormsController {
 			$id = FrmAppHelper::get_param( 'id', '', 'get', 'absint' );
 		}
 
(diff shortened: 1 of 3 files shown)


truongwp:
I'm checking why phpunit tests failed.

stephywells (approved):
Looks good to me! I move the code to get the fallback so we don't have to add it in other places too.

Actually... Looks like the unit tests are failing. Looking into it.

truongwp:
@stephywells We have a different fallback edit message. Can you update it? I’m away from desk now.

stephywells:
Thanks for checking this! I think we need a hook or something to cover that. For now, I think it's fine if the fallback uses the same message.

PR #1092, opened 2023-02-25
Move frm_show_form_after_edit filter to lite

truongwp:
I moved that filter to lite since it's get run after this change https://github.com/Strategy11/formidable-pro/pull/4102/files#diff-0b783bc2474a31b50b78b8e63872466196b161f94bd6c7a01f4cd72205e8f060R1132

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -2499,11 +2499,17 @@ class FrmFormsController {
 	 * Prepare to show the success message and empty form after submit
 	 *
 	 * @since 2.05
+	 * @since 6.1.1 Moved `frm_show_form_after_edit` filter to this method.
 	 */
 	public static function show_message_after_save( $atts ) {
 		$atts['message'] = self::prepare_submit_message( $atts['form'], $atts['entry_id'], $atts );
+		$show_form       = ( isset( $atts['form']->options['show_form'] ) ) ? $atts['form']->options['show_form'] : true;
 
-		if ( ! isset( $atts['form']->options['show_form'] ) || $atts['form']->options['show_form'] ) {
+		if ( 'edit' === $atts['success_opt'] && version_compare( FrmProDb::$plug_version, '6.1.1', '>=' ) ) {
+			$show_form = apply_filters( 'frm_show_form_after_edit', $show_form, $atts['form'] );
+		}
+
+		if ( $show_form ) {
 			self::show_form_after_submit( $atts );
 		} else {
 			self::show_lone_success_messsage( $atts );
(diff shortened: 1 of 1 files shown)


classes/controllers/FrmFormsController.php
```diff
@@ -2499,11 +2499,17 @@ private static function get_redirect_message( $success_url, $success_msg, $args
...
 	 *
 	 * @since 2.05
+	 * @since 6.1.1 Moved `frm_show_form_after_edit` filter to this method.
 	 */
 	public static function show_message_after_save( $atts ) {
 		$atts['message'] = self::prepare_submit_message( $atts['form'], $atts['entry_id'], $atts );
+		$show_form       = ! isset( $atts['form']->options['show_form'] ) || $atts['form']->options['show_form'];
 
-		if ( ! isset( $atts['form']->options['show_form'] ) || $atts['form']->options['show_form'] ) {
+		if ( 'edit' === $atts['success_opt'] && version_compare( FrmProDb::$plug_version, '6.1.1', '>=' ) ) {
```
Crabcyborg, line 2508: @truongwp We would probably want to add a check here that `FrmProDb::$plug_version` is accessible, right? Or is it not possible to have the `edit` condition when Pro isn't active?

truongwp: I added the check for `FrmProDb` to make it safer. Currently, `edit` check is enough, but we don't know what will happen in the future.

classes/controllers/FrmFormsController.php
```diff
@@ -2499,11 +2499,17 @@ private static function get_redirect_message( $success_url, $success_msg, $args
 	 * Prepare to show the success message and empty form after submit
 	 *
 	 * @since 2.05
+	 * @since 6.1.1 Moved `frm_show_form_after_edit` filter to this method.
 	 */
 	public static function show_message_after_save( $atts ) {
 		$atts['message'] = self::prepare_submit_message( $atts['form'], $atts['entry_id'], $atts );
+		$show_form       = ! isset( $atts['form']->options['show_form'] ) || $atts['form']->options['show_form'];
```
stephywells, line 2506: It looks like we should add a new filter here instead of moving code, so we can keep as much edit-specific code in pro.

stephywells (changes requested):

classes/controllers/FrmFormsController.php
truongwp: But with the new code in pro. That filter won’t run. Do we make it deprecated?

stephywells: I'm not sure I understand. Can we add a hook here that is more generic that runs when not editing too? I'm thinking we could pass $attr as the second parameter and check if editing in pro when we use the hook. 

Is there a reason it wouldn't work that way?

truongwp: I meant this filter is deprecated since we used the code in Lite and that method wasn't reached. So I moved it to Lite plugin. https://github.com/Strategy11/formidable-pro/blob/master/classes/controllers/FrmProEntriesController.php#L1078

If not, we should mark it as deprecated in docs.

truongwp:
I close this since #1106 solve it.

PR #1095, opened 2023-03-01
Enable duplication of collapsed sections

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4087

This PR is aimed at making collapsed sections duplicatable, fix style issues in the context dropdown and also show the number of fields to be deleted in the confirmation box that is shown when trying to delete a section.
BEFORE:
image

NOW:
image

[issue_4087.webm](https://user-images.githubusercontent.com/41271840/222132436-603ba44f-4a68-4856-af69-dc4cdf2e77fc.webm)

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -3531,7 +3531,7 @@ function frmAdminBuildJS() {
 				}
 			});
 			if ( fieldIdsToDelete ) {
-				confirmMsg = confirmFieldsDeleteMessage( fieldIdsToDelete );
+				confirmMsg = confirmFieldsDeleteMessage( ++fieldIdsToDelete );
 			}
 		}
 
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -2044,7 +2044,7 @@ function frmAdminBuildJS() {
 
 		$field = jQuery( this ).closest( 'li.form-field' );
 
-		if ( $field.hasClass( 'frm-section-collapsed' ) || $field.hasClass( 'frm-page-collapsed' ) ) {
+		if ( $field.hasClass( 'frm-page-collapsed' ) ) {
 			return false;
 		}
 
@@ -3520,7 +3520,7 @@ function frmAdminBuildJS() {
 			fieldIdsToDelete = 0;
 
 		if ( field.data( 'ftype' ) === 'divider' ) {
-			fieldBoxes = document.querySelector( '.frm-field-group-hover-target' )?.querySelector( '.start_divider')?.querySelectorAll( '.frm_field_box' );
+			fieldBoxes = document.querySelector( '.frm-field-group-hover-target' )?.querySelector( '.start_divider' )?.querySelectorAll( '.frm_field_box' );
 			if ( fieldBoxes && fieldBoxes.length ) {
 				fieldBoxes.forEach( fieldBox => {
 					fieldsInsideFieldBox = fieldBox.querySelectorAll( 'li.form-field' );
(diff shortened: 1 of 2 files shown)


garretlaxton:
This looks good, but could we make the dropdown close after we duplicate? When I hover the duplicated section the dropdown from the above section gets in the way.
image

AbdiTolesa:
> This looks good, but could we make the dropdown close after we duplicate? When I hover the duplicated section the dropdown from the above section gets in the way. image

@garretlaxton Thanks for catching that! I just fixed that, you can give it another shot.

garretlaxton (approved):
Functions nice!

js/formidable_admin.js
```diff
@@ -3516,6 +3521,22 @@ function frmAdminBuildJS() {
 			field = jQuery( this ).closest( 'li.form-field' ),
 			fieldId = field.data( 'fid' );
 
+		if ( field.data( 'ftype' ) === 'divider' ) {
+			const fieldBoxes = document.querySelector( '.frm-field-group-hover-target' )?.querySelector( '.start_divider' )?.querySelectorAll( '.frm_field_box' );
```
Crabcyborg, line 3525: @AbdiTolesa Let's try to avoid the `?` operator.

You can really just do some `querySelectorAll` here still like `document.querySelectorAll( '.frm-field-group-hover-target .start_divider .frm_field_box' );`. It's shorter and doesn't require all of these nullable checks that way.

js/formidable_admin.js
```diff
@@ -3516,6 +3521,22 @@ function frmAdminBuildJS() {
...
+			if ( fieldBoxes && fieldBoxes.length ) {
```
Crabcyborg, line 3526: I think this is one of the weird things about using the `?`. When you just use `querySelectorAll`, you always end up with a NodeList, but when you use `?` sometimes you get a NodeList and sometimes you get undefined.

If we don't use the `?` we don't really even have to check if there are fieldBoxes, but just iterate over them all right away instead.

AbdiTolesa: Thank you for the insight, definitely not a good spot to use the optional chaining operator! Just removed that now.

Crabcyborg:
@AbdiTolesa I think we should count the section field in the count as well. Technically I'm deleting the section, the radio buttons, and the text fields. I expect the count to be 3, but I get 2.

Screen Shot 2023-03-07 at 3 58 48 PM

Crabcyborg (changes requested):
I left this comment on the issue by mistake https://github.com/Strategy11/formidable-pro/issues/4087#issuecomment-1466254146

Crabcyborg:
I'm also seeing this layering issue when you have a collapsed section under another collapsed section and open the dropdown.

Screen Shot 2023-03-13 at 11 35 06 AM

But it doesn't seem too easy to fix. But maybe we can use a class that happens for the active hover target to make sure the section we're opening a dropdown for has a higher z-index. The hard coded 3 we have here isn't great as another collapsed section would just have the same z-index and still be in the way.

AbdiTolesa:
> I'm also seeing this layering issue when you have a collapsed section under another collapsed section and open the dropdown.
> 
> Screen Shot 2023-03-13 at 11 35 06 AM
> 
> But it doesn't seem too easy to fix. But maybe we can use a class that happens for the active hover target to make sure the section we're opening a dropdown for has a higher z-index. The hard coded 3 we have here isn't great as another collapsed section would just have the same z-index and still be in the way.
> 
> Or maybe we could just listen for the `show.bs.dropdown` event.

Thanks, that makes sense indeed. Updated accordingly.

Crabcyborg (approved):

PR #1101, opened 2023-03-06
move functions to global scope

AbdiTolesa:


js/formidable_admin.js
```diff
@@ -423,19 +423,6 @@ function frmAdminBuildJS() {
 		return false;
 	}
 
-	function infoModal( msg ) {
```
Crabcyborg, line 426: It shouldn't be necessary to move this.

Check out this code in the ACF add on where Truong is calling it. This function can be called through the `frmAdminBuild` object.
https://github.com/Strategy11/formidable-acf/blob/77a85556dbe3edfba8cf43c5f1aa7e467686d0c4/js/post-action.js

AbdiTolesa: Thanks for letting me know @Crabcyborg. I think we can close this PR then.

PR #1102, opened 2023-03-06
show all table content on print preview

AbdiTolesa:


css/frm_admin.css
```diff
@@ -8769,6 +8769,11 @@ Responsive Design
 		display: block;
 		width: 100%;
 	}
+
+	#form_reports_page .google-visualization-table > div {
```
Crabcyborg, line 8773: @AbdiTolesa Maybe we should just include this update as part of the date filter update?

That way we can include this CSS in Pro instead of having to add more reports CSS into Lite.

AbdiTolesa: Sure, done!

PR #1104, opened 2023-03-07
Only show add new entry when pro is installed

Crabcyborg:
I discussed this the other day in Slack.

We're showing the import/add new buttons on the Entries page when Pro is inactive but the add new button does nothing as it links to a page that we only support in Pro.

**Before (or when Pro is installed)**
Screen Shot 2023-03-07 at 2 47 23 PM

**After (when Pro is not installed)**
Screen Shot 2023-03-07 at 2 47 14 PM


garretlaxton (approved):
Looks good! The 2 buttons now only show when Pro is activated.

PR #1108, opened 2023-03-09
overwrite .columns-2 styles for view page

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4098

BEFORE:
image

NOW:
image

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -723,14 +723,18 @@ ul.frm_form_nav > li {
 	overflow: hidden;
 }
 
+.frm-white-body .columns-2 > div {
+	overflow-y: auto;
+	box-sizing: border-box;
+}
+
 .frm-white-body.post-type-frm_display .columns-2 {
 	display: block;
 	overflow: visible;
 }
 
-.frm-white-body .columns-2 > div {
-	overflow-y: auto;
-	box-sizing: border-box;
+.frm-white-body.post-type-frm_display .columns-2 > div {
+	overflow-y: hidden;
 }
 
 .frm_wrap .columns-2 .frm-right-panel,
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -737,6 +737,10 @@ ul.frm_form_nav > li {
 	overflow-y: hidden;
 }
 
+.frm-white-body.post-type-frm_display #titlediv #title-prompt-text {
+	padding: 3px 0 0 10px;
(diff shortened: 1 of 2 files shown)


css/frm_admin.css
```diff
@@ -728,6 +728,19 @@ ul.frm_form_nav > li {
 	box-sizing: border-box;
 }
 
+.frm-white-body.post-type-frm_display .columns-2 {
```
Crabcyborg, line 731: @AbdiTolesa What do you think about trying to put this behind a feature check so we can avoid making `frm_admin.css` bigger (it's already too big).

I think it may make sense to use `wp_add_inline_style` https://developer.wordpress.org/reference/functions/wp_add_inline_style/) if we detect that someone is on the legacy views editor.

And if you can find any other CSS in this file related to the legacy views editor (I think there's probably quite a bit), that isn't still being used in Visual Views, we could try moving it there as well.

If it's a lot of classes, we may just want to add another CSS file and enqueue it.

garretlaxton:
Thanks Mike! This looks pretty good, but a few things I see.

The fields IDs and Keys buttons are a bit too far to the right.
image

The search bar on the Advanced tab is all the way to the left, it should be centered I believe.
image

AbdiTolesa:
> Thanks Mike! This looks pretty good, but a few things I see.
> 
> The fields IDs and Keys buttons are a bit too far to the right. image
> 
> The search bar on the Advanced tab is all the way to the left, it should be centered I believe. image

I was comparing how this panel looks in other pages like email action content and have made the search box fit into the width.
image

But it looks the ID and keys are aligned to the right most there as well. Not sure if we need to introduce extra right padding here, what do you think about that @Crabcyborg?

css/frm_admin.css
AbdiTolesa: I have changed this to enqueue those new styles conditionally but couldn't really figure out how to get obsolete styles that were used in the old version but not current.

Crabcyborg:
@AbdiTolesa Since it's the legacy views I don't think we need to focus too much on getting it to look too nice. I think this is enough.

I added a small rule so the`#advanced-search-input` also has the flex styling.

And I've moved all of this legacy views assets logic into a new function to keep the functions smaller and the legacy views logic more contained.

Thanks for working on this @AbdiTolesa and thanks for testing @garretlaxton!

PR #1112, opened 2023-03-14
Do not run invalid confirmation actions

truongwp:


classes/controllers/FrmFormsController.php
```diff
@@ -2297,6 +2301,28 @@ public static function get_met_on_submit_actions( $args, $event = 'create' ) {
...
+	 * @return bool
+	 */
+	private static function is_valid_on_submit_action( $action ) {
+		$action_type = FrmOnSubmitHelper::get_action_type( $action );
+
+		if ( 'redirect' === $action_type && empty( $action->post_content['success_url'] ) ) {
+			return false;
+		}
+
+		if ( 'page' === $action_type && ( empty( $action->post_content['success_page_id'] ) || ! get_post( $action->post_content['success_page_id'] ) ) ) {
```
Crabcyborg, line 2319: Thanks @truongwp!

What do you think about adding a get check the posts status like I had in my update? I don't think we should try pulling something from trash.

truongwp: Thanks for pointing that out. I added that check.

Crabcyborg (approved):
Thanks @truongwp!

PR #1117, opened 2023-03-15
Code/remove-deprecated-functions

stephywells:
There is a bit of code that has been deprecated, but didn't have the deprecated messages. Anything left with a 2.x deprecated version, didn't have helpful messages so we can't remove them yet.


Crabcyborg (approved):
Thanks @stephywells! Exciting to have so much cleaned out out here! 🥳

PR #1123, opened 2023-03-23
Fix gravity importer issue 18 and style

AbdiTolesa:


classes/controllers/FrmAddonsController.php
```diff
@@ -971,6 +971,27 @@ protected static function maybe_show_cred_form() {
...
+	/**
+	 * Checks if an addon download url is whitelisted.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $url
+	 *
+	 * @return bool
+	 */
+	private static function is_whitelisted( $url ) {
```
Crabcyborg, line 983: @AbdiTolesa The term "whitelisted" is controversial so we should avoid it and use a term like "allowed_list" or something else instead https://core.trac.wordpress.org/ticket/50413.

classes/controllers/FrmAddonsController.php
```diff
@@ -971,6 +971,27 @@ protected static function maybe_show_cred_form() {
...
+		$whitelisted_urls = array(
+			'https://downloads.wordpress.org/plugin/formidable-gravity-forms-importer.zip',
+			'https://downloads.wordpress.org/plugin/formidable-import-pirate-forms.zip',
+		);
+
+		if ( in_array( $url, $whitelisted_urls, true ) ) {
```
Crabcyborg, line 989: This return true/return false is pretty verbose. We could just do a `return in_array( $url, $whitelisted_urls, true );`

classes/controllers/FrmAddonsController.php
```diff
@@ -982,7 +1003,7 @@ protected static function install_addon() {
 
 		$download_url = self::get_current_plugin();
 
-		if ( ! FrmAppHelper::validate_url_is_in_s3_bucket( $download_url, 'zip' ) ) {
+		if ( ! ( self::is_whitelisted( $download_url ) || FrmAppHelper::validate_url_is_in_s3_bucket( $download_url, 'zip' ) ) ) {
```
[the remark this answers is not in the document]
AbdiTolesa: @Crabcyborg All comments addressed! Thank you for your suggestions!

garretlaxton:
Could you upload this to the QA site since I don't have access to gravity forms?

stephywells:
@AbdiTolesa you can add the 'beta deploy' tag to this pr to push automatically to QA. 

Also, can we also test to make sure installing WP mail smtp from our SMTP page works? I think it would be fine to allow any wordpress.org plugin.

AbdiTolesa:
> > 
> 
> I tried it and it seems a bit heavy to upload install it from my zip file. I will try again later.

@garretlaxton The update is deployed to QA now, you can activate the latest lite on your site and try it. https://qa.formidableforms.com/abdi/wp-admin/admin.php?page=formidable-import

AbdiTolesa:
> @AbdiTolesa you can add the 'beta deploy' tag to this pr to push automatically to QA.
> 
> Also, can we also test to make sure installing WP mail smtp from our SMTP page works? I think it would be fine to allow any wordpress.org plugin.

Thanks for the deployment tip @stephywells! Although I couldn't figure out how to test WP Mail SMTP installation, I assume that updating the `$download_url` allow function I added in this update to allow all urls like `https://downloads.wordpress.org/plugin` would suffice your suggestion. I'm I right?

classes/controllers/FrmAddonsController.php
```diff
@@ -971,6 +971,19 @@ protected static function maybe_show_cred_form() {
...
+	 * Checks if an addon download url is allowed.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $url
+	 *
+	 * @return bool
+	 */
+	private static function is_allowed( $url ) {
+		return strpos( $url, 'https://downloads.wordpress.org/plugin' ) !== false;
```
Crabcyborg, line 984: @AbdiTolesa Checking for `0` was pretty intentional. We don't want someone faking a URL here with something like `https://unsafeurl.com?empty_my_whole_wallet&something_to_bypass_this_check=https://downloads.wordpress.org/plugin`.

stephywells:
> Although I couldn't figure out how to test WP Mail SMTP installation, I assume that updating the `$download_url` allow function I added in this update to allow all urls like `https://downloads.wordpress.org/plugin` would suffice your suggestion. I'm I right?

I would think so, but we should definitely test to be sure. This can be tested on a site without WP Mail SMTP installed. If it's already there you'll need to delete it first.

classes/controllers/FrmAddonsController.php
```diff
@@ -971,6 +971,19 @@ protected static function maybe_show_cred_form() {
...
+	 * Checks if an addon download url is allowed.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $url
+	 *
+	 * @return bool
+	 */
+	private static function is_allowed( $url ) {
+		return strpos( $url, 'https://downloads.wordpress.org/plugin' ) === 0 && substr_compare( $url, '.zip', -4 ) === 0;
```
Crabcyborg, line 984: @AbdiTolesa What do you think about including `FrmAppHelper::validate_url_is_in_s3_bucket( $download_url, 'zip' )` inside of this function check. With a name like `is_allowed` it's confusing that it's only half of the "is allowed" check.

AbdiTolesa: @Crabcyborg That sounds good, I have just done updated that.

Crabcyborg (approved):
Thanks @AbdiTolesa! 🚀

PR #1124, opened 2023-03-24
switch old field ids in form conditional settings

AbdiTolesa:
This update just makes sure the submit button conditionals are copied over to the new form successfully by replacing old ids to new ids.

### DIFF
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -138,7 +138,7 @@ class FrmForm {
 		if ( ! empty( $new_opts['submit_conditions']['hide_field'] ) ) {
 			global $frm_duplicate_ids;
 			foreach ( $new_opts['submit_conditions']['hide_field'] as $key => $val ) {
-				if ( array_search( $val, array_keys( $frm_duplicate_ids ) ) !== false ) {
+				if ( array_search( $val, array_keys( (array) $frm_duplicate_ids ) ) !== false ) {
 					$new_opts['submit_conditions']['hide_field'][ $key ] = $frm_duplicate_ids[ $val ];
 				}
 			}
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -135,6 +135,10 @@ class FrmForm {
 			$new_opts['success_msg'] = FrmFieldsHelper::switch_field_ids( $new_opts['success_msg'] );
 		}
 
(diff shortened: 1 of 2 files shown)


classes/models/FrmForm.php
```diff
@@ -135,6 +135,15 @@ public static function after_duplicate( $form_id, $values ) {
 			$new_opts['success_msg'] = FrmFieldsHelper::switch_field_ids( $new_opts['success_msg'] );
 		}
 
+		if ( ! empty( $new_opts['submit_conditions']['hide_field'] ) ) {
+			global $frm_duplicate_ids;
+			foreach ( $new_opts['submit_conditions']['hide_field'] as $key => $val ) {
+				if ( array_search( $val, array_keys( (array) $frm_duplicate_ids ) ) !== false ) {
```
Crabcyborg, line 141: @AbdiTolesa This seems like a really inefficient line.

Can't you just check if `isset( $frm_duplicate_ids[ $val ] )` without having to call `array_keys` in a loop or an expensive `array_search` call? It's already indexed.

AbdiTolesa: @Crabcyborg So weird I was blind to that!! I don't know what I was thinking when I wrote it... just fixed it now

classes/models/FrmForm.php
```diff
@@ -135,6 +135,15 @@ public static function after_duplicate( $form_id, $values ) {
...
+		if ( ! empty( $new_opts['submit_conditions']['hide_field'] ) ) {
+			global $frm_duplicate_ids;
+			foreach ( $new_opts['submit_conditions']['hide_field'] as $key => $val ) {
+				if ( isset( $frm_duplicate_ids[ $val ] ) ) {
+					$new_opts['submit_conditions']['hide_field'][ $key ] = $frm_duplicate_ids[ $val ];
+				}
+			}
+		}
+
 		$new_opts = apply_filters( 'frm_after_duplicate_form_values', $new_opts, $form_id );
```
Crabcyborg, line 147: @AbdiTolesa One other thing. Submit button conditions are a Pro feature. Can we move this update into Pro using this `frm_after_duplicate_form_values` filter here?

Crabcyborg:
Closing this as it has been replaced with https://github.com/Strategy11/formidable-pro/pull/4162

PR #1130, opened 2023-03-28
Use checkbox role for multiselect buttons and add aria checked attribute

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2196994272/156886/

The multiselect dropdowns (used for field visiblity settings, and global permission settings, and registration action role settings) wouldn't state to a screen reader whether the nested input was checked or not.

I'm fixing this with a `role="checkbox"` attribute and syncing the value of `aria-checked` when the dropdown shows and when the value changes.

### DIFF
--- a/js/admin/dom.js
+++ b/js/admin/dom.js
@@ -185,6 +185,10 @@
 					const $dropdownItem = $dropdown.find( 'input[value="' + optionValue + '"]' ).closest( 'button.dropdown-item' );
 					if ( $dropdownItem.length ) {
 						$dropdownItem.attr( 'aria-checked', checked ? 'true' : 'false' );
+
+						// Delay a focus event so the screen reader reads the option value again.
+						// Without this, and without the setTimeout, it only reads "checked" or "unchecked".
+						setTimeout( () => $dropdownItem.get( 0 ).focus(), 0 );
 					}
 				}
 			});
--- a/js/admin/dom.js
+++ b/js/admin/dom.js
@@ -141,12 +141,11 @@
 
 	const multiselect = {
 		init: function() {
-			let $select, id, labelledBy;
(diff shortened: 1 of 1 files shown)


garretlaxton:
This sounds good now, it will say `checkbox checked` or `checkbox not checked`. 

I did have issues using the arrow keys to navigate the different options though. Also If I unchecked an option I couldn't check it again unless I tab to a different option and come back. Not sure if this is just my screen reader though.

Crabcyborg:
> I did have issues using the arrow keys to navigate the different options though.

Personally I navigate it with screen reader controls. On my mac, `ctrl` + `option` + `left`/`right` moves up the down the dropdown. Arrow keys may be a good idea but I don't want to put too much code on top of this as this mutliselect dropdown is a dependency that will update in the future so anything we put on top of it may break.

> Also If I unchecked an option I couldn't check it again unless I tab to a different option and come back.

That's interesting. I can't replicate that in my screen reader. I just added a small update though where I've added a focus event with a slight delay. I noticed when I checked/unchecked that it would just say "checked" or "unchecked". Now it says, "checked, Author" or "unchecked, Author" when "Author" is my option, at least in my screen reader. Maybe that would help in your case as well if it's losing focus after clicking. In my case I'm using the space key which I haven't seen any issues with.

garretlaxton (approved):
Nice, the focus now fixes the select issue I mentioned.

It sounds like I use the same keys as you do, but for some reason my left/right keys aren't navigating to the different options. I just use an extension for chrome though so it's probably that.

Crabcyborg:
Thanks @garretlaxton! If it comes up in support after this again I think we could keep improving on it.

For now I think this is a good step forward 🚀.

PR #1131, opened 2023-03-30
Add submit with ajax into Lite

Crabcyborg:
**Pro branch** https://github.com/Strategy11/formidable-pro/pull/4164

This update:
- Replaces the placeholder AJAX submit setting with a real one. If Pro is active but not up to date, it doesn't display it and lets Pro render it isn't. This includes a new `ajax_create` function that is copied from Pro but with all of the Pro features stripped away to make it more simple. When Pro is active, the Pro `ajax_create` function is used instead.
- Adds the `frm_ajax_submit` unless Pro is active and not up to date. Otherwise it lets Pro handle it to avoid it appearing twice.
- Adds a specialized `FrmFormState` class that doesn't add a state field if the title attributes match the form settings, and doesn't add the state field when AJAX submit isn't enabled. This way it is only added when it is required. When Pro is active, the `FrmProFormState` is used instead.


garretlaxton (approved):
This looks good! Tested with only updating lite and only with the updated pro.

classes/controllers/FrmEntriesAJAXSubmitController.php
```diff
@@ -0,0 +1,186 @@
...
+		$response = array(
+			'errors'  => array(),
+			'content' => '',
+			'pass'    => false,
+		);
+
+		$form_id = FrmAppHelper::get_post_param( 'form_id', 0, 'absint' );
+		if ( ! $form_id ) {
+			echo json_encode( $response );
+			wp_die();
```
omidakhavans, line 43: I usually go with `wp_send_json` since it has `wp_die` for ajax within it. But not important.

omidakhavans (approved):
Definitely we could at least cut off and deprecate the `FrmProFormState` class from pro in future versions.
Scripts on footer, form validation, Error, HTML success message etc checked and passed! Great job @Crabcyborg. It works awesome.

Crabcyborg:
Thanks @omidakhavans! I plan to re-use more code soon in an upcoming release. I have a plan in mind. For now I'm going to leave it as-is to avoid any possibly last minute bugs as it's been tested already.

Thanks @garretlaxton for testing too!

🚀

PR #1135, opened 2023-04-04
make dropdown height match date inputs

AbdiTolesa:
Make Date options dropdown height match date inputs.
**Before**:
image

**After**:
image

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -5924,6 +5924,10 @@ span.howto {
 	max-width: 100%;
 }
 
+#form_reports_page form select {
+    line-height: 2;
+}
+
 .frm_wrap .google-visualization-table .gradient,
 .frm_wrap .google-visualization-table-tr-head,
 .frm_wrap .google-visualization-table-tr-odd,


Crabcyborg:
@AbdiTolesa I don't think this is the proper fix for this issue at all.

It looks like the height is already good on Firefox. I only see the inconsistent height in Chrome.

If you do try this fix in Firefox, it looks worse than it does in Chrome without this fix,
Screen Shot 2023-04-05 at 9 59 40 AM

We also don't want any references to `#form_reports_page` in the lite plugin.

I played with it for a moment and I found I had the best results by using a smaller font for date fields.
```
.frm_wrap input[type="date"] {
	font-size: var(--text-sm);
}
```

The font size difference bumps it down 2px which I don't find too noticeable.
Screen Shot 2023-04-05 at 10 01 17 AM

What do you think?

Crabcyborg:
@stephywells What do you think about bumping down the font size on the date input?

Or do you think there's a better solution we should go with?

stephywells:
This PR isn't needed. I already fixed this here:
https://github.com/Strategy11/formidable-forms/pull/1134

Crabcyborg:
> This PR isn't needed. I already fixed this here: #1134

@stephywells  The select is 30px high and the date fields are 33.71px.

Screen Shot 2023-04-05 at 12 18 15 PM
Screen Shot 2023-04-05 at 12 18 56 PM
Screen Shot 2023-04-05 at 12 19 02 PM

If that doesn't matter, I'll just close this.

stephywells:
I see it now. Looks like this needs to be overridden. New PR incoming.

```
::-webkit-datetime-edit {
    line-height: 1.85714286;
}
```

PR #1136, opened 2023-04-05
Add AI field in Lite

stephywells:
@Crabcyborg did you have a good way of not showing this field until the plugin is released?


Crabcyborg:
@stephywells Yeah! I just pushed a commit that uses `FrmAppHelper::show_new_feature`. It takes a plugin key.

stephywells:
Awesome! Thanks so much for adding that!

PR #1142, opened 2023-04-06
Fix shortcodes aren't parsing in redirect URL

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/4174

I ran `FrmEntriesController::prepare_redirect_url` before processing shortcodes, so shortcode tags are encoded.


Crabcyborg (approved):
Looks great! Thanks @truongwp!

PR #1144, opened 2023-04-06
support custom validity message

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4150

Field setting:
image

**Before:**
image

**After:**
image


AbdiTolesa:
@Crabcyborg  I see this error in the workflow, I tried to run the same `npm install eslint` command locally and that seems successfully executing. Do you any idea what could be causing this probably?

image

Crabcyborg:
@AbdiTolesa This is what I was asking you to look into yesterday https://strategy11.slack.com/archives/G01KDLFJTJS/p1680797320948139?thread_ts=1680786023.720269&cid=G01KDLFJTJS

It's broken in all of our plugins I think. Either one of the dependencies here is deprecated and removed, or something isn't locked and it changed versions and now requires things that are incompatible.

I don't really know the root of the issue. It likely requires updating a lot of these dependencies to versions that work together. There are a ton of them though.

It seems to stem from `@wordpress/eslint-plugin`.

AbdiTolesa:
> @AbdiTolesa This is what I was asking you to look into yesterday https://strategy11.slack.com/archives/G01KDLFJTJS/p1680797320948139?thread_ts=1680786023.720269&cid=G01KDLFJTJS
> 
> It's broken in all of our plugins I think. Either one of the dependencies here is deprecated and removed, or something isn't locked and it changed versions and now requires things that are incompatible.
> 
> I don't really know the root of the issue. It likely requires updating a lot of these dependencies to versions that work together. There are a ton of them though.
> 
> It seems to stem from `@wordpress/eslint-plugin`.

I see, I'll take more time to investigate it then. Thanks!

AbdiTolesa:
@Crabcyborg I get to fix the workflow dependencies issue but the linter task is failing seemingly because of the same `camelcase` rule being broken.
`'frm_admin_js' is not in camel case`.

 Should we update that as well or probably turn off the rule to pass this check?

Crabcyborg:
Thanks @AbdiTolesa!

`frm_admin_js` is a global so we shouldn't update it. We should add exceptions for these specific variables if possible so it can still check for camel case for other variables.

AbdiTolesa:
> Thanks @AbdiTolesa!
> 
> `frm_admin_js` is a global so we shouldn't update it. We should add exceptions for these specific variables if possible so it can still check for camel case for other variables.

@Crabcyborg I don't know if there is a better way to add exceptions but I at least know we can do that adding ` // eslint-disable-line camelcase` inline with the lines of code with the rules broken. Shall we just do that?

Crabcyborg:
@AbdiTolesa I don't know if there's a better way than that. Psalm is greater at this but these tools all have different strengths.

The comments are fine.

AbdiTolesa:
> @AbdiTolesa I don't know if there's a better way than that. Psalm is greater at this but these tools all have different strengths.
> 
> The comments are fine.

Got it all passed finally!

js/formidable.js
```diff
@@ -1347,6 +1348,24 @@ function frmFrontFormJS() {
 		});
 	}
 
+	function setCustomValidityMessage( e ) {
+		let selector;
```
Crabcyborg, line 1352: There are a couple uses of `let` here.

js/formidable.js
```diff
@@ -1347,6 +1348,24 @@ function frmFrontFormJS() {
...
+		let textInputSelector = 'input[type=text]:not(.frm_verify)';
+		if ( typeof e !== 'undefined' && e.type === 'frmAfterAddRow' ) {
+			selector = `.frm_repeat_sec:last-of-type ${textInputSelector}`;
```
Crabcyborg, line 1355: We should also avoid template literals for now in front end code.

AbdiTolesa: Done, good to know that as well!

garretlaxton:
This looks good!

If you mark the field as required and select the validate this with JS option, the notice does show twice though. Not sure if that's really important though.
image
image

js/formidable.js
```diff
@@ -1347,6 +1348,25 @@ function frmFrontFormJS() {
...
+		var selector,
+			textInputSelector = 'input[type=text]:not(.frm_verify)';
+
+		if ( typeof e !== 'undefined' && e.type === 'frmAfterAddRow' ) {
+			selector = '.frm_repeat_sec:last-of-type ' + textInputSelector;
+		} else {
+			selector = '.frm_fields_container ' + textInputSelector;
+		}
+
+		document.querySelectorAll( selector ).forEach( function( el ) {
```
Crabcyborg, line 1361: @AbdiTolesa I'm concerned this might not be the most optimized query to have on every page load.

Do you think there's a way we could do this more efficiently? Maybe we could just set it the first time the field is changed? We could probably use a pre-existing listener or something.

Crabcyborg:
Thanks @garretlaxton it looks like that's consistent to how it worked before. We probably wouldn't want to hide the red labels just to be consistent.

garretlaxton (approved):
Thanks, this looks good then!

js/formidable.js
AbdiTolesa: @Crabcyborg I think setting the validity message when the field is changed the first time would require similar query, right? I couldn't see how it may perform better if so. I couldn't find pre-existing listener we could reuse either. I am thinking we could probably avoid the query that immediately runs on page load by doing it in PHP, something like `oninvalid="..."` but not sure if that attribute would pass through our kses check.

Crabcyborg: @AbdiTolesa I think using `oninvalid` isn't really the more consistent syntax to begin with.

What I'm trying to avoid is any query selector logic on load. A page with a lot of HTML could be pretty slow to query selector every element on a page. Especially with a fairly complex selector like `.frm_repeat_sec:last-of-type  input[type=text]:not(.frm_verify)` which isn't cheap to process.

I'm thinking something more like this https://github.com/Strategy11/formidable-forms/pull/1176

In this case I'm querying Formidable forms instead and listening to invalid events on the forms. Most of the logic happens now when it's invalid, rather than on form load. So the time it takes for the browser to fully load should be quicker this way.

Crabcyborg:
I'm just going to merge this as-is so we can get linting working on our other pull requests. This could get cleaned up more but I think it's fine.

PR #1145, opened 2023-04-07
Also apply frm_redirect_url filter when paypal url is set in args

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-paypal/issues/42

The issue is that it's exiting here for a PayPal action and not running `frm_redirect_url` which is required.

@truongwp Can you come up with something better than this? We probably want this behind `ran_redirect_url_filter` - this is still a bit of a rough draft.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -2343,6 +2343,9 @@ class FrmFormsController {
 	 */
 	private static function run_redirect_url_filter( $url, $form, $args ) {
 		if ( empty( $args['action'] ) || ! is_string( $args['action'] ) ) {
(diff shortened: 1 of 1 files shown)


truongwp:
@Crabcyborg I see that Stripe also uses the `frm_redirect_url`. I'm afraid that there might be other add-ons that use it too.

I think we should make that filter always run in the `run_success_action()` method as before. When getting met actions, we will run that filter to check if the redirect action is valid, not assign the filtered URL. This might affect the performance because that filter is called twice. (the URL is assigned once)

Crabcyborg:
@truongwp Thanks. Good point about other add ons.

I'm good with whatever solution you think is best for this issue.

PR #1146, opened 2023-04-07
Fix PayPal action redirecting to a blank page

truongwp:
Fixes https://github.com/Strategy11/formidable-paypal/issues/42

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -5,15 +5,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 class FrmFormsController {
 
-	/**
-	 * Track the form and action that ran frm_redirect_url filter. Each item in array is {form_id}_create or {form_id}_update.
-	 *
-	 * @since 6.2
-	 *
-	 * @var array
-	 */
-	private static $ran_redirect_url_filter = array();
-
 	public static function menu() {
 		$menu_label = __( 'Forms', 'formidable' );
 		if ( ! FrmAppHelper::pro_is_installed() ) {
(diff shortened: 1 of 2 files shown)


Crabcyborg:
@garretlaxton Can you help test this out? It would be good to test with confirmation actions that use the message option as well as ones that use the redirect. And to test with PayPal, and to test with Stripe.

I'm hoping to release this today in the next couple of hours as it's a pretty important issue.

Thank you!

Crabcyborg:
@truongwp Is it possible to add unit tests to cover the PayPal issue?

truongwp:
@Crabcyborg Can it be added later? I'm not sure I will be able to work in the next couple of hours.
I also see one failing unit test. I think the test inside the `if (headers_sent())` is enough, right?

Crabcyborg:
> I think the test inside the `if (headers_sent())` is enough, right?

@truongwp That may only be enough if headers_sent is always true. Would it work in a `assertTrue( headers_sent() )` first or are there cases where it would be false while running the unit test?

I'm also concerned why it would be broken with this update when that unit test passes in master. Why it this update causing a breaking test?

Crabcyborg:
@truongwp I think it's failing because this line here doesn't filter the success_url https://github.com/Strategy11/formidable-forms/blob/7ec1189ff7a0f56643b51f9560bd3593166a055b/classes/controllers/FrmFormsController.php#L2582

garretlaxton:
I just finished looking at this and PayPal worked fine for me. 

I do have an issue with Stripe when using a redirect though, I'm taken to my wp-admin page instead of the URL set in the confirm action. Should I log that in the Stripe repo?

Crabcyborg:
> I do have an issue with Stripe when using a redirect though, I'm taken to my wp-admin page instead of the URL set in the confirm action. Should I log that in the Stripe repo?

@garretlaxton If that with Stripe link on only, and for a URL that isn't to your own site?

I believe I've seen this happen - because I'm using `wp_safe_redirect` which doesn't let you go to an external site. I should update this but I haven't prioritized it as I haven't seen it come up in support yet.

truongwp:
> I think it's failing because this line here doesn't filter the success_url

@Crabcyborg I think confirmation actions aren't run when you run `show_form()`.

PR #1150, opened 2023-04-10
add 'Select All' button to forms export

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/3715

Just adding a checkbox that allows selecting/deselecting forms for in the export table.

image

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -10267,6 +10267,10 @@ function frmAdminBuildJS() {
 			});
 
 			showOrHideRepeaters( getExportOption() );
+
+			document.querySelector( '#cb-import-select-all' ).addEventListener( 'change', event => {
+				document.querySelectorAll( '[name="frm_export_forms[]"]' ).forEach( cb => cb.checked = event.target.checked );
+			});
 		},
 
 		inboxBannerInit: function() {
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -8133,10 +8133,10 @@ function frmAdminBuildJS() {
 	}
 
 	function exportTypeChanged( event ) {
-		var target = event.target;
-		showOrHideRepeaters( target.value );
-		checkExportTypes.call( target );
-		checkSelectedAllFormsCheckbox( target.value );
+		var value = event.target;
+		showOrHideRepeaters( value );
+		checkExportTypes.call( event.target );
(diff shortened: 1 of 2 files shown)


garretlaxton:
This looks pretty good. 

One thing to point out though is if I toggle select all and switch the format to CSV for example, the select all is still toggled but the forms aren't.
image

AbdiTolesa:
> This looks pretty good.
> 
> One thing to point out though is if I toggle select all and switch the format to CSV for example, the select all is still toggled but the forms aren't. image
> 
> Also, it seems the select all doesn't work with the CSV format as it only exports entries from the first form available.

Thanks for feedback! I just pushed a little update, would you give it a shot again please?

garretlaxton (approved):
Looks great now!

Switching formats unchecks the select all. And the select all isn't available to CSV exports.

js/formidable_admin.js
```diff
@@ -8133,8 +8133,19 @@ function frmAdminBuildJS() {
...
+		var value = event.target.value;
+		showOrHideRepeaters( value );
 		checkExportTypes.call( event.target );
+		checkSelectedAllFormsCheckbox( value );
+	}
+
+	function checkSelectedAllFormsCheckbox( exportType ) {
+		if ( exportType === 'csv' ) {
+			document.querySelector( '#frm-export-select-all' ).checked = false;
+			document.querySelector( '#frm-export-select-all' ).disabled = true;
```
Crabcyborg, line 8145: @AbdiTolesa I think `getElementById` is generally a better function to use when querying for things by id. It's likely a bit faster, but it's also a bit more readable.

Crabcyborg: We should also avoid getting the element twice. It should be set to a variable.

js/formidable_admin.js
```diff
@@ -8133,8 +8133,19 @@ function frmAdminBuildJS() {
 	}
 
 	function exportTypeChanged( event ) {
-		showOrHideRepeaters( event.target.value );
+		var value = event.target.value;
```
Crabcyborg, line 8136: This is an admin script so let's use `let` or `const`.

Crabcyborg (approved):
Thanks @AbdiTolesa and @garretlaxton! 🚀

PR #1151, opened 2023-04-11
Fix common XML import issues

Crabcyborg:
I was looking in HelpScout for errors and noticed a pretty common pattern with import errors.

It's common enough that there's information on how to avoid it in the documentation https://formidableforms.com/knowledgebase/import-forms-entries-and-views/#kb-simplexml_parse_error

In a few cases the customers included XML import examples

In https://secure.helpscout.net/conversation/2061631158/116223 it's because of an extra `<meta name="generator" content="Equity 1.7.13" />` tag added by the Equity WP theme. I was able to find the problematic code here in GitHub https://github.com/idxbroker/Equity-Framework

```
add_filter( 'the_generator', 'equity_generator' );
function equity_generator( $generator ) {
    $generator .= "\r\n" . '<meta name="generator" content="' . PARENT_THEME_NAME . ' ' . PARENT_THEME_VERSION . '" />';
    return $generator;
}
```

In https://secure.helpscout.net/conversation/1663593319/83217 the issue is a carriage return at the start of the file instead which is easy to strip out.

This helps eliminate two common XML parser errors:
"XML declaration allowed only at the start of the document"
"Extra content at the end of the document"

### DIFF
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -36,10 +36,14 @@ class FrmXMLHelper {
 			return new WP_Error( 'SimpleXML_parse_error', __( 'Your server does not have XML enabled', 'formidable' ), libxml_get_errors() );
 		}
 
+		$xml_string = file_get_contents( $file );
+		self::maybe_fix_xml( $xml_string );
+
 		$dom = new DOMDocument();
+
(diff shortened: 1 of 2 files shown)


Crabcyborg:
Thanks for testing @garretlaxton! 🚀

PR #1152, opened 2023-04-12
Stop calling utf8_encode when wrapping cdata

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4177

I spent some more time looking into how this is going to be handled in WordPress and found https://core.trac.wordpress.org/ticket/55603. It's still in-progress, set now for v6.3 but it's already been delayed multiple versions.

Referring to this comment https://core.trac.wordpress.org/ticket/55603#comment:11 this person at least is also suggesting that the use of `utf8_encode` may not be required at all.

> As for the two utf8_encode calls in wp_read_image_metadata and wxr_cdata functions, they are wrapped with !seems_utf8() clauses, which make believe core's utf8_encode calls are wrong too because there is no guarantee the source strings are always ISO-8859-1.

### DIFF
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -1573,8 +1573,6 @@ class FrmXMLHelper {
 		FrmAppHelper::unserialize_or_decode( $str );
 		if ( is_array( $str ) ) {
 			$str = json_encode( $str );
-		} elseif ( seems_utf8( $str ) === false ) {
-			$str = utf8_encode( $str );
 		}
 
 		if ( is_numeric( $str ) ) {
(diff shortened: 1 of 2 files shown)


stephywells (approved):
Nice find! This would be nice in a separate function.

Also, in the unit tests, I'm not sure which line is the ISO encoding. Maybe a comment there could help?

Crabcyborg:
Thanks @stephywells! I thought about the new function but figured I'd run the idea by your first.

I pushed a commit with a new `FrmAppHelper::maybe_utf8_encode` function.

I also added another assert for another ISO-8859-1 and added some extra comments. Basically, `\xC1` converts to `Á` and `\xE9` converts to `é`.

PR #1153, opened 2023-04-12
Avoid using arrays for array keys when unserializing

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4179


Crabcyborg:
It seems like the issue resolved itself. This seems cache related.

PR #1157, opened 2023-04-13
Add option to open the redirect page in another tab

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/4127

This PR adds "Open in new tab" toggle in the Redirect Confirmation action. When that action runs, the redirect URL is opened in a new tab, and the "Open in new tab" message (which is a new setting in Global settings > Message Defaults) is shown in the current tab. Since it opens a new tab automatically, it will be blocked by default, and users need to unblock it.

When the "Click here ..." text shows and "Open in new tab" is enabled, the "Click here" link will also be opened in a new tab when users click on it.

This is the screencast for this feature:

https://user-images.githubusercontent.com/19748318/235640769-395a425f-a930-46ef-8b8d-b4d52adc2cdf.mov


garretlaxton:
When AJAX submit is on, the form gets stuck and doesn't show the confirm message. The new tab opens, just the form is stuck.
image


Also, is it possible to get the message set in the show message section? It currently shows the default message even if I change it.
image

Crabcyborg:
> Also, is it possible to get the message set in the show message section?

@garretlaxton You should be able to use a second Confirmation action for that. I don't think it makes a lot of sense to try supporting settings in an inactive tab.

classes/controllers/FrmFormsController.php
```diff
@@ -2538,12 +2553,25 @@ private static function redirect_after_submit( $args ) {
...
+					'redirect'        => $success_url,
+					'open_in_new_tab' => $open_in_new_tab,
+				)
+			);
 			wp_die();
 		}
 
 		if ( ! headers_sent() && empty( $args['force_delay_redirect'] ) ) { // Not AJAX submit, no headers sent, and there is just one Redirect action runs.
+			if ( $open_in_new_tab ) {
+				echo "<script>window.open('" . esc_url_raw( $success_url ) . "', '_blank');</script>";
```
Crabcyborg, line 2570: I wonder if we'd want to escape this differently. `esc_js` may be a better function in this context as this is being used in JavaScript?

https://developer.wordpress.org/reference/functions/esc_js/

js/formidable.js
```diff
@@ -616,7 +616,12 @@ function frmFrontFormJS() {
 				}
 
 				jQuery( document ).trigger( 'frmBeforeFormRedirect', [ object, response ]);
-				window.location = response.redirect;
+
+				if ( response.open_in_new_tab ) {
```
Crabcyborg, line 620: Maybe we should call it `openInNewTab` so we can avoid snake case in our JavaScript.

classes/controllers/FrmFormsController.php
```diff
@@ -14,6 +14,16 @@ class FrmFormsController {
 	 */
 	private static $ran_redirect_url_filter = array();
 
+	/**
+	 * Track the form that opened the redirect URL in a new tab. This is used to check if we should show the default
+	 * message in the currect tab.
+	 *
+	 * @since 6.2
+	 *
+	 * @var array Keys are form IDs and values are boolean.
```
Crabcyborg, line 23: This says the values are boolean but it looks like it's a 1. It may be less ambiguous to just put that value is 1 if it is set? It looks like it's always 1, right?

truongwp:
@garretlaxton @Crabcyborg Thanks for checking this PR. I fixed the issue Garret reported.

truongwp:
I'm adding unit test for this case.

garretlaxton:
That looks good, but if I add a second confirmation to change the message the form will still get stuck with AJAX submit enabled.
image

truongwp:
@stephywells Garret and Mike found that we can set up another Confirmation action to show the message after opening the URL in a new tab. Do we need to show the default message anymore?

stephywells:
> @stephywells Garret and Mike found that we can set up another Confirmation action to show the message after opening the URL in a new tab. Do we need to show the default message anymore?

@truongwp I'm not quite sure what you mean. Maybe you could attach a screencast to show?

truongwp:
@stephywells Sorry. I edited my comment, but look like it wasn't successful. 

In case there aren't any other confirmation actions, we'll show the default message (`Your responses were successfully submitted. Thank you!`). But I think users will want to show another message, like `You will be redirected in a new tab`. To do that, users need to create another confirmation action to show that message.

If we show that new message by default, users don't need to add the new action. If users want to change it globally, they can translate it. If users want to change it per form, they can add new confirmation action. What do you think?

stephywells:
> In case there aren't any other confirmation actions, we'll show the default message (`Your responses were successfully submitted. Thank you!`). But I think users will want to show another message, like `You will be redirected in a new tab`. To do that, users need to create another confirmation action to show that message.
> 
> If we show that new message by default, users don't need to add the new action. If users want to change it globally, they can translate it. If users want to change it per form, they can add new confirmation action. What do you think?

Thanks for clearing this up! That sounds like a good plan to adjust the default when the new tab is used, and add this in the global settings. We could combine the regular default message and the tab message.

Thanks so much for adding this!

Crabcyborg:
@truongwp It looks like phpcs is failing some checks.

Screen Shot 2023-04-26 at 10 27 56 AM

truongwp:
@garretlaxton Please check this again. I fixed the bug you reported, and added the `Open In New Tab Message` settings in the Message Defaults of Global settings. Now when opening the URL in a new tab, that message will be shown in the tab contains form.

@Crabcyborg I added unit tests (for AJAX submit as well in the PR I mentioned above) and fixed the PHPCS errors. Since you followed this PR from the first, I don't request review from Abdi. But you can request him if you need.

garretlaxton (approved):
This looks great now!

js/formidable.js
```diff
@@ -616,8 +616,16 @@ function frmFrontFormJS() {
...
 
 				jQuery( document ).trigger( 'frmBeforeFormRedirect', [ object, response ]);
-				window.location = response.redirect;
-			} else if ( response.content !== '' ) {
+
+				if ( response.openInNewTab ) {
+					window.open( response.redirect, '_blank' );
+				} else {
+					window.location = response.redirect;
+					return;
```
Crabcyborg, line 624: @truongwp This `return` here is throwing me off. If you're returning on the path where things shouldn't continue, I think it's usually a good moment to just exit early.

I'm thinking it may be useful to add more context here. It's a bit funny at a glance that one would return early and the other wouldn't - though it makes sense as you would will need to load the response content when opening a new tab because the old tab never disappears.

```
if ( ! response.openInNewTab ) {
    // Because we're redirecting there is no need to update content.
    // Unless we're opening in a new tab because the old page will still update.
    window.location = response.redirect;
	return;
}
```

Crabcyborg: I don't really have any other comments though. I think this all looks great! 🥳

truongwp:
@garretlaxton Thanks for testing.
@Crabcyborg Thank you. I updated the code.

I'm preparing the screencast for Steph to review this.

Crabcyborg (approved):
🎉

classes/helpers/FrmOnSubmitHelper.php
```diff
@@ -94,6 +94,27 @@ public static function show_redirect_settings( $args ) {
...
+				$name_attr,
+				array(
+					'div_class' => 'with_frm_style frm_toggle',
+					'checked'   => ! empty( $args['form_action']->post_content['open_in_new_tab'] ),
+					'echo'      => true,
+				)
+			);
+			?>
+			<label for="<?php echo esc_attr( $id_attr ); ?>">
+				<?php esc_html_e( 'Open in new tab', 'formidable' ); ?>
```
stephywells, line 115: Let's add a tooltip here to hopefully answer questions quickly. I'm thinking something like:
"This option will open the link in a new browser tab. Please note that some popup blockers may prevent this from happening, in which case the link will be displayed."

js/formidable.js
```diff
@@ -616,8 +616,18 @@ function frmFrontFormJS() {
...
-			} else if ( response.content !== '' ) {
+
+				if ( ! response.openInNewTab ) {
+					// We return here because we're redirecting there is no need to update content.
+					window.location = response.redirect;
+					return;
+				}
+
+				// We don't return here because we're opening in a new tab, the old tab will still update.
+				window.open( response.redirect, '_blank' );
```
stephywells, line 627: When I tried this, The new tab didn't open because the popup was blocked. Maybe we can add a link to open the url if it doesn't open automatically? We could add the same 'Click here if you are not automatically redirected' message, but it would open in a new tab.

```
var newTab = window.open( response.redirect, '_blank' );
if (newTab == null) {
  // Popup blocked
  // Display a message with the fallback link
}
```

stephywells (changes requested):
Great work on this! I have a couple requests to make it a bit more usable and cut down on support questions.

js/formidable.js
truongwp: @stephywells Should we append the 'Click here if you are not automatically redirected' text to the new tab message (users will always see that text, even if it opens in new tab successfully)?

stephywells: @truongwp I think it's okay not to add it if we know the new tab has been opened, since it opens in front. Let's just add it as a fallback.

truongwp:
@stephywells Sorry. This will be ready on Monday.

truongwp:
@garretlaxton Can you test this again. I did a feature that Steph mentioned here https://github.com/Strategy11/formidable-forms/pull/1157#discussion_r1182764450

You need to remove your local site from allowed popup sites in your browser settings.

garretlaxton:
I'm still seeing popup blocked on Firefox and Chrome.

truongwp:
@garretlaxton Yes. But in case it’s blocked, you should see the “Click here” link.

garretlaxton (approved):
Ohh, okay. This looks good then!

truongwp:
@garretlaxton Thank you for testing this.

@stephywells Can you please take a look? I just added that fallback message for cases with 1 confirmation action because we had the message like that when multiple confirmation actions run.

stephywells:
Thanks for adding this @truongwp !

Here's what I'm seeing now when it's blocked:
Screenshot 2023-05-23 at 5 38 01 PM

Can the extra message be included inside the success message instead of outside? This doesn't look as good as it could.

truongwp:
@stephywells When multiple actions run, the Click here message is outside. So I make it outside for consistency.

stephywells:
@truongwp can it be inside if there is only one action with a message?

truongwp:
@stephywells Yes, I did it.

@garretlaxton Sorry. I think it's worth testing once again. Now if just one redirect action runs, it has Open in new tab enabled, and your test site is blocked from opening a popup, the "Click here..." message should be in the same div with "The page has been opened in new tab" message.

stephywells:
Thanks so much @truongwp !

garretlaxton (approved):
This looks good still!

stephywells (approved):
This is great! Nice job on this @truongwp !

PR #1158, opened 2023-04-13
Remove moveTopBanner function

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4176

The banner is being added in the proper place, then this JavaScript is breaking it.

Do we need this anymore @stephywells? It looks like it was added in v5.4.3 but the v6.0 changes appear to break it.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -6131,22 +6131,6 @@ function frmAdminBuildJS() {
 		}
 	}
 
-	// Move the top banner above the screen options to prevent overlap.
-	function moveTopBanner() {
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This looks great now!
image

stephywells:
@Crabcyborg looks like we don't need it now. Does this cover the Views pages too?

Crabcyborg:
> Does this cover the Views pages too?

Yeah, the same issue happens on the views editor and this update covers it.

**Before**
Screen Shot 2023-04-14 at 9 44 00 AM

**After**
Screen Shot 2023-04-14 at 9 43 49 AM

I also tested full screen mode and it's the same issue there. This update fixes that too.

PR #1159, opened 2023-04-13
Add a warning to the styler to a detect a broken style

Crabcyborg:
Following up from Steph's comment here https://github.com/Strategy11/formidable-forms/pull/1141#pullrequestreview-1380009049

This update doesn't include https://github.com/Strategy11/formidable-forms/pull/1141. It's easier to test this when you can still break it with `(` or `)` like in my example here,

I added a referer check so it wouldn't interfere with the front end. It also only happens through the `frmpro_css` action but that is the fallback for when the CSS file doesn't work.

gpvqvEdH0o

@stephywells What do you think we should use for the warning's message body? I put something but it could probably use more information.

### DIFF
--- a/classes/views/styles/_style-preview-container.php
+++ b/classes/views/styles/_style-preview-container.php
@@ -22,6 +22,9 @@ if ( $sample_form_is_on ) {
 		include FrmAppHelper::plugin_path() . '/classes/views/shared/errors.php';
 		?>
 		<?php FrmTipsHelper::pro_tip( 'get_styling_tip', 'p' ); // If Pro is not active, this will show an upsell. ?>
+		<div id="frm_css_error" class="frm_warning_style">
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
This looks good! 

It may be helpful to include something like `One or more of your styles may contain invalid characters`.

classes/views/styles/_style-preview-container.php
```diff
@@ -22,6 +22,9 @@
 		include FrmAppHelper::plugin_path() . '/classes/views/shared/errors.php';
 		?>
 		<?php FrmTipsHelper::pro_tip( 'get_styling_tip', 'p' ); // If Pro is not active, this will show an upsell. ?>
+		<div id="frm_css_error" class="frm_warning_style">
+			<?php esc_html_e( 'If you can see this warning it means that one or more of your style settings are incorrectly formatted!', 'formidable' ); ?>
```
stephywells, line 26: How about this?

One or more of your style settings may contain invalid characters that are breaking form styling.

PR #1161, opened 2023-04-17
Avoid fatal woocommerce error

Crabcyborg:
Requires https://github.com/Strategy11/formidable-pro/pull/4192 when Pro is active.

@garretlaxton This fix may not be final. I think I could avoid the duplicate code better. But the fix will be the same, and this is an important bug so testing it as soon as possible would be great.


Crabcyborg:
This is now fixed with https://github.com/Strategy11/formidable-forms/pull/1163 instead.

PR #1163, opened 2023-04-17
Fix woocommerce conflict in one place

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-stripe/issues/220
Fixes https://strategy11.slack.com/archives/CJFQ599V5/p1681620515610519

Improves upon https://github.com/Strategy11/formidable-forms/pull/1161 by fixing the issue in both Lite and Pro to avoid multiple patch releases and duplicate code.

It's also a private function so it will be easy to remove if WooCommerce fixes this on their end in the future.


stephywells (approved):
Great call! Looks good.

PR #1167, opened 2023-04-18
Add Dismissible Warning Message Functionality

shervElmi:
This PR introduces dismissible warning message functionality. It adds the `FrmAppHelper::add_dismissable_warning_message` method, which can be used to easily display dismissible warning messages.

### Related Issues:
https://github.com/Strategy11/formidable-pro/issues/4134

### Before:
image

### After:
image

### After - RTL:
image

### DIFF
--- a/classes/controllers/FrmHooksController.php
+++ b/classes/controllers/FrmHooksController.php
@@ -5,6 +5,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 class FrmHooksController {
 
+	const DISMISS_IP_NOTICE_ACTION = 'frm_dismiss_custom_header_ip_notice';
+
 	/**
 	 * Trigger plugin-wide hook loading
 	 *
@@ -237,7 +239,7 @@ class FrmHooksController {
 
 		// Settings.
 		add_action( 'wp_ajax_frm_lite_settings_upgrade', 'FrmSettingsController::settings_cta_dismiss' );
-		add_action( 'wp_ajax_frm_dismiss_custom_header_ip_notice', 'FrmAppHelper::dismiss_warning_message' );
+		add_action( 'wp_ajax_' . self::DISMISS_IP_NOTICE_ACTION, 'FrmAppHelper::dismiss_warning_message' );
 
 		// Styles Controller.
 		add_action( 'wp_ajax_frm_settings_reset', 'FrmStylesController::reset_styling' );
--- a/classes/controllers/FrmHooksController.php
+++ b/classes/controllers/FrmHooksController.php
@@ -5,8 +5,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 class FrmHooksController {
 
-	const DISMISS_IP_NOTICE_ACTION = 'frm_dismiss_custom_header_ip_notice';
-
 	/**
 	 * Trigger plugin-wide hook loading
 	 *
@@ -239,7 +237,6 @@ class FrmHooksController {
 
 		// Settings.
 		add_action( 'wp_ajax_frm_lite_settings_upgrade', 'FrmSettingsController::settings_cta_dismiss' );
-		add_action( 'wp_ajax_' . self::DISMISS_IP_NOTICE_ACTION, 'FrmAppHelper::dismiss_warning_message' );
 
 		// Styles Controller.
 		add_action( 'wp_ajax_frm_settings_reset', 'FrmStylesController::reset_styling' );
--- a/classes/controllers/FrmHooksController.php
+++ b/classes/controllers/FrmHooksController.php
@@ -237,6 +237,7 @@ class FrmHooksController {
 
 		// Settings.
 		add_action( 'wp_ajax_frm_lite_settings_upgrade', 'FrmSettingsController::settings_cta_dismiss' );
+		add_action( 'wp_ajax_frm_dismiss_custom_header_ip_notice', 'FrmAppHelper::dismiss_warning_message' );
 
 		// Styles Controller.
 		add_action( 'wp_ajax_frm_settings_reset', 'FrmStylesController::reset_styling' );
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -3857,11 +3857,13 @@ class FrmAppHelper {
 			return;
 		}
 
+		$ajax_callback = function() use ( $option ) {
+			self::dismiss_warning_message( $option );
+		};
+
 		// We're handling JS codes with `doJsonPost` and it adds 'frm_' to the beginning of the action.
 		// To prevent any issues, we add 'frm_' from the beginning of the action.
-		add_action( 'wp_ajax_frm_' . $option, function() use ( $option ) {
-			self::dismiss_warning_message( $option );
-		});
+		add_action( 'wp_ajax_frm_' . $option, $ajax_callback);
 
 		add_filter(
 			'frm_message_list',
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -3878,4 +3878,23 @@ class FrmAppHelper {
 			}
 		);
 	}
+
+	/**
+	 * Dismiss a warning message and update the dismissal state.
+	 *
+	 * @since x.x
+	 *
+	 * @return void
+	 */
+	public static function dismiss_warning_message() {
+		self::permission_check( 'frm_change_settings' );
+		check_ajax_referer( 'frm_ajax', 'nonce' );
+
+		if ( ! empty( $_POST['action'] ) ) {
+			$option = sanitize_text_field( wp_unslash( $_POST['action'] ) );
+			update_option( $option, true );
+		}
+
+		wp_die();
+	}
 }
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -3849,30 +3849,43 @@ class FrmAppHelper {
 	 * @since x.x
 	 *
 	 * @param string $message The warning message to display.
-	 * @param string $action The WP Ajax action.
-	 *                       The unique identifier for the dismissal state of the message.
+	 * @param string $option  The unique identifier for the dismissal state of the message and the WP Ajax action.
 	 * @return void
 	 */
-	public static function print_dismissable_warning_message( $message, $action ) {
+	public static function add_dismissable_warning_message( $message = '', $option = '' ) {
+		if ( ! $message || ! $option ) {
+			return;
+		}
+
+		// Check if the option doesn't start with 'frm_'
+		if ( strpos( $option, 'frm_' ) !== 0 ) {
+			// Add the 'frm_' prefix to the option
+			$option = 'frm_' . $option;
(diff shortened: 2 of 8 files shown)


js/formidable_admin.js
```diff
@@ -3455,6 +3455,28 @@ function frmAdminBuildJS() {
...
+	 * Dismiss a warning message and send an AJAX request to update the dismissal state.
+	 *
+	 * @since x.x
+	 *
+	 * @param {Event} event The event object associated with the click on the dismiss icon.
+	 */
+	function dismissWarningMessage( event ) {
+		event.preventDefault();
+
+		const $this = jQuery( this );
```
Crabcyborg, line 3468: @shervElmi Can you write this without any jQuery? We're trying to move away from using it. Ideally we'll be able to never load it in the future - any new jQuery code will make it more difficult to stop using it in the future.

In this case `.closest` can be used this `this` and instead of `.data` you could probably just use `.dataset.action` instead.

I also wouldn't use `jQuery.post` for the same reason. We've been using `fetch` now on our admin pages. Ideally you would use the `doJsonPost` function loading to the `frmDom.ajax` global namespace defined by the `dom.js` JS dependency file instead.

classes/helpers/FrmAppHelper.php
```diff
@@ -3842,4 +3842,80 @@ public static function jquery_ui_base_url() {
...
+	public static function dismiss_warning_message() {
+		self::permission_check( 'frm_change_settings' );
+		check_ajax_referer( 'frm_ajax', 'nonce' );
+
+		if ( ! empty( $_POST['action'] ) ) {
+			$option = sanitize_text_field( wp_unslash( $_POST['action'] ) );
+			update_option( $option, true );
+		}
+
+		wp_die();
```
Crabcyborg, line 3898: It would be good to still use something like `wp_send_json_success` here rather than dying without any confirmation in the Network tab.

classes/helpers/FrmAppHelper.php
```diff
@@ -3842,4 +3842,80 @@ public static function jquery_ui_base_url() {
...
+	 * @since x.x
+	 *
+	 * @return void
+	 */
+	public static function dismiss_warning_message() {
+		self::permission_check( 'frm_change_settings' );
+		check_ajax_referer( 'frm_ajax', 'nonce' );
+
+		if ( ! empty( $_POST['action'] ) ) {
+			$option = sanitize_text_field( wp_unslash( $_POST['action'] ) );
```
Crabcyborg, line 3894: In Formidable we have this all covered with a `FrmAppHelper::get_post_param( 'action' )` line. It handles the `isset`, `wp_unslash`, and `sanitize_text_field` calls.

So you'd end up with something like this instead:

```
$action = FrmAppHelper::get_post_param( 'action' );
if ( $action ) {
    update_option( $action, true );
}
```

Next, `update_option( $option, true );` is a huge red flag. We should never update an option using a name from user input without very heavy validation. I think we should avoid sending the option name entirely and possible consider a hard coded key that holds an array of dismissing warning keys instead?

js/formidable_admin.js
shervElmi: Certainly! I have developed several builders, including form builders, using Vanilla JavaScript, and I truly enjoy working with Vanilla JS.

classes/controllers/FrmHooksController.php
```diff
@@ -5,6 +5,8 @@
 
 class FrmHooksController {
 
+	const DISMISS_IP_NOTICE_ACTION = 'frm_dismiss_custom_header_ip_notice';
```
Crabcyborg, line 8: This const isn't used anywhere else, and const values are public so this would be difficult to deprecate.

I think we're probably fine just keeping the action name in the string like we do for all of the other actions. It's more consistent.

classes/helpers/FrmAppHelper.php
```diff
@@ -3842,4 +3842,80 @@ public static function jquery_ui_base_url() {
...
+	 * Display a dismissable warning message and save its dismissal state.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $message The warning message to display.
+	 * @param string $action The WP Ajax action.
+	 *                       The unique identifier for the dismissal state of the message.
+	 * @return void
+	 */
+	public static function print_dismissable_warning_message( $message, $action ) {
```
Crabcyborg, line 3856: `print` isn't a great name in this case as this function isn't really printing anything itself but just setting filters.

I think a name more like `add_dismissing_warning_message` may make more sense.

classes/controllers/FrmHooksController.php
Crabcyborg: Nevermind, I see this is used somewhere else. Not sure why my search missed it the first time.

Maybe we should make a new class or something. I just don't love this in the `HooksController`. It doesn't feel like the right place for it.

Crabcyborg:
@shervElmi It looks like the dismiss icon gets forced to a new line in Firefox.

Screen Shot 2023-04-19 at 2 59 25 PM

classes/helpers/FrmAppHelper.php
```diff
@@ -3842,4 +3842,80 @@ public static function jquery_ui_base_url() {
...
+	 * @param array $allowed_html The array of allowed HTML tags and attributes.
+	 * @return array The updated array of allowed HTML tags and attributes.
+	 */
+	public static function add_allowed_dismiss_icon_tags( $allowed_html ) {
+		$allowed_html['span']['data-action'] = true;
+		$allowed_html['svg'] = array(
+			'class'      => true,
+			'aria-label' => true,
+		);
+		$allowed_html['use'] = array(
```
Crabcyborg, line 3915: In this case we could probably just add `svg` and `use` to the list of allows types on the line where we output the message like `FrmAppHelper::kses( $m, array( 'a', 'br', 'span', 'p', 'svg', 'use' )`.

Since `data-action` could be re-used for another warning I think we could in this case add the `data-action` exception filter right before we loop `$show_messages` and remove it after we're done the loop. The hard coding does probably make sense in this case.

classes/helpers/FrmAppHelper.php
```diff
@@ -3842,4 +3842,80 @@ public static function jquery_ui_base_url() {
...
+				if ( get_option( $action, false ) ) {
+					return $show_messages;
+				}
+
+				add_filter( 'frm_striphtml_allowed_tags', 'FrmAppHelper::add_allowed_dismiss_icon_tags' );
+
+				$dismiss_icon = self::icon_by_class(
+					'frmfont frm_close_icon',
+					array(
+						'aria-label' => 'Dismiss',
```
Crabcyborg, line 3869: We should translate "Dismiss".

css/frm_admin.css
```diff
@@ -1891,6 +1891,43 @@ div.frm_updated_message {
...
+    position: relative;
+    padding-right: calc(var(--gap-md) + 15px);
+}
+
+body.rtl .frm_warning_style:has(.frm-warning-dismiss) {
+    padding-right: var(--gap-md);
+    padding-left: calc(var(--gap-md) + 15px);
+}
+
+.frm_warning_style li:has(.frm-warning-dismiss) {
```
Crabcyborg, line 1905: This is probably why Firefox isn't supporting the dismiss icon in the corner. We should avoid experimental features like `:has` for now.

shervElmi:
Thank you for the feedback @Crabcyborg. I will return shortly with all the necessary solutions.

Crabcyborg:
Thanks @shervElmi! This looks good as a first draft once we figure out a better solution for handling the option. Thanks for cleaning some things up!

shervElmi:
@Crabcyborg, I have resolved the issues that you pointed out. The PR is ready for review.

js/formidable_admin.js
```diff
@@ -9549,6 +9567,11 @@ function frmAdminBuildJS() {
 				thisFormId = jQuery( document.getElementById( 'form_id' ) ).val();
 			}
 
+			// Add event listener for dismissible warning messages.
+			document.querySelectorAll( '.frm-warning-dismiss' )?.forEach( ( dismissIcon ) => {
+				dismissIcon.addEventListener( 'click', dismissWarningMessage );
```
Crabcyborg, line 9572: This isn't really a necessary change but you can hide some of the duplicate `preventDefault` code by using `frmDom.util.onClickPreventDefault` instead. It doesn't pass around `this`, but you can always use `event.target` instead. I generally try to avoid `this` in most cases anyway.

js/formidable_admin.js
```diff
@@ -9549,6 +9567,11 @@ function frmAdminBuildJS() {
 				thisFormId = jQuery( document.getElementById( 'form_id' ) ).val();
 			}
 
+			// Add event listener for dismissible warning messages.
+			document.querySelectorAll( '.frm-warning-dismiss' )?.forEach( ( dismissIcon ) => {
```
Crabcyborg, line 9571: Also, this optional chaining isn't necessary. Removing the `?` will have the same effect anyway since `querySelectorAll` always returns a NodeList array.

garretlaxton:
This looks good! Would it be worth adding to the file upload protection warning as well?
image

classes/helpers/FrmAppHelper.php
```diff
@@ -3842,4 +3842,68 @@ public static function jquery_ui_base_url() {
...
+	 *
+	 * @param string $option The unique identifier for the dismissal state of the message.
+	 * @return void
+	 */
+	public static function dismiss_warning_message( $option = '' ) {
+		self::permission_check( 'frm_change_settings' );
+		check_ajax_referer( 'frm_ajax', 'nonce' );
+
+		if ( $option ) {
+			update_option( $option, true );
```
Crabcyborg, line 3904: @shervElmi Would we possibly want to use `$autoload = 'no'` here?

shervElmi: @Crabcyborg, Could you please clarify when I should use this code? Also, would you like me to add this code after update_option( $option, true );?

shervElmi: Never mind, I got that!

classes/helpers/FrmAppHelper.php
```diff
@@ -3842,4 +3842,68 @@ public static function jquery_ui_base_url() {
...
+				$dismiss_icon = self::icon_by_class(
+					'frmfont frm_close_icon',
+					array(
+						'aria-label' => _x( 'Dismiss', 'warning message: close icon label', 'formidable' ),
+						'echo' => false,
+					)
+				);
+
+				$show_messages['warning_msg'] = $message;
+				$show_messages['dismiss_icon'] = '<span class="frm-warning-dismiss frmsvg" data-action="' . esc_attr( $option ) . '">' . $dismiss_icon . '</span>';
```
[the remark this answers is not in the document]
shervElmi: Apologies for the funny key name! 😅

We used a key name in the previous version of the warning message, as shown in the attached screenshot of the code. However, you're right, it's better not to use a key. Thank you!

image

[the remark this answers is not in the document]
Crabcyborg: Right, makes sense why you used keys then. Thanks for updating that!

PR #1168, opened 2023-04-20
Improve Readability of Import XML Error Messages

shervElmi:
In this pull request, we are improving the readability of import XML error messages by combining WP_Error details with the message.

### Related Issue:
https://github.com/Strategy11/formidable-pro/issues/4189

---

## Before:
image

### After:
image

### DIFF
--- a/classes/helpers/FrmXMLHelper.php
+++ b/classes/helpers/FrmXMLHelper.php
@@ -1280,7 +1280,7 @@ class FrmXMLHelper {
 		if ( is_wp_error( $result ) ) {
 			$errors[] = $result->get_error_message();
 
-			// Remove the SimpleXML_parse_error from a WP_Error object to avoid
+			// Remove the SimpleXML_parse_error from the WP_Error object to avoid
 			// displaying duplicate error messages from $result->get_error_message()
 			$error_codes = $result->get_error_codes();
 			$error_details = array();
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This looks good for me!
image

classes/helpers/FrmXMLHelper.php
```diff
@@ -1279,6 +1279,9 @@ private static function maybe_update_stylesheet( $imported ) {
 	public static function parse_message( $result, &$message, &$errors ) {
 		if ( is_wp_error( $result ) ) {
 			$errors[] = $result->get_error_message();
+			$errors[] = '<br />' . _x( 'Error details:', 'import xml message', 'formidable' ) . '<br />' . htmlentities( print_r( $result, 1 ) );
```
Crabcyborg, line 1282: Is `htmlentities` the best function here? Would `esc_html` make more sense?

It may also make sense to use `esc_html` around `_x` as well, similar to using `esc_html__` over `__`.

shervElmi: You're right, using `esc_html` would indeed be a better choice in this case. I initially chose `htmlentities` because I noticed it was used elsewhere in the code. However, I agree that esc_html is more appropriate. Thank you for the suggestion!

Crabcyborg:
Thank you @shervElmi!

One more thought.

I think it is kind of funny that we're showing the exact same "There was an error when reading this XML file" twice.
Screen Shot 2023-04-20 at 4 05 36 PM

Maybe we want to remove `SimpleXML_parse_error` from the object when we dump it?

shervElmi:
Thank you, @Crabcyborg! I think the changes are complete, and it's ready now.

Crabcyborg (approved):
Thank you @shervElmi! This looks great! 🚀

PR #1171, opened 2023-04-24
Fix blank value is selected by default for radio and checkbox

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/3385

This fix applies to the radio and checkbox fields. Users need to re-save the old form to apply this fix.


Crabcyborg:
@truongwp It looks like there may be other things that also need to change.

If I select the empty value, it doesn't appear as selected in the entry page.
R40OsCDDNp

It looks like that actually works fine on master, so this update appears to break that.

PR #1172, opened 2023-04-24
Refined Embed Modal Footer Styles

shervElmi:
In this PR, we have enhanced the styling of the Embed Modal Footer on the first step.

### Related Issue:
https://github.com/Strategy11/formidable-pro/issues/3657

### Testing Instructions:
1. Navigate to WP Admin > Formidable > Forms
2. Open an existing form or create a new one
3. Access the Embed Modal through the top navigation

### Before:
image

### After:
image


css/frm_admin.css
```diff
@@ -1635,6 +1635,12 @@ a.frm_hidden,
 	padding: 0;
 }
 
+.frm-entry-container .frm_start_over {
```
Crabcyborg, line 1638: @shervElmi I'm not sure if we'd want this styling to be specific to admin pages as the start over button is a front end feature.

Screen Shot 2023-04-25 at 2 26 00 PM

But adding space now could also cause funny issues for people. @truongwp What is the best way to space these apart at the moment?

truongwp: @Crabcyborg I don't think we should update the frontend. Users might add a bit of margin with custom CSS if they need the spacing. If we add our own spacing here, I think that will cause problems for users.

Crabcyborg: @shervElmi that sounds good. Changing the button spacing may not be worth it a sit just makes things inconsistent.

shervElmi:
@Crabcyborg, I think it's okay for you to review this PR instead of Omid. I'll request Omid's input on future PRs. Thank you.

Crabcyborg (approved):
Thanks @shervElmi! This looks good.

While I was looking at this, I noticed and logged an overflow issue with this modal https://github.com/Strategy11/formidable-pro/issues/4224

I assigned it to you. When you have a chance, could you look into that as well?

PR #1173, opened 2023-04-25
Add abandonment event to trigger_actions besides draft

omidakhavans:
There is a condition to avoid triggering actions whenever the draft column is set and the event is not draft to prevent the actions to be used while is_draft is set and event is different. This PR will add the abandonment event besides the draft event since we are using is_draft column for abandonment too.

### DIFF
--- a/classes/controllers/FrmFormActionsController.php
+++ b/classes/controllers/FrmFormActionsController.php
@@ -505,7 +505,7 @@ class FrmFormActionsController {
 				$entry = FrmEntry::getOne( $entry, true );
 			}
 
-			if ( empty( $entry ) || ( $entry->is_draft && $event != 'draft' ) ) {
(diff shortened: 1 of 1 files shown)


classes/controllers/FrmFormActionsController.php
```diff
@@ -505,7 +505,7 @@ public static function trigger_actions( $event, $form, $entry, $type = 'all', $a
 				$entry = FrmEntry::getOne( $entry, true );
 			}
 
-			if ( empty( $entry ) || ( $entry->is_draft && $event != 'draft' ) ) {
+			if ( empty( $entry ) || ( $entry->is_draft && $event !== 'draft' && $event !== 'abandoned' ) ) {
```
Crabcyborg, line 508: @omidakhavans Is it possible to just use `frm_skip_form_action` above instead so we can contain the abandonment update in the add on itself?

Crabcyborg:
Closing this as I believe we'll be able to keep this in the abandonment add on.

PR #1176, opened 2023-04-25
Try a more optimized version of issue 4150

Crabcyborg:
Trying to come up with something faster than https://github.com/Strategy11/formidable-forms/pull/1144

See https://github.com/Strategy11/formidable-forms/pull/1144#discussion_r1176725717 for more information.

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1348,23 +1348,39 @@ function frmFrontFormJS() {
 		});
 	}
 
-	function setCustomValidityMessage( e ) {
-		var selector,
-			textInputSelector = 'input[type=text]:not(.frm_verify)';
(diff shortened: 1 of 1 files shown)


AbdiTolesa (commented):
@Crabcyborg Thank you for reminding me of this capturing phase technique, which I have read about it a bit ago but never used it and now I know how to practically use it! Do you think we better put the if logics into a single block than multiple if...returns? I can do that if so.

Crabcyborg:
@AbdiTolesa I don't think one big check would really help readability but I'm all for improving it - I do agree it's a lot of ifs.

I'd begin with a new function like shouldChangeValidityMessage. There we can return true/false. That should at least break it up better.

PR #1181, opened 2023-04-26
Stop trying to display entry descriptions in the entry list

Crabcyborg:
This doesn't totally fix the root of the issue in https://github.com/Strategy11/formidable-pro/issues/3986 but it helps to avoid the fatal error if you used "description" as a field key.

I don't really understand why `'description'` is here in the switch/case statement. We don't include "Entry Description" as one of our options. It doesn't work here if it's an array, which is common as we store other data in this like user journeys.
Screen Shot 2023-04-26 at 2 53 53 PM

@stephywells Do you see an issue with removing `description` from the switch/case here? When I do, the description uses its meta normally and doesn't cause the fatal error,

Screen Shot 2023-04-26 at 2 56 44 PM

### DIFF
--- a/classes/helpers/FrmEntriesListHelper.php
+++ b/classes/helpers/FrmEntriesListHelper.php
@@ -122,7 +122,7 @@ class FrmEntriesListHelper extends FrmListHelper {
 
 		$colspan = $this->get_column_count();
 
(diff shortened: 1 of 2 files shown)


stephywells (approved):
That seems like it should be fine. That must have been copied from the form list class, but I'm not positive. I don't see a reason to keep it there.

PR #1182, opened 2023-04-27
Include the license in the POST request for dynamic urls

stephywells:
See more detail here about how to test:
https://github.com/Strategy11/formidable-pro/issues/4219

This only updates the license activation process. This adds a block to wait 2 minutes before trying to activate the same license again.


garretlaxton (approved):
This looks good!
image
image

One question though, when it says wait 2 minutes I can still enter my real license key. Is the 2 minutes only when an invalid license is used?

stephywells:
Thanks for double checking this!

The two minute limit is only for the same license. It doesn't block it if you deactivate and reactivate a license either.

Crabcyborg (approved):
Looks good! I can't think of anything else to change.

🚀

PR #1187, opened 2023-05-02
make sure to be specific about search inputs to exclude from searchContent

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4180

It looks like `template-search-input` is used as html id in different pages that are not even template searches, like the import/export page forms list container. I tried to make sure `searchContent` is not called in new form modal templates search input in this update.


Crabcyborg (approved):
Thanks @AbdiTolesa!

PR #1192, opened 2023-05-04
avoid whitespaces for dropdown field options

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4184

### DIFF
--- a/classes/helpers/FrmHtmlHelper.php
+++ b/classes/helpers/FrmHtmlHelper.php
@@ -38,7 +38,7 @@ class FrmHtmlHelper {
 	 * This is useful to avoid closing and opening PHP to echo <option> tags which leads to extra whitespace.
 	 * Avoiding whitespace saves 5KB of HTML for an international address field with a country dropdown with 252 options.
 	 *
-	 * @since 6.3
+	 * @since 6.3.1
 	 *
 	 * @param string $option   The string used as the option label.
 	 * @param bool   $selected True if the option should be selected.
--- a/classes/helpers/FrmHtmlHelper.php
+++ b/classes/helpers/FrmHtmlHelper.php
@@ -32,4 +32,25 @@ class FrmHtmlHelper {
 			isset( $args['echo'] ) ? $args['echo'] : false
 		);
 	}
+
+	/**
+	 * Echo a dropdown option.
+	 * This is useful to avoid closing and opening PHP to echo <option> tags which leads to extra whitespace.
+	 * Avoiding whitespace saves 5KB of HTML for an international address field with a country dropdown with 252 options.
+	 *
+	 * @since 6.3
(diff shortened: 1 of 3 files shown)


classes/views/frm-fields/front-end/dropdown-field.php
```diff
@@ -56,11 +56,16 @@
 		if ( FrmFieldsHelper::is_other_opt( $opt_key ) ) {
 			$option_params['class'] = 'frm_other_trigger';
 		}
-		?>
-		<option <?php FrmAppHelper::array_to_html_params( $option_params, true ); ?>>
-			<?php echo esc_html( $opt == '' ? ' ' : $opt ); ?>
-		</option>
-		<?php
+
+		if ( is_callable( 'FrmProHtmlHelper::echo_dropdown_option' ) ) {
```
Crabcyborg, line 60: @AbdiTolesa I wouldn't call a Pro function to Lite.

I think adding a lite equivalent of this function to `FrmHtmlHelper` is fine. In Pro, we can call the Lite function and eventually remove the other code there instead.

Crabcyborg: Thanks for considering the option @AbdiTolesa.

As I mentioned in the other pull request I don't think we should get rid of `FrmProHtmlHelper::echo_dropdown_option` immediately. Instead it makes sense to just use it as an interface to call this function, with the other logic as a fallback. This way we can support it backwards better, and can eventually just remove that other code.

This all gets harder to do when you put it in a view though. Views should ideally have no logic - ideally you'll have no conditions in a view at all.

This branch looks good to go as it is. Thanks for working on it!

Crabcyborg (approved):

classes/helpers/FrmHtmlHelper.php
```diff
@@ -32,4 +32,25 @@ function() use ( $id, $name, $args ) {
 			isset( $args['echo'] ) ? $args['echo'] : false
 		);
 	}
+
+	/**
+	 * Echo a dropdown option.
+	 * This is useful to avoid closing and opening PHP to echo <option> tags which leads to extra whitespace.
+	 * Avoiding whitespace saves 5KB of HTML for an international address field with a country dropdown with 252 options.
+	 *
+	 * @since 6.3
```
Crabcyborg, line 41: This since will be off though. Either `x.x` or `6.3.1` is fine to put in this case.

AbdiTolesa: @Crabcyborg Thanks for catching that, it is fixed now and I apologize for not following our new PR workflow as I learned about it lately!

garretlaxton (approved):
This looks good! I noticed my size went down a bit.

PR #1194, opened 2023-05-05
Introduce Floating Links in the Formidable Admin Panel

shervElmi:
This PR introduces the Floating Links feature to the Formidable Admin Panel, which provides users with a set of links for quick navigation to the **Upgrade**, **Support**, and **Documentation** pages.

## QA URL:
https://qa.formidableforms.com/sherv/wp-admin/admin.php?page=formidable

## Testing Instructions:
2. Check the perfect pixel implementation.
  - **Figma File:** https://www.figma.com/file/YcSWkrjVGroC1zyxhuHH6W/Floating-Links?node-id=1-117&t=Xf5EKO8Mvs1JwuFj-4
3. Verify that all functionality works well.


garretlaxton:
This functions good, but there are a few things that could be improved.

For smaller screens this can block the date columns. Not sure if that's a big deal or not.
image

One suggestion I have is when the pro version is installed we keep support and documentation separate. That way we can link to the premium support: https://formidableforms.com/new-topic/

shervElmi:
Thank you, @garretlaxton!

@Crabcyborg, could you please provide your insights on Garret's comment? I think we could consider hiding this feature on smaller screens for a better user experience.

Crabcyborg:
@shervElmi I think it's likely to get in the way a lot there.

I'm seeing it overlaps the row merging options.
Screen Shot 2023-05-24 at 9 43 28 AM

And it overlaps the visual styler toggles.
Screen Shot 2023-05-24 at 9 46 10 AM

It can overlap a lot of things.
Screen Shot 2023-05-24 at 9 46 49 AM

It makes it hard for me to identify my Stripe webhook URL,
Screen Shot 2023-05-24 at 9 47 39 AM

I wonder if we could check if it's on top of something, and hide it when it is?

What do you think @tuguirazvan? Can we do something to this floating button so it isn't just on top of everything? Should we be avoiding this on certain pages?

It also looks like it isn't on every page. It could probably go on the applications page as well (`/wp-admin/admin.php?page=formidable-applications`).

tuguirazvan:
@shervElmi @Crabcyborg I'm reviewing this right now and will come back with feedback as soon as poossible.

tuguirazvan:
@shervElmi @Crabcyborg 

I would keep the floating links on the top-level pages only: 

- Forms
- Entries
- Views
- Applications
- Import/Export
- Global Settings
- Inbox
- Add-ons
- Upgrade
- SMTP

I wouldn't show it on our form editor, not on pages like Build, Style and Settings because I don't think it will be used that much in this area since as a user you have different needs while editing/building a form. 

On the mobile breakpoint, on the other hand, I'm thinking of the following options:
1. Make it smaller (around 50x50px) and reduce the left and bottom margins, to take less of the central space of the screen. (see the image below)
2. Hide it when scrolling down and show it only when scrolling up again. This allows them to see what's going on, but still have access to the floating links. 
3. We might want to remove it from the phone breakpoint and leave it on for tablets and desktops only... 

@stephywells Would we lose much if the floating links won't show on mobile breakpoint at all? I'm concerned that it might get a bit too much in the way that doing good for the user. If they really need those options we have plenty of other ways of accessing them. Just a thought!

image

shervElmi:
@tuguirazvan, I've just implemented show or hide the floating menu based on the scroll position feature. From my perspective, it looks like this feature could enhance user experience even on desktop devices.

Could you please take a moment to evaluate this functionality on the desktop and share your feedback with me?
Here's the link for your reference:
https://qa.formidableforms.com/sherv/wp-admin/admin.php?page=formidable-applications.

tuguirazvan:
@shervElmi I tried to give it a try but it doesn't seem to work for me.

shervElmi:
> @shervElmi I tried to give it a try but it doesn't seem to work for me.

I recommend clearing your cache to resolve the issue. If you're using Chrome, I suggest utilizing the "Clear Cache" extension available [here](https://chrome.google.com/webstore/detail/clear-cache/cppjkneekbjaeellbfkmgnhonkkjfpdn).

tuguirazvan:
@shervElmi Thanks for the extension. It worked now! 
Personally, I love the idea of being able to hide the floating logo even on desktops, considering that not all people have big screen resolutions, which otherwise will be difficult for them to browse and see anything in the bottom right of the screen.

shervElmi:
> @shervElmi Thanks for the extension. It worked now! Personally, I love the idea of being able to hide the floating logo even on desktops, considering that not all people have big screen resolutions, which otherwise will be difficult for them to browse and see anything in the bottom right of the screen.

Sure, I feel the same way, Razvan! Thanks for your feedback. I'll keep this on all screens.

shervElmi (commented):
@omidakhavans, since this PR involves a package intended for use across all our plugins, I've written the code in a modular fashion to facilitate easy copying and pasting. Therefore, if you notice CSS written directly in the JS file, it's due to this approach.

js/packages/floating-links/s11-floating-links.js
```diff
@@ -0,0 +1,518 @@
...
+ * WordPress dependencies
+ */
+const { __ } = wp.i18n;
+
+/**
+ * Class representing a floating action button for showing links.
+ *
+ * @class S11FloatingLinks
+ */
+class S11FloatingLinks {
```
omidakhavans, line 11: Since I have problems using class convention in JS when it's unnecessary. I will not talk about it since it could be categorized as a fanatic conversation about approaches. but you did great on this.

js/packages/floating-links/s11-floating-links.js
```diff
@@ -0,0 +1,518 @@
...
+				<path fill="#fff" d="M0 0h40v40H0z"/>
+				</clipPath>
+			</defs>
+		</svg>
+	`
+};
+
+// Determine the appropriate links and initialize the S11FloatingLinks class
+const links = s11FloatingLinksData.proIsInstalled ? proVersionLinks : freeVersionLinks;
+new S11FloatingLinks( links, options );
```
omidakhavans, line 518: What do you think to check for dom ready before initiation?

js/packages/floating-links/s11-floating-links.js
```diff
@@ -0,0 +1,518 @@
...
+	{
+		title: __( 'Support & Docs', 'formidable' ),
+		icon: supportIcon,
+		url: 'https://wordpress.org/support/plugin/formidable/',
+		openInNewTab: true
+	}
+];
+
+// Define options
+const options = {
```
omidakhavans, line 498: For better readability I would prefer to load options etc from other files.

js/packages/floating-links/s11-floating-links.js
```diff
@@ -0,0 +1,518 @@
...
+
+/**
+ * Initialize Floating Links.
+ * Setup links and options parameters and send them to the S11FloatingLinks class.
+ *
+ * @class S11FloatingLinks
+ */
+
+// Define link icons
+const upgradeIcon = `
```
omidakhavans, line 440: Like my other comment it would be great if you call these const from other separated module.

js/packages/floating-links/s11-floating-links.js
```diff
@@ -0,0 +1,518 @@
...
+	/**
+	 * Create a new S11FloatingLinks instance.
+	 *
+	 * @constructor
+	 *
+	 * @param {Array<Object>} links - An array of link objects to display, each containing 'title', 'icon', 'url', and an optional 'openInNewTab' property.
+	 * @param {Object} options - Configuration options for the floating links component.
+	 * @param {string} options.logoIcon - The SVG icon of the plugin logo that used for the floating action button.
+	 */
+	constructor( links, options ) {
```
omidakhavans, line 22: I'm concerned about accepting arguments like this which is an object in a JS file. I know this one is safe but I don't feel safe at all with this approach however I'm not sure what is the threat but it would be great if you load your necessary data in your class without accepting any arguments.

If you prefer to make your code extendable please consider using `wp.hooks.doAction`

js/packages/floating-links/s11-floating-links.js
```diff
@@ -0,0 +1,518 @@
...
+
+	/**
+	 * Create the wrapper element and add it to the DOM.
+	 *
+	 * @memberof S11FloatingLinks
+	 */
+	createWrapper() {
+		// Create the wrapper element
+		this.wrapperElement = document.createElement( 'div' );
+		this.wrapperElement.classList.add( 's11-floating-links', 's11-fadein' );
```
omidakhavans, line 79: I'm sure we prefer to use `frm` prefix instead of s11. What do you think?

js/packages/floating-links/s11-floating-links.js
```diff
@@ -0,0 +1,518 @@
...
+			linkElement.href = link.url;
+			linkElement.title = link.title;
+
+			if ( link.openInNewTab ===  true ) {
+				linkElement.target = '_blank';
+				linkElement.rel = 'noopener noreferrer';
+			}
+
+			linkElement.innerHTML = `
+				<span class="s11-floating-links-nav-text">${link.title}</span>
```
omidakhavans, line 113: Also we have `span` in dom.js it would be great to use them in your code.

omidakhavans (commented):
@shervElmi  Thank you very much for working on this, Overall I could see this is great and code quality is great too. Thank you for utilizing modern approaches in your JS. Most of my reviews related to small enhancement and I couldn't find any big issues. Also some of my reviews remain and I will do it next time after the changes are made.

omidakhavans:
Thank you for clarifying @shervElmi. I would rather register scripts `wp_register_script` and styles in my main PHP method and enqueue them in other plugins depending on my needs. and use `wp.hooks.doAction` for expandability. I have noted them in my comments. However your approach also works great. So with that fact you could skip some of my comments or you could use @Crabcyborg 's opinion too.

classes/controllers/FrmAppController.php
```diff
@@ -638,6 +638,17 @@ public static function admin_js() {
 
 		$page = FrmAppHelper::simple_get( 'page', 'sanitize_title' );
 
+		// Enqueue Floating Links.
+		$action           = FrmAppHelper::simple_get( 'frm_action', 'sanitize_title' );
+		$is_allowed_pages = strpos( $page, 'formidable' ) === 0 && $page !== 'formidable-styles' && ! $action;
```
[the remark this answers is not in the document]
shervElmi: @omidakhavans, I greatly appreciate your kind words. Actually, we plan to utilize this package across all plugins. In order to ensure easy maintenance and updates, I've avoided using specific plugin functions. Therefore, everything has been developed to be independent of the FF plugin.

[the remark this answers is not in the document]
shervElmi: @stephywells and @Crabcyborg, I'd like to gather your insights regarding these proposed changes.

We aim to employ these codes across all S11 plugins. To facilitate this, I've crafted my code independently, using the `s11-` prefix, and have refrained from utilizing any FF methods. Furthermore, I am considering separating the PHP codes into a different file, which can then be specifically utilized within the `FrmAppController.php` file. Your feedback would be greatly appreciated.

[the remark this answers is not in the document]
Crabcyborg: @shervElmi I think it's fine to keep it self-contained. I'm okay with the s11 prefix.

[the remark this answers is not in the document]
Crabcyborg: I have some other comments on the code here though,

`$is_allowed_pages` sounds like a confusing variable name. It also seems too generic for something in the middle of a large function. It would be nicer if that was also defined outside of this function somewhere else as it isn't related to this other parts of this function.

The `strpos( $page, 'formidable' ) === 0 && $page !== 'formidable-styles' && ! $action;` check as well is confusing. I think this should be refactored to be more readable. I think Omid's suggestion of using `FrmAppHelper::is_style_editor_page` would help.

classes/controllers/FrmAppController.php
```diff
@@ -638,6 +638,17 @@ public static function admin_js() {
...
+		self::enqueue_floating_links(
+			array(
+				'plugin_url'       => $plugin_url,
```
Crabcyborg, line 646: You could use `compact` in this case to reduce the code to something more simple.

`self::enqueue_floating_links( compact( 'plugin_url', 'version', ... ) );`

js/packages/floating-links/s11-floating-links.js
```diff
@@ -0,0 +1,518 @@
...
+		});
+	}
+
+	/**
+	 * Apply styles to the component elements.
+	 *
+	 * @memberof S11FloatingLinks
+	 */
+	applyComponentStyles() {
+		const css = `
```
Crabcyborg, line 256: @shervElmi What do you think about pulling this out into a CSS file? This is a ton of CSS for JavaScript. I think it would be a lot more efficient if it could just load a cached CSS file.

js/packages/floating-links/s11-floating-links.js
shervElmi: Actually, I'm creating the DOM with JavaScript and I'm not using any PHP or HTML code to do so. I believe that without checking for DOM readiness, floating links will appear faster. However, if you think I need to implement the DOM readiness check, please let me know and I'll do so. Thank you.

classes/controllers/FrmAppController.php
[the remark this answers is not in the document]
shervElmi: @Crabcyborg, following our discussion with Razvan, I need to configure the system to display Floating Links on all parent pages, excluding the styles page. However, I've noticed that the `FrmAppHelper::is_style_editor_page` doesn't yield the condition I'm seeking.

Specifically, we need to include the floating links on the pages listed below, as Razvan has previously indicated:

> I would keep the floating links on the top-level pages only:
> 
> Forms
> Entries
> Views
> Applications
> Import/Export
> Global Settings
> Inbox
> Add-ons
> Upgrade
> SMTP

js/packages/floating-links/config.js
```diff
@@ -0,0 +1,106 @@
...
+];
+
+/**
+ * Define links for the "pro" version of the plugin
+ */
+const proVersionLinks = [
+	{
+		title: __( 'Support & Docs', 'formidable' ),
+		icon: supportIcon,
+		url: 'https://wordpress.org/support/plugin/formidable/',
```
Crabcyborg, line 76: @shervElmi I think in Pro we want to link to https://formidableforms.com/new-topic/ instead. https://wordpress.org/support/plugin/formidable/ is only the support URL for Lite.

I also don't see why this is Support & Docs. The Lite link to documentation makes sense for both Lite and Pro. I think it should always be defined the same way for both.

shervElmi: Could you please check Steve's comment about the links here?
https://app.clickup.com/t/9006050445/S11-179?comment=90060013166980

Thank you, Mike.

shervElmi: Do you still think we need to change this?

Crabcyborg: I think it's wrong.

@stephywells When Pro is active, I think we should have two links - one to our documentation, and one to our Pro support, right? Right now there's one link called "Support & Docs" that points to our Lite support, which makes no sense to me. But it does line up with Steve's comment in ClickUp.

stephywells: @Crabcyborg you're completely right. The nav items should change when pro is active. Here's a link to the design:
https://www.figma.com/file/YcSWkrjVGroC1zyxhuHH6W/Floating-Links?type=design&node-id=1-129&mode=design&t=jU40NcVMRWWhTQOf-0

In Lite, the documentation link should go to our docs, and support should go to wp.org.
In Pro, this is combined into one and it should go to our docs page.

Crabcyborg: > In Pro, this is combined into one and it should go to our docs page.

Thanks @stephywells!

@shervElmi We should be fine just changing this URL for this option to https://formidableforms.com/knowledgebase/ instead then.

shervElmi: Done, Mike!

Crabcyborg (approved):

PR #1196, opened 2023-05-10
Surveys Issue 190 - scale display options core update

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-surveys/issues/190


classes/helpers/FrmFieldsHelper.php
```diff
@@ -1928,6 +1928,14 @@ public static function show_add_field_buttons( $args ) {
...
 	 */
 	public static function show_radio_display_format( $field ) {
+		$options = self::get_display_format_options( $field );
+
+		$args = self::get_display_format_args( $field, $options );
+
+		include FrmAppHelper::plugin_path() . '/classes/views/frm-fields/back-end/radio-display-format.php';
+	}
+
+	public static function get_display_format_options( $field ) {
```
truongwp, line 1938: Please add a doc comment here, with `@since` tag.

classes/helpers/FrmFieldsHelper.php
```diff
@@ -1974,7 +1981,7 @@ public static function show_radio_display_format( $field ) {
 	 * @param array $options Options array.
 	 * @return array
 	 */
-	private static function get_display_format_args( $field, $options ) {
+	public static function get_display_format_args( $field, $options ) {
```
truongwp, line 1984: Please add a `@since` tag to note that this method is public since this version.

AbdiTolesa: @truongwp We already have a `@since` tag here, I think you might have missed that?

truongwp (approved):
The code looks good now. Thanks @AbdiTolesa!

Crabcyborg (approved):
This looks good.

I'll wait to merge this for now until the others are also approved and merged. Since a release just went out yesterday, I don't want to merge this into master immediately if anything does come up.

classes/models/fields/FrmFieldType.php
```diff
@@ -523,8 +523,16 @@ public function displayed_field_type( $field ) {
...
-			'scale'    => FrmField::is_field_type( $field, 'scale' ),
 		);
+
+		/**
+		 * @since x.x
+		 *
+		 * @param array $displayed_type
+		 * @param array $field
+		 */
+		$display_type = apply_filters( 'frm_displayed_field_type', $display_type, compact( 'field' ) );
```
Crabcyborg, line 534: @AbdiTolesa Is it possible to do this in some way that doesn't introduce a new filter?

This may be adding too much flexibility here.

Are we able to just overwrite `FrmFieldType::has_field_choices` from the `models/fields/Scale` class with a function that returns true?

css/custom_theme.css.php
```diff
@@ -914,8 +914,10 @@
 }
 
 .with_frm_style .frm_radio input[type=radio]:checked:before,
+.with_frm_style .frm_scale input[type=radio]:checked:before,
 .with_frm_style .frm_checkbox input[type=checkbox]:checked:before {
 	transform: scale(1);
+	box-shadow: inset 10px 10px var(--progress-active-bg-color);
```
Crabcyborg, line 920: @AbdiTolesa I don't love this line. `--progress-active-bg-color` is a Pro style so it shouldn't really be here in Lite code.

It should be in Pro.

css/custom_theme.css.php
```diff
@@ -875,6 +875,7 @@
 }
 
 .with_frm_style .frm_radio input[type=radio],
+.with_frm_style .frm_scale input[type=radio],
```
Crabcyborg, line 878: @AbdiTolesa Maybe we could at least check if Pro is installed first before including this?

We want the CSS to be as small as possible. People using Lite only have no use for this string, so it shouldn't be included for them.

I get why it's included here, to keep the styling consistent, but we don't really want references to scale fields in people's generated CSS files if they're only using Lite.

AbdiTolesa: That makes sense and better!

PR #1199, opened 2023-05-17
Enhancing Accessibility and Keyboard Navigation for Radio Buttons and Checkboxes

shervElmi:
This PR aims to enhance the accessibility of our `Radio Buttons` and `Checkboxes` fields, primarily addressing issues encountered with the JAWS screen reader.

Furthermore, if `Formidable PRO` and `Formidable Surveys` are activated, this PR addresses the `Dynamic`, `Product`, and `Likert Scale` fields.

The key enhancements incorporated in this PR include:

1. Optimizing the `role` attribute to enrich the semantic context and achieve alignment with ARIA standards.
2. Including `tabindex=0` in the group to allow focus via the tab key, thereby improving keyboard navigation.
3. Shifting the `aria-required` attribute from individual radio and checkbox elements to the group element to better align with ARIA guidelines.

## Related Issue:
https://github.com/Strategy11/formidable-pro/issues/4204

## Testing Instructions for Radio Buttons and Checkboxes Fields:
2. Open an existing form or create a new one, adding the `Radio Buttons` and `Checkbox` fields to the form.
3. Enable the Required setting for fields.
4. Confirm that the rendered HTML matches the provided `after` screenshots.

## Testing Instructions for Dynamic, Product, and Likert Scale Fields:
1. Activate the `Formidable PRO` and `Formidable Surveys` plugins.
2. Follow the above instructions, changing the `Display Types`/`Product Type` setting options. Confirm that the rendered HTML matches the provided `after` screenshots.

#### Before
image

#### After
image

#### Before
image

#### After
image

#### Before
image

#### After
image

#### Before
image

#### After
image

#### Before
image

#### After
image

#### Before
image

#### After
image

#### Before
image

#### After
image

#### Before
image

#### After
image

#### Before
image

#### After
image


classes/controllers/FrmFieldsController.php
```diff
@@ -753,10 +753,20 @@ public static function pull_custom_error_body_from_custom_html( $form, $field, $
 	 * @param array $add_html
 	 */
 	private static function maybe_add_html_required( $field, array &$add_html ) {
-		if ( in_array( $field['type'], array( 'file', 'data', 'lookup' ), true ) ) {
+		if ( in_array( $field['type'], array( 'file', 'nps', 'radio', 'checkbox' ), true ) ) {
 			return;
 		}
 
+		if ( in_array( $field['type'], array( 'data', 'lookup', 'product' ), true ) ) {
```
stephywells, line 760: Could the FrmField::is_checkbox() and FrmField::is_radio() functions would cover this?

classes/models/fields/FrmFieldType.php
```diff
@@ -199,8 +199,19 @@ protected function input_html() {
...
+	 * [multiple-input-attributes] acts as a placeholder for any additional attributes that an input field might require.
+	 *
+	 * @see FrmFieldFormHtml->get_html() for the function handling the dynamic replacement.
+	 *
+	 * @return string The template HTML string for a div container with multiple input fields. This string is
+	 *                prepared for dynamic replacement of the placeholders [key], [input], and [multiple-input-attributes].
+	 */
 	protected function multiple_input_html() {
-		return '<div class="frm_opt_container" aria-labelledby="field_[key]_label" role="group">[input]</div>';
+		return '<div class="frm_opt_container" aria-labelledby="field_[key]_label" tabindex="0" [multiple-input-attributes]>[input]</div>';
```
stephywells, line 214: Maybe we could leave role="group" and use it with str_replace later instead? It would be nice to avoid adding a new shortcode in the HTML, and creating a need for a big migration that would be unreliable.

stephywells (changes requested):
Great detail on this!

I think we can simplify this quite a bit to make it more readable with less room for error. Can you take a look at my comments and see what you think?

classes/models/fields/FrmFieldType.php
shervElmi: Thanks for your review, Steph!

I used `[multiple-input-attributes]` to add `aria-required` and `role`. It's the only suitable method I found, as we don't have access to the field object in the `multiple_input_html()` method.

What do you think? Should I only revert the `role="group"`, or do you believe I need to refactor some methods so that I can provide the field object in the `multiple_input_html()` method? I feel that refactoring could be riskier than adding a shortcode.

shervElmi: Another crucial point is that the `multiple_input_html()` method is only called when we create a new field. After updating the field, we don't call this method. Consequently, with the current functionality, we cannot update the role and other HTML attributes when updating a field.

classes/controllers/FrmFieldsController.php
shervElmi: I will check that if we decide to use my current code and determine that we don't need to refactor it.

classes/models/fields/FrmFieldType.php
Crabcyborg: @shervElmi I think that's the problem.

Since `multiple_input_html` is only used for `default_html` we'd need to modify existing fields that do not have default HTML. Ideally this update would work on fields that already existed, with no migrations. Migrating thousands of fields should be avoided unless totally necessary.

We'll want to modify the fields later in the process, like before it's printed. Possibly around the `FrmFIeldFormHtml::replace_shortcodes_before_input` code. We have a line `$this->html = apply_filters( 'frm_before_replace_shortcodes', $this->html, $this->field_obj->get_field(), $this->pass_args['errors'], $this->form );` that we may want to go before or after. I'm not sure at the moment if it would be more beneficial before or after but either is probably fine.

shervElmi: Following your feedback, @Crabcyborg, regarding my recent refactoring, I will proceed with writing the PHPUnit tests. :)

classes/controllers/FrmFieldsController.php
shervElmi: Thank you for your valuable suggestion, Steph. I've implemented the refactor on the `maybe_add_html_required()` method accordingly. You can view the changes at this URL:
https://github.com/Strategy11/formidable-forms/pull/1199/files

shervElmi:
@Crabcyborg, I have been working on writing PHPUnit tests for my new method, namely `add_multiple_input_attributes()`.

However, there are two points I would like to discuss where I believe we could forgo writing PHPUnit tests for this method:
1. As we currently do not have a `test_FrmFieldFormHtml.php` file, creating one would involve passing attributes to `FrmFieldFormHtml::__construct( $atts )`, establishing a `field_obj`, and so on. This process can be quite time-consuming, and it might not be the most efficient use of our resources.
2. Within `test_FrmFieldFormHtml.php`, we are checking several conditions that may be subject to future changes. For instance, the **Likert Scale** is managed in a separate file and does not use the `default_html` method. There are many similar conditional scenarios that could likely be refactored in the future.

Crabcyborg:
@shervElmi

For `#1` is it any easier to use `set_private_property` to only set what's necessary, and call functions privately? If it isn't, it's probably acceptable to skip it for now if it's it isn't worth the time cost.

I'm not I totally understand the `#2` item. Could you provide an example?

shervElmi:
> @shervElmi
> 
> For `#1` is it any easier to use `set_private_property` to only set what's necessary, and call functions privately? If it isn't, it's probably acceptable to skip it for now if it's it isn't worth the time cost.
> 
> I'm not I totally understand the `#2` item. Could you provide an example?

Thank you for your feedback, Mike.

What I'm suggesting is that establishing tests for the methods in the `FrmFieldFormHtml.php` file could prove to be time-consuming. Considering that my method is part of this file, it could potentially require a full day of work just to write these tests.

Take a look at my conditions:

image

For instance, the `Lookup` field doesn't conform to standard field patterns and doesn't utilize the `default_html` method.

image

Therefore, writing a test for a situation that isn't particularly stable could potentially lead to future complications!

Crabcyborg:
Thanks @shervElmi. I think it's safe to trust your gut on this.

We can skip the tests if it's going to take a whole day. If you can add some small partial test to cover just your update, that's fine too. We don't need to have 100% coverage all at once - we can always build up.

Like you mention, things that could break in the future may not be worth adding tests for. A test that breaks whenever new features are added is generally not a great type of test.

garretlaxton:
For some reason, I'm not seeing any of the changes on the page. Any ideas why?
image

shervElmi:
> For some reason, I'm not seeing any of the changes on the page. Any ideas why? image

I've set up a QA site, and everything seems to be working as expected.
**QA URL:** https://qa.formidableforms.com/sherv6/wp-admin/admin.php?page=formidable&frm_action=edit&id=1

garretlaxton:
Thanks @shervElmi. Does JAWS need to be installed to see these? I still don't see it on my end for some reason.

classes/models/FrmFieldFormHtml.php
```diff
@@ -484,4 +486,50 @@ private function process_wp_shortcodes() {
...
+	 */
+	private function add_multiple_input_attributes() {
+		$field      = (array) $this->field_obj->get_field();
+		$field_type = $this->field_obj->get_field_column( 'type' );
+		$attributes = array();
+
+		// Check if the field type is one of the following.
+		$is_valid_field_type = in_array( $field_type, array( 'radio', 'checkbox', 'data', 'product', 'scale' ), true );
+
+		if ( $is_valid_field_type ) {
```
Crabcyborg, line 503: @shervElmi This is a lot of code to have indented. If we just checked `if ( ! $is_valid_field_type ) {` we could exit early and outdent all of this.

shervElmi: Done. Thank you, Mike!

garretlaxton (approved):
This looks and sounds good!

tests/fields/test_FrmFieldFormHtml.php
```diff
@@ -0,0 +1,84 @@
...
+			$instance = new FrmFieldFormHtml( $atts );
+
+			$this->run_private_method( array( $instance, 'add_multiple_input_attributes' ) );
+			$html = $this->get_private_property( $instance, 'html' );
+
+			// Assert html has the correct attributes
+			$this->assertStringContainsString( 'aria-required="true"', $html );
+
+			if ( 'radio' === $field_type['type'] || 'scale' === $field_type['type'] || ( isset( $data_type ) && 'radio' === $data_type ) ) {
+				$this->assertStringContainsString( 'role="radiogroup"', $html );
```
Crabcyborg, line 78: Since the only part of this check that changes is the actual role, why not separate it out like that?

```
$use_radio_group = 'radio' === $field_type['type'] || 'scale' === $field_type['type'] || ( isset( $data_type ) && 'radio' === $data_type );
$expected_role = $use_radio_group ? 'radiogroup' : 'group';
$this->assertStringContainsString( 'role="' . $expected_role . '"', $html );
```

classes/models/FrmFieldFormHtml.php
```diff
@@ -484,4 +486,52 @@ private function process_wp_shortcodes() {
...
+			if ( 'radio' === $data_type ) {
+				$is_radio = true;
+			}
+		}
+
+		// Add 'role' attribute to the field.
+		$attributes['role'] = $is_radio ? 'role="radiogroup"' : 'role="group"';
+
+		// Add 'aria-required' attribute to the field if required.
+		if ( $is_needed_aria_required && '1' === $field['required'] ) {
```
Crabcyborg, line 529: `$is_needed_aria_required` reads funny.

I think something like `$type_requires_aria_required` reads better grammatically and explains what this variable's purpose is more.

classes/models/FrmFieldFormHtml.php
```diff
@@ -484,4 +486,52 @@ private function process_wp_shortcodes() {
...
+				$is_needed_aria_required = false;
+			}
+			// Check if data type is 'radio'
+			if ( 'radio' === $data_type ) {
+				$is_radio = true;
+			}
+		}
+
+		// Add 'role' attribute to the field.
+		$attributes['role'] = $is_radio ? 'role="radiogroup"' : 'role="group"';
```
Crabcyborg, line 526: If we just set `radiogroup` and `group` on the right side, and updated the aria-required code below we could use `FrmAppHelper::array_to_html_params` instead of the custom implode logic.

shervElmi:
Thank you so much for your excellent review, @Crabcyborg! 🙇🏻
I've just addressed the feedback.

omidakhavans (commented):
@shervElmi Thank you for working on this, I can not make the `tabindex` work! Also radio buttons don't seem to work fine without tab indexes.

classes/models/FrmFieldFormHtml.php
```diff
@@ -484,4 +486,52 @@ private function process_wp_shortcodes() {
...
+
+		if ( ! $is_valid_field_type ) {
+			return;
+		}
+
+		$is_radio = 'radio' === $field_type || 'scale' === $field_type;
+		$type_requires_aria_required = true;
+
+		// Check if the field type is 'data' or 'product'.
+		$is_data_or_product = in_array( $field_type, array( 'data', 'product' ), true );
```
omidakhavans, line 511: Dear @shervElmi  This method could splitted into two different methods.

classes/models/FrmFieldFormHtml.php
```diff
@@ -484,4 +486,52 @@ private function process_wp_shortcodes() {
...
+		if ( $is_data_or_product ) {
```
omidakhavans, line 512: @shervElmi I would prefer to implement this condition in this way.
`if ( in_array( $field_type, array( 'data', 'product' ), true ) ) `
1. No memory usage on variables.
2. Code readability.

shervElmi:
> @shervElmi Thank you for working on this, I can not make the `tabindex` work! Also radio buttons don't seem to work fine without tab indexes.

Thanks for your review, @omidakhavans.

I've attached a gif that shows everything working correctly. Could you please check it again?


I've also researched the tab index and other topics, and the current HTML looks good!

classes/models/FrmFieldFormHtml.php
shervElmi: Thank you @omidakhavans for your suggestion.

We use variables inside this condition that are defined outside. I think it's more flexible for future updates if we keep it as one method. Please share your thoughts again after reading my comment.

classes/models/FrmFieldFormHtml.php
shervElmi: Thank you @omidakhavans.

Crabcyborg (approved):
Thanks for working on this @shervElmi!

@omidakhavans I'm not sure why you're having issues tabbing through the radio buttons but this works fine for me.

I plan to release this tomorrow so I'm going to merge this now.

shervElmi:
@Crabcyborg, could you also review the related PR? I believe it's ready to merge. Thanks!
https://github.com/Strategy11/formidable-pro/pull/4249

Crabcyborg:
@shervElmi I'm looking at https://github.com/Strategy11/formidable-pro/pull/4249.

But now I see what Omid is talking about. You have a `tabindex` on your radio group there, but not here. I don't tab into the radio group with a regular radio button, but I do with lookups?

Crabcyborg:
@shervElmi I made a small pull request https://github.com/Strategy11/formidable-forms/pull/1252

What do you think?

shervElmi:
> @shervElmi I made a small pull request #1252
> 
> What do you think?

Thanks, Mike. I'm checking it out now.

PR #1200, opened 2023-05-19
make scale field dependent logic update in builder before saving form

AbdiTolesa:
Fixes  Scale field showing wrong values in Conditional logic.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -5338,12 +5338,47 @@ function frmAdminBuildJS() {
 		adjustConditionalLogicOptionOrders( fieldId );
 	}
 
-	function adjustConditionalLogicOptionOrders( fieldId ) {
+
+	function updateConditionalLogicsDependentOnThis() {
+		let that = this;
+		setTimeout( function() {
+			var fieldId, liContainer = that.closest( '.frm-single-settings' );
+
+			fieldId = liContainer.dataset['fid'];
+			if ( ! fieldId ) {
+				return;
+			}
+
+			adjustConditionalLogicOptionOrders( fieldId, 'scale' );
+		}, 0 );
+
+	}
+
+	function getScaleFieldOptions( fieldId ) {
+		let opts = [];
+		let label, saved, i, optObj;
+		const optVals = document.querySelectorAll( 'input[name^="item_meta[' + fieldId + ']"]' );
+
+		optVals.forEach( ( opt, index ) => {
+			opts.push( opt.value );
+		});
+
+		return opts;
+	}
+
+	function adjustConditionalLogicOptionOrders( fieldId, type ) {
 		var row, opts, logicId, valueSelect, rowOptions, expectedOrder, optionLength, optionIndex, expectedOption, optionMatch,
 			rows = document.getElementById( 'frm_builder_page' ).querySelectorAll( '.frm_logic_row' ),
 			rowLength = rows.length,
-			fieldOptions = getFieldOptions( fieldId ),
-			optionLength = fieldOptions.length;
+			fieldOptions, optionLength;
+
+		if ( type && 'scale' === type ) {
+			fieldOptions = getScaleFieldOptions( fieldId );
+		} else {
+			fieldOptions = getFieldOptions( fieldId );
+		}
+
+		optionLength = fieldOptions.length;
 
 		for ( rowIndex = 0; rowIndex < rowLength; rowIndex++ ) {
 			row = rows[ rowIndex ];
@@ -9913,6 +9948,7 @@ function frmAdminBuildJS() {
 			$builderForm.on( 'focusin', '.frm_single_option input[type=text]', maybeClearOptText );
 			$builderForm.on( 'click', '.frm_add_opt', addFieldOption );
 			$builderForm.on( 'change', '.frm_single_option input', resetOptOnChange );
+			$builderForm.on( 'change', '.frm_scale_opt', updateConditionalLogicsDependentOnThis );
 			$builderForm.on( 'change', '.frm_image_id', resetOptOnChange );
 			$builderForm.on( 'change', '.frm_toggle_mult_sel', toggleMultSel );
 			$builderForm.on( 'focusin', '.frm_classes', showBuilderModal );


js/formidable_admin.js
```diff
@@ -5338,12 +5338,43 @@ function frmAdminBuildJS() {
 		adjustConditionalLogicOptionOrders( fieldId );
 	}
 
-	function adjustConditionalLogicOptionOrders( fieldId ) {
-		var row, opts, logicId, valueSelect, rowOptions, expectedOrder, optionLength, optionIndex, expectedOption, optionMatch,
+
+	function updateConditionalLogicsDependentOnThis( e ) {
+		setTimeout( function() {
+			let fieldId = e.target.closest( '.frm-single-settings' ).dataset['fid'];
```
truongwp, line 5344: There is an eslint error on this line. You should use `dataset.fid`.

js/formidable_admin.js
```diff
@@ -5338,12 +5338,43 @@ function frmAdminBuildJS() {
 		adjustConditionalLogicOptionOrders( fieldId );
 	}
 
-	function adjustConditionalLogicOptionOrders( fieldId ) {
-		var row, opts, logicId, valueSelect, rowOptions, expectedOrder, optionLength, optionIndex, expectedOption, optionMatch,
+
+	function updateConditionalLogicsDependentOnThis( e ) {
+		setTimeout( function() {
```
truongwp, line 5343: Can you explain why we need `setTimeout()` here? Does this code need to run after something?

js/formidable_admin.js
```diff
@@ -5338,12 +5338,43 @@ function frmAdminBuildJS() {
...
+		return opts;
+	}
+
+	function adjustConditionalLogicOptionOrders( fieldId, type ) {
+		var row, opts, logicId, valueSelect, optionLength, optionIndex, expectedOption, optionMatch,
 			rows = document.getElementById( 'frm_builder_page' ).querySelectorAll( '.frm_logic_row' ),
 			rowLength = rows.length,
-			fieldOptions = getFieldOptions( fieldId ),
-			optionLength = fieldOptions.length;
+			fieldOptions, optionLength;
```
truongwp, line 5369: We should move these 2 variables above the `rows = `. Variables without assignments should be declared first, before the assigned ones. This makes the code easier to read.

js/formidable_admin.js
AbdiTolesa: Yeah this code needed to run after the code that updates the radio buttons in the scale field after adjusting the range, which is from another script.

truongwp:
@AbdiTolesa When I change the Scale field to have other options, the old options aren't cleared in the Conditional Logic of other fields. See the screencast below, the option `2` should be deleted after I changing the Scale field.


https://github.com/Strategy11/formidable-forms/assets/19748318/818c15d9-3589-4589-bc88-6dcdf6b26a27

AbdiTolesa:
> @AbdiTolesa When I change the Scale field to have other options, the old options aren't cleared in the Conditional Logic of other fields. See the screencast below, the option `2` should be deleted after I changing the Scale field.
> 
>  Screen.Recording.2023-05-26.at.13.46.02.mov

@truongwp I have noticed that even in the main branch. It looks also the issue is not just limited to Scale fields but also other option fields have this bug. I have logged this separately a few days ago (https://github.com/Strategy11/formidable-pro/issues/4254)

truongwp:
@AbdiTolesa Ok. Thank you for doing that.

js/formidable_admin.js
```diff
@@ -5338,12 +5338,42 @@ function frmAdminBuildJS() {
...
+
+		optVals.forEach( opt => {
+			opts.push( opt.value );
+		});
+
+		return opts;
+	}
+
+	function adjustConditionalLogicOptionOrders( fieldId, type ) {
+		var row, opts, logicId, valueSelect, optionLength, optionIndex, expectedOption, optionMatch, fieldOptions, optionLength,
```
truongwp, line 5366: I see `optionLength` is declared twice.

AbdiTolesa: Oops! Just removed that!

truongwp (approved):
The code is good to me. 

For the `getScaleFieldOptions()` function, we can use min/max/step to get the options array. But I think Abdi's solution is also fine.

Thank you @AbdiTolesa!

AbdiTolesa:
@Crabcyborg I have asked Garret to test this in the other branch in Pro and this is tested along that.

css/formidableforms1.css
```diff
@@ -0,0 +1,2 @@
+/* WARNING: Any changes made to this file will be lost when your Formidable settings are updated */
```
Crabcyborg, line 1: @AbdiTolesa It looks like this file was committed accidentally.

js/formidable_admin.js
```diff
@@ -5338,12 +5338,42 @@ function frmAdminBuildJS() {
 		adjustConditionalLogicOptionOrders( fieldId );
```
Crabcyborg, line 5338: Is it a problem at all that `adjustConditionalLogicOptionOrders` is called here without a `type` parameter?

And is it possible to move more of this into the Pro `builder.js` file? Scale fields are a Pro feature so it would be nice if we could have the fix in Pro instead.

AbdiTolesa: @Crabcyborg I have not found an issue with this line. But your suggestion is really worth it, I have moved more of this code here to `builder.js`

js/formidable_admin.js
```diff
@@ -5338,12 +5338,13 @@ function frmAdminBuildJS() {
...
-		var row, opts, logicId, valueSelect, rowOptions, expectedOrder, optionLength, optionIndex, expectedOption, optionMatch,
+	function adjustConditionalLogicOptionOrders( fieldId, type ) {
+		var row, opts, logicId, valueSelect, optionLength, optionIndex, expectedOption, optionMatch, fieldOptions,
 			rows = document.getElementById( 'frm_builder_page' ).querySelectorAll( '.frm_logic_row' ),
-			rowLength = rows.length,
-			fieldOptions = getFieldOptions( fieldId ),
-			optionLength = fieldOptions.length;
+			rowLength = rows.length;
+
+		fieldOptions = wp.hooks.applyFilters( 'frm_field_options', getFieldOptions( fieldId ), { type, fieldId });
```
Crabcyborg, line 5346: `frm_field_options` sounds too generic for what this is. I think we should mention this is only for conditional logic options.

Maybe `frm_conditional_logic_field_options`?

Crabcyborg:
@AbdiTolesa So I merged https://github.com/Strategy11/formidable-pro/pull/4253 but I didn't test this part of it. Is it meant to update the conditional logic options dynamically as soon as I update my scale field options? Or what is this update doing?

It looks like my options don't update.
Screen Shot 2023-07-04 at 5 10 30 PM

AbdiTolesa:
> It looks like my options don't update.

@Crabcyborg Is that both when increasing and decreasing the options? The options don't update if you are removing existing options for all option fields actually (I have logged this in https://github.com/Strategy11/formidable-pro/issues/4254) but they update if you add more options except Scale field before this update. This update is just meant to at least make sure Scale field is consistent with other option fields so that adding more options would also update the dependent conditional logic field options. When the issue I logged is addresses, it would fix the same issue but when removing options from the fields.

Crabcyborg:
Thanks @AbdiTolesa. I thought I tried that too, but I just tried again and at least that part does appear to work.

Crabcyborg (approved):

🚀

PR #1201, opened 2023-05-20
Remove Back Button from 'Leave your email address' Modal

shervElmi:
This PR involves the removal of the back button from the 'Leave your email address' modal. The proposed changes are part of the Interface Reviews conducted by Razvan.

## QA URL:
https://qa.formidableforms.com/sherv3/wp-admin/admin.php?page=formidable&triggerNewFormModal=1&free-templates=1

## Testing Instructions:
1. Navigate to `WP Admin > Formidable > Inbox`.
2. Click on the `Get Now` button in the `Get 10+ Free Form Templates` message.
3. Verify that the back button has been removed.

### Before
image

### After
image

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -2652,7 +2652,7 @@ li .frm-hover-icons .frm-preview-form,
 	display: none;
 }
 
-#frm_new_form_modal:not([frm-page="create"]):not([frm-page="email"]) .frm-modal-back,
+#frm_new_form_modal:not([frm-page="create"]) .frm-modal-back,
 #frm_new_form_modal[frm-page="create"] #frm-create-title,
 #frm-create-title[frm-type="template"] span[frm-type="template"],
 #frm_new_form_modal[frm-page="preview"] #frm-preview-title,
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -2652,7 +2652,7 @@ li .frm-hover-icons .frm-preview-form,
 	display: none;
 }
 
-#frm_new_form_modal:not([frm-page="create"]) .frm-modal-back,
+#frm_new_form_modal:not([frm-page="create"]):not([frm-page="email"]) .frm-modal-back,
 #frm_new_form_modal[frm-page="create"] #frm-create-title,
 #frm-create-title[frm-type="template"] span[frm-type="template"],
 #frm_new_form_modal[frm-page="preview"] #frm-preview-title,
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -8896,6 +8896,13 @@ function frmAdminBuildJS() {
 				if ( firstLockedTemplate.length ) {
 					showFreeTemplatesForm( firstLockedTemplate );
 				}
+
+				// Hides the back button in the Free Template Modal and shows it when the cancel button is clicked
+				$modalBackButton = $modal.find( '.frm-modal-back' );
+				$modalBackButton.hide();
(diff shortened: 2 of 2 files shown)


Crabcyborg:
@tuguirazvan What is the purpose of removing the option to go back when you click on a locked template?

I'm not sure if I agree with this decision. This means that you need to close the modal and then go to add a new form again to get back.

tuguirazvan:
@Crabcyborg Will review it tomorrow when I'm back from the holidays. Pretty sure I had a reason.

garretlaxton (approved):
This looks good, but we may want to remove this from other premium modals to  make this consistent.
image

shervElmi:
> This looks good, but we may want to remove this from other premium modals to make this consistent. image

@Crabcyborg, what's your take on Garret's comment?

tuguirazvan:
@shervElmi @Crabcyborg Can you guys wait until tomorrow to merge this? It may require a custom logic here because this modal can be triggered from different locations. In some of them, the back button doesn't make sense.

tuguirazvan:
That's the reasons of why I want it to be removed, but only in some places. Will get back to you asap tomorrow. I'm on my way back to home.

Crabcyborg:
Thanks for clearing that up @tuguirazvan!

That makes way more sense. We can just add a check for that.

shervElmi:
> @shervElmi @Crabcyborg I recorded a quick loom showing you what I was thinking by removing the back arrow. Take a look here: https://www.loom.com/share/7078132a44224f0694e5d9e17f7a46d6
> 
> The bottom line is that the same component you have (I suppose is the same) gets triggered from different locations:
> 
> **1. From the "Inbox" screen**
> 
> Here, the modal should not contain the back icon, because you have nowhere to go back, therefore it makes sense to have only cancel and "get code". Moreover, I am not very sure if it is ok to redirect them to the "Forms" page after they have accessed "get now". I would rather open the modal on the same page, i.e. on "Inbox" so that the whole experience happens in one place.
> 
> **2. While clicking on a form template which is not available on my current plan.**
> 
> Here, the back button makes sense because you have accessed a premium option, and the possibility to go back is useful in this case and I would not remove it.
> 
> Hope it makes more sense now?
> 
> image

@tuguirazvan, I truly appreciate your detailed clarification. You're truly exceptional! 😎

Crabcyborg:
@shervElmi It looks like this hasn't changed since the one commit.

I think you're missing one use case here with the email page. It can happen the same way as the upgrade page. If you don't have Pro active, we offer a few free templates once you provide your email and enter the code that gets sent to your inbox.

c3pT75tOkh

When we're hiding the back button, we'd want it to be hidden only when the email page is triggered from the inbox. Not when it's triggered the same way as the upgrade page is.

Crabcyborg:
Thanks @shervElmi!

It may make sense if the cancel option on the email page doesn't go back to the new form page if you're sent there from the inbox as well?

dBv9k0MOIn

shervElmi:
> Thanks @shervElmi!
> 
> It may make sense if the cancel option on the email page doesn't go back to the new form page if you're sent there from the inbox as well?
> 
> dBv9k0MOIn

Thank you, Mike, for your review. I agree with your perspective. Accordingly, I'll ensure that the modal closes when the cancel button is clicked.

Crabcyborg (approved):
Thanks for updating this @shervElmi! It looks good to me now! 🚀

And thanks @garretlaxton @omidakhavans and @tuguirazvan for helping with this as well! 🥳

PR #1208, opened 2023-05-30
maybe clear custom validity message on input value change

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4267

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1348,6 +1348,47 @@ function frmFrontFormJS() {
 		});
 	}
 
+	function shouldUpdateValidityMessage( target ) {
+		if ( 'INPUT' !== target.nodeName ) {
+			return false;
+		}
+
+		if ( ! target.dataset.invmsg ) {
+			return false;
+		}
+
+		if ( 'text' !== target.getAttribute( 'type' ) ) {
+			return false;
+		}
+
+		if ( target.classList.contains( 'frm_verify' ) ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	function maybeClearCustomValidityMessage( event, field, fieldId, e ) {
+		var target = field,
+			isInvalid = false;
+
+		if ( ! shouldUpdateValidityMessage( field ) ) {
+			return;
+		}
+
+		for ( let key in target.validity ) {
+			if ( 'customError' === key ) {
+				continue;
+			}
+			if ( 'valid' !== key && target.validity[ key ] === true ) {
+				isInvalid = true;
+			}
+		}
+		if ( ! isInvalid ) {
+			target.setCustomValidity( '' );
+		}
+	}
+
 	function setCustomValidityMessage() {
 		var forms, length, index;
 
@@ -1360,19 +1401,7 @@ function frmFrontFormJS() {
 				function( event ) {
 					var target = event.target;
 
-					if ( 'INPUT' !== target.nodeName ) {
-						return;
-					}
-
-					if ( ! target.dataset.invmsg ) {
-						return;
-					}
-
-					if ( 'text' !== target.getAttribute( 'type' ) ) {
-						return;
-					}
-
-					if ( target.classList.contains( 'frm_verify' ) ) {
+					if ( ! shouldUpdateValidityMessage( target ) ) {
 						return;
 					}
 
@@ -1423,6 +1452,7 @@ function frmFrontFormJS() {
 
 			jQuery( document ).on( 'frmAfterAddRow', setCustomValidityMessage );
 			setCustomValidityMessage();
+			jQuery( document ).on( 'frmFieldChanged', maybeClearCustomValidityMessage );
 		},
 
 		getFieldId: function( field, fullID ) {


AbdiTolesa:
@garretlaxton The steps to reproduce this are listed clearly in the issue (https://github.com/Strategy11/formidable-pro/issues/4267#issue-1726118353). The form should be submitted once the regex validation error is gone with this update.

garretlaxton (approved):
This worked great for me!

js/formidable.js
```diff
@@ -1361,6 +1361,47 @@ function frmFrontFormJS() {
...
+	}
+
+	function maybeClearCustomValidityMessage( event, field ) {
+		var isInvalid = false;
+
+		if ( ! shouldUpdateValidityMessage( field ) ) {
+			return;
+		}
+
+		Object.keys( field.validity ).forEach( function( key ) {
```
truongwp, line 1391: We can use `every()` instead of `forEach()` here, it helps break the loop when `isInvalid` is `true`, which is enough for the latter check.

js/formidable.js
```diff
@@ -1361,6 +1361,47 @@ function frmFrontFormJS() {
...
+			if ( 'customError' === key ) {
```
truongwp, line 1392: I'm not sure why `customError` is handled differently. In this case, if `field.validity.customError = true`, `isInvalid` isn't set to `true`. Can you explain or give me the case that needs it?

AbdiTolesa: @truongwp I think `customError` is always true as long as a custom validity message is set with `setCustomValidity()` function and checking that isn't relevant here.

js/formidable.js
AbdiTolesa: @truongwp I don't think we can use `every` for non array-like objects, which is returned from `Object.keys`. Have you used that before or know of similar use case in such scenario?

truongwp: We will use `every()` on array: `Object.keys( field.validity ).every()`. It's almost the same as `forEach()`, but we can break the loop when we need (when `isInvalid` is `true`). See more at https://www.w3schools.com/jsref/jsref_every.asp

js/formidable.js
truongwp: Are there any cases that the `customError` is `true` but the field is valid?

js/formidable.js
AbdiTolesa: I think the issue I tried to mention is that `Object.keys( field.validity )` doesn't return an Array or Array-like

truongwp: I have just checked. `field.validity` is a `ValidateState` object: 

`ValidityState {valueMissing: false, typeMismatch: false, patternMismatch: false, tooLong: false, tooShort: false, …}`

But `Object.keys( field.validity )` is an empty array. So I don't think `forEach()` actually runs here.

truongwp: `for ( var x in field.validity )` can loop through the `ValidityState` object.

js/formidable.js
AbdiTolesa: Oh, shoot! I used `for..in` at first but I don't even remember why I replaced it in this commit (https://github.com/Strategy11/formidable-forms/pull/1208/commits/fcbc48b86a9b163ae0d1ab6636bea59ffec9a02c). I have fixed that now. Thanks!

AbdiTolesa: > Are there any cases that the `customError` is `true` but the field is valid?

Yes, as long as there is a custom validity set.

truongwp: Instead of looping through this, can we just check `field.validity.valid`? I see that if I set a custom validity, or any other validities to `true`, the `validity.valid` is changed to `false` automatically.

truongwp: Oh. Ignore my last comment. I got it now.

truongwp:
@AbdiTolesa Thanks for working on this. Your solution solves the issue. 

I wonder if we can just need to clear the `customError` on field change. The `validityState` is updated on input change. Look like the code in `change` event runs before the `validityState` is updated. I think this way will be more natural, if it solves the problem here.

@Crabcyborg What do you think?

js/formidable.js
AbdiTolesa: > Are there any cases that the `customError` is `true` but the field is valid?

Yes, as long as there is a custom validity set.

> 

Is it possible to update the `customError` in the ValidityState? It says that is readonly in the MDN page.

AbdiTolesa:
> I wonder if we can just need to clear the `customError` on field change.

Is it possible to update the `customError` in the ValidityState? It says that is readonly in the MDN page.

Crabcyborg:
> What do you think?

@truongwp It looks like it's triggered on `frmFieldChanged` events already, which happens on change events. What are you picturing changing?

truongwp:
@AbdiTolesa @Crabcyborg I think we don't need to run this `for`: https://github.com/Strategy11/formidable-forms/pull/1208/files#diff-78fbedd7c685979d81b5d742a47782f812c44f7474dd258f6835ea074ccec165R1392

We just need to call `setCustomValidity( '' )` there, to clear any custom validity we set before. The custom validity is still added back in the `invalid` event if field is invalid.

AbdiTolesa:
> @AbdiTolesa @Crabcyborg I think we don't need to run this `for`: https://github.com/Strategy11/formidable-forms/pull/1208/files#diff-78fbedd7c685979d81b5d742a47782f812c44f7474dd258f6835ea074ccec165R1392
> 
> We just need to call `setCustomValidity( '' )` there, to clear any custom validity we set before. The custom validity is still added back in the `invalid` event if field is invalid.

@truongwp Wouldn't that cause weird bugs if someone wants to hook into `frmFieldChanged` after our code and access the Validity text for some reason, if we clear the custom validity while the field is still invalid?

truongwp (approved):
@AbdiTolesa Thank you. Your explain makes sense. This looks good to me.

Crabcyborg (approved):
This looks good. Thanks @AbdiTolesa @truongwp and @garretlaxton!

🚀

PR #1213, opened 2023-06-01
use admin_enqueue_scripts hook instead of admin_init

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4273

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -160,8 +160,7 @@ class FrmAppController {
 	 * @return void
 	 */
 	public static function load_wp_admin_style() {
-		_deprecated_function( __METHOD__, '6.x', __CLASS__ . '::admin_enqueue_scripts' );
-		self::admin_enqueue_scripts();
+		FrmAppHelper::load_font_style();
 	}
 
 	/**
@@ -726,7 +725,7 @@ class FrmAppController {
 	 * @return void
 	 */
 	public static function admin_enqueue_scripts() {
-		FrmAppHelper::load_font_style();
+		self::load_wp_admin_style();
 		self::maybe_force_formidable_block_on_gutenberg_page();
 	}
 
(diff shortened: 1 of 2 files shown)


Crabcyborg:
Thanks @garretlaxton.

I don't know exactly what is required to replicate the headers sent issues. If you can't replicate the issue, it's fine to just confirm nothing is weird with the auto-scroll when you inject a form into an existing page using the embed modal.

garretlaxton (approved):
Thanks! This still embeds as it should!
image

classes/controllers/FrmHooksController.php
```diff
@@ -117,6 +117,7 @@ public static function load_admin_hooks() {
 		add_action( 'admin_enqueue_scripts', 'FrmAppController::load_wp_admin_style' );
 		add_action( 'admin_notices', 'FrmAppController::pro_get_started_headline' );
 		add_action( 'admin_init', 'FrmAppController::admin_init', 11 );
+		add_action( 'admin_enqueue_scripts', 'FrmAppController::admin_enqueue_scripts' );
```
Crabcyborg, line 120: @AbdiTolesa

Could we avoid adding a second hook?

There's already a `add_action( 'admin_enqueue_scripts', 'FrmAppController::load_wp_admin_style' );` line slightly above this.

We should add one action that handles both of these functions to avoid some overhead on the hooks.

classes/controllers/FrmAppController.php
```diff
@@ -160,7 +160,8 @@ private static function is_white_page() {
 	 * @return void
 	 */
 	public static function load_wp_admin_style() {
-		FrmAppHelper::load_font_style();
+		_deprecated_function( __METHOD__, '6.x', __CLASS__ . '::admin_enqueue_scripts' );
```
Crabcyborg, line 163: It looks like this function is still being called on https://github.com/Strategy11/formidable-forms/blob/3e2eb8eac63af29b84fa53b67fc1873ee63c7d5f/classes/controllers/FrmElementorController.php#L32

There's also a `test_load_wp_admin_style` unit test. Maybe we should rename it.

AbdiTolesa: I think it is better not to deprecate the function. I just called it from the new function.

PR #1214, opened 2023-06-02
Fix PHP 8.2 deprecations

Crabcyborg:
I'm just adding a new factory class that extends the WordPress one so I can define the factory properties to bypass the dynamic property deprecation messages.

Since it's working in PHP 8.2 I added a few more workflow PHP versions so we're covering PHP 8.0, 8.1, and 8.2.

> Deprecated: Creation of dynamic property WP_UnitTest_Factory::$form is deprecated in .../formidable/tests/base/FrmUnitTest.php on line 57
> Deprecated: Creation of dynamic property WP_UnitTest_Factory::$field is deprecated in .../formidable/tests/base/FrmUnitTest.php on line 58
> Deprecated: Creation of dynamic property WP_UnitTest_Factory::$entry is deprecated in .../formidable/tests/base/FrmUnitTest.php on line 59

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -23,6 +23,12 @@ jobs:
           - php: 8.0
             wordpress: master
             multisite: 0
+          - php: 8.1
+            wordpress: master
+            multisite: 0
+          - php: 8.2
+            wordpress: master
+            multisite: 0
 
     name: PHP ${{ matrix.php }} tests in WP ${{ matrix.wordpress }}
     steps:
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -18,16 +18,13 @@ jobs:
       matrix:
         include:
           - php: 7.4
-            wordpress: master
+            wordpress: trunk
             multisite: 0
           - php: 8.0
-            wordpress: master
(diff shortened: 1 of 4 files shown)


Crabcyborg:
@stephywells Do you see anything wrong with this approach to avoiding dynamic properties on the factory object?

stephywells:
@Crabcyborg Does this break the WP elements in the factory? Like $this->factory->user?

Crabcyborg:
> Does this break the WP elements in the factory? Like $this->factory->user?

@stephywells It appears to be fine. I wondered the same thing. I believe this all just gets set in the constructor, which still gets called.

The code normally for setting the factory is pretty simple. It's just a singleton with no parameters.
```
protected static function factory() {
	static $factory = null;
	if ( ! $factory ) {
		$factory = new WP_UnitTest_Factory();
	}
	return $factory;
}
```

I tried creating a user with the user factory after I set the factory with:
```
$user_id = $this->factory->user->create();
```

And I got a user id back.

I've run this against Pro and Stripe unit tests, and they pass.

stephywells (approved):
Thanks for checking! Sounds good to me!

Crabcyborg:
Thanks @stephywells!

PR #1215, opened 2023-06-02
make sure js validation is consistent with server side validation

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/3488

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -395,10 +395,8 @@ function frmFrontFormJS() {
 		}
 
 		options = field.querySelectorAll( 'option' );
-		if ( options.length ) {
-			if ( options.length === 1 && options[0].value === '' ) {
-				return false;
-			}
+		if ( options.length === 1 && options[0].value === '' ) {
+			return false;
 		}
 
 		return true;
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -389,8 +389,13 @@ function frmFrontFormJS() {
 	}
 
 	function shouldAddRequiredValidationError( fieldID, errors, field ) {
-		var options = field.querySelectorAll( 'option' );
-		if ( ! ( fieldID in errors ) && options.length ) {
+		var options;
+		if ( fieldID in errors ) {
+			return false;
+		}
+
+		options = field.querySelectorAll( 'option' );
+		if ( options.length ) {
 			if ( options.length === 1 && options[0].value === '' ) {
 				return false;
 			}
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -306,7 +306,7 @@ function frmFrontFormJS() {
 	}
 
 	function checkRequiredField( field, errors ) {
-		var checkGroup, options, tempVal, i, placeholder,
+		var checkGroup, tempVal, i, placeholder,
 			val = '',
 			fieldID = '',
 			fileID = field.getAttribute( 'data-frmfile' );
@@ -380,15 +380,23 @@ function frmFrontFormJS() {
 				fieldID = getFieldId( field, true );
 			}
 
-			options = field.querySelectorAll( 'option' );
-			if ( ! ( fieldID in errors ) && options.length ) {
-				if ( ! ( options.length === 1 && options[0].value === '' ) ) {
-					errors[ fieldID ] = getFieldValidationMessage( field, 'data-reqmsg' );
-				}
+			if ( shouldAddRequiredValidationError( fieldID, errors, field ) ) {
+				errors[ fieldID ] = getFieldValidationMessage( field, 'data-reqmsg' );
 			}
 		}
 
 		return errors;
+	}
+
+	function shouldAddRequiredValidationError( fieldID, errors, field ) {
+		var options = field.querySelectorAll( 'option' );
+		if ( ! ( fieldID in errors ) && options.length ) {
+			if ( options.length === 1 && options[0].value === '' ) {
+				return false;
+			}
+		}
+
+		return true;
 	}
 
 	function isSignatureField( field ) {
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -306,7 +306,7 @@ function frmFrontFormJS() {
 	}
 
 	function checkRequiredField( field, errors ) {
-		var checkGroup, tempVal, i, placeholder,
+		var checkGroup, options, tempVal, i, placeholder,
 			val = '',
 			fieldID = '',
 			fileID = field.getAttribute( 'data-frmfile' );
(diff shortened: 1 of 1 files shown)


garretlaxton:
I'm still able to submit the form for this.

AbdiTolesa:
> I'm still able to submit the form for this.

@garretlaxton This one is not to be tested independently as you have tested the other part and left a comment for js based form validation (https://github.com/Strategy11/formidable-pro/pull/4278#issuecomment-1572608120), that is just a followup. Could you test this checking out both branches: this branch and https://github.com/Strategy11/formidable-pro/pull/4278?

garretlaxton:
It works fine when `Validate this form with javascript` is selected, but I can still submit a blank form when this isn't selected.

AbdiTolesa:
> It works fine when `Validate this form with javascript` is selected, but I can still submit a blank form when this isn't selected.

It is allowed to submit an empty option as long as there is no options in the required lookup field in the repeater field, like the following example:
image

Can you attach a screencast here if you are able to submit empty value for the required lookup field while it has options please?

garretlaxton (approved):
Ahh, okay that explains it. This looks good then!

js/formidable.js
```diff
@@ -379,14 +379,26 @@ function frmFrontFormJS() {
...
+			if ( shouldAddRequiredValidationError( fieldID, errors, field ) ) {
 				errors[ fieldID ] = getFieldValidationMessage( field, 'data-reqmsg' );
 			}
 		}
 
 		return errors;
 	}
 
+	function shouldAddRequiredValidationError( fieldID, errors, field ) {
+		var options = field.querySelectorAll( 'option' );
```
truongwp, line 392: I think we should check `fieldID in errors` before this to exit earlier.

js/formidable.js
```diff
@@ -379,14 +379,31 @@ function frmFrontFormJS() {
...
 	}
 
+	function shouldAddRequiredValidationError( fieldID, errors, field ) {
+		var options;
+		if ( fieldID in errors ) {
+			return false;
+		}
+
+		options = field.querySelectorAll( 'option' );
+		if ( options.length ) {
```
truongwp, line 398: We don't need this check because we check `options.length === 1` below.

AbdiTolesa: Thanks, I overlooked that indeed!

truongwp (approved):
Thank @AbdiTolesa. This looks good to me now!

js/formidable.js
```diff
@@ -379,14 +379,29 @@ function frmFrontFormJS() {
...
 	}
 
+	function shouldAddRequiredValidationError( fieldID, errors, field ) {
+		var options;
+		if ( fieldID in errors ) {
+			return false;
+		}
+
+		options = field.querySelectorAll( 'option' );
+		if ( options.length === 1 && options[0].value === '' ) {
```
Crabcyborg, line 398: @AbdiTolesa I don't think we want this check for all dropdowns.

A regular dropdown field does not have this exception. The PHP rejects a required dropdown with just an empty option. The JS also catches it, but not with this update.

8fKDJI8v3H

Crabcyborg: Ideally we'd only have this check for a lookup, and it would only be in Pro code.

AbdiTolesa: @Crabcyborg I tried both the PHP and JS validations for a dropdown with empty option and it looks to me that this update hasn't changed that behaviour. The validation error would get set in both cases. Not sure why you are seeing a different behaviour.

I also tried to see if I can add a check for the field type but I couldn't figure out how to do that also I didn't find an example code. Regarding moving this to Pro, I was thinking to use new hooks to achieve that but it looks we are not using hooks in the front end js and I lack confidence to try that now.

Crabcyborg: I'm not sure why you're having issues replicating this. It just requires a form with JavaScript validation enabled and a required dropdown with just a blank option.

I set up an example on my QA site,

_Edit link_ https://qa.formidableforms.com/mike2/wp-admin/admin.php?page=formidable&frm_action=edit&id=326
_Preview link_ https://qa.formidableforms.com/mike2/wp-admin/admin-ajax.php?action=frm_forms_preview&form=requireddropdownexample

This is the master branch. You can see that the JavaScript validation doesn't allow a submission for my regular dropdown field,
AcKpLyTS9Z

However, in this example, I have this branch checked out. Now it submits and fails server-side instead. The "There was a problem with your submission. Errors are marked below." message gets added, which doesn't happen when validation validation catches it early.
oD2dUukRko

AbdiTolesa: I see it now @Crabcyborg. I was having similar issue yesterday with another branch and it was a caching plugin that I didn't realise was turned on. Giving this a try now make the difference apparently clear. I was thinking to check the field type but I couldn't figure out how to do that in the js file, if possible. I looked around in the same file for a similar check but no luck yet. I'm thinking to add a data attribute to the lookup fields to differentiate them from other dropdown if that sounds good to you as well.

Crabcyborg: @AbdiTolesa I'm not sure if it's worth trying to make an update for this. It seems problematic to allow an empty value for a required field even if there is no possible value to submit. I don't think this impacts a lot of people as it is.

If we were to try something, we could possibly try overwriting `frmFrontForm.getAjaxFormErrors` in Pro, and then running additional post-processing in Pro. Something like:

```
var liteGetAjaxFormErrors = frmFrontForm.getAjaxFormErrors;
frmFrontForm.getAjaxFormErrors = function( object ) {
    var jsErrors = liteGetAjaxFormErrors( object );
    // Modify the jsErrors here.
    return jsErrors;
}
```

As for confirming if a field is a lookup, I've done this in conversational forms. There's a function in `chat.js` you could look at, `isLookupList`. It determines if a field is a lookup field by checking the `__FRMLOOKUP` global data.

AbdiTolesa: Good to know about `isLookupList` function. This update was just supposed to make the js and php validations consistent as we are allowing empty value for Lookup fields if there is no options ( see `FrmProFieldLookup::maybe_remove_error`). Should we probably change the PHP side validations so that an error is triggered if empty value is submitted whether there are options or not?

Crabcyborg: I'd probably leave that alone, to avoid breaking things.
It just doesn't seem worth it to add the check as well in JavaScript since `FrmProFieldLookup::maybe_remove_error` feels more like a patch solution.

For now I think we could just close this, and not worry about making an update.

AbdiTolesa: Thanks!

Crabcyborg:
I'm closing this for now to get it off of the list. I don't think we need this update.

PR #1216, opened 2023-06-02
Avoid PHP Notice that get_current_screen is called incorrectly when u…

Crabcyborg:
…sing WooCommerce

Fixes https://secure.helpscout.net/conversation/2261261948/163616/

> Function get_current_page was called incorrectly. Current page retrieval should be called on or after the `current_screen` hook. Please see [Debugging in WordPress](https://wordpress.org/documentation/article/debugging-in-wordpress/) for more information. (This message was added in version 0.16.0.)


garretlaxton (approved):
This looks good! I don't get any notice in my logs with this update.

Crabcyborg:
Thanks @garretlaxton!

PR #1217, opened 2023-06-05
Add a modal link in the blocks

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/4281

Screencast: 

https://github.com/Strategy11/formidable-forms/assets/19748318/1e3dd24a-3897-4b7e-83b6-1f229cfff6ef

@stephywells Please check the Modal addon URL. I'm not sure that's what you mentioned.


js/src/form/modal.js
```diff
@@ -0,0 +1,78 @@
...
+		</svg>
+	},
+	category: 'widgets',
+	keywords: [
+		'modal',
+		'formidable',
+	],
+
+	edit: ( { setAttributes, attributes } ) => {
+		const hasAccess = -1 !== [ 'elite', 'business', 'personal', 'grandfathered' ].indexOf( formidable_form_selector.licenseType );
```
Crabcyborg, line 36: @truongwp Instead of doing a license type check ourselves it may be better to just leave that to the API and get what is available to us.

We can use `FrmFormApi` and use the data from `get_api_info` at the bootstrap modal ID key (`1850131`). If the `url` key is set and not empty, it means that they do have access. Otherwise they do not.

By leaving it out from here we can change it in one place on our server without having to update this code, and we're less likely to get an incorrect match.

truongwp:
@Crabcyborg Thanks for checking. I fixed those requests.

classes/controllers/FrmSimpleBlocksController.php
```diff
@@ -33,12 +33,18 @@ public static function block_editor_assets() {
...
+		$modal_addon = self::get_modal_addon_info();
+
 		$script_vars = array(
 			'forms' => self::get_forms_options(),
 			'icon'  => $icon,
 			'name'  => $block_name,
 			'link'  => FrmAppHelper::admin_upgrade_link( 'block' ),
 			'url'   => FrmAppHelper::plugin_url(),
+			'modalAddon' => array(
+				'link'      => FrmAppHelper::admin_upgrade_link( 'addons', $modal_addon['link'] ),
```
stephywells, line 45: @truongwp I was just looking through the stats for this, and it looks like this will need an update. Right now, this is the same as the addons page, so we can't differentiate.

Can you create another PR and change 'addons' here to 'block'? Thank you!

truongwp: I created a new PR for this https://github.com/Strategy11/formidable-forms/pull/1253

PR #1218, opened 2023-06-08
remove crons when lite is deactivated

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4292

This small update just makes sure the events scheduled in lite are unscheduled when it is deactivated!

This could be tested with any cron management plugin. I'm using Advanced Cron Manager.

### DIFF
--- a/classes/controllers/FrmCronController.php
+++ b/classes/controllers/FrmCronController.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * Cron controller
+ *
+ * @since x.x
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This looks good!

Crabcyborg (approved):
Thanks @AbdiTolesa @truongwp and @garretlaxton! I think this looks good. It's consistent with our other cron implementations now.

🚀

PR #1219, opened 2023-06-09
make sure  param to FrmHtmlHelper::echo_dropdown_option is always boolean

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4286

The "other" option should be now selected when editing an entry either in front/back end, if it has been selected when submitting the entry.


garretlaxton (approved):
This fix worked great!

truongwp (approved):

Crabcyborg (approved):

This fix seems sensible. The `FrmFieldsHelper::prepare_other_input` changing the value to `'checked="checked" '` is definitely not great, but I'd be afraid to change it at the risk of breaking something.

PR #1226, opened 2023-06-23
trim captcha secrets before save

AbdiTolesa:
@Crabcyborg I'm not sure if I should link this to the issue as the issue is public but I have updated the captcha secrets so that they are trimmed before save.


garretlaxton (approved):
This looks good!

Crabcyborg (approved):

🚀

PR #1229, opened 2023-06-26
Live update SVG icons when changing Styles settings

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/4367


truongwp:
@garretlaxton Please test this PR. I mentioned the necessary PR in the description. But you should also test without it to make sure there won't be any problems if users use the old Pro version.

garretlaxton:
This looks good with the pro version, but with an older version some of the icons are cut off at the bottom.

Example, the bottom is flat and not rounded.
image

truongwp:
@garretlaxton It doesn't happen with me. Can you test it with the `master` branch?

Do you see any problems else?

garretlaxton (approved):
Hmmm, for some reason I don't see it anymore either so it must've been cache. Everything else looks good!

Crabcyborg:
@truongwp Some updates have been made to the icons branch.

https://github.com/Strategy11/formidable-pro/issues/4348 now includes https://github.com/Strategy11/formidable-pro/pull/4353.

It looks like this update no longer works with those new changes.

Could you look into updating it please?

truongwp:
@Crabcyborg I'm going to create a new PR for the live updating on the Styles page. This PR is based on #3877, which was closed and wasn't merged.

truongwp:
@Crabcyborg I found that we can reuse much code in this PR, so I continue on this. Please check it.

Why is the separate SVG icons code in Pro plugin? I think it should be in the Lite plugin.

Crabcyborg:
> Why is the separate SVG icons code in Pro plugin

@truongwp It's there because it's only required for Pro. We can continue using the `icons.svg` file on the back end and we only require icons for the front end in Pro because it's only used for Pro fields (star rating, password, repeater, collapsible sections, signatures).

If we put it in Lite we'd need to still have backward compatibility support for when Lite isn't updated and Pro is so it's easier to just keep it contained to Pro.

Is it possible to actually include this update in Pro as well? We could put it in the Pro style settings JS instead https://github.com/Strategy11/formidable-pro/blob/master/js/admin/settings/style-settings.js

I don't think we need this in Lite as the icons are only in Pro styling settings.

truongwp:
@Crabcyborg Ok. I will check and move this change to the Pro plugin.

truongwp:
I moved the code to Pro plugin in https://github.com/Strategy11/formidable-pro/pull/4381

PR #1230, opened 2023-06-29
Enhancement of Category Headings Visibility and Search Results

shervElmi:
This PR introduces two enhancements:

The first is synchronizing the visibility of category headings with their associated fields' search results. When all fields under a heading are hidden due to not matching the search criteria, the category heading will also be hidden.

The second enhancement addresses an issue where advanced fields were always shown in the field list, regardless of the search value. With the changes introduced in this PR, advanced fields won't show by default, they will only appear if the search string contains the field name.

## Related Issue:
https://github.com/Strategy11/formidable-pro/issues/4283
https://github.com/Strategy11/formidable-pro/issues/4289

## QA URL:
https://qa.formidableforms.com/sherv5/wp-admin/admin.php?page=formidable&frm_action=edit&id=1

## Testing Instructions:
2. Open an existing form or create a new one.
3. Perform a search using the search functionality.
4. Notice that category headings are hidden when all their associated fields have no search matches.
5. Ensure that advanced fields don't appear in the search results unless the search string contains the field name.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -7593,6 +7593,11 @@ input[disabled],
 	display: flex !important;
 }
 
+/* Prevents advanced fields from always appearing in search results */
+#frm-insert-fields li.frm_show_upgrade.frm_hidden {
+	display: none !important;
+}
+
 .frmbutton.ui-draggable-dragging a {
 	border: 1px solid var(--grey-300);
 	box-shadow: var(--box-shadow-md);
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -9218,6 +9218,9 @@ function frmAdminBuildJS() {
 
 		headingElements.forEach( heading => {
 			const fieldsListElement = heading.nextElementSibling;
+			if ( ! fieldsListElement ) {
+				return;
+			}
 			const listItemElements = fieldsListElement.querySelectorAll( ':scope > li.frmbutton' );
 			const allHidden = Array.from( listItemElements ).every( li => li.classList.contains( 'frm_hidden' ) );
 
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -9199,7 +9199,31 @@ function frmAdminBuildJS() {
 			}
 		}
 
+		// Updates the visibility of category headings based on search results.
+		updateCatHeadingVisibility();
+
 		jQuery( this ).trigger( 'frmAfterSearch' );
+	}
+
+	/**
+	 * Updates the visibility of category headings based on search results.
+	 * If all associated fields are hidden (indicating no search matches),
+	 * the heading is hidden.
+	 *
+	 * @since x.x
+	 */
+	function updateCatHeadingVisibility() {
+		const insertFieldsElement = document.querySelector( '#frm-insert-fields' );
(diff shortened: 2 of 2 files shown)


garretlaxton (approved):
I feel honored to be featured in your demo! This also looks great!

shervElmi:
> I feel honored to be featured in your demo! This also looks great!

Thank you, Garret! 🎉

js/formidable_admin.js
```diff
@@ -9199,9 +9199,36 @@ function frmAdminBuildJS() {
...
+		const insertFieldsElement = document.querySelector( '#frm-insert-fields' );
+		const headingElements = insertFieldsElement.querySelectorAll( ':scope > .frm-with-line' );
+
+		headingElements.forEach( heading => {
+			const fieldsListElement = heading.nextElementSibling;
+			if ( ! fieldsListElement ) {
+				return;
+			}
+			const listItemElements = fieldsListElement.querySelectorAll( ':scope > li.frmbutton' );
+			const allHidden = Array.from( listItemElements ).every( li => li.classList.contains( 'frm_hidden' ) );
```
omidakhavans, line 9225: @shervElmi  Thank you for woking on this. `Array.from` has a second argument in which you could pass a map function and `every` is not necessary in this case I believe.

css/frm_admin.css
```diff
@@ -7593,6 +7593,11 @@ input[disabled],
 	display: flex !important;
 }
 
+/* Prevents advanced fields from always appearing in search results */
+#frm-insert-fields li.frm_show_upgrade.frm_hidden {
+	display: none !important;
```
omidakhavans, line 7598: Dear @shervElmi , How we could avoid this `important`?

js/formidable_admin.js
shervElmi: I don't require a new list to iterate over, @omidakhavans. My primary requirement is to ensure that all items have the `frm_hidden` class.

css/frm_admin.css
shervElmi: We're faced with a situation where we need to override existing general CSS that uses the `!important` keyword, @omidakhavans. Unfortunately, without using `!important` ourselves, we're unable to achieve the desired override.

js/formidable_admin.js
omidakhavans: `every` will return when something is false or falsy https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every.

I mean second argument of `from Array.from( first argument, second argument )` not new map to iterate over.

I hope you find it useful.

shervElmi: My intention is to retrieve a boolean value, @omidakhavans. Could you please provide me with an example of how I can receive a boolean value without using `every`? I'm having trouble aligning your approach with my objective.

Crabcyborg (approved):
This looks good! Thanks @shervElmi @omidakhavans and @garretlaxton!

🚀

shervElmi:
Thanks, @Crabcyborg! Could you please also close the following issue? I've addressed it in this PR:
https://github.com/Strategy11/formidable-pro/issues/4283

PR #1231, opened 2023-07-03
Prevent Form Submission in WP Block Editor

shervElmi:
This PR addresses an issue that allows forms to be submitted in the WP Block Editor, causing the page to reload. The proposed changes prevent this from happening.

## QA URL:
https://qa.formidableforms.com/sherv/wp-admin/post.php?post=1&action=edit

## Related Issue:
Fixes https://github.com/Strategy11/formidable-pro/issues/4246

## Testing Instructions:
1. Navigate to the block editor and add a Form using the Formidable Form block.
2. Try to submit the form on the block editor page.
3. Confirm that the form cannot be submitted and the page does not reload.


garretlaxton (approved):
This looks good. 

Just a question though, is it possible to disable the fields and submit button hover actions? I can still select and fill out the fields, and the submit button still has the hover affect so it looks like it can be submitted.

shervElmi:
> This looks good.
> 
> Just a question though, is it possible to disable the fields and submit button hover actions? I can still select and fill out the fields, and the submit button still has the hover affect so it looks like it can be submitted.

Thanks for your review, @garretlaxton!

@Crabcyborg, could you share your thoughts on Garret's comment?

Crabcyborg:
@shervElmi I'm not sure if we need the form to be interactive at all. In visual views, the whole preview uses `pointer-events: none;`.

@stephywells What do you think makes sense for the Gutenberg block form preview? Would we want it to remain interactive, or does it make more sense to just disable all of the inputs the buttons like we do in visual views?

stephywells:
> @shervElmi I'm not sure if we need the form to be interactive at all. In visual views, the whole preview uses `pointer-events: none;`.
> 
> @stephywells What do you think makes sense for the Gutenberg block form preview? Would we want it to remain interactive, or does it make more sense to just disable all of the inputs the buttons like we do in visual views?

Yep, exactly. This form isn't intended to be interactive. That sounds like a perfect solution.

Crabcyborg:
@garretlaxton This needs re-testing as we've changed the solution. Now everything in the form preview is disabled, not just the submit button. There won't be hover effects as the mouse is ignored on the form elements.

Thank you!

shervElmi:
> @garretlaxton This needs re-testing as we've changed the solution. Now everything in the form preview is disabled, not just the submit button. There won't be hover effects as the mouse is ignored on the form elements.
> 
> Thank you!

I learn from you every day, Mike! Thanks for your comment. If I have similar changes in the future, I'll do that.

garretlaxton (approved):
This looks great now!

omidakhavans (approved):
If the approach confirmed well. I could confirm this looks good. Thank you. @shervElmi

Crabcyborg (approved):
Thanks @shervElmi for updating that!

And thank you @garretlaxton and @omidakhavans for reviewing!

🚀

PR #1232, opened 2023-07-04
add repeater delete confirmation message to global setting

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4318

### DIFF
--- a/classes/models/FrmSettings.php
+++ b/classes/models/FrmSettings.php
@@ -130,6 +130,8 @@ class FrmSettings {
 			// Normally custom CSS is a string. A false value is used when nothing has been set.
 			// When it is false, we try to use the old custom_key value from the default style's post_content array.
 			'custom_css' => false,
+
+			'repeater_row_delete_confirmation' => __( 'Are you sure you want to delete this row?', 'formidable' ),
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


classes/models/FrmSettings.php
```diff
@@ -130,6 +130,8 @@ public function default_options() {
 			// Normally custom CSS is a string. A false value is used when nothing has been set.
 			// When it is false, we try to use the old custom_key value from the default style's post_content array.
 			'custom_css' => false,
+
+			'repeater_row_delete_confirmation' => __( 'Are you sure you want to delete this row?', 'formidable' ),
```
Crabcyborg, line 134: @AbdiTolesa Maybe we could just save this to Pro settings? It would be great if we could leave this line out of the free plugin.

classes/views/frm-settings/messages.php
```diff
@@ -57,6 +57,15 @@ class="frm_with_left_label"
 			class="frm_with_left_label"
 			value="<?php echo esc_attr( $frm_settings->unique_msg ); ?>"/>
 	</p>
+	<p>
```
Crabcyborg, line 60: @AbdiTolesa Can we also get this into Pro?

The Stripe plugin adds a custom message to global settings by saving the setting to the Stripe global object, and including the field with the `frm_messages_settings_form` action.

We don't really want Pro updates in Lite.

AbdiTolesa:
Closing this now since it is not needed anymore.

PR #1233, opened 2023-07-05
Disallow editing form in trash

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4296

This update prevents a deleted form from being edited. It could use some styling if required for some alignment and color stuff.

I also added more text to the form preview message as to why the form cannot be previewed for deleted form case.

In addition to that there is a minor style fix as the message from is hidden when using `wp_die` for cases like attempting to edit non-existent form or deleted one.

Sreenshots:
Attempting to edit a deleted form,
image

Attempting to preview a deleted form,
image

Steps to test:
1. On the form list page, open the 'Trash' link in new tab so that the list page doesn't reload with the deleted form gone.
image

2. Try to click on the edit and preview buttons and test.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1977,12 +1977,12 @@ class FrmFormsController {
 	public static function show_form( $id = '', $key = '', $title = false, $description = false, $atts = array() ) {
 		$form = self::maybe_get_form_by_id_or_key( $id, $key );
 
-		if ( is_object( $form ) && $form->status === 'trash' ) {
-			wp_die( esc_html__( 'You cannot preview this item because it is in the Trash. Please restore it and try again.', 'formidable' ) );
-		}
-
 		if ( ! $form ) {
 			return __( 'Please select a valid form', 'formidable' );
+		}
+
+		if ( is_object( $form ) && $form->status === 'trash' ) {
+			wp_die( esc_html__( 'You cannot preview this item because it is in the Trash. Please restore it and try again.', 'formidable' ) );
 		}
 
 		if ( 'auto' === $title ) {
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1090,6 +1090,10 @@ class FrmFormsController {
 			wp_die( esc_html__( 'You are trying to edit a form that does not exist.', 'formidable' ) );
 		}
 
+		if ( $form->status === 'trash' ) {
+			wp_die( esc_html__( 'You cannot edit this item because it is in the Trash. Please restore it and try again.', 'formidable' ) );
+		}
+
 		if ( $form->parent_form_id ) {
 			/* translators: %1$s: Start link HTML, %2$s: End link HTML */
 			wp_die( sprintf( esc_html__( 'You are trying to edit a child form. Please edit from %1$shere%2$s', 'formidable' ), '<a href="' . esc_url( FrmForm::get_edit_link( $form->parent_form_id ) ) . '">', '</a>' ) );
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1977,6 +1977,10 @@ class FrmFormsController {
 	public static function show_form( $id = '', $key = '', $title = false, $description = false, $atts = array() ) {
 		$form = self::maybe_get_form_by_id_or_key( $id, $key );
 
+		if ( is_object( $form ) && $form->status === 'trash' ) {
+			wp_die( esc_html__( 'You cannot preview this item because it is in the Trash. Please restore it and try again.', 'formidable' ) );
+		}
+
 		if ( ! $form ) {
 			return __( 'Please select a valid form', 'formidable' );
 		}
@@ -2028,7 +2032,7 @@ class FrmFormsController {
 
 		if ( ! empty( $id ) ) { // form id or key is set
 			$form = FrmForm::getOne( $id );
-			if ( ! $form || $form->parent_form_id || $form->status === 'trash' ) {
+			if ( ! $form || $form->parent_form_id ) {
 				$form = false;
 			}
 		}
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -2040,6 +2040,13 @@ class FrmFormsController {
 		return $form;
 	}
 
+	public static function get_form_status() {
+		// check_ajax_referer( 'frm_ajax', 'nonce' );
+		$form_id = FrmAppHelper::get_param( 'form_id', '', 'post', 'absint' );
+		$form    = FrmForm::getOne( $form_id );
+		wp_send_json_success( $form->status );
+	}
+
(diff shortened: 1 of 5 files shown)


garretlaxton:
Actually, I just noticed when I try and add a field to a form I get an Internal Server Error.
image

garretlaxton (approved):

classes/controllers/FrmFormsController.php
```diff
@@ -1973,6 +1977,10 @@ private static function maybe_get_form_by_id_or_key( $id, $key ) {
 	public static function show_form( $id = '', $key = '', $title = false, $description = false, $atts = array() ) {
 		$form = self::maybe_get_form_by_id_or_key( $id, $key );
 
+		if ( is_object( $form ) && $form->status === 'trash' ) {
+			wp_die( esc_html__( 'You cannot preview this item because it is in the Trash. Please restore it and try again.', 'formidable' ) );
+		}
+
 		if ( ! $form ) {
```
truongwp, line 1984: I think we can move the `$form->status === 'trash'` to after this check.

classes/controllers/FrmFormsController.php
```diff
@@ -1973,6 +1977,10 @@ private static function maybe_get_form_by_id_or_key( $id, $key ) {
 	public static function show_form( $id = '', $key = '', $title = false, $description = false, $atts = array() ) {
 		$form = self::maybe_get_form_by_id_or_key( $id, $key );
 
+		if ( is_object( $form ) && $form->status === 'trash' ) {
```
truongwp, line 1980: I think we should move this check to after https://github.com/Strategy11/formidable-forms/pull/1233/files#diff-1b880a2d11e84622570383ea0503da990c04c1ccbffe4f5f81ab21d68fa00280R1984 or https://github.com/Strategy11/formidable-forms/pull/1233/files#diff-1b880a2d11e84622570383ea0503da990c04c1ccbffe4f5f81ab21d68fa00280R2033

classes/controllers/FrmFormsController.php
```diff
@@ -1977,6 +1981,10 @@ public static function show_form( $id = '', $key = '', $title = false, $descript
 			return __( 'Please select a valid form', 'formidable' );
 		}
 
+		if ( is_object( $form ) && $form->status === 'trash' ) {
```
truongwp, line 1984: We don't need `is_object()` check because `$form` can only be `object` or `false`.

AbdiTolesa: Oh, that is indeed extra check, thanks!

truongwp (approved):
Thank @AbdiTolesa. This looks good now.

classes/controllers/FrmFormsController.php
```diff
@@ -1090,6 +1090,10 @@ private static function get_edit_vars( $id, $errors = array(), $message = '', $c
 			wp_die( esc_html__( 'You are trying to edit a form that does not exist.', 'formidable' ) );
 		}
 
+		if ( $form->status === 'trash' ) {
+			wp_die( esc_html__( 'You cannot edit this item because it is in the Trash. Please restore it and try again.', 'formidable' ) );
```
Crabcyborg, line 1094: Is there anything we could do to make this error look nicer?

Screen Shot 2023-07-14 at 1 27 24 PM

Crabcyborg: @tuguirazvan How do you think we should handle someone trying to load a form that is in trash?

We're going to show an error to prevent working on something in the trash.

Should we have a dedicated page for the error? What do you think we should make that page look like? The above screenshot is what is it is doing now (just white background with some text that is lined up against the left of the screen - it doesn't look good).

I think we could consider redirecting to the forms list, and showing an error message at the top there.

What direction should we take?

tuguirazvan: @Crabcyborg I'm thinking that we can show a modal when clicking on "edit", explaining the fact that this current form is in the trash alongside the option to restore it if they want/need it. Giving them this option at this step is helpful because it is contextual and they don't need to go somewhere else to restore the form. 

Let me know your thoughts! 

CleanShot 2023-07-17 at 10 07 15@2x

AbdiTolesa: @Crabcyborg What is your take on Razvan's input here?

Crabcyborg: @AbdiTolesa I think we should give it a shot when you're back from your leave.

AbdiTolesa: @Crabcyborg Can we progress on this now since I'm back from my leave?

Crabcyborg: @AbdiTolesa Sure. Could you give that modal a try?

AbdiTolesa: Sure, I have just updated it. I think we should go back and start over the PR flow as the update is a significant one.
image

AbdiTolesa: @Crabcyborg I prefer to leave the PHP side check that is already in this update as it is as a backup to edge cases. Like if a user trashed a form and immediately clicked edit before the ajax response has arrived. Any thoughts?

AbdiTolesa:
@garretlaxton Could you give this another round of test as it has been updated so that a pop up is displayed when a user attempts to edit a deleted form from the Form listing page. Clicking on the 'Restore form' button should restore the form from the trash now.

image

classes/controllers/FrmFormsController.php
Crabcyborg: @AbdiTolesa Why is the modal with a JS check? I was picturing we'd just load a blank page with just a modal error - Basically just this error with more styling.

AbdiTolesa: @Crabcyborg My understanding from Razvan's comment above was that the user should be able to have that modal open without leaving the page (being redirected). In fact, I can try doing the same thing after the page has redirected to form edit page but that could become more complicated as it involves stuffs that are not straight forward, like loading `formidable_admin.js`, which has a bunch of scripts as dependency. That is (enqueuing it) just something I gave a shot now.

We could give that a shot anyway but I personally believe doing the modal here on the Form list page is better for the user experience.

garretlaxton:
The popup works great! Clicking the restore button restores the form.

classes/controllers/FrmFormsController.php
AbdiTolesa: @Crabcyborg I found no luck in showing the popup after redirect, getting to load the required scripts has become a bit tricky for me due to hooks firing sequence primarily. I believe it would require changing a bunch of functions if it is possible to make it work at all.

Crabcyborg: @AbdiTolesa I gave it a shot. It's a bit messy but this is working fairly well https://github.com/Strategy11/formidable-forms/pull/1280. I left a checklist there if you could try working from that.

Crabcyborg:
I'm going to close this in favour of https://github.com/Strategy11/formidable-forms/pull/1280.

PR #1235, opened 2023-07-05
add read only setting to name field

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/4306
Just adding a "read only" field setting to Name field.

image

### DIFF
--- a/classes/models/fields/FrmFieldName.php
+++ b/classes/models/fields/FrmFieldName.php
@@ -40,6 +40,13 @@ class FrmFieldName extends FrmFieldCombo {
 		);
 	}
 
+	protected function field_settings_for_type() {
+		$settings = parent::field_settings_for_type();
+		$settings['read_only'] = true;
+
+		return $settings;
+	}
+
 	/**
 	 * Gets processed sub fields.
 	 * This should return the list of sub fields after sorting or show/hide based of some options.


garretlaxton (approved):
This looks great!

truongwp (approved):
This looks great! Thank @AbdiTolesa!

Crabcyborg (changes requested):
@AbdiTolesa As far as I can tell, `read_only` is actually a Pro feature.

If you disable Pro, and go to all of the basic fields, there is no read only option. When you enable Pro, they appear.

See my recording. The setting never sticks.
U4M36KySvh

Can we move this update to Pro so it's consistent with the other fields?

AbdiTolesa:
> @AbdiTolesa As far as I can tell, `read_only` is actually a Pro feature.
> 
> If you disable Pro, and go to all of the basic fields, there is no read only option. When you enable Pro, they appear.
> 
> See my recording. The setting never sticks. Because Pro is inactive. U4M36KySvh
> 
> Can we move this update to Pro so it's consistent with the other fields?

@Crabcyborg I never thought about that but I have created a similar update in pro now, (https://github.com/Strategy11/formidable-pro/pull/4334) thanks!

PR #1236, opened 2023-07-10
Use a true arrow up instead of using transform

truongwp:
Steps to replicate:
- Edit a style
- Open the `Section fields` settings
- Under `Collapse Icon > Icons`, there is one option with 2 arrow down icons. This affects the frontend too.

I asked Răzvan to recreate this icon.

**Before:**
Screen Shot 2023-07-10 at 21 03 19

**After:**
Screen Shot 2023-07-10 at 21 01 56


garretlaxton:
This looks good. I'm not sure if this matters much, but the up arrows seem a tiny bit bigger than the down arrow.
image

truongwp:
Thank @garretlaxton. I'm checking why it's a bit smaller. Maybe we will need to modify the icon.

truongwp:
@garretlaxton I fixed the small icon issue. Can you take a look again?

garretlaxton (approved):
Nice, this looks good!

PR #1237, opened 2023-07-10
Entry statuses

omidakhavans:
### Summary

This PR modifies is_draft column in entries list table and also add entry status to [entry details](https://gyazo.com/f9a2eaaf9c779b4a3f61205b5f6d75ba).

### Related PRs
https://github.com/Strategy11/formidable-pro/pull/4352
https://github.com/Strategy11/formidable-abandonment/pull/27

### Checklist

- [ ] Import/Export with old and new exported files.
- [ ] Disable abandonment add-on and test the remaining "In Progress" and abandoned entries.

### DIFF
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -46,8 +46,8 @@ class FrmEntriesController {
 			}
 
 			if ( $include_entries_status ) {
-				add_filter( 'manage_' . $base . '_columns', 'FrmEntriesController::add_status_column', 99 );
-				add_filter( 'frm_entries_post_status_column', 'FrmEntriesController::status_column_value', 10, 2 );
+				// add_filter( 'manage_' . $base . '_columns', 'FrmEntriesController::add_status_column', 99 );
+				// add_filter( 'frm_entries_post_status_column', 'FrmEntriesController::status_column_value', 10, 2 );
 			}
 		} else {
 			add_filter( 'screen_options_show_screen', __CLASS__ . '::remove_screen_options', 10, 2 );
@@ -104,6 +104,8 @@ class FrmEntriesController {
 			$columns[ $form_id . '_user_id' ] = __( 'Created By', 'formidable' );
 		}
 
+		self::add_status_column( $form_id, $columns );
+
 		$columns[ $form_id . '_created_at' ] = __( 'Entry creation date', 'formidable' );
 		$columns[ $form_id . '_updated_at' ] = __( 'Entry update date', 'formidable' );
 		self::maybe_add_ip_col( $form_id, $columns );
@@ -128,27 +130,14 @@ class FrmEntriesController {
 	/**
 	 * Add "Entry status" column to entries table.
 	 *
-	 * @since 1.0
+	 * @since x.x
 	 *
 	 * @param array<string> $columns Entries table columns.
 	 *
 	 * @return array<string>
 	 */
-	public static function add_status_column( $columns ) {
-		global $frm_vars;
-		$form_id = FrmForm::get_current_form_id();
-
-		$columns[ $form_id . '_post_status' ] = __( 'Entry Status', 'formidable' );
-		// Change order of draft col, Since we added post status in line above it could be accessible by count -1.
-		$columns = array_merge(
-			array_splice( $columns, 0, 4 ),
-			array_splice( $columns, count( $columns ) - 1, 1 ),
-			$columns
-		);
-
-		$frm_vars['cols'] = $columns;
-
-		return $columns;
+	public static function add_status_column( $form_id, &$columns ) {
+		$columns[ $form_id . '_entry_status' ] = esc_html__( 'Entry Status', 'formidable' );
 	}
 
 	/**
@@ -164,7 +153,14 @@ class FrmEntriesController {
 	public static function status_column_value( $val, $args ) {
 		$entry = $args['item'];
 
-		return FrmEntriesHelper::get_entry_status( $entry->is_draft );
+		$entry_status_label = FrmEntriesHelper::get_entry_status( $entry->is_draft );
+
+		return sprintf(
+			/* translators: %1$s: Status class name %2$s: Status name */
+			'<span class="frm-entry-status frm-entry-status-%s">%s</span>',
+			sanitize_html_class( strtolower( str_replace( ' ', '-', $entry_status_label ) ) ),
+			$entry_status_label
+		);
 	}
 
 	private static function get_columns_for_form( $form_id, &$columns ) {
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -86,8 +86,7 @@ class FrmEntriesController {
 			$columns[ $form_id . '_user_id' ] = __( 'Created By', 'formidable' );
 		}
 
-		self::entry_status_column( $form_id, $columns );
-
+		$columns[ $form_id . '_is_draft' ] = esc_html__( 'Entry Status', 'formidable' );
 		$columns[ $form_id . '_created_at' ] = __( 'Entry creation date', 'formidable' );
 		$columns[ $form_id . '_updated_at' ] = __( 'Entry update date', 'formidable' );
 		self::maybe_add_ip_col( $form_id, $columns );
@@ -107,19 +106,6 @@ class FrmEntriesController {
 		}
 
 		return $columns;
-	}
-
-	/**
-	 * Add "Entry status" column to entries table.
-	 *
-	 * @since x.x
-	 *
-	 * @param array<string> $columns Entries table columns.
-	 *
-	 * @return array<string>
-	 */
-	private static function entry_status_column( $form_id, &$columns ) {
-		$columns[ $form_id . '_entry_status' ] = esc_html__( 'Entry Status', 'formidable' );
 	}
 
 	private static function get_columns_for_form( $form_id, &$columns ) {
@@ -297,7 +283,7 @@ class FrmEntriesController {
 			$form_id . '_updated_at'    => 'updated_at',
 			$form_id . '_ip'            => 'ip',
 			$form_id . '_item_key'      => 'item_key',
-			$form_id . '_entry_status'  => 'is_draft',
+			$form_id . '_is_draft'      => 'is_draft',
 		);
 
 		foreach ( $fields as $field ) {
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -86,7 +86,7 @@ class FrmEntriesController {
 			$columns[ $form_id . '_user_id' ] = __( 'Created By', 'formidable' );
 		}
 
-		self::add_status_column( $form_id, $columns );
+		self::entry_status_column( $form_id, $columns );
 
 		$columns[ $form_id . '_created_at' ] = __( 'Entry creation date', 'formidable' );
 		$columns[ $form_id . '_updated_at' ] = __( 'Entry update date', 'formidable' );
@@ -118,7 +118,7 @@ class FrmEntriesController {
 	 *
 	 * @return array<string>
 	 */
-	public static function add_status_column( $form_id, &$columns ) {
+	public static function entry_status_column( $form_id, &$columns ) {
 		$columns[ $form_id . '_entry_status' ] = esc_html__( 'Entry Status', 'formidable' );
 	}
 
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -31,24 +31,6 @@ class FrmEntriesController {
 			add_filter( 'manage_' . $base . '_columns', 'FrmEntriesController::manage_columns' );
 			add_filter( 'get_user_option_' . self::hidden_column_key( $menu_name ), 'FrmEntriesController::hidden_columns' );
 			add_filter( 'manage_' . $base . '_sortable_columns', 'FrmEntriesController::sortable_columns' );
-
-			/**
-			 * Allows include entry status column to entries list.
-			 *
-			 * @since x.x
-			 *
-			 * @param bool $include_entries_status Excluded fields type.
-			 */
-			$include_entries_status = apply_filters( 'frm_entries_add_status_column', false );
-
-			if ( ! is_bool( $include_entries_status ) ) {
-				_doing_it_wrong( __METHOD__, esc_html__( 'You need to return bool to enable/disable entry status column.', 'formidable' ), 'x.x' );
-			}
-
-			if ( $include_entries_status ) {
-				// add_filter( 'manage_' . $base . '_columns', 'FrmEntriesController::add_status_column', 99 );
-				// add_filter( 'frm_entries_post_status_column', 'FrmEntriesController::status_column_value', 10, 2 );
-			}
 		} else {
 			add_filter( 'screen_options_show_screen', __CLASS__ . '::remove_screen_options', 10, 2 );
 		}
@@ -138,29 +120,6 @@ class FrmEntriesController {
 	 */
 	public static function add_status_column( $form_id, &$columns ) {
 		$columns[ $form_id . '_entry_status' ] = esc_html__( 'Entry Status', 'formidable' );
-	}
-
-	/**
-	 * Return entry status based on is_draft column.
-	 *
-	 * @since x.x
-	 *
-	 * @param string       $val Column value.
-	 * @param array<mixed> $args Column args.
-	 *
-	 * @return string
-	 */
-	public static function status_column_value( $val, $args ) {
-		$entry = $args['item'];
-
-		$entry_status_label = FrmEntriesHelper::get_entry_status( $entry->is_draft );
-
-		return sprintf(
-			/* translators: %1$s: Status class name %2$s: Status name */
-			'<span class="frm-entry-status frm-entry-status-%s">%s</span>',
-			sanitize_html_class( strtolower( str_replace( ' ', '-', $entry_status_label ) ) ),
-			$entry_status_label
-		);
 	}
 
 	private static function get_columns_for_form( $form_id, &$columns ) {
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -31,6 +31,24 @@ class FrmEntriesController {
 			add_filter( 'manage_' . $base . '_columns', 'FrmEntriesController::manage_columns' );
 			add_filter( 'get_user_option_' . self::hidden_column_key( $menu_name ), 'FrmEntriesController::hidden_columns' );
 			add_filter( 'manage_' . $base . '_sortable_columns', 'FrmEntriesController::sortable_columns' );
+
+			/**
+			 * Allows include entry status column to entries list.
+			 *
+			 * @since x.x
+			 *
+			 * @param bool $include_entries_status Excluded fields type.
+			 */
+			$include_entries_status = apply_filters( 'frm_entries_add_status_column', false );
+
+			if ( ! is_bool( $include_entries_status ) ) {
+				_doing_it_wrong( __METHOD__, esc_html__( 'You need to return bool to enable/disable entry status column.', 'formidable' ), 'x.x' );
+			}
+
+			if ( $include_entries_status ) {
+				add_filter( 'manage_' . $base . '_columns', 'FrmEntriesController::add_status_column', 99 );
+				add_filter( 'frm_entries_post_status_column', 'FrmEntriesController::status_column_value', 10, 2 );
+			}
 		} else {
 			add_filter( 'screen_options_show_screen', __CLASS__ . '::remove_screen_options', 10, 2 );
 		}
@@ -105,6 +123,48 @@ class FrmEntriesController {
 		}
 
 		return $columns;
+	}
+
+	/**
+	 * Add "Entry status" column to entries table.
+	 *
+	 * @since 1.0
+	 *
+	 * @param array<string> $columns Entries table columns.
+	 *
+	 * @return array<string>
+	 */
+	public static function add_status_column( $columns ) {
+		global $frm_vars;
+		$form_id = FrmForm::get_current_form_id();
+
+		$columns[ $form_id . '_post_status' ] = __( 'Entry Status', 'formidable' );
+		// Change order of draft col, Since we added post status in line above it could be accessible by count -1.
+		$columns = array_merge(
+			array_splice( $columns, 0, 4 ),
+			array_splice( $columns, count( $columns ) - 1, 1 ),
+			$columns
+		);
+
+		$frm_vars['cols'] = $columns;
+
+		return $columns;
+	}
+
+	/**
+	 * Return entry status based on is_draft column.
+	 *
+	 * @since x.x
+	 *
+	 * @param string       $val Column value.
+	 * @param array<mixed> $args Column args.
+	 *
+	 * @return string
+	 */
+	public static function status_column_value( $val, $args ) {
+		$entry = $args['item'];
+
+		return FrmEntriesHelper::get_entry_status( $entry->is_draft );
 	}
 
 	private static function get_columns_for_form( $form_id, &$columns ) {
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -118,7 +118,7 @@ class FrmEntriesController {
 	 *
 	 * @return array<string>
 	 */
-	public static function entry_status_column( $form_id, &$columns ) {
+	private static function entry_status_column( $form_id, &$columns ) {
 		$columns[ $form_id . '_entry_status' ] = esc_html__( 'Entry Status', 'formidable' );
 	}
 
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -278,12 +278,12 @@ class FrmEntriesController {
 		$fields  = FrmField::get_all_for_form( $form_id );
 
 		$columns = array(
-			$form_id . '_id'            => 'id',
-			$form_id . '_created_at'    => 'created_at',
-			$form_id . '_updated_at'    => 'updated_at',
-			$form_id . '_ip'            => 'ip',
-			$form_id . '_item_key'      => 'item_key',
-			$form_id . '_is_draft'      => 'is_draft',
+			$form_id . '_id'         => 'id',
+			$form_id . '_created_at' => 'created_at',
+			$form_id . '_updated_at' => 'updated_at',
+			$form_id . '_ip'         => 'ip',
+			$form_id . '_item_key'   => 'item_key',
+			$form_id . '_is_draft'   => 'is_draft',
 		);
 
 		foreach ( $fields as $field ) {
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -292,12 +292,12 @@ class FrmEntriesController {
 		$fields  = FrmField::get_all_for_form( $form_id );
 
 		$columns = array(
-			$form_id . '_id'         => 'id',
-			$form_id . '_created_at' => 'created_at',
-			$form_id . '_updated_at' => 'updated_at',
-			$form_id . '_ip'         => 'ip',
-			$form_id . '_item_key'   => 'item_key',
-			$form_id . '_is_draft'   => 'is_draft',
+			$form_id . '_id'            => 'id',
+			$form_id . '_created_at'    => 'created_at',
+			$form_id . '_updated_at'    => 'updated_at',
+			$form_id . '_ip'            => 'ip',
+			$form_id . '_item_key'      => 'item_key',
+			$form_id . '_entry_status'  => 'is_draft',
 		);
 
 		foreach ( $fields as $field ) {
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -86,7 +86,7 @@ class FrmEntriesController {
 			$columns[ $form_id . '_user_id' ] = __( 'Created By', 'formidable' );
 		}
 
-		$columns[ $form_id . '_is_draft' ] = esc_html__( 'Entry Status', 'formidable' );
+		$columns[ $form_id . '_is_draft' ]   = esc_html__( 'Entry Status', 'formidable' );
 		$columns[ $form_id . '_created_at' ] = __( 'Entry creation date', 'formidable' );
 		$columns[ $form_id . '_updated_at' ] = __( 'Entry update date', 'formidable' );
 		self::maybe_add_ip_col( $form_id, $columns );
--- a/classes/controllers/FrmFormActionsController.php
+++ b/classes/controllers/FrmFormActionsController.php
@@ -505,7 +505,7 @@ class FrmFormActionsController {
 				$entry = FrmEntry::getOne( $entry, true );
 			}
 
-			if ( empty( $entry ) || ( FrmEntriesHelper::DRAFT_ENTRY_STATUS === $entry->is_draft && 'draft' != $event ) ) {
+			if ( empty( $entry ) || ( FrmEntriesHelper::DRAFT_ENTRY_STATUS === (int) $entry->is_draft && 'draft' != $event ) ) {
 				continue;
 			}
 
--- a/classes/controllers/FrmFormActionsController.php
+++ b/classes/controllers/FrmFormActionsController.php
@@ -505,7 +505,7 @@ class FrmFormActionsController {
 				$entry = FrmEntry::getOne( $entry, true );
 			}
 
-			if ( empty( $entry ) || ( FrmEntriesHelper::DRAFT_ENTRY_STATUS === (int) $entry->is_draft && 'draft' != $event ) ) {
+			if ( empty( $entry ) || ( FrmEntriesHelper::DRAFT_ENTRY_STATUS === (int) $entry->is_draft && 'draft' !== $event ) ) {
 				continue;
 			}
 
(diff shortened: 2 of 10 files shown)


classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +755,58 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+		 * "3" is reserved for abandonment-addon.
+		 *
+		 * @since x.x
+		 *
+		 * @param array<string> $extended_entry_status Entry statuses.
+		 */
+		$extended_entry_status = apply_filters( 'frm_register_entry_statuses', array() );
+
+		if ( ! is_array( $extended_entry_status ) ) {
+			_doing_it_wrong( __METHOD__, esc_html__( 'Entry status must be return in array format.', 'formidable' ), 'x.x' );
```
Crabcyborg, line 806: @omidakhavans Since this doesn't correct the invalid entry status, it would still break when it tries to array_merge.

What I've been doing is setting a variable like `$filtered_value = apply_filters( ..., $non_filtered_value, ...` and then if `$filtered_value` is valid, I set it to the `$non_filtered_value` variable then, instead of overwriting the valid value.

I realize there is no default value though for `$extended_entry_status` so in this case I think you could you just do a `$extended_entry_status = array();` here to correct it.

classes/controllers/FrmEntriesController.php
```diff
@@ -31,6 +31,24 @@ private static function load_manage_entries_hooks() {
...
+			/**
+			 * Allows include entry status column to entries list.
+			 *
+			 * @since x.x
+			 *
+			 * @param bool $include_entries_status Excluded fields type.
+			 */
+			$include_entries_status = apply_filters( 'frm_entries_add_status_column', false );
+
+			if ( ! is_bool( $include_entries_status ) ) {
```
stephywells, line 44: Could this be moved to a new function so we aren't making the existing function so much longer?

classes/helpers/FrmEntriesHelper.php
```diff
@@ -5,6 +5,30 @@
...
+	 */
+	const DRAFT_ENTRY_STATUS = 1;
+
+	/**
+	 * "In Progress" entry status.
+	 *
+	 * @since x.x
+	 * @var int
+	 */
+	const IN_PROGRESS_ENTRY_STATUS = 2;
```
stephywells, line 30: Let's not included the in progress status in lite since we won't be using it here.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +755,58 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+	 * @return string
+	 */
+	public static function get_entry_status( $status, $include_label = true ) {
+		$statuses = self::get_entry_statuses();
+
+		if ( array_key_exists( $status, $statuses ) ) {
+			return $include_label ? $statuses[ $status ] : $status;
+		}
+
+		return $include_label ? esc_html__( 'In Progress', 'formidable' ) : self::IN_PROGRESS_ENTRY_STATUS;
```
stephywells, line 776: Let's move the label to a new function too. This is a bit complex when it's trying to do both things at the same time. Something more like this. Then if we want to label, we use this function.
```
function get_status_label( $status ) {
$labels = self::get_entry_statuses();
return isset( $labels[ $status ] ) ? $labels[ $status ] : $labels[1]; 
}
```

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +755,58 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+	 * Get all entry statuses.
+	 *
+	 * @since x.x
+	 *
+	 * @return array<string>
+	 */
+	private static function get_entry_statuses() {
+
+		$default_entry_statuses = array(
+			0 => esc_html__( 'Submitted', 'formidable' ),
```
stephywells, line 789: In this spot, the statuses don't need escaping. They will need to be escaped when they are displayed instead.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +755,58 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+			1 => esc_html__( 'Draft', 'formidable' ),
+			2 => esc_html__( 'In Progress', 'formidable' ),
```
stephywells, line 791: Let's remove in progress from here since it isn't needed in Lite.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +755,58 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+	 *
+	 * @param int $status is_draft column.
+	 * @param bool $include_label Return label.
+	 *
+	 * @return string
+	 */
+	public static function get_entry_status( $status, $include_label = true ) {
+		$statuses = self::get_entry_statuses();
+
+		if ( array_key_exists( $status, $statuses ) ) {
```
stephywells, line 772: If the status is empty, let's default to 0. If it has a value that isn't in the array, let's default to 1. There may be old entries that don't have a value for is_draft.

classes/views/frm-entries/_sidebar-shared-pub.php
```diff
@@ -12,9 +12,11 @@
 	<span id="timestamp">
 	<?php
 	$date_format = __( 'M j, Y @ G:i', 'formidable' );
+	$status = isset( $entry->is_draft ) ? FrmEntriesHelper::get_entry_status( $entry->is_draft ) : esc_html__( 'In Progress', 'formidable' );
 	printf(
-		/* translators: %1$s: The date */
-		esc_html__( 'Submitted: %1$s', 'formidable' ),
+		/* translators: %1$s: Entry status %2$s: The date */
+		esc_html__( '%1$s: %2$s', 'formidable' ),
+		$status,
```
stephywells, line 19: We want to be able to handle this from the get_status_label function I mentioned in my comments. So this would be more like:
`FrmEntriesHelper::get_status_label( isset( $entry->is_draft ) ? $entry->is_draft : 0 )`
(A trailing comma here will cause fatal errors in PHP 5.6)

Are there cases when $entry->is_draft is not available?

classes/controllers/FrmEntriesController.php
```diff
@@ -107,6 +125,48 @@ public static function manage_columns( $columns ) {
...
+	/**
+	 * Add "Entry status" column to entries table.
+	 *
+	 * @since 1.0
+	 *
+	 * @param array<string> $columns Entries table columns.
+	 *
+	 * @return array<string>
+	 */
+	public static function add_status_column( $columns ) {
```
stephywells, line 137: Looks like we already have a 'is_draft' column. Let's use that instead of adding a new column. We can change the name of the existing column too.

stephywells (changes requested):
Thanks @omidakhavans ! This is headed in the right direction, but could use a few additions and changes.

Looks like this line in the FrmFormActionsController could also use an update:
if ( empty( $entry ) || ( $entry->is_draft && $event != 'draft' ) ) {

And maybe the get_is_draft_value() function in FrmEntry.

This is in FrmEntryMeta. Looks like it may need to run through your new get_entry_status() function:
```
if ( ! $args['is_draft'] ) {
				$where['e.is_draft'] = 0;
			} elseif ( $args['is_draft'] == 1 ) {
				$where['e.is_draft'] = 1;
			}
```

And:
```
if ( ! $args['is_draft'] ) {
			$draft_where = $wpdb->prepare( ' AND e.is_draft=%d', 0 );
		} elseif ( $args['is_draft'] == 1 ) {
			$draft_where = $wpdb->prepare( ' AND e.is_draft=%d', 1 );
		}
```

I only searched in the Lite version, so we'll also need to check Pro for anything that needs to run through these new functions. Could you also search Pro for "is_draft" to see what needs to change there?

omidakhavans:
@stephywells Yes, I'm already aware of the Pro plugin and other add-ons too. Just wanted to get confirmation on the approach so I could proceed with the others. Thank you.

classes/helpers/FrmEntriesHelper.php
omidakhavans: My bad it seems, I didn't commit the abandoned part and my test value remains for notice but yes `$extended_entry_status = array();` this is fixing the fatal. Thank you very much.

classes/controllers/FrmEntriesController.php
omidakhavans: Yes, I already knew that but my approach in this case was : implement the `apply_filters( 'frm_entries_add_status_column', false ) `to disable entry status column and make the entry statuses hidden by default and enable this column by other add-ons which adds new statuses when is_draft has more values. So I will delete that filter and enable the entry status column by default for newer versions. Like you said.

classes/views/frm-entries/_sidebar-shared-pub.php
omidakhavans: Sure, Thank you. It seems you reviewed the older version! However I'm sure I have pushed my changes and there is no comma there since my IDE warns me about php 5.6.

classes/helpers/FrmEntriesHelper.php
omidakhavans: Sure. Thank you.

classes/helpers/FrmEntriesHelper.php
omidakhavans: Yes, I could use i18n only here. Thank you very much sorry for this mistake.

classes/helpers/FrmEntriesHelper.php
omidakhavans: True. I used 2 methods before. but merged it for simplicity but you are right this is better. Thank you very much.

classes/helpers/FrmEntriesHelper.php
omidakhavans: Sure, Done, but In progress is something that we may use in the future by other add-ons and it's not something that has only use cases in abandonment. I decided to place it here. I recommend if we have a plan to use in progress in future in other addons, let's keep this in progress here otherwise I could add this status in abandonment. What do you think? Please let me know to make it back in case we have a plan to use it across other add-ons. Thank you.

classes/controllers/FrmEntriesController.php
omidakhavans: I have replied this in the other comment blew https://github.com/Strategy11/formidable-forms/pull/1237#discussion_r1263678424 Following your comment in blew I'm deleting part of this code.

classes/helpers/FrmEntriesHelper.php
stephywells: I don’t have anything in mind that would use that in progress status right now, so I think we can keep it in the add-on to keep it simple.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +747,99 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+	 * @since x.x
+	 *
+	 * @param int $status is_draft column.
+	 *
+	 * @return string
+	 */
+	public static function get_entry_status_label( $status ) {
+		$statuses = self::get_entry_statuses();
+
+		return isset( $statuses[ $status ] ) ? $statuses[ $status ] : $statuses[1];
```
stephywells, line 787: I think we would need to run this through get_entry_status( $status ) first so we are following the same logic.

`$status = self::get_entry_status( $status ) ;`

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +747,99 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+		 */
+		$extended_entry_status = apply_filters( 'frm_register_entry_statuses', array() );
+
+		if ( ! is_array( $extended_entry_status ) ) {
+			_doing_it_wrong( __METHOD__, esc_html__( 'Entry status must be return in array format.', 'formidable' ), 'x.x' );
+			$extended_entry_status = array();
+		}
+
+		// Check for registered statuses.
+		$registered_entry_statuses = get_option( 'frm_entry_statuses', array() );
```
stephywells, line 821: I'm not quite sure why this is needed. Can you explain why we need a DB call here in addition to the hook? This looks like extra complication and weight to me. What is the benefit of saving it?

classes/controllers/FrmEntriesController.php
stephywells: I think that may be overcomplicating it. Instead of using yes or no in the is_draft column, let's show the status. This way, we don't need two columns.

stephywells (changes requested):
Thank you! This is much more simplified and easy to read now.

I think we can simplify it even further. The more simple, the less hiding places for bugs.

classes/controllers/FrmEntriesController.php
```diff
@@ -107,6 +109,19 @@ public static function manage_columns( $columns ) {
...
+	/**
+	 * Add "Entry status" column to entries table.
+	 *
+	 * @since x.x
+	 *
+	 * @param array<string> $columns Entries table columns.
+	 *
+	 * @return array<string>
+	 */
+	public static function entry_status_column( $form_id, &$columns ) {
```
Crabcyborg, line 121: @omidakhavans This doesn't need to be public.

I also don't think this function needs to be a function. We could just inline this in `manage_columns` and it would make it easier to follow what is happening there.

classes/helpers/FrmEntriesHelper.php
Crabcyborg, line 787: `$statuses[1]` would seem like like a magic value if you replaced `1` with `self::DRAFT_ENTRY_STATUS`

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +747,99 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+		/**
+		 * Register entry status.
+		 *
+		 * "3" is reserved for abandonment-addon.
+		 *
+		 * @since x.x
+		 *
+		 * @param array<string> $extended_entry_status Entry statuses.
+		 */
+		$extended_entry_status = apply_filters( 'frm_register_entry_statuses', array() );
```
Crabcyborg, line 813: I'm not sure if we need the `register` in the name of this filter. It seems unnecessary.

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -287,7 +287,16 @@ private function column_value( $item ) {
 				$val  = '<abbr title="' . esc_attr( FrmAppHelper::get_formatted_time( $item->{$col_name}, '', 'g:i:s A' ) ) . '">' . $date . '</abbr>';
 				break;
 			case 'is_draft':
```
Crabcyborg, line 289: Do we still want the `is_draft` column? We should be replacing it with with entry status, not keeping it and having both options, right?

classes/helpers/FrmEntriesHelper.php
omidakhavans: Sorry for the late reply, it seems my email doesn't deliver mentions anymore. 

Sure, I changed this in my commits days before. Sure, When we deactivate or disable the entry statuses hook which is currently abandonment add-on we will lose the statuses which we submitted entries based on those extra statuses! So in processes like import/export, etc. we would lose the is_draft column value so it may introduce a bug to save the in_progress abandoned or whatever status entries as submitted.

classes/helpers/FrmEntriesListHelper.php
omidakhavans: I thought about this and eventually I decided to merge it with the entry_status column, and I want to keep this because if the pro add-on didn't update to the latest version that would make the table look weird. https://github.com/Strategy11/formidable-forms/pull/1237/commits/c2666b8277a549a006eab92a0de30f8e701da3e0

If you think It's ok please let me know I could delete it.

classes/helpers/FrmEntriesHelper.php
omidakhavans: Done. Thank you.

classes/helpers/FrmEntriesHelper.php
omidakhavans: Sure. Thank you.

classes/controllers/FrmEntriesController.php
omidakhavans: Yes. Sure Thank you.

classes/helpers/FrmEntriesHelper.php
stephywells: I see. The is_draft value wouldn't change in the database, right? What if we adjust the import/export to use the exact saved value? Would that solve the problem? In other places, that is why we fallback to saying it's a draft entry. Can you think of any other possible issues with this way, and removing the DB part?

css/frm_admin.css
```diff
@@ -4039,6 +4039,35 @@ label input[type="checkbox"], label input[type="radio"] {
 	font-size: inherit;
 }
 
+/* Start entry statuses */
+.frm-entry-status {
+	padding: 4px 8px;
+	align-items: center;
+	border-radius: 4px;
```
stephywells, line 4046: @omidakhavans Let's use css vars where possible so this will match better now and going forward. This works for me for a closer match:

```
.frm-entry-status {
    padding: var(--gap-2xs) var(--gap-xs);
    border-radius: var(--small-radius);
}
```

@tuguirazvan do you have any recommendations for this? I'm having a bit of trouble with adding this style when we have a different style for the application tags on the form listing page. Should we stick with this design? https://www.figma.com/file/AkQRRLqb8qcf1S5APxyzsG/Formidable-Forms-Design-System?node-id=3387%3A30550&mode=dev

classes/helpers/FrmEntriesListHelper.php
stephywells: I'm still not seeing the benefit to adding a new entry_status column and removing is_draft. Can you help me understand this?

stephywells (changes requested):

classes/helpers/FrmEntriesHelper.php
omidakhavans: Right, The submitted entry like "in progress" etc would still remain the corresponding indexes 2, 3 in is_draft column when hook is disabled.
Yes we could adjust import/export to export exactly the same values.

`Can you think of any other possible issues with this way, and removing the DB part?`

This was the reason I have spent time testing each part we have utilized is_draft, still I can not 100% grantee that I could succeed to discover the possible issues! That was the reason I made this decision. This is much safer than the other ways.

Also may I know what's the downside of using `get_option( 'frm_entry_statuses', array() )`? If CRUD time execution is the problem I could provide the benchmark and it wouldn't hurt to have this.

omidakhavans: BTW 
image

We have many functions that need the label of the status and when we make them work with only a filter and when we lose them hook by disabling the plugin or whatever reason they will return nothing! or whatever we hardcoded or false information which will damage to user trust since entry status is one of the important entity of our product! Also if we hard code something what's the benefit of making entry status extendable?

garretlaxton:
This looks good on my end. Let me know if anything changes.

classes/helpers/FrmEntriesHelper.php
stephywells: I think if the export keeps the actual value, it would be just fine if in progress and abandoned entries become drafts. I haven't tested this, but it looks to me like the saved array would be updated without the extra statuses when the from abandonment is disabled. 

If Abandoned is no longer an available status, it would show as draft in the status column. The code left behind wouldn't know how to handle in progress and abandoned entries, so I think it would make the most sense to tell the plugin to treat them as drafts if the status no longer exists.

css/frm_admin.css
omidakhavans: Sure, The reason was I wanted to follow the Figma. We don't have the exact vars for radius however I could --gap-2xs for it Also the tags color. Your point about tag integrity with the project design system is my question too. Please let me know what I should do.

stephywells: Using a gap css var wouldn't keep it the same. Let's use the radius var for corners so it'll be consistent throughout the plugin. We can wait on Razvan's input on what we need to change beyond that.

classes/helpers/FrmEntriesListHelper.php
omidakhavans: Dear @stephywells I confirmed that there is no benefit in changing the is_draft to entry_status. I saw you have marked the other code in some other comments and I misunderstood and missed the point. my apologies. Now it's revert back the is_draft as main https://github.com/Strategy11/formidable-forms/pull/1237/commits/05f4d01c2958949816854f791d9491afc7a2f813 also in Pro pr  https://github.com/Strategy11/formidable-pro/pull/4352/files I have removed the is_draft column hook too. So we could have the entry status in Lite.

stephywells: Thank you for updating that! I think this will be better for reverse compatibility. 

Also, it looks like some extra space got left behind in lines like this:
`$form_id . '_is_draft' => 'is_draft',`

classes/helpers/FrmEntriesHelper.php
stephywells: Right. But why? What is the benefit of showing those statuses if they are still handled the same way?

omidakhavans: Yes the code left behind wouldn't know how they treat the abandoned. But they do know how to submit it right? When an entry is abandoned the only way would be editing the entry and submitting it! So status will be changed!

stephywells: How is that different from the regular draft status?

omidakhavans: In action nothing. They are both considered as no submitted entries. But for a customer it may become a problem and questions. Also false information! when it's an abandoned entry but now it's considered as a draft!

omidakhavans: The one other point is for the reports page What we should consider an abandoned entry there?

omidakhavans: What if we consider a feature request to make a quick edit of the status on the table?
What if customers made views for abandoned entries? When they are displayed as draft at once.

stephywells: If we no longer have Form Abandonment, then I see all of this as a non-issue. For example, if someone has a date field and switches to Lite, all date fields are then text fields. Any field type that isn't supported in Lite becomes a different field type that is supported in Lite. It's more about a graceful fallback. 

If the abandonment plugin is not installed, we wouldn't consider abandoned entries in the reports at all. We would treat them as drafts. As long as we keep the 2 or 3 in the DB, the status would show again once the code is there to support it.

stephywells: As for the quick edit option, if they do a quick edit and select to change all selected entries to draft, then it would change. Otherwise, it would remain unchanged, just like in the WP quick edit.

omidakhavans: Sure I will change the code accordingly, Thank you very much for helping me with this.

omidakhavans: @stephywells Changes has been made to the PRs and I'm updating this PR description. Thank you.

tuguirazvan: @omidakhavans @stephywells 

It seems like those tags are outdated anyways. I haven't redesigned those up until now, because we redesigned the way form templates are presented, thus the need to show the "Business+" tag won't be needed once we ship the redesign. 

But, until then, I quickly redesigned the tags here:
https://www.figma.com/file/DQbZMgeNmJWkxOWYfHjDZX/Formidable-System?type=design&node-id=557%3A2789&mode=design&t=yu0yuIHkKOJHgvwH-1

As you will see, the "business+" tag and the "abandoned" ones are not using the same color, which I think was the issue you ran into here. 

And what are the statuses we'll be using? Only submitted, draft and abandoned? 

 Here's a screenshot of what's inside the Figma file
 
CleanShot 2023-08-03 at 16 04 23@2x

classes/controllers/FrmEntriesController.php
```diff
@@ -277,12 +278,12 @@ public static function sortable_columns() {
 		$fields  = FrmField::get_all_for_form( $form_id );
 
 		$columns = array(
-			$form_id . '_id'         => 'id',
-			$form_id . '_created_at' => 'created_at',
-			$form_id . '_updated_at' => 'updated_at',
-			$form_id . '_ip'         => 'ip',
-			$form_id . '_item_key'   => 'item_key',
-			$form_id . '_is_draft'   => 'is_draft',
+			$form_id . '_id'            => 'id',
```
truongwp, line 281: We don't need this extra space for these 6 lines.

classes/controllers/FrmFormActionsController.php
```diff
@@ -505,7 +505,7 @@ public static function trigger_actions( $event, $form, $entry, $type = 'all', $a
 				$entry = FrmEntry::getOne( $entry, true );
 			}
 
-			if ( empty( $entry ) || ( $entry->is_draft && $event != 'draft' ) ) {
+			if ( empty( $entry ) || ( FrmEntriesHelper::DRAFT_ENTRY_STATUS === $entry->is_draft && 'draft' != $event ) ) {
```
truongwp, line 508: I think we should convert `$entry->is_draft` to `int` when comparing with `DRAFT_ENTRY_STATUS`.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +747,79 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
 		}
 	}
+
+	/**
+	 * Return entry status based on is_draft column value.
+	 *
+	 * @since x.x
+	 *
+	 * @param int $status is_draft column.
+	 * @param bool $include_label Return label.
```
truongwp, line 757: We don't have this param anymore.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +747,79 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+	 *
+	 * @return string
```
truongwp, line 759: This might return `int` too.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +747,79 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+	 */
+	public static function get_entry_status( $status ) {
+		$statuses = self::get_entry_statuses();
+
+		if ( array_key_exists( $status, $statuses ) ) {
+			return $status;
+		}
+
+		if ( empty( $status ) ) {
+			return self::SUBMITTED_ENTRY_STATUS; //If the status is empty, let's default to 0.
```
truongwp, line 769: We should have a space after `//`.

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -287,7 +287,13 @@ private function column_value( $item ) {
...
-				$val = empty( $item->is_draft ) ? esc_html__( 'No', 'formidable' ) : esc_html__( 'Yes', 'formidable' );
+				$entry_status = FrmEntriesHelper::get_entry_status_label( $item->is_draft );
+				$val = sprintf(
+					/* translators: %1$s: Status class name %2$s: Status name */
+					'<span class="frm-entry-status frm-entry-status-%s">%s</span>',
+					sanitize_html_class( strtolower( str_replace( ' ', '-', $entry_status ) ) ),
```
truongwp, line 294: I think we should use the entry status key instead of the label here. Because the label is translatable, the styling won't work if users translate the label.

omidakhavans: Nice point. Thank you.

classes/controllers/FrmEntriesController.php
omidakhavans: Thank you. https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#indentation

classes/controllers/FrmFormActionsController.php
omidakhavans: Yes, Thank you very much.

classes/helpers/FrmEntriesHelper.php
omidakhavans: Thank you

classes/helpers/FrmEntriesHelper.php
omidakhavans: Thank you very much

classes/controllers/FrmEntriesController.php
stephywells: I'm still seeing this. An extra 3 spaces have been added on each line.

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -287,7 +287,13 @@ private function column_value( $item ) {
 				$val  = '<abbr title="' . esc_attr( FrmAppHelper::get_formatted_time( $item->{$col_name}, '', 'g:i:s A' ) ) . '">' . $date . '</abbr>';
 				break;
 			case 'is_draft':
-				$val = empty( $item->is_draft ) ? esc_html__( 'No', 'formidable' ) : esc_html__( 'Yes', 'formidable' );
+				$entry_status = FrmEntriesHelper::get_entry_status_label( $item->is_draft );
+				$val = sprintf(
+					/* translators: %1$s: Status class name %2$s: Status name */
```
stephywells, line 292: This line isn't needed since it isn't followed by a translatable string.

images/icons.svg
```diff
@@ -843,5 +843,6 @@
...
 <symbol id="frm_sample_form_icon" viewBox="1.5 1.5 16.8 16.8" fill="none"><path stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" d="M15.833 2.5H4.167c-.92 0-1.667.746-1.667 1.667v11.666c0 .92.746 1.667 1.667 1.667h11.666c.92 0 1.667-.746 1.667-1.667V4.167c0-.92-.746-1.667-1.667-1.667ZM2.5 7.5h15M7.5 17.5v-10" stroke-width="1.5"/></symbol>
 <symbol id="frm_globe_icon" viewBox="1.1 1.1 21.8 21.8" fill="none"><path d="M12 22a10 10 0 1 0 0-20 10 10 0 0 0 0 20ZM2 12h20" stroke="currentColor" stroke-width="1.8" stroke-linecap="round" stroke-linejoin="round"/><path d="M12 2c2.5 2.7 4 6.3 4 10s-1.5 7.3-4 10c-2.5-2.7-4-6.3-4-10s1.5-7.3 4-10v0Z" stroke="currentColor" stroke-width="1.8" stroke-linecap="round" stroke-linejoin="round"/></symbol>
+<symbol id="frm_abandoned_icon" viewBox="0 0 20 20" fill="none"><g stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" clip-path="url(#a)"><path d="M7 19H3a2 2 0 0 1-2-2V3a2 2 0 0 1 2-2h4M14 15l5-5-5-5M19 10H7"/></g></symbol>
```
stephywells, line 846: Looks like clip-path="url(#a)" should be removed here. I'm not seeing anything to match this.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -731,4 +747,78 @@ public static function maybe_render_captcha_score( $entry_id ) {
...
+
+		$default_entry_statuses = array(
+			self::SUBMITTED_ENTRY_STATUS => __( 'Submitted', 'formidable' ),
+			self::DRAFT_ENTRY_STATUS     => __( 'Draft', 'formidable' ),
+		);
+
+		/**
+		 * Register entry status.
+		 *
+		 * "3" is reserved for abandonment-addon.
```
stephywells, line 806: Let's also mention that 2 is used in abandonment for "in progress".

stephywells (changes requested):
This is so close! We're just down to the minor details now. Nice job!

classes/helpers/FrmEntriesHelper.php
stephywells: Thanks @tuguirazvan . My concern is that we have a different tag for applications. https://www.figma.com/file/zGdcUmMi0KOF8BNAqbVZOa/Applications?type=design&node-id=494-18123&mode=design&t=TDw8YjzwJkQPY5pM-0
Screenshot 2023-08-04 at 1 59 13 PM

Do you think we should make these new tags match the existing ones?

classes/controllers/FrmFormActionsController.php
```diff
@@ -505,7 +505,7 @@ public static function trigger_actions( $event, $form, $entry, $type = 'all', $a
 				$entry = FrmEntry::getOne( $entry, true );
 			}
 
-			if ( empty( $entry ) || ( $entry->is_draft && $event != 'draft' ) ) {
+			if ( empty( $entry ) || ( FrmEntriesHelper::DRAFT_ENTRY_STATUS === (int) $entry->is_draft && 'draft' != $event ) ) {
```
Crabcyborg, line 508: We could do a strict string comparison here so it's `'draft' !== $event`.

images/icons.svg
omidakhavans: Sure. Thank you.

classes/controllers/FrmFormActionsController.php
omidakhavans: Yes, I didn't implement it since I'm not sure if this could start a bug or not. That part is the old code.

classes/helpers/FrmEntriesListHelper.php
omidakhavans: Sure, Thank you.

classes/helpers/FrmEntriesHelper.php
omidakhavans: Done, Thank you.

classes/controllers/FrmFormActionsController.php
Crabcyborg: Any string comparisons are fine to make strict. Nothing other than 'draft' would ever match 'draft' with a soft check.

omidakhavans: Done, Thank you.

Crabcyborg (approved):
This looks perfect now. Thanks @omidakhavans!

And thank you @truongwp @stephywells @tuguirazvan and @garretlaxton for all of your input on this update as well!

🚀

PR #1241, opened 2023-07-17
Required for Date calculation

truongwp:
This PR includes:
- Ignore `formidableforms1.css` file.
- Improve `dom.tag()` to accept a string as one of children.
- Add JS action when show inline modal (in the backend).
- Add `inside_class` to the inline modal args (to fix the ugly styling of nested modals inside the `<div class="inside">` tag).
- Bump version for testing.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -16,7 +16,7 @@ class FrmAppHelper {
 	/**
 	 * @since 2.0
 	 */
-	public static $plug_version = '6.4-date-calculation';
+	public static $plug_version = '6.4.1';
 
 	/**
 	 * @since 1.07.02
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -16,7 +16,7 @@ class FrmAppHelper {
 	/**
 	 * @since 2.0
 	 */
-	public static $plug_version = '6.4';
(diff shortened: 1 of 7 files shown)


truongwp:
@Crabcyborg These are required codes for Date calculation. The PR in Pro hasn't been ready yet. But I think this is ready for review. Can you take a look when you have time?

Crabcyborg:
@truongwp These changes all look fine to me!

classes/helpers/FrmAppHelper.php
```diff
@@ -1024,6 +1024,7 @@ public static function kses_icon( $icon ) {
 	public static function add_allowed_icon_tags( $allowed_html ) {
 		$allowed_html['svg']['data-open'] = true;
 		$allowed_html['svg']['title']     = true;
+		$allowed_html['svg']['data-fid']  = true;
```
truongwp, line 1027: @Crabcyborg Do you think we should put a filter here then add `data-fid` in the dates plugin?

```
$allowed_svg_attrs = apply_filters( 'frm_allowed_svg_icon_attrs', array( 'data-open', 'title' ) );
```

truongwp: Using `frm_striphtml_allowed_tags` works. I updated the code. Thank you.

Crabcyborg (approved):
I'll merge this as well so we're ready for a datepicker release.

It looks good to me!

🚀

PR #1242, opened 2023-07-18
Add a submit element html reset button

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4335

This update just adds a reset button to Submit element html box to allow resetting the value.

Screenshot:
image

### DIFF
--- a/classes/controllers/FrmHooksController.php
+++ b/classes/controllers/FrmHooksController.php
@@ -143,6 +143,9 @@ class FrmHooksController {
 		add_action( 'admin_head-toplevel_page_formidable', 'FrmFormsController::head' );
 		add_action( 'frm_after_field_options', 'FrmFormsController::logic_tip' );
 
+		// Forms Helper.
+		add_action( 'wp_ajax_frm_get_default_submit_html', 'FrmFormsHelper::get_default_submit_html' );
+
 		add_filter( 'set-screen-option', 'FrmFormsController::save_per_page', 10, 3 );
 		add_action( 'admin_footer', 'FrmFormsController::insert_form_popup' );
 
--- a/classes/controllers/FrmHooksController.php
+++ b/classes/controllers/FrmHooksController.php
@@ -143,9 +143,6 @@ class FrmHooksController {
 		add_action( 'admin_head-toplevel_page_formidable', 'FrmFormsController::head' );
 		add_action( 'frm_after_field_options', 'FrmFormsController::logic_tip' );
 
-		// Forms Helper.
-		add_action( 'wp_ajax_frm_get_default_submit_html', 'FrmFormsHelper::get_default_submit_html' );
-
 		add_filter( 'set-screen-option', 'FrmFormsController::save_per_page', 10, 3 );
 		add_action( 'admin_footer', 'FrmFormsController::insert_form_popup' );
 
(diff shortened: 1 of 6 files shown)


garretlaxton (approved):
This works great!

classes/controllers/FrmHooksController.php
```diff
@@ -143,6 +143,9 @@ public static function load_admin_hooks() {
 		add_action( 'admin_head-toplevel_page_formidable', 'FrmFormsController::head' );
 		add_action( 'frm_after_field_options', 'FrmFormsController::logic_tip' );
 
+		// Forms Helper.
+		add_action( 'wp_ajax_frm_get_default_submit_html', 'FrmFormsHelper::get_default_submit_html' );
```
Crabcyborg, line 147: @AbdiTolesa Since this is an AJAX hook it should be in the `load_ajax_hooks` function.

classes/helpers/FrmFormsHelper.php
```diff
@@ -443,6 +443,19 @@ public static function get_default_html( $loc ) {
 		return $default_html;
 	}
 
+	/**
+	 * @since x.x
+	 *
+	 * @return void
+	 */
+	public static function get_default_submit_html() {
+		check_ajax_referer( 'frm_ajax', 'nonce' );
```
Crabcyborg, line 452: We should also add a permission check.

Probably `FrmAppHelper::permission_check( 'frm_edit_forms' );`. Someone without edit permissions shouldn't need this, right?

js/formidable_admin.js
```diff
@@ -10188,6 +10188,23 @@ function frmAdminBuildJS() {
 			jQuery( document ).on( 'frm-action-loaded', onActionLoaded );
 
 			initOnSubmitAction();
+
+			document.getElementById( 'frm_reset_submit_html' ).addEventListener( 'click', ( e ) => {
+				jQuery.ajax({
```
Crabcyborg, line 10193: Is it possible to use `frmDom.ajax.doJsonPost` or `frmDom.ajax.doJsonFetch` instead?

It will make it easier to remove jQuery eventually if we don't add more uses of `jQuery.ajax` to our code.

AbdiTolesa: @Crabcyborg I have addressed all the concerns now.

classes/controllers/FrmHooksController.php
```diff
@@ -258,6 +258,9 @@ public static function load_ajax_hooks() {
 
 		// Submit with AJAX.
 		add_action( 'wp_loaded', 'FrmEntriesAJAXSubmitController::ajax_create', 5 ); // Trigger before process_entry.
+
+		// Forms Helper.
+		add_action( 'wp_ajax_frm_get_default_submit_html', 'FrmFormsHelper::get_default_submit_html' );
```
Crabcyborg, line 263: Since this is an AJAX action, we should really be calling `FrmFormsController` instead of `FrmFormsHelper`.

Crabcyborg: I think this is in the wrong file.

Crabcyborg:
So it turns out that you can clear the submit button HTML and it will reset to defaults then.

That seems like an issue. We were adding this action because I believed that we were supporting a blank input for submit button HTML (so that the button could be easily removed).

This update may not be necessary after all.

shervElmi:
@Crabcyborg, thank you for working on this and for your review. Do I need to review this PR?

Crabcyborg:
No @shervElmi I don't believe we'll be moving forward with this update right now.

PR #1243, opened 2023-07-19
Display "(no title)" for Forms Without a Title in Formidable Block

shervElmi:
For forms without a title, the Formidable block used to display a blank space. This PR updates this behavior to show "(no title)" instead.

## Related Issue:
https://github.com/Strategy11/formidable-pro/issues/4261

## Testing Instructions:
2. Create a form without a title.
3. Add this form using the Formidable block to a page.
4. Check that forms without titles now display "(no title)".


garretlaxton (approved):
This looks great!

js/src/form/block.js
```diff
@@ -28,7 +28,10 @@ registerBlockType( 'formidable/simple-form', {
 	edit: function( { setAttributes, attributes, isSelected } ) {
 		const { formId } = attributes;
 
-		const forms = formidable_form_selector.forms;
+		const forms = formidable_form_selector.forms.map( form => ({
+			...form,
+			label: form.label.trim() === '' ? '(no title)' : form.label
```
AbdiTolesa, line 33: @shervElmi Thank you for working on this! I think we can use a bit of i18n here.

AbdiTolesa (changes requested):

js/src/form/block.js
shervElmi: @AbdiTolesa, thanks for your review. In this case, the label `(no title)` indicates a form for which the user hasn't set a name, so we might not need translatable text.

@Crabcyborg, what are your thoughts on Abdi's input?

AbdiTolesa: > @AbdiTolesa, thanks for your review. In this case, the label `(no title)` indicates a form for which the user hasn't set a name, so we might not need translatable text.

@shervElmi We have been doing that for the same text in all of our js scripts and PHP and I believe it is worth being consistent.

Crabcyborg: @shervElmi We translate most strings. I think `(no title)` would make sense getting translated to the active language as well.

Crabcyborg:
I've updates this to use the PHP `FrmFormsHelper::edit_form_link_label` function instead. This feels more consistent. If we ever changed the "(no title)" string we don't need to change it in as many places.

shervElmi:
Thanks so much, @Crabcyborg for working on this! You truly are the best! ❤️

PR #1244, opened 2023-07-20
Add horizontal margin to form preview when template is not found

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4223

image

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -397,6 +397,7 @@ class FrmFormsController {
 		add_filter( 'is_active_sidebar', '__return_false' );
 		FrmStylesController::enqueue_css( 'enqueue', true );
 
+		$p = get_query_template( 'page' );
 		if ( false === get_template_part( 'page' ) ) {
 			self::fallback_when_page_template_part_is_not_supported_by_theme();
 		}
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This looks good!

css/custom_theme.css.php
```diff
@@ -110,6 +110,10 @@
 <?php } ?>
 }
 
+.frm-preview {
```
Crabcyborg, line 113: @AbdiTolesa We don't want to add any special preview code into our front end CSS. We want the CSS to be as small as possible.

Crabcyborg: I also don't love that this is always applied. This update should really only change how a block template is rendered.

classes/controllers/FrmFormsController.php
```diff
@@ -397,6 +397,7 @@ private static function load_theme_preview() {
 		add_filter( 'is_active_sidebar', '__return_false' );
 		FrmStylesController::enqueue_css( 'enqueue', true );
 
+		$p = get_query_template( 'page' );
```
Crabcyborg, line 400: @AbdiTolesa What is up with this line? `$p` is never used.

shervElmi:
@AbdiTolesa, I'm relatively new to the FF team. Could you please provide the steps to reproduce the issue in this PR?
Detailed steps would be incredibly helpful and time-saving for me. Thank you!

Crabcyborg:
I'm going to close this for now as I don't think we'll re-use any of these updates in the final solution.

PR #1258, opened 2023-08-02
Add phpcs check for php 5.6 and 8.1

stephywells:
I noticed when reviewing another PR that there was a php 5.6 syntax error that didn't get caught. @Crabcyborg do you want to make any modifications to this workflow before we merge it?


Crabcyborg (approved):
This looks fine. It might be nicer with a matrix but it isn't a big deal if this is copied/pasted.

PR #1261, opened 2023-08-10
get server os

Liviu-p:


Liviu-p:
> Thanks for working on this @Liviu-p!
> 
> Just one more thing. We're referencing `php_uname` in this plugin still (
> 
> https://github.com/Strategy11/formidable-forms/blob/29237b1305b593cd20b5ae0e03428da17e9bff9e/classes/models/FrmUsage.php#L73
> 
> ).
> We should be calling `FrmAppHelper::get_server_os` there as well or there would still be fatal errors when the function isn't available.

Thank you for noticing it @Crabcyborg. Totally agree,  I've just updated it.

Crabcyborg (approved):
Thanks @Liviu-p for updating that! This looks good now.

🚀

PR #1262, opened 2023-08-14
save and reload fix when form is submited via ajax

Liviu-p:


Crabcyborg (approved):
This solution looks good. Thanks @Liviu-p!

garretlaxton (approved):
This works great!

Crabcyborg (approved):
This looks good.

Thanks @Liviu-p and @garretlaxton!

PR #1266, opened 2023-08-17
Pro issue 4302

Liviu-p:


js/formidable_admin.js
```diff
@@ -7697,8 +7698,21 @@ function frmAdminBuildJS() {
 		}
 	}
 
+	function reloadIfAddonActivatedAjaxSubmitOnly() {
```
Crabcyborg, line 7701: This isn't part of this update. Branches should be made from master, not from other unmerged branches.

It makes it more difficult to review and there can be unintended mistakes.

js/formidable.js
```diff
@@ -1882,3 +1882,10 @@ function frm_resend_email( entryId, formId ) { // eslint-disable-line camelcase
 		}
 	});
 }
+
+( function() {
+	// Elementor popup show event. Fix Elementor Popup && FF Captcha field conflicts
+	jQuery( document ).on( 'elementor/popup/show', () => {
```
truongwp, line 1888: Can we move this to inside of `frmFrontFormJS()`? We don't need an extra `( function() {}() )` here.

Liviu-p: sure thing @truongwp, I've just updated it.

Crabcyborg (approved):
I just changed the `frmCaptcha` function back to what it was. I don't think this change was necessary.

Trying to initialize all `.g-recaptcha` could possibly cause conflicts other other captchas. Only checking the `frm-` prefixed class is safer.

I didn't notice any difference when changing this back.

I also reduced the code a bit. I removed the arrow function. It's down to a single line update which is nice.

Thanks for working on this @Liviu-p!

And thank you @truongwp for your help reviewing it as well.

PR #1272, opened 2023-08-25
Fix whitelabel name and icon of Modal block

truongwp:
This is related to https://github.com/Strategy11/formidable-modal/issues/26


Crabcyborg (approved):
Thanks @truongwp! This looks great!

PR #1276, opened 2023-08-31
Offset verication field vertically

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-wpml/issues/104

The issue is caused by the positioning of `frm_verify` element. I believe we can offset it vertically to avoid the extra 3000px wide whitespace.

### DIFF
--- a/css/frm_grids.css
+++ b/css/frm_grids.css
@@ -17,7 +17,7 @@
 
 .frm_verify{
 	position:absolute;
-	left:-3000px;
+	top:-3000px;
 }
 
 .frm_clear_none{


garretlaxton (approved):
This looks good for me!

css/frm_grids.css
```diff
@@ -17,7 +17,7 @@
 
 .frm_verify{
 	position:absolute;
-	left:-3000px;
+	top:-3000px;
```
Crabcyborg, line 20: In a tall enough page I think the honeypot could end up being visible if we used top.

Does `right` work?

AbdiTolesa: @Crabcyborg `right` is not working for me, I tried in both directions and I still see issues with that in responsive. I didn't get how the height of the page could reveal the honeypot if used `top` since the element looks to be placed absolute with reference to the body. I tried making my test page very long and it doesn't seem to pull down the honeypot. Should we make the offset much bigger than this?

shervElmi: @AbdiTolesa or @garretlaxton, could you please provide me with a QA URL for further testing? I'm encountering some difficulties in reproducing the issue on my end. Thank you!

AbdiTolesa: @shervElmi There you go: https://qa.formidableforms.com/abdi/ar/translation-to-arabic-responsive-test/

shervElmi (approved):
Thank you for working on this, @AbdiTolesa.

css/frm_grids.css
shervElmi: Thanks so much!

Crabcyborg: I'm just not sold on this solution, especially not in Lite for every form. It's fairly likely that the honeypot could become visible this way.

The conflict we're trying to fix is a really uncommon use case. I'd maybe consider this fix only under certain conditions and keep the default behaviour for other forms.

I'm thinking it may be worth trying to an alternative way to hide the honeypot field.

For now we can probably just hold on this fix. I don't think this solution is worth merging.

Crabcyborg:
I'm going to close this for now. I don't think this is worth fixing this way.

PR #1278, opened 2023-09-04
Form abandonment settings placeholder

omidakhavans:
### Summary
This PR is following [this](https://github.com/Strategy11/formidable-abandonment/pull/1#issuecomment-1702641053) comment by @tuguirazvan to change the form abandonment icon in form settings. also adds a placeholder for form abandonment on lite.

image

### Checklist
- [x] Change abandonment icon in form settings.
- [x] Add abandonment placeholder in form settings.


omidakhavans:
@tuguirazvan Yes the problem was we have renamed those options so I couldn't export them. Thank you very much. 

That change is already done in the other branch.

classes/controllers/FrmFormsController.php
```diff
@@ -1276,9 +1276,22 @@ private static function get_settings_tabs( $values ) {
...
 			),
+			'abandonment'   => array(
+				'name'       => __( 'Form Abandonment', 'formidable' ),
+				'icon'       => 'frm_icon_font frm_abandoned_icon',
+				'html_class' => 'frm_show_upgrade_tab frm_noallow',
+				'data'       => FrmAppHelper::get_upgrade_data_params(
+					'abandonment',
+					array(
+						'upgrade'    => __( 'Form Abandonment', 'formidable' ),
+						'message'    => __( 'Capture form data before it\'s submitted to save more leads and learn to optimize forms.', 'formidable' ),
```
omidakhavans, line 1287: @stephywells @Crabcyborg  The plugin description is written by myself. Do you confirm?
'Capture form data before it\'s submitted to save more leads and learn to optimize forms.'

garretlaxton:
This looks great, but some of the text needs to be changed. I think it would be better to use activated instead of installed since the plugin is already installed, just not activated.

`Form Abandonment are not installed` to `Form Abandonment is not activated`

`That add-on is not installed. Would you like to install it now? ` to `This add-on is not activated. Would you like to activate it now?`

omidakhavans:
@garretlaxton This is not part of my commits. I will follow this in a different branch since it may be that this view is not utilized here only.

garretlaxton (approved):
This looks good then.

tuguirazvan (approved):
This looks okay to me. I see that you changed the icon and added the placeholder image there. 

I agree with Garret about the copy that needs to be changed. Thanks @omidakhavans for working on this!

classes/controllers/FrmFormsController.php
stephywells: Here's what ChatGPT gave me:
> Unlock the power of data capture to boost lead generation and master the art of form optimization.

Here's my prompt so you can do the same in other places:

> Keep this short, but make it sound more appealing.
> "Capture form data before it's submitted to save more leads and learn to optimize forms."

Since "are" is trickier to change, let's adjust this to sound right. In other places we added "settings", like "Form scheduling settings are not installed".

omidakhavans: Thank you very much. Done.

stephywells: This hadn't been done so I went ahead and updated it.

stephywells (approved):
This looks ready to merge now.

Crabcyborg (approved):

PR #1279, opened 2023-09-04
Override tinymce full screen position when admin sidebar is shown

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4417

### Testing

- Add code snippet to show admin menu
- Add an HTML field to a form and try going full screen

### Before

image

### After
image

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -140,7 +140,7 @@ a, .widget .widget-top, .stuffbox h3, .frm-collapsed {
 	display: none;
 }
 
-body.frm-admin-page-formidableedit:not(.frm-full-screen) .mce-fullscreen {
+body.frm-white-body:not(.frm-full-screen) .mce-fullscreen {
 	top: 32px;
 	left: 160px;
 }
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -140,7 +140,7 @@ a, .widget .widget-top, .stuffbox h3, .frm-collapsed {
 	display: none;
 }
 
-body:not(.frm-full-screen) .mce-fullscreen {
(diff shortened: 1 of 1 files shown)


css/frm_admin.css
```diff
@@ -140,6 +140,11 @@ a, .widget .widget-top, .stuffbox h3, .frm-collapsed {
 	display: none;
 }
 
+body:not(.frm-full-screen) .mce-fullscreen {
```
Crabcyborg, line 143: @AbdiTolesa I think we'll need this to be more detailed or this could have an effect on TinyMCE on other pages that we don't want it to be changing.

It looks like this is only an issue on the form builder page, so we should probably just target that page.

garretlaxton (approved):
This looks great now!

css/frm_admin.css
```diff
@@ -140,6 +140,11 @@ a, .widget .widget-top, .stuffbox h3, .frm-collapsed {
 	display: none;
 }
 
+body.frm-admin-page-formidableedit:not(.frm-full-screen) .mce-fullscreen {
```
Crabcyborg, line 143: Thanks for updating that @AbdiTolesa.

I think we're too specific now though. There are some other rich text fields on other Formidable pages.

For example, a confirmation action.
Screen Shot 2023-09-06 at 10 08 44 AM

Maybe we want to target `.frm-white-body`?

AbdiTolesa: Oh, I should have thought about that! Thanks!

Crabcyborg (approved):
Thanks @AbdiTolesa and @garretlaxton for working on this!

This looks good to go now.

🚀

PR #1280, opened 2023-09-05
Load error in a dialog

Crabcyborg:
This can be improved a lot but this is what I'm picturing instead of https://github.com/Strategy11/formidable-forms/pull/1233

Screen Shot 2023-09-05 at 3 13 09 PM

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1090,6 +1090,15 @@ class FrmFormsController {
 			wp_die( esc_html__( 'You are trying to edit a form that does not exist.', 'formidable' ) );
 		}
 
+		if ( 'trash' === $form->status ) {
+			$error = array(
+				'title' => __( 'You can\'t edit the form', 'formidable' ),
+				'body'  => __( 'The form you\'re trying to edit is in trash. You must restore it first before you can make changes', 'formidable' ),
+			);
+			include FrmAppHelper::plugin_path() . '/classes/views/frm-forms/error-modal.php';
+			return;
+		}
+
 		if ( $form->parent_form_id ) {
 			/* translators: %1$s: Start link HTML, %2$s: End link HTML */
 			wp_die( sprintf( esc_html__( 'You are trying to edit a child form. Please edit from %1$shere%2$s', 'formidable' ), '<a href="' . esc_url( FrmForm::get_edit_link( $form->parent_form_id ) ) . '">', '</a>' ) );
--- a/classes/views/frm-forms/error-modal.php
+++ b/classes/views/frm-forms/error-modal.php
@@ -1,0 +1,38 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+?>
+<div id="frm_error_modal" class="frm-modal frm_common_modal" style="display: block;">
+	<div class="metabox-holder">
+		<div class="inside">
+			<div style="padding: 16px;">
+				<div style="font-size: var(--text-xl); font-weight: 400; color: var(--grey-900);"><?php echo esc_html( $error['title'] ); ?></div>
+				<?php echo esc_html( $error['body'] ); ?>
+				<script src="<?php echo esc_attr( includes_url() ); ?>js/jquery/ui/core.js?ver=1.13.2" id="jquery-ui-core-js"></script>
+				<script src="<?php echo esc_attr( includes_url() ); ?>js/jquery/ui/dialog.js?ver=1.13.2" id="jquery-ui-dialog-js"></script>
+				<script>
+					jQuery( document ).ready(
+						function() {
+							const modal = document.querySelector( '#frm_error_modal' );
+							jQuery( modal ).dialog(
+								{
+									dialogClass: 'frm-dialog',
+									modal: true,
+									autoOpen: true,
+									closeOnEscape: false,
+									width: '550px',
+									resizable: false,
+									draggable: false,
+									open: function() {
+										jQuery( '.ui-dialog-titlebar' ).addClass( 'frm_hidden' ).removeClass( 'ui-helper-clearfix' );
+									}
+								}
+							);
+						}
+					);
+				</script>
+			</div>
+		</div>
+	</div>
+</div>


Crabcyborg:
@AbdiTolesa It looks like you edited the description instead of posting a comment. Let's try to keep to comments as I didn't see a notification this way.

The screen recording looks good. It seems like a huge upgrade over the blank white page with no styling.

We don't need to rush with all errors at once. Maybe we can start with making a checklist, and working on individual errors a piece at a time. It doesn't all have to be in one update.

AbdiTolesa:
@Crabcyborg I'm thinking it might be a good idea to add a filter that allows customising the icon. Currently I have made it possible to hide it by passing icon => '' to the error args. Do you want to add that to the check list? In the screencast below I have tried to replace the die message in the Formidable settings page when the nonce verification is failed with thew new error modal. What do you think about it before I go ahead and get rid of more die messages?

AbdiTolesa:
Note: The PHP codestyle check is failing but the line it failed on doesn't have the style issue when I check it manually (whitespace). I couldn't figure out a solution to this.

Crabcyborg:
> @Crabcyborg I'm thinking it might be a good idea to add a filter that allows customising the icon.

@AbdiTolesa I'm confused why you would need a filter for this. The parameter argument sounds like all we'd need.

Crabcyborg:
> Note: The PHP codestyle check is failing but the line it failed on doesn't have the style issue when I check it manually (whitespace). I couldn't figure out a solution to this.

@AbdiTolesa When I ran it locally I got the right line number. I'm not sure why it's so off here in the workflow.

I pushed a small commit to fix that https://github.com/Strategy11/formidable-forms/pull/1280/commits/0ea9b72d69e50707d6e9e74527b1fc18cb7ecf3b

```
FILE: .../wp-content/plugins/formidable/classes/controllers/FrmAppController.php
-----------------------------------------------------------------------------------------------------------------------------------------
-----------------------------------------------------------------------------------------------------------------------------------------
 1063 | ERROR | [x] Whitespace found at end of line (Squiz.WhiteSpace.SuperfluousWhitespace.EndLine)
-----------------------------------------------------------------------------------------------------------
```

AbdiTolesa:
> @AbdiTolesa When I ran it locally I got the right line number. I'm not sure why it's so off here in the workflow.
> 
> I pushed a small commit to fix that [0ea9b72](https://github.com/Strategy11/formidable-forms/commit/0ea9b72d69e50707d6e9e74527b1fc18cb7ecf3b)

Thanks @Crabcyborg! I lately realised my PHPCS extension for VSCode didn't highlight the error line for me as it was not working since I have bumped up my PHP version and it wasn't compatible with it.

garretlaxton (approved):
This works great for me with both editing a form in the trash and updating the settings with an invalid key.

classes/controllers/FrmEntriesController.php
```diff
@@ -483,7 +483,13 @@ public static function show( $id = 0 ) {
 	public static function destroy() {
 		$permission_error = FrmAppHelper::permission_nonce_error( 'frm_delete_entries', '_wpnonce', -1 );
 		if ( false !== $permission_error ) {
-			wp_die( esc_html( $permission_error ) );
+			$error_args = array(
+				'title'       => __( 'Verification failed', 'formidable' ),
+				'body'        => $permission_error,
+				'cancel_url'  => admin_url( 'admin.php?page=formidable-entries' ),
+				'cancel_text' => __( 'Cancel', 'formidable' ),
```
truongwp, line 490: I see that we use `Cancel` as `cancel_text` in almost all places. We can make it as default, so we don't need to declare `cancel_text` anymore.

js/frmstrp.min.js
```diff
@@ -0,0 +1,40 @@
+(function(){var thisForm=false,formID=0,event=false,frmstripe,running=100,elements,isStripeLink=false,linkAuthenticationElementIsComplete=false,stripeLinkElementIsComplete=false,triggerCustomEvent;triggerCustomEvent=function(el,eventName,data){frmFrontForm.triggerCustomEvent(el,eventName,data)};function validateForm(e){thisForm=this;formID=jQuery(thisForm).find('input[name="form_id"]').val();if(shouldProcessForm()){e.preventDefault();event=e;processForm();return}frmFrontForm.submitFormManual(e,thisForm);
```
truongwp, line 1: Is this file added accidentally?

classes/controllers/FrmEntriesController.php
AbdiTolesa: I kind of want to give users the flexibility to leave that out as I noticed it would be applicable in other places as we replace more die() calls with this error dialog in the future. Avoiding that option from the error args, it means the user wants to prevent any text from being displayed.

Crabcyborg: @AbdiTolesa What about using "Cancel" as the default value for `cancel_text` only when `cancel_url` is also set?

AbdiTolesa: That one makes sense to me, updated it already.

classes/controllers/FrmAppController.php
```diff
@@ -1059,6 +1059,39 @@ public static function add_admin_footer_links() {
...
+			'body'             => '',
+			'cancel_url'       => '',
+			'cancel_classes'   => '',
+			'continue_url'     => '',
+			'continue_classes' => '',
+			'icon'             => 'frm_lock_simple',
+		);
+
+		$error_args = wp_parse_args( $error_args, $defaults );
+		if ( empty( $error_args['cancel_text'] ) && ! empty( $error_args['cancel_url'] ) ) {
```
truongwp, line 1083: I think we should use `asset()` check for `cancel_text` because someone might want not to show the Cancel button. The same thing for `continue_text` below.

classes/controllers/FrmSettingsController.php
```diff
@@ -248,7 +248,12 @@ public static function process_form( $stop_load = false ) {
 		$process_form = FrmAppHelper::get_post_param( 'process_form', '', 'sanitize_text_field' );
 
 		if ( ! wp_verify_nonce( $process_form, 'process_form_nonce' ) ) {
-			wp_die( esc_html( $frm_settings->admin_permission ) );
+			$error_args = array(
+				'title' => __( 'Verification failed', 'formidable' ),
+				'body'  => $frm_settings->admin_permission,
+				'cancel_text' => __( 'Cancel', 'formidable' ),
```
truongwp, line 254: This can be removed too.

AbdiTolesa: I don't think this can be removed as it its cancel_url is not set. If both cancel_url and cancel_text are omitted, it wouldn't show the button.

truongwp (approved):
Thanks @AbdiTolesa. I got it. This looks good now.

Crabcyborg:
I'm not sure why but my modals look pretty off.

I'm seeing this in Firefox and in Chrome,

Screen Shot 2023-10-26 at 11 17 53 AM

I also pushed a small update to include `button.js` from the jQuery UI JS files as well. I was seeing a console error that `.button` didn't exist without that.

AbdiTolesa:
> I'm not sure why but my modals look pretty off.
> 
> I'm seeing this in Firefox and in Chrome,

It shows correctly in all major browsers I have. I just deployed this to QA to see if the issue is just in your local environment. (https://qa.formidableforms.com/abdi/wp-admin/admin.php?page=formidable&frm_action=edit&id=23) Do you see the same issue when you follow that link?

Crabcyborg:
Thanks @AbdiTolesa. It looks fine there. I'll try to dig deeper into what the issue is for me locally.

What can do we do about the blue outline on the X icon? I think we've avoided this in other modals. It doesn't look great.

Screen Shot 2023-10-27 at 9 17 11 AM

AbdiTolesa:
> What can do we do about the blue outline on the X icon? I think we've avoided this in other modals. It doesn't look great.

@Crabcyborg I have just removed the focus with js on modal open.

Crabcyborg:
Thanks for working on this @AbdiTolesa @truongwp and @garretlaxton!

I merged some markup changes to make it more consistent with other modals (https://github.com/Strategy11/formidable-forms/pull/1403).

I was seeing weird styling issues before (https://github.com/Strategy11/formidable-forms/pull/1280#issuecomment-1781227228) which I tried to fix. I don't see it anymore, but it's the main reason I've been waiting to merge this.

I think this is good to go now. I'm going to merge it now.

PR #1282, opened 2023-09-11
Leave email the same way on the inbox page as we do for the new form …

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4436

The inbox page was a big inconsistent. It was loading the active campaign JS and trying to simulate an active campaign form.

Whereas the modal was loading a form with AJAX.

I moved out some of the modal logic so it could be done the same way on the inbox page.


Crabcyborg:
I may have got this wrong. I don't know if this form has anything to do with Active Campaign.

PR #1285, opened 2023-09-12
Change start and end time when changing step unit

truongwp:


truongwp:
I'm closing this because this is moved to builder.js in Pro.

PR #1286, opened 2023-09-12
Issue 4436

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4436

This update re-uses code from the free templates form (part of the new form modal) to submit to our Active Campaign form that is now set up on our sandbox site. During QA, it would be good to test that form as well.

8u1AnipPAQ

I don't have access to the sandbox site to confirm the entries are actually coming through as expected though.


garretlaxton:
I get 2 warnings when I visit the inbox page.
`Warning:  Undefined variable $user in \wp-content\plugins\formidable\classes\views\frm-forms\new-form-overlay\leave-email.php on line 16`

`Warning:  Attempt to read property "user_email" on null in \wp-content\plugins\formidable\classes\views\frm-forms\new-form-overlay\leave-email.php on line 16`

Crabcyborg:
Thanks @garretlaxton for catching that! I just noticed it a moment ago as well and hadn't returned to fixing it.

I just pushed an update for that. The user variable not being there meant that there was no default email value in the input. That's fixed now.

garretlaxton:
For some reason, I can't click the subscribe button now.

Crabcyborg:
@garretlaxton I think I fixed the issue. If you loaded the inbox with no messages already, it wasn't calling some necessary code - I was only called it after the "dismiss all" action.

garretlaxton (approved):
Nice, this works great for me now!

PR #1297, opened 2023-09-21
delete all entries modal confirmation

Liviu-p:
Related update https://github.com/Strategy11/formidable-pro/pull/4448
Related issue https://github.com/Strategy11/formidable-pro/issues/4317


css/frm_admin.css
```diff
@@ -9296,3 +9297,31 @@ Responsive Design
 	color: var(--primary-color);
 }
 /* End On Submit action */
+
+/* Delete All Entries Modal Confirmation */
+.frm-delete-all-entires-modal-confirmation {
```
Crabcyborg, line 9302: There is a typo here. It says `entires` but it should be `entries`, right?

js/formidable_admin.js
```diff
@@ -382,10 +382,122 @@ function frmAdminBuildJS() {
...
 			return true;
 		} else {
-			return confirmModal( link );
+			return new confirmDeleteAllEntriesModal( link );
 		}
 	}
 
+	function confirmDeleteAllEntriesModal( link ) {
+
+		var self = this;
```
Crabcyborg, line 391: Since this is an admin file we can stick to `const` and `let` here.

js/formidable_admin.js
```diff
@@ -382,10 +382,122 @@ function frmAdminBuildJS() {
 		if ( message === null || link.id === 'frm-confirmed-click' ) {
 			return true;
 		} else {
-			return confirmModal( link );
```
Crabcyborg, line 385: I think this confirm modal may be used for other actions. We should make sure this is still called I think if this isn't the delete all entries button.

js/formidable_admin.js
```diff
@@ -382,10 +382,122 @@ function frmAdminBuildJS() {
...
+
+		this.modalOptions = {
+			heading: 'Delete all %entriesCount% entries?',
+			headingSingleEntry: 'Delete the entry?',
+			copy: '',
+			inputPlaceholder: 'Type in "DELETE ALL" to delete all entries',
+			entriesCount: 0
+		};
+
+		this.countEntries = function() {
```
Crabcyborg, line 410: I don't think we can get the count in JavaScript. The delete all action can delete entries from other pages as well so we need a total entry count from all pages.

Liviu-p: hey @Crabcyborg, I double-checked this one. It appears to me that the delete action occurs exclusively on the active page, rather than affecting multiple pages. When navigating through pages, the active status of a section of checkboxes is not retained. Since this modal happens before form submission, it appears to be related to the frontend.

css/frm_admin.css
Liviu-p: should be fixed from now.

js/formidable_admin.js
Liviu-p: updated

js/formidable_admin.js
Liviu-p: I added an extra check for the entries list page. However, it looks to me like the anchor(link param) is displayed on the entries list page only.

js/formidable_admin.js
```diff
@@ -377,13 +377,130 @@ function frmAdminBuildJS() {
...
 		}
+
+		if ( 'entries-list' === loadedFrom ) {
+			return new confirmDeleteAllEntriesModal( link );
+		}
+
+		return confirmModal( link );
+	}
+
+	function confirmDeleteAllEntriesModal( link ) {
```
Crabcyborg, line 394: @Liviu-p Is there any chance we can get this function into Pro? The bulk deleting is really only a feature in Pro so it would be nice to leave this out of Lite where possible.

We can use WordPress hooks to communicate between the two functions with JS. You can see in this file there are some calls to `wp.hooks.doAction` and `wp.hooks.applyFilters`.

js/formidable_admin.js
Crabcyborg: @Liviu-p So there are two types of bulk deleting.

1 uses the checkboxes. That's when you use this "Bulk Action". With that, we can count checkboxes.
Screen Shot 2023-09-22 at 3 11 49 PM

2 deletes all entries in all pages and doesn't involve the checkboxes. That's this button.
Screen Shot 2023-09-22 at 3 11 07 PM

It looks like with this update at the moment that the #2 delete all button way isn't working anymore. It seems to require that some entries are checked, but that isn't how it works.

Screen Shot 2023-09-22 at 3 11 07 PM

Liviu-p: Got it @Crabcyborg, makes sense. Quick question about the "Delete All Entries" button, does it come with a Download CSV add-on? I'm not able to see it so far, on my end, on the entries page.

js/formidable_admin.js
Liviu-p: sure thing @Crabcyborg, I will move it to PRO.

js/formidable_admin.js
Liviu-p: nevermind, I found it.

Liviu-p: hey @Crabcyborg, I believe this is fixed from now.

js/formidable_admin.js
Liviu-p: @Crabcyborg I moved this function as requested to PRO

css/frm_admin.css
```diff
@@ -9296,3 +9297,31 @@ Responsive Design
 	color: var(--primary-color);
 }
 /* End On Submit action */
+
+/* Delete All Entries Modal Confirmation */
```
Crabcyborg, line 9301: Can we also add this CSS into Pro?

Liviu-p: sure @Crabcyborg

Crabcyborg (approved):
This branch looks good to me.

I left a comment on the Pro PR though. The styling seems off, but otherwise that looks good too.

Thanks @Liviu-p!

garretlaxton (approved):

PR #1298, opened 2023-09-22
Make sure all search terms are found in addon terms during plugin search

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4407

I tried to make sure all search terms are found in the terms for an addon before it is matched when searching for a plugin.

### Steps to test
1. Go to the Plugins -> Add New and search for "Download monitor"
2. Our Campaign monitor plugin should not show up in the result
3. Do another search with "Campaign monitor" and our "Campaign monitor" should show up
Basically, you should not see any of our plugins in the search suggestion unless all of the terms in the addon name are found in the query term.

### Screenshots
This is the suggestion we are talking about in the search result.
CleanShot 2023-09-22 at 11 40 14


garretlaxton (approved):
This looks good!

Crabcyborg (approved):
Thanks @AbdiTolesa and @garretlaxton for testing it!

PR #1299, opened 2023-09-22
Hide honeypot using overflow and randomize class and name

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-wpml/issues/104 by changing how the Honeypot is hidden using overflow style rules and a 0 size.

I also made the honeypot more difficult to predict, giving it a randomly generated class/name value.


garretlaxton (approved):
This looks good!

stephywells (commented):
This looks good. Do you think this will make a difference in the false positives? I thought we were adding autocomplete=false. Did you decide it's better not to add that?

PR #1301, opened 2023-09-26
Add a new frm_skip_draft_action

stephywells:
This is to avoid duplicate code in the abandonment plugin


stephywells:
Nevermind! The existing hook will work.

PR #1303, opened 2023-09-27
Fix Views page content button size and position

stephywells:
Pass params like class to FrmAddonsController::conditional_action_button

See related pr https://github.com/Strategy11/formidable-pro/pull/4455


Crabcyborg:
@stephywells With this update I'm seeing the activate button really close up, touching, to the video.

Screen Shot 2023-09-27 at 11 31 44 AM

stephywells:
@Crabcyborg the linked pro update fixes that. The upgrade button has space when pro is disabled. Do you think we should fix that without the pro update requirement?

Crabcyborg (approved):
Sorry about that @stephywells! I thought all the buttons would have this issue but when I disable Pro the button shown in Lite looks fine.

I don't see any other issues!

PR #1306, opened 2023-09-28
skip js validation for readonly/disabled fields

Liviu-p:
Fixes https://github.com/Strategy11/formidable-pro/issues/4447

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -211,8 +211,8 @@ function frmFrontFormJS() {
 		).filter( ':not(.frm_optional)' );
 		if ( requiredFields.length ) {
 			for ( r = 0, rl = requiredFields.length; r < rl; r++ ) {
-				if ( hasClass( requiredFields[r], 'ed_button' ) ) {
-					// skip rich text field buttons.
+				if ( hasClass( requiredFields[r], 'ed_button' ) || requiredFields[r].disabled ) {
+					// skip rich text field buttons + readonly fields.
 					continue;
 				}
 				errors = checkRequiredField( requiredFields[r], errors );


garretlaxton:
Should I still get a error when the time field is blank even though it's read-only?
image

Liviu-p:
> Should I still get a error when the time field is blank even though it's read-only? image

I'm not sure about this, it makes sense to me to trigger the error as long it is set up as required + readonly. @Crabcyborg   what do you think?

Crabcyborg:
Thanks @Liviu-p and @garretlaxton. This seems like a weird conflict of settings the more I think about this.

I don't understand how something would make sense as read only, and also required, with no default value. It seems like something that should display a warning in the form builder.

@stephywells What do you think makes sense here? Should we really validate anything disabled with JS? When it's disabled, no value gets sent anyway so I don't know why we'd validate it before its sent.

stephywells:
@Crabcyborg it seems like a validation message isn't helpful if the user can't do anything about it. But we also don't want to allow people to manually clear out the read only value and then get past validation. 

This does seem to me like more of an error on the builder end that we should notify of the issue. Is it possible to handle this type of error with a message on the builder page like you mentioned?

Crabcyborg:
Thanks @stephywells! We'll look into adding a warning.

Crabcyborg:
@Liviu-p Could you try to work on that when you have the chance?

Liviu-p:
> @Liviu-p Could you try to work on that when you have the chance?

sure @Crabcyborg, I will take care of it.

Crabcyborg:
I'm going to close this PR for now.

I think we can leave the validation as-is for now.

I fixed this issue in another way (See https://github.com/Strategy11/formidable-forms/pull/1946), by addressing the JS error instead.

PR #1315, opened 2023-10-04
Fix PHP Notice in form action

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/4469

I can't replicate this. After checking the code, I think that the entry ID we use might be mismatched or the entry is deleted. So I added one more check for the `$entry`.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -2281,7 +2281,11 @@ class FrmFormsController {
 			return array( FrmOnSubmitHelper::get_fallback_action_after_open_in_new_tab( $event ) );
 		}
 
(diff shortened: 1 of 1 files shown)


truongwp:
@Crabcyborg Thanks for pointing out that setting. I'm checking it, so I made this PR draft. I think if that is the reason, we will need to do something to process the entry instead of just checking and then returning.

truongwp:
I checked again and can't find out the problem. This might come from a custom code or another addon that makes some form action code run after entry is deleted. I'm waiting for user's reply.

Crabcyborg:
It doesn't look like the user got back.

This isn't too urgent.

For now I'll unassign you @truongwp, and we can close this, until we have a better idea of what the cause of the issue is.

Thanks for your help!

PR #1319, opened 2023-10-04
Remove changelog item

Crabcyborg:
This seems to be incomplete still. It was working for me locally, but testing on QA I'm seeing more issues.

> The payment method type ideal cannot be used with subscriptions that have the collection_method set to charge_automatically.


Crabcyborg:
Nevermind. It was something else.

PR #1322, opened 2023-10-04
form preview - block theme padding

Liviu-p:
Fixes https://github.com/Strategy11/formidable-pro/issues/4223


Crabcyborg (commented):
@Liviu-p I added a check for `wp_is_block_theme` as well so we don't add this when we don't need to.

Do you see any issues with that?

Crabcyborg:
@Liviu-p One more thing.

I made a new issue for this since it isn't new to this PR.

I'm seeing a couple deprecation messages.
https://github.com/Strategy11/formidable-pro/issues/4478

Could you look into those as well?

garretlaxton (approved):
This looks good!

PR #1328, opened 2023-10-06
Check that overlay controller exists just in case

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2382420517/175264

This shouldn't really be possible, but I've seen errors like this come up that seem to happen when updating the plugin, when some code has loaded and other code hasn't.

> Error Message: Uncaught Error: Class 'FrmOverlayController' not found
> File Path: .../plugins/formidable/classes/controllers/FrmAppController.php


Crabcyborg:
@Liviu-p If I don't call this function, nothing weird would happen, right?

Liviu-p:
@Crabcyborg, I find it safe if not called, It won't break anything.

PR #1331, opened 2023-10-11
Switch field key by field id

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-api/issues/51

The root cause for the linked issue is that trying to store a string (field key in this case) into a an integer database field. Switching that to the field id fixes the issue.

### DIFF
--- a/classes/models/FrmEntryMeta.php
+++ b/classes/models/FrmEntryMeta.php
@@ -6,15 +6,13 @@ if ( ! defined( 'ABSPATH' ) ) {
 class FrmEntryMeta {
 
 	/**
-	 * @since x.x    Added $field parameter.
 	 *
 	 * @param int    $entry_id
 	 * @param int    $field_id
 	 * @param string $meta_key usually set to '' as this parameter is no longer used.
 	 * @param mixed  $meta_value
-	 * @param object|null $field
 	 */
-	public static function add_entry_meta( $entry_id, $field_id, $meta_key, $meta_value, $field = null ) {
+	public static function add_entry_meta( $entry_id, $field_id, $meta_key, $meta_value ) {
 		global $wpdb;
 
 		if ( FrmAppHelper::is_empty_value( $meta_value ) ) {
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This works for me now.

classes/models/FrmEntryMeta.php
```diff
@@ -138,7 +145,7 @@ public static function update_entry_metas( $entry_id, $values ) {
 				}
 			} else {
 				// if value does not exist, then create it
-				self::add_entry_meta( $entry_id, $field_id, '', $meta_value );
+				self::add_entry_meta( $entry_id, $field_id, '', $meta_value, $field );
```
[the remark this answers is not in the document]
AbdiTolesa: That also works, just need to add a check if `field` is not empty. Thanks!

classes/models/FrmEntryMeta.php
```diff
@@ -136,9 +136,9 @@ public static function update_entry_metas( $entry_id, $values ) {
 					// if value exists, then update it
 					self::update_entry_meta( $entry_id, $field_id, '', $meta_value );
 				}
-			} else {
+			} elseif ( ! empty( $field ) ) {
```
truongwp, line 139: We can miss the case `$field` is empty but `$field_id` is a number. I think we should keep the `else`. In the `add_entry_meta(` line, we check if `$field_id` is number, use it, else if `$field` isn't empty, use `$field->id`, else do not add meta.

AbdiTolesa: @truongwp If you look a couple of lines above, `$field` if just queried with `field_id`. So if `field_id` is a valid field id, `$field` cannot be empty.

truongwp: I checked them. `$field_id` comes from `$values`. Maybe one of the field ID in `$values` doesn't exist, which will make `$field` empty.

But I'm not sure if we allow adding an entry meta for a field that doesn't exist. I'll approve this PR but leave this conversation so @Crabcyborg will take a look.

truongwp (approved):
Thanks @AbdiTolesa.

classes/models/FrmEntryMeta.php
Crabcyborg: This feels incomplete if we're updating `self::add_entry_meta( $entry_id, $field_id, '', $meta_value );` but not the `self::update_entry_meta( $entry_id, $field_id, '', $meta_value );` above that.

It seems to me that this issue would persist in other parts of this function.

Crabcyborg: Is it possible to try changing this mapping in the API add on as well?

I don't think we need to support keys like this in Lite.

The API add on could possibly map keys to ids before it calls the create function.

Crabcyborg:
I'm going to close this because we're getting close in the API add on instead to keep the fix contained to the add on with the issue.

PR #1333, opened 2023-10-12
Remove unconditional view permission set from global settings

AbdiTolesa:
Related issue https://github.com/Strategy11/formidable-pro/issues/3591
Related pull request https://github.com/Strategy11/formidable-pro/pull/4490

The View permission removed from lite to be handled conditionally in pro.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1674,6 +1674,11 @@ class FrmAppHelper {
 		 */
 		$pro_cap = apply_filters( 'frm_pro_capabilities', $pro_cap );
 
+		if ( ! array_key_exists( 'frm_edit_displays', $pro_cap ) && is_callable( 'FrmProAppHelper::views_is_installed' ) && FrmProAppHelper::views_is_installed() ) {
+			// For backward compatibility, add the Add/Edit Views permission if Pro is not up to date.
+			$pro_cap['frm_edit_displays'] = __( 'Add/Edit Views', 'formidable' );
+		}
+
 		if ( 'pro_only' === $type ) {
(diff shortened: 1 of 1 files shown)


classes/helpers/FrmAppHelper.php
```diff
@@ -1666,7 +1666,6 @@ public static function frm_capabilities( $type = 'auto' ) {
 			'frm_create_entries' => __( 'Add Entries from Admin Area', 'formidable' ),
 			'frm_edit_entries'   => __( 'Edit Entries from Admin Area', 'formidable' ),
 			'frm_view_reports'   => __( 'View Reports', 'formidable' ),
-			'frm_edit_displays'  => __( 'Add/Edit Views', 'formidable' ),
```
Crabcyborg, line 1669: We should probably only remove this when Pro is updated.

Otherwise we may lose the permission when versions don't match up.

AbdiTolesa: It looks impossible to check anything from Pro/Views and conditionally update the permissions at this point as they haven't been loaded, so it may be good to hold this until the pro update is released first indeed.

Crabcyborg: @AbdiTolesa I just made a commit https://github.com/Strategy11/formidable-forms/pull/1333/commits/14a15d08b23d47f9091750a19cff64ac274df341

I think this is really all we need to do here. We can reflect the https://github.com/Strategy11/formidable-pro/pull/4490 update in Lite as well, for now, and remove it later.

Crabcyborg (approved):
I'll go ahead and merge this for next release.

Thanks @AbdiTolesa!

PR #1335, opened 2023-10-13
Stop using the Stripe Lite payments table when only PayPal has migrated

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2389539238/175949/

> WordPress database error Table 'wp_frm_subscriptions' doesn't exist for query SELECT COUNT(*) FROM wp_frm_subscriptions made by do_action('forms_page_formidable-payments'), WP_Hook->do_action, WP_Hook->apply_filters, FrmTransLitePaymentsController::route, FrmTransLiteListsController::route, FrmTransLiteListsController::display_list, include('/plugins/formidable/stripe/views/lists/list.php'), FrmListHelper->views, FrmTransLiteListHelper->get_views, FrmTransLiteDb->get_count [12-Oct-2023 14:55:13 UTC] PHP Warning: Undefined property: stdClass::$status in ../wp-content/plugins/formidable/stripe/helpers/FrmTransLiteAppHelper.php on line 74

The problem here is that PayPal only has the one payments table, not a subscriptions table.

So the UI of Stripe Lite kind of gets in the way.

With this update, if the PayPal migrations have run and not Stripe, I avoid the menu item and the routing, and let the PayPal add on take over as a fallback.

Once Stripe migrations have run, it will use Stripe Lite again.

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -136,7 +136,7 @@ class FrmAppController {
 			'formidable-applications',
 		);
 
-		if ( ! class_exists( 'FrmTransHooksController', false ) ) {
+		if ( ! class_exists( 'FrmTransHooksController', false ) && ! FrmTransLiteAppHelper::should_fallback_to_paypal() ) {
 			// Only consider the payments page as a "white page" when the Payments submodule is off.
 			// Otherwise this causes a lot of styling issues when the Stripe add-on (or Authorize.Net) is active.
 			$white_pages[] = 'formidable-payments';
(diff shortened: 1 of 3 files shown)


Crabcyborg:
@stephywells This makes sense as the fix here, right? I'm not sure if it's worth trying to use the Stripe Lite payments table for just the PayPal add on.

stephywells:
@Crabcyborg what happens if a form uses PayPal and Stripe lite? Would this approach still work?

Crabcyborg:
@stephywells Since the PayPal add on doesn't use the `frm_subscriptions` table, any PayPal subscription appears in "Payments" and not "Subscriptions".

This is how it works when you use the Stripe Add on and the PayPal add on. Using Stripe Lite and the PayPal add on works the same way. But since the migrations for Stripe Lite only run when you've opted in, showing the "Subscriptions" section is just confusing since it won't include PayPal subscriptions, and since that's all you've opted into, "Subscriptions" will always be empty.

Crabcyborg:
@stephywells the Stripe migrations run once you've connected to Stripe Connect, in either test or live mode.

The payment cron also doesn't run until Stripe Connect is configured.

PR #1337, opened 2023-10-16
Stop showing the add new entry button without permissions to add entries

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3395

The "Add New" button on the entries list would appear for users who don't have the permissions to create entries.

This button logic was moved here with the v6.0 update. But I tried v5.5.7, and the button appears then as well so this is an old issue.


garretlaxton (approved):
This looks great!

PR #1341, opened 2023-10-18
Fix address field format issue in form and submit actions

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-geo/issues/57


Crabcyborg:
@AbdiTolesa Since address fields are only in Pro, can we make this work in Pro instead?

PR #1348, opened 2023-10-23
Fix: Display "Add Fields Here" Message After Deleting All Fields

shervElmi:
This PR fixes the issue where the "Add Fields Here" message was not displaying after all form fields were deleted.

## Related Issue:
["Add Fields Here" Message Not Displaying After Deleting All Fields](https://github.com/Strategy11/formidable-pro/issues/4509)

## QA URL
https://qa.formidableforms.com/sherv10/wp-admin/admin.php?page=formidable&frm_action=edit&id=1

## Testing Instructions:
1. Go to the form builder
2. Delete all form fields
3. Verify that the "Add Fields Here" message displays correctly

PR #1349, opened 2023-10-23
Fix: Resolve Overlapping Issue After Deleting a Section in Form Builder

shervElmi:
This PR addresses the overlapping problem that occurs in the form builder after a section with fields is deleted.

## Related Issue:
[Funny overlapping in form builder after deleting a section](https://github.com/Strategy11/formidable-pro/issues/3259)

## Depends on:
[Fix: Display "Add Fields Here" Message After Deleting All Fields](https://github.com/Strategy11/formidable-forms/pull/1348)

## QA URL
https://qa.formidableforms.com/sherv11/wp-admin/admin.php?page=formidable&frm_action=edit&id=1

## Testing Instructions:
1. Navigate to the form builder
2. Add a text field to a section
3. Save to get the update message
4. Delete the section
5. Verify that no funny overlapping occurs and the layout looks correct


garretlaxton (approved):
This looks great! 

Just an opinion though, I think it would be great if this banner disappeared after a few seconds.

js/formidable_admin.js
```diff
@@ -4606,17 +4606,20 @@ function frmAdminBuildJS() {
...
-					} else if ( $section.length ) {
-						toggleOneSectionHolder( $section );
-					}
 					if ( $adjacentFields.length ) {
 						syncLayoutClasses( $adjacentFields.first() );
 					} else {
 						$liWrapper.remove();
 					}
 
+					if ( jQuery( '#frm-show-fields li' ).length === 0 ) {
```
Crabcyborg, line 4615: @shervElmi Is there a reason this code was moved? Does it work on the previous lines as well?

shervElmi: The base branch for this PR is `issue-pro-3259`, which resolves a different issue. Here, you can see that I relocated the omitted code to follow the `$liWrapper.remove()` code: https://github.com/Strategy11/formidable-forms/pull/1348

Crabcyborg: Thanks. I changed your base branch then. It's good to exclude other updates from the scope of this one.

Screen Shot 2023-10-30 at 9 48 17 AM

Crabcyborg:
@shervElmi I don't know about the hard coded value here. The placeholder is in totally different places before and after.

What is the real reason for this? Can we just return it to the previous HTML structure it had on load?

**When a form is loaded**
Screen Shot 2023-10-30 at 10 10 43 AM

**After deleting a field**
Screen Shot 2023-10-30 at 10 10 52 AM

Crabcyborg:
@shervElmi I updated this so it only offsets when there is a success message. The extra margin looks bad when there is no message.

I also switched it to `top` since it's using absolute positioning. Overwriting the default margin feels wrong.
https://github.com/Strategy11/formidable-forms/pull/1349/commits/f76e1143884776629843d26357851ac7e14226ad

Crabcyborg (approved):

shervElmi:
Thank you so much for your help in wrapping up this issue.

PR #1352, opened 2023-10-25
Remove extra permission checks that break API

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4485

These permission checks don't appear to be necessary.

The function `can_install_addon_api` is called from our API, so the user is not logged in and doesn't have this permission. Instead the token is verified using `hash_equals` against the DB record.

The `install_addon` function is called in one function only, `download_and_activate`.

`download_and_activate` is only called when the API token is valid, or via `ajax_install_addon` which goes through `install_addon_permissions` which handles the permission check, so this permission check happens twice.

### DIFF
--- a/classes/controllers/FrmAddonsController.php
+++ b/classes/controllers/FrmAddonsController.php
@@ -1024,8 +1024,6 @@ class FrmAddonsController {
 	 * @since 3.04.02
 	 */
 	protected static function install_addon() {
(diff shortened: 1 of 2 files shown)


omidakhavans (approved):
It's looking good.

Crabcyborg:
Thanks @omidakhavans!

PR #1304, opened 2023-09-27
Introduce Form Templates Page

shervElmi:
This PR introduces a new "Form Templates" page within the Formidable Admin Panel. The Form Templates page allows users to either start from scratch with a blank form or choose from a selection of pre-built templates.

## QA URL:
https://qa.formidableforms.com/sherv/wp-admin/admin.php?page=formidable-form-templates

## Testing Instructions:
1. Make sure you have Formidable PRO installed and activated on the **[`introduce-form-templates` branch](https://github.com/Strategy11/formidable-pro/tree/introduce-form-templates)**.
    - **Formidable PRO PR**: [https://github.com/Strategy11/formidable-pro/pull/4483](https://github.com/Strategy11/formidable-pro/pull/4483)
2. Confirm that the designs on the Form Templates page match those in the Figma file. **[Figma Design File](https://www.figma.com/file/x8Epd7CJLLY6JIyPoCMPvc/Create-Form?type=design&node-id=1864-17378&mode=dev)**
4. Review the video introductions (Part One and Part Two) and verify that every part of the feature works as shown.

### Updated Custom Templates and Ready-Made Solutions
https://www.loom.com/share/4657fb8d47c2437d879e807ee851bacb?sid=93082411-a37c-4d31-a762-6cfc400418cb

### Part One
https://www.loom.com/share/d8ed343652dd4ed3b817a8d2d54ae284

### Part Two
https://www.loom.com/share/7fa9d217b98e448b8e1419531f86b710

## Checklist
- [x] Write PHPUnit Tests.
- [x] Test real licenses, including (Basic, Pro, Business, Elite, and expired)

PR #1356, opened 2023-10-30
Dashboard UI

Liviu-p:
QA Link
https://qa.formidableforms.com/liviu/wp-admin/admin.php?page=formidable-dashboard

More details: are available in this PR: https://github.com/Strategy11/formidable-pro/pull/4548


classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,497 @@
...
+					'display' => self::display_counter_cta( 'entries', $counters_value['entries'], $latest_available_form ),
+					'title'   => esc_html__( 'Add Entry', 'formidable' ),
+					'link'    => $add_entry_cta_link,
+				),
+				'counter' => $counters_value['entries'],
+			),
+		);
+
+		$pro_counters_placeholder = array(
+			array(
```
Crabcyborg, line 128: Could we add a function to cover the defaults here to cut back on some duplicate code?

I picture it returning one of these arrays, with a few of the values already set by default: `'counter' => 0`, `'type' => 'default'`, `'disabled' => true`. It could accept a `$heading` and `$tooltip` parameter to finish the array.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,497 @@
...
+
+	/**
+	 * Init view args for entries placeholder.
+	 *
+	 * @param array $forms_count The total forms count. If there are no any forms yet, we'll have CTA pointing to creating a form.
+	 * @return array
+	 */
+	private static function view_args_entries_placeholder( $forms_count ) {
+
+		if ( 0 === (int) $forms_count ) {
```
Crabcyborg, line 216: For a check like this, I'd generally just reduce it to `if ( ! $forms_count ) {`. I don't think the strict checking is necessary here.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,497 @@
...
+		return true;
+	}
+
+	/**
+	 * Hide 3rd party notifications from dashboard
+	 *
+	 * @return void
+	 */
+	public static function remove_admin_notices_on_dashboard() {
+		if ( 'formidable-dashboard' !== FrmAppHelper::simple_get( 'page', 'sanitize_title' ) ) {
```
Crabcyborg, line 271: Can we use `FrmAppHelper::is_admin_page( 'formidable-dashboard' )` here instead?

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,497 @@
...
+		return false;
+	}
+
+	/**
+	 * Check if is dashboard page.
+	 *
+	 * @return boolean
+	 */
+	public static function is_dashboard_page() {
+		if ( 'formidable-dashboard' === FrmAppHelper::simple_get( 'page', 'sanitize_title' ) ) {
```
Crabcyborg, line 362: Can we use FrmAppHelper::is_admin_page( 'formidable-dashboard' ) here instead?

classes/controllers/FrmDashboardController.php
Crabcyborg: After looking at more code, I think this should call `self::is_dashboard_page` instead.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,497 @@
...
+	 * @return void
+	 */
+	public static function ajax_requests() {
+		$dashboard_action = FrmAppHelper::get_post_param( 'dashboard_action', '', 'sanitize_text_field' );
+
+		switch ( $dashboard_action ) {
+
+			case 'welcome-banner-cookie':
+				if ( true === self::ajax_set_cookie_banner( FrmAppHelper::get_post_param( 'banner_has_closed' ) ) ) {
+					echo wp_json_encode( array( 'success' => true ) );
```
Crabcyborg, line 294: I think this can be reduced to `wp_send_json_success();`.

classes/models/FrmYoutubeFeedApi.php
```diff
@@ -0,0 +1,106 @@
...
+	 *
+	 * @param string $api_endpoint. The YouTube API endpoint.
+	 * @param string $cache_key The cache key used to store data to wp cache.
+	 *
+	 * @return array
+	 */
+	private function get_feed_by_api_endpoint( $api_endpoint, $cache_key ) {
+		$response = wp_remote_get( $this->get_api_url( $api_endpoint ) );
+		if ( is_wp_error( $response ) ) {
+			return false;
```
Crabcyborg, line 94: If there's an error, can we still set a transient in some way to avoid too many requests?

We've had issues where a failed API would mean every request would make another API request, which would basically cause a DDOS effect.

classes/views/FrmDashboardView.php
```diff
@@ -0,0 +1,415 @@
...
+
+	/**
+	 * Dashboard - top counters card widgets template.
+	 *
+	 * @param boolean $echo
+	 *
+	 * @return void|string Echo or return the widgets's HTML
+	 */
+	public function get_counters( $echo = true ) {
+		if ( false === $echo ) {
```
Crabcyborg, line 160: We have a function for handling this `echo` vs. `return` behaviour.

It's `FrmAppHelper::clip`.

You basically use it like this:

```
return FrmAppHelper::clip(
    function() {
        echo wp_kses_post( $this->load_counters_template( $this->view['counters'] ) );
    },
    $echo
);
```

classes/views/dashboard/templates/inbox.php
```diff
@@ -0,0 +1,84 @@
...
+			<li class="frm-active"><?php echo esc_html__( 'Inbox', 'formidable' ); ?></li>
+			<li><?php echo esc_html__( 'Dismissed', 'formidable' ); ?></li>
+		</ul>
+	</div>
+	<div class="frm-tabs-delimiter">
+		<span class="frm-tabs-active-underline"></span>
+	</div>
+	<div class="frm-tabs-container">
+		<div class="frm-tabs-slide-track frm-flex-box">
+			<div class="frm-active" data-mdb-perfect-scrollbar ata-mdb-suppress-scroll-x='true'>
```
Crabcyborg, line 18: Can we use double quotes here for the `'true'` attribute?

Also, is `ata-mdb-suppress-scroll-x` a typo? It looks like it may be missing a `d` on the front.

classes/views/dashboard/templates/pro-features-list.php
```diff
@@ -0,0 +1,29 @@
...
+				<li><?php echo esc_html__( 'Form Templates', 'formidable' ); ?></li>
+				<li><?php echo esc_html__( 'Calculated Fields and Math', 'formidable' ); ?></li>
+				<li><?php echo esc_html__( 'Quizzes', 'formidable' ); ?></li>
+				<li><?php echo esc_html__( 'Save and Continue', 'formidable' ); ?></li>
+				<li><?php echo esc_html__( 'Ecommerce pricing fields', 'formidable' ); ?></li>
+			</ul>
+		</li>
+		<li>
+			<ul>
+				<li><?php echo esc_html__( 'Customize Form HTML', 'formidable' ); ?></li>
```
Crabcyborg, line 20: Instead of `echo esc_html` you can use `esc_html_e`.

images/dashboard/entries-placeholder.jpg
Crabcyborg, line 1: Can we optimize these files? This JPG seems a lot larger than it needs to be.

I'm not sure if we have another more official solution documented somewhere but I've generally used https://imagecompressor.com/ to optimize images.

js/formidable_admin.js
```diff
@@ -10121,6 +10122,8 @@ function frmAdminBuildJS() {
 			});
 		},
 
+		initNewFormModal: initNewFormModal,
```
Crabcyborg, line 10125: It's fine to just reduce this line to `initNewFormModal,` since we don't need to support an older syntax here.

js/src/components/class-counter.js
```diff
@@ -0,0 +1,58 @@
+export class FrmCounter {
+
+	constructor( element, options ) {
+		if ( ! element instanceof Element || null === element.getAttribute( 'data-counter' ) ) {
+			return;
+		}
+
+		this.template = 'default';
+		if ( null !== element.getAttribute( 'data-type' ) ) {
```
Crabcyborg, line 9: Does it work to access `element.dataset.type` instead? It looks nicer and seeing `.getAttribute( 'data-type' )` looks off - I'd generally set it to a variable.

js/src/components/class-counter.js
```diff
@@ -0,0 +1,58 @@
...
+			return number.toLocaleString( this.locale, { minimumFractionDigits: 2 });
+		}
+		return number;
+	}
+
+	animate() {
+		if ( this.activeCounter < this.value ) {
+			this.activeCounter += this.valueStep;
+			this.element.innerText = this.formatNumber( this.activeCounter );
+			setTimeout( () => {
```
Crabcyborg, line 49: You could reduce this to just `setTimeout( this.animate, this.timeoutInterval );` and skip the extra function.

js/src/components/class-tabs-navigator.js
```diff
@@ -0,0 +1,63 @@
...
+		this.slides.forEach( slide => {
+			new PerfectScrollbar( slide );
+		});
+	}
+
+	changeSlide( index ) {
+		this.removeActiveClassnameFromSlides();
+		const translate = index == 0 ? '0px' : `calc( ( ${( index * 100 )}% + ${this.flexboxSlidesGap} ) * -1 )`;
+		this.slideTrack.style.transform = `translateX(${translate})`;
+		if ( 'undefined' !== typeof this.slides[ index ]) {
```
Crabcyborg, line 51: I think this can be simplified to `if ( index in this.slides )`

Crabcyborg:
The styling on this "Check all features" button is a bit funny on a small screen.

Screen Shot 2023-12-07 at 4 16 01 PM

Crabcyborg:
My count in the sidebar also seems kind of off center.

Screen Shot 2023-12-07 at 4 17 12 PM

css/admin/dashboard.css
```diff
@@ -0,0 +1,597 @@
...
+}
+.frm-inbox-wrapper div.frm-tabs-container div.frm-inbox-message-container a.frm-button-primary,
+.frm-inbox-message-container a {
+	font-weight: 600;
+	color: var(--primary-500);
+	font-size: 12px;
+}
+.frm-inbox-wrapper div.frm-tabs-container div.frm-inbox-message-container a.frm-button-primary {
+	background-color: unset !important;
+	height: auto;
```
stephywells, line 574: This could use a bit more css to override the height of the learn more link:
```suggestion
	height: auto;
	min-height: 0;
```

The hover could also use some work. Or maybe we could remove the button classes so we don't have to override so much?:
Screenshot 2023-12-07 at 9 00 47 PM

js/src/components/class-counter.js
Liviu-p: hey @Crabcyborg, I'm afraid that this won't work as it will loose the "this" scope from es6 class. 
I see at the moment 2 ways of writing this:
`setTimeout( () => this.animate(), this.timeoutInterval )` ( I believe this one will be shorter, which I can update to ) or
`setTimeout( this.animate.bind( this ), this.timeoutInterval )`

Crabcyborg: The `.bind( this )` version still avoids an extra function call so I think that's probably the better option.

classes/controllers/FrmDashboardController.php
Liviu-p: Hey @Crabcyborg, this on should be updated in the latest commits.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,497 @@
...
+	 *
+	 * @return array
+	 */
+	private static function view_args_license() {
+		if ( class_exists( 'FrmProDashboardController' ) ) {
+			return FrmProDashboardController::view_args_license();
+		}
+		return array(
+			'heading' => 'License Key',
+			'copy'    => esc_html__( 'You\'re using Formidable Forms Lite - no license needed. Enjoy!', 'formidable' ) . ' 🙂',
```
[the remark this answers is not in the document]
Liviu-p: Hey @Crabcyborg & @stephywells I moved the LITE license copy into a separate function that will be used by both layouts from now. 
@stephywells I fixed the lite copy issue too when an older version of PRO is active

classes/controllers/FrmDashboardController.php
Liviu-p: updated!

classes/controllers/FrmDashboardController.php
Liviu-p: updated

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,497 @@
...
+					wp_die();
+				}
+				echo wp_json_encode( array( 'success' => false ) );
+				wp_die();
+				break;
+
+			case 'save-subscribed-email':
+				$email = FrmAppHelper::get_post_param( 'email' );
+				self::save_subscribed_email( $email );
+				echo wp_json_encode( array( 'success' => true ) );
```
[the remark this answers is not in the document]
Liviu-p: updated

classes/views/dashboard/templates/inbox.php
Liviu-p: fixed

classes/views/dashboard/templates/pro-features-list.php
Liviu-p: updated

images/dashboard/entries-placeholder.jpg
Liviu-p: I believe Steph has fixed this.

js/formidable_admin.js
Liviu-p: updated

js/src/components/class-counter.js
Liviu-p: Updated

js/src/components/class-counter.js
Liviu-p: I updated it to `.bind(this)`

js/src/dashboard.js
```diff
@@ -0,0 +1,143 @@
...
+		this.initTooltips();
+	}
+
+	initInbox() {
+		new FrmTabsNavigator( '.frm-inbox-wrapper' );
+		const userEmailInput  = document.querySelector( '#frm_leave_email' );
+		const subscribeButton = document.querySelector( '#frm-add-my-email-address' );
+
+		subscribeButton.addEventListener( 'click', () => {
+			this.saveSubscribedEmail( userEmailInput.value ).then( data => {});
```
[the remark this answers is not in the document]
Liviu-p: sure, it's updated now.

js/src/dashboard.js
```diff
@@ -0,0 +1,143 @@
...
+				headers: {
+					'Content-Type': 'application/x-www-form-urlencoded'
+				},
+				body: new URLSearchParams({
+					action: this.options.ajax.action,
+					dashboard_action: this.options.ajax.dashboardActions.saveSubscribedEmail,
+					email: email
+				})
+			}).then( result => {
+				result.json().then( ( data ) => {
```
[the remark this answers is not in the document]
Liviu-p: yep, it's updated now

Liviu-p:
> The styling on this "Check all features" button is a bit funny on a small screen.
> 
> Screen Shot 2023-12-07 at 4 16 01 PM

It should be fixed from now @Crabcyborg. Please let me know if you're still able to replicate it.

Liviu-p:
> My count in the sidebar also seems kind of off center.
> 
> Screen Shot 2023-12-07 at 4 17 12 PM

I couldn't replicate this one on my end. However, the layout is coming as a default from WordPress. 
Screenshot 2023-12-10 at 20 39 03

classes/controllers/FrmAppController.php
```diff
@@ -655,6 +658,11 @@ public static function admin_js() {
 
 		FrmAppHelper::load_admin_wide_js();
 
+		if ( class_exists( 'FrmDashboardController' ) ) {
```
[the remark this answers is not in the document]
Liviu-p: Hey @stephywells, there is no special reason for checking this. I clean it up.

classes/controllers/FrmAppController.php
```diff
@@ -1120,12 +1128,13 @@ public static function add_admin_footer_links() {
 		$post_type = FrmAppHelper::simple_get( 'post_type', 'sanitize_title' );
 
 		// Check admin privileges, screen mode, and branding
-		$is_not_admin = ! FrmAppHelper::is_formidable_admin() && $post_type !== 'frm_logs';
-		$is_full_screen = FrmAppHelper::is_full_screen();
+		$is_not_admin               = ! FrmAppHelper::is_formidable_admin() && $post_type !== 'frm_logs';
+		$is_full_screen             = FrmAppHelper::is_full_screen();
 		$is_not_formidable_branding = ! FrmAppHelper::is_formidable_branding();
+		$is_dashboard_page          = 'formidable-dashboard' === FrmAppHelper::simple_get( 'page', 'sanitize_title' );
```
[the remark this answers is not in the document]
Liviu-p: I updated it, from now it will be handled from `FrmDashboardController` via `remove_action`

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,497 @@
...
+			'copy'    => esc_html__( 'You\'re using Formidable Forms Lite - no license needed. Enjoy!', 'formidable' ) . ' 🙂',
+			'buttons' => array(
+				array(
+					'label'  => esc_html__( 'Connect Account', 'formidable' ),
+					'link'   => FrmAddonsController::connect_link(),
+					'action' => 'default',
+					'type'   => 'primary',
+				),
+				array(
+					'label'  => esc_html__( 'Get Formidable PRO', 'formidable' ),
```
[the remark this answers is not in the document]
Liviu-p: Updated

classes/views/FrmDashboardView.php
```diff
@@ -0,0 +1,415 @@
...
+	}
+
+	/**
+	 * Dashboard - welcome banner template.
+	 *
+	 * @param boolean $echo
+	 *
+	 * @return void|string Echo or return the widget's HTML
+	 */
+	public function get_welcome_banner( $echo = true ) {
```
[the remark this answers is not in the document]
Liviu-p: Updated. It shouldn't show from now for users that have at least 1 form.

classes/views/dashboard/templates/pro-features-list.php
```diff
@@ -0,0 +1,29 @@
...
+				<li><?php echo esc_html__( 'Smart Forms with Conditional Logic', 'formidable' ); ?></li>
+				<li><?php echo esc_html__( 'Schedule Forms & Limit Responses', 'formidable' ); ?></li>
+				<li><?php echo esc_html__( 'Display Form Data With Views', 'formidable' ); ?></li>
+				<li><?php echo esc_html__( 'And much more...', 'formidable' ); ?></li>
+			</ul>
+		</li>
+	</ul>
+	<a target="_blank" href="<?php echo esc_url( FrmAppHelper::admin_upgrade_link( '', 'lite-upgrade' ) ); ?>" title="Upgrade to PRO & Get 50% OFF" class="frm-button-primary"><?php echo esc_html__( 'Upgrade to PRO & Get 50% OFF', 'formidable' ); ?></a>
```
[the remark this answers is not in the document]
Liviu-p: updated

css/admin/dashboard.css
```diff
@@ -0,0 +1,597 @@
...
+}
+.frm-tabs-navs ul li {
+	color: var( --grey-500 );
+	font-weight: 400;
+	font-size: 16px;
+	cursor: pointer;
+}
+.frm-tabs-navs ul li.frm-active {
+	color: var(--primary-500);
+	font-weight: 700;
```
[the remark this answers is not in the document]
Liviu-p: updated

css/admin/dashboard.css
```diff
@@ -0,0 +1,597 @@
...
+/* Inbox */
+.frm-inbox-wrapper {
+	padding: 16px 24px;
+	height: 612px;
+}
+.frm-tabs-navs ul {
+	margin: 0px 0px 11px 0px;
+	position: relative;
+}
+.frm-tabs-navs ul li {
```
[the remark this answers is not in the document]
Liviu-p: updated

css/admin/dashboard.css
```diff
@@ -0,0 +1,597 @@
...
+	color: var( --grey-500 );
+	font-weight: 400;
+	font-size: 16px;
```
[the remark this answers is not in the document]
Liviu-p: I fixed this, I changed more values with CSS variables.

css/admin/dashboard.css
```diff
@@ -0,0 +1,597 @@
...
+}
+.frm-white-body .frm-dashboard-widget table.wp-list-table .row-actions{
+	display: none;
+}
+.frm-white-body .frm-dashboard-widget table.wp-list-table th,
+.frm-white-body .frm-dashboard-widget table.wp-list-table td,
+.frm-dashboard-widget table.wp-list-table th,
+.frm-dashboard-widget table.wp-list-table td {
+	padding: 12px 24px;
+	font-weight: 500;
```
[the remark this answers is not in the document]
Liviu-p: updated

css/admin/dashboard.css
```diff
@@ -0,0 +1,597 @@
...
+.frm-dashboard-wrapper .frm-dashboard-widget h2.frm-widget-heading {
+	color: #101828;
+	font-size: 16px;
+	font-weight: 600;
+	line-height: 24px;
+	position: relative;
+	z-index: 22;
+}
+.frm-dashboard-wrapper .frm-dashboard-widget .frm-widget-cta {
+	font-weight: 600;
```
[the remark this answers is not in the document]
Liviu-p: updated

css/admin/dashboard.css
```diff
@@ -0,0 +1,597 @@
...
+	top: 0;
+	left: 0;
+	width: 469px;
+	height: 247px;
+	border-radius: 469px;
+	background: #8B5CF6;
+	mix-blend-mode: overlay;
+	filter: blur(50px);
+}
+.frm-dashboard-container {
```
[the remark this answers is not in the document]
Liviu-p: I added a breakpoint from 960. It should fit better now on small screens

css/admin/dashboard.css
```diff
@@ -0,0 +1,597 @@
...
+.frm-pro-features-list ul > li ul li {
+	position: relative;
+	display: block;
+	padding-left: 24px;
+	margin-bottom: 12px;
+	color: #656F7C;
+	font-weight: 400;
+	font-size: 14px;
+}
+.frm-pro-features-list ul > li ul li:before {
```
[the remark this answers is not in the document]
Liviu-p: fixed

images/dashboard/close.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="24" height="25" viewBox="0 0 24 25" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
Liviu-p: Reusing the same SVG icon will require including **icons.svg** file which is about 100kb. I don't find it valuable to use it just for one icon.

css/admin/dashboard.css
```diff
@@ -0,0 +1,597 @@
...
+} */
+.frm-tabs-slide-track > div.frm-active {
+	opacity: 1.0;
+	transition: 0.35s opacity linear;
+}
+.frm-inbox-wrapper div.frm-tabs-container div.frm-inbox-message-container a.frm-button-primary,
+.frm-inbox-message-container a {
+	font-weight: 600;
+	color: var(--primary-500);
+	font-size: 12px;
```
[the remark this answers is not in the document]
Liviu-p: fixed

css/admin/dashboard.css
Liviu-p: fixed, I just removed all the classes via backend.

stripe/helpers/FrmTransLiteAppHelper.php
```diff
@@ -281,6 +281,89 @@ public static function formatted_amount( $payment ) {
...
+	}
+
+	/**
+	 * Gets payments data.
+	 *
+	 * @param string $from_date From date.
+	 * @param string $to_date   To date.
+	 * @return array            Contains `count` and `total`.
+	 */
+	public static function get_payments_data( $from_date = null, $to_date = null ) {
```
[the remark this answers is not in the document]
Liviu-p: thank you for letting me know @Crabcyborg, I will update it soon.

Crabcyborg:
I'm seeing no entries in this report.

When I go to add more entries, none still appear. I'm not sure if it's caching or something, but it would be nice if we cleared that cache when a new entry is created so we could see something.

Is this likely to always look empty on Monday? Would we want to show the previous week or something on Monday?

Screen Shot 2023-12-11 at 4 50 14 PM

classes/models/FrmYoutubeFeedApi.php
```diff
@@ -0,0 +1,106 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+
+class FrmYoutubeFeedApi {
```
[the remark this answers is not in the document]
Liviu-p: I updated the YT API feed to use FrmFormApi.

classes/models/FrmYoutubeFeedApi.php
Liviu-p: I updated the YT feed to extend the FrmFormApi. All the requests are handled by FrmFormApi now.

Liviu-p:
> I'm seeing no entries in this report.
> 
> When I go to add more entries, none still appear. I'm not sure if it's caching or something, but it would be nice if we cleared that cache when a new entry is created so we could see something.
> 
> Is this likely to always look empty on Monday? Would we want to show the previous week or something on Monday? If we're checking - 0 DAYs, we're probably not checking anything?
> 
> Screen Shot 2023-12-11 at 4 50 14 PM

Hey @Crabcyborg, I'm wondering if you are still able to replicate it with the latest PRO updates?

Crabcyborg (approved):
I skimmed through the code again, and everything looks good to me.

Once we get the conflicts resolved, I think we're good.

I have a couple small comments in Pro as well (https://github.com/Strategy11/formidable-pro/pull/4548#discussion_r1424220895 and https://github.com/Strategy11/formidable-pro/pull/4548#discussion_r1424223236), but otherwise I think the code is good.

I'll keep testing it out, see if I can find anything else.

classes/controllers/FrmAppController.php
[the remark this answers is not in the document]
stephywells: @Liviu-p I ended up redoing this in the [templates page PR](https://github.com/Strategy11/formidable-forms/pull/1304/commits/0961031bd0ce116d2127dcac225b2738fc710668#diff-039c4f3ae8d934c84e54afe21380b1772b3789671c15484efafdccdab8a7c155R148). Now, all you need to do is add this line after checking the page:
`add_filter( 'frm_show_footer_links', '__return_false' );`

css/frm_admin.css
```diff
@@ -1586,6 +1586,9 @@ input.frm_insert_in_template {
 	display: flex;
 	gap: var(--gap-sm);
 }
+.frm-vertical-align-middle {
```
stephywells, line 1589: We've been trying to stay close to Tailwind for classes. It would be nice to switch over at some point.

So can we call this .frm-items-center ?

classes/views/dashboard/templates/pro-features-list.php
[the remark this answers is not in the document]
stephywells: Thanks! Let's go ahead and remove 'lite-upgrade' from this so we can control it from one place. We want to use the default.

images/dashboard/close.svg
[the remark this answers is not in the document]
stephywells: Thank you!

css/admin/dashboard.css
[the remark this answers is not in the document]
stephywells: I'm still seeing this. The values in the entries table feel very heavy when they are all bold.

classes/controllers/FrmHooksController.php
```diff
@@ -163,7 +167,7 @@ public static function load_admin_hooks() {
 		add_action( 'frm_after_duplicate_form', 'FrmForm::after_duplicate', 10, 2 );
 
 		// Inbox Controller.
-		add_action( 'admin_menu', 'FrmInboxController::menu', 50 );
+		// add_action( 'admin_menu', 'FrmInboxController::menu', 50 );
```
stephywells, line 170: What's the plan here?

classes/controllers/FrmSettingsController.php
```diff
@@ -12,6 +12,15 @@ public static function menu() {
 		add_submenu_page( 'formidable', 'Formidable | ' . __( 'Global Settings', 'formidable' ), __( 'Global Settings', 'formidable' ), 'frm_change_settings', 'formidable-settings', 'FrmSettingsController::route' );
 	}
 
+	/**
+	 * Lite license copy.
+	 * Used in FrmDashboardController & self::license_box
+	 *
+	 * @return string
+	 */
+	public static function copy_for_lite_license() {
```
stephywells, line 21: Can we move this to a helper file instead? If possible, it's nice if functions called from view files use helpers.

classes/controllers/FrmEntriesController.php
```diff
@@ -705,6 +705,22 @@ public static function entry_sidebar( $entry = false ) {
...
 	}
 
+	/**
+	 * Get entries count.
+	 *
+	 * @since 6.x
+	 *
+	 * @return int
+	 */
+	public static function get_entries_count() {
```
stephywells, line 715: Since this function interacts with the database, it would be great if we could move it to FrmEntry.php

Crabcyborg:
@Liviu-p How much time do you think you may need?

I'm wondering if we need to keep this update out of the release or not if it needs more time.

classes/views/FrmDashboardView.php
stephywells, line 1: How would you feel about moving this into the models folder? This feels like it doesn't fit here.

css/admin/dashboard.css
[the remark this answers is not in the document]
stephywells: I'm still seeing a few that would be nice to update. I'll replace all the colors to use existing colors.

Liviu-p:
> @Liviu-p How much time do you think you may need?
> 
> I'm wondering if we need to keep this update out of the release or not if it needs more time.
> 
> I generally don't make releases much later than a couple hours from now.

@Crabcyborg I see around 2hrs from now.

classes/views/FrmDashboardView.php
```diff
@@ -0,0 +1,464 @@
...
+					'link'  => '',
+				),
+				array(
+					'label' => '',
+					'link'  => '',
+				),
+			),
+		),
+		'entries'  => array(
+			'widget-heading'   => 'Latest Entries',
```
stephywells, line 37: Let's make all of these new strings translatable.

classes/views/FrmDashboardView.php
Liviu-p: I see the models more related to DB queries. Maybe we can move it to helpers' folder?

classes/views/FrmDashboardView.php
```diff
@@ -0,0 +1,464 @@
...
+		return $widget_heading . self::load_entries_list_template();
+	}
+
+	/**
+	 * Dashboard - load the entries list template.
+	 *
+	 * @return string HTML template
+	 */
+	private static function load_entries_list_template() {
+		add_filter(
```
stephywells, line 428: I'm seeing the "screen options" tab when I scroll all the way down. Can we remove that for this page?

classes/views/FrmDashboardView.php
Liviu-p: These are used as defaults in view class and can it be empty. At the moment adding it won't allow to add a translation function( or any other function) into the class property. I can remove all the default labels there and keep the defaults empy?

classes/controllers/FrmHooksController.php
Liviu-p: We're removing the inbox submenu with the dashboard update. I will remove the line.

classes/controllers/FrmDashboardController.php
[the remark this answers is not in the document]
stephywells: I think a left a comment in the wrong spot. I meant for the label in the sidebar box to be updated, but not the features list at the bottom of the page.
Screenshot 2023-12-12 at 12 21 02 PM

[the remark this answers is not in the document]
stephywells: Actually... nevermind. I'm making an update to the width so it works

classes/views/dashboard/templates/license-management.php
```diff
@@ -0,0 +1,20 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+?>
+<div class="frm-dashboard-license-management">
```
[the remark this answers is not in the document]
stephywells: Now I'm seeing this:
Screenshot 2023-12-12 at 12 22 52 PM

Closer, but we still need to text back.

classes/views/FrmDashboardView.php
stephywells: Can we move this into a function so we can translate it? It looks like there's some duplication with this too. It gets confusing to have the same text in multiple places.

classes/views/FrmDashboardView.php
stephywells: That would be great! Thank you!

js/formidable_admin.js
```diff
@@ -9681,6 +9683,8 @@ function frmAdminBuildJS() {
 			});
 		},
 
+		initNewFormModal,
```
stephywells, line 9686: This line is causing a js error since the merge.

js/src/dashboard.js
```diff
@@ -0,0 +1,100 @@
...
+		const userEmailInput  = document.querySelector( '#frm_leave_email' );
+		const subscribeButton = document.querySelector( '#frm-add-my-email-address' );
+
+		subscribeButton.addEventListener( 'click', () => {
+			this.saveSubscribedEmail( userEmailInput.value ).then();
+		});
+	}
+
+	initIntroWidgetAnimation() {
+		const widgets = document.querySelectorAll( '.frm-dashboard-widget.frm-animate' );
```
stephywells, line 40: Looks like we're doing these animations in a couple different ways. Can we merge these into 1? I think it's in the expired license warning, the new templates page, and here.

js/formidable_admin.js
Crabcyborg: This function is gone now. There is no new form modal.

Liviu-p: Steph, I reverted the modal js functions. The license modal depends on this function at the moment.

classes/controllers/FrmHooksController.php
stephywells: Sounds good. Looks like we need to deprecate FrmInboxController::menu too.

js/formidable_admin.js
stephywells: Hmm... this is a lot of extra js that we don't need. Can we switch this to work in a more standard way now?

images/dashboard/Frame.svg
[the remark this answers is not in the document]
stephywells: No problem. I'm on it.

classes/views/FrmDashboardView.php
stephywells: Also, it looks like we should switch from using esc_html__() to just __() for these. Then we escape when echoing the value.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,576 @@
...
+			),
+		);
+		require FrmAppHelper::plugin_path() . '/classes/views/dashboard/dashboard.php';
+	}
+
+	/**
+	 * Init top counters widgets view args used to construct FrmDashboardView.
+	 *
+	 * @param object|false $latest_available_form. If a form is availble, we utilize its ID to direct the 'Create New Entry' link of the entries counter CTA when no entries exist.
+	 * @param array $counters_value. The counter values for "Total Forms" & "Total Entries"
```
stephywells, line 123: I'm seeing quite a few phpcs issues with the doc blocks. For example:
- Doc comment for parameter $latest_available_form. does not match actual variable name $latest_available_form (there can't be punctuation here)

js/formidable_admin.js
Liviu-p: sure @stephywells, I'll update it.

package.json
```diff
@@ -31,6 +31,7 @@
 		"classnames": "^2.2.6",
 		"google-closure-compiler": "^20181210.0.0",
 		"npm": "^10.2.5",
+		"perfect-scrollbar": "^1.5.5",
```
stephywells, line 34: What is the benefit of using Perfect Scrollbar over just css?

Liviu-p: The scrollbar's appearance and functionality seem to be more consistently styled using JavaScript rather than CSS, especially considering the variety of devices, operating systems, and browsers. Regarding the CSS scrollbar, it still isn't supported by a few browsers like IE 11 and Firefox. 
https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,576 @@
...
+
+	}
+
+	/**
+	 * Check if user has closed the welcome banner. The status of banner is saved in a cookie: self::$banner_closed_cookie_name.
+	 *
+	 * @return boolean
+	 */
+	public static function welcome_banner_has_closed() {
+		if ( isset( $_COOKIE[ self::$banner_closed_cookie_name ] ) ) {
```
stephywells, line 397: Let's use the standard dismiss notification system for a user instead of a cookie. We don't want it coming back when clearing cookies or switching between browsers or devices.

classes/views/dashboard/templates/license-management.php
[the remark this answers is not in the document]
Liviu-p: Hey @stephywells, this one should be fixed

css/frm_admin.css
Liviu-p: I updated it to .frm-items-center

images/dashboard/Frame.svg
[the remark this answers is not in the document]
Liviu-p: Thank you @stephywells!

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -15,6 +15,7 @@ class FrmEntriesListHelper extends FrmListHelper {
 
 	/**
 	 * @return void
+	 * @param array $args
```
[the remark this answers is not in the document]
Liviu-p: I've removed it

classes/controllers/FrmHooksController.php
Liviu-p: I removed it and added deprecated message to FrmInboxController::menu.

classes/controllers/FrmEntriesController.php
Liviu-p: I moved it to FrmEntry

classes/views/FrmDashboardView.php
Liviu-p: I renamed the `FrmDashboardView` to `FrmDashboardProHelper` and moved it to the helpers folder. I did the same for PRO

classes/views/FrmDashboardView.php
Liviu-p: I removed the default args from the view property, I thought it would help better to find the possible values that DasboardView can take. Also, I fixed the translations in the controllers

classes/views/FrmDashboardView.php
Liviu-p: I removed it

package.json
Liviu-p: I removed the perfect scrollbar from npm package.

classes/controllers/FrmDashboardController.php
Liviu-p: I believe it should be fixed now.

js/formidable_admin.js
Liviu-p: I fixed the license modal js error that was triggered by `initNewFormModal` removal. Also, cleared back the `initNewFormModal`  and its related functions.

package.json
stephywells: Thanks. I'm fine with switching to something like this, but I think we'd need to make it a separate PR so we cover all the scrollbars so we can keep them all consistent. For now, I'd rather not have a bunch of different scrollbar styles.

js/src/components/class-counter.js
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	initTimeoutInterval() {
+		if ( this.value < 10 ) {
+			return 160;
+		}
+		if ( this.value < 70 ) {
+			return 40;
+		}
+		return 4;
```
stephywells, line 35: I have 45 forms and this counter feels a little slow. For 330 entries, it's just fine and 330 is done counting a couple seconds before the 45 forms.

Maybe it would be better to have a calculation here so the counters finish at the same time? Maybe something like 2 seconds / value?

classes/models/FrmEntry.php
```diff
@@ -1087,4 +1087,20 @@ public static function get_id_by_key( $key ) {
 
 		return (int) $entry_id;
 	}
+
+	/**
+	 * Get entries count.
+	 *
+	 * @since 6.x
```
stephywells, line 1094: I think there's a search and replace for x.x, but I don't think 6.x will be replaced automatically. Is that right @Crabcyborg ?

classes/controllers/FrmFormsController.php
```diff
@@ -2995,6 +2995,48 @@ function() {
 		);
 	}
 
+	/**
+	 * Count and return total forms.
+	 *
+	 * @since 6.x
+	 * @return int
+	 */
+	public static function get_forms_count() {
```
stephywells, line 3004: Can these two functions move the FrmForm too since it's a database interaction?

css/frm_admin.css
```diff
@@ -1680,6 +1696,31 @@ input.frm_insert_in_template {
...
+
+.frm-gap-md {
+	gap: var(--gap-md);
+}
+
+.frm-transition-ease {
+	transition: all .2s ease;
+}
+
+.frm_invisible {
```
stephywells, line 1715: Looks like we ended up with some duplication here. Must have happened in a merge conflict?

js/src/components/class-counter.js
Liviu-p: Of course, I can update that. Just double-checking on the new expected behavior: should all the counters complete in 2s?

classes/views/dashboard/templates/license-management.php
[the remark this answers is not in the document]
stephywells: Awesome! Looks good. Thank you!

package.json
Liviu-p: Sure, I just didn't notice that custom scroll bars were already implemented. On MAC it left to me like it was used default one.

classes/views/dashboard/templates/inbox.php
```diff
@@ -0,0 +1,84 @@
...
+					<?php endforeach; ?>
+				</div>
+
+			</div>
+
+			<div class="frm-scrollbar-wrapper">
+				<div class="frm-dismissed-inbox-messages">
+				<?php foreach ( $template['dismissed'] as $key => $message ) : ?>
+					<div class="frm-inbox-message-container frm-flex-col">
+						<div class="frm-inbox-message-heading frm-flex-box frm-justify-between">
```
stephywells, line 60: Looks like this block is duplicated. Would it work to combine these into one function?

js/src/components/class-counter.js
stephywells: .getAttribute is actually faster than .dataset. @Crabcyborg what's your preference here?
https://www.measurethat.net/Benchmarks/Show/14432/0/getattribute-vs-dataset#latest_results_block

package.json
stephywells: Most of it was added in Sherv's form template page PR. No worries!

classes/controllers/FrmDashboardController.php
stephywells: I'm still seeing it in a few spots like this:
`@param string $heading,`

classes/controllers/FrmSettingsController.php
stephywells: That looks good. Since we're using it across multiple areas, I think that works great.

classes/controllers/FrmHooksController.php
stephywells: It looks like the FrmInboxController::inbox() function is now unused too and the views/inbox/list.php.

We might need to trigger these from the dashboard page now?:
```
self::add_tracking_request();
self::add_free_template_message();
```

images/dashboard/Frame.svg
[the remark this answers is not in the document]
stephywells: This is mostly done. I think we can keep the bell icon as is to make it easier to use from the css.

I still need to check the email icon since we've got others.

classes/controllers/FrmDashboardController.php
stephywells: I think we can take this a step further and either use this:
```
public static function is_dashboard_page() {
	return FrmAppHelper::is_admin_page( self::$assets_handle_name );
}
```

Or we could remove this function and use FrmAppHelper::is_admin_page( self::$assets_handle_name ) directly. This is super minimal, so not a big deal if it stays as is either.

classes/views/FrmDashboardView.php
stephywells: We can totally do a default, but in a function so we can avoid the duplication. Or a doc block might cover what you had in mind?

classes/controllers/FrmAppController.php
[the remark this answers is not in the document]
Liviu-p: I believe this one was fixed.

classes/views/dashboard/templates/pro-features-list.php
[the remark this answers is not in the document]
Liviu-p: I believe it's updated now.

css/admin/dashboard.css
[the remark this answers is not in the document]
Liviu-p: I believe it's fixed now.

classes/controllers/FrmHooksController.php
Liviu-p: I believe these are triggered in `FrmInboxController::get_inbox_messages`

classes/controllers/FrmAppController.php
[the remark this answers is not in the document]
stephywells: Yep, you're right. Sorry I missed this.

classes/views/dashboard/templates/pro-features-list.php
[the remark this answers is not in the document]
stephywells: Yep, this is fixed too.

css/admin/dashboard.css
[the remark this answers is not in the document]
stephywells: Oops, yes, I got this one.

classes/controllers/FrmHooksController.php
stephywells: Yep, you're right. That part is covered.

We'll still need to remove more code related to showing the old inbox like the FrmInboxController::inbox() function.

classes/views/FrmDashboardView.php
Liviu-p: Steph, I added a comment block, do you think that will work?

js/src/dashboard.js
Liviu-p: I added a `FrmAnimate` module that will handle all these 3 animations. Also, it will add dynamically through Javascript(it will be included only once per page), the animation.css file.

classes/controllers/FrmDashboardController.php
Liviu-p: I think I fixed those, it looks like my IDE is not picking all the phpcs issues. I guess is related to local repo phpcs configs.

classes/controllers/FrmDashboardController.php
Liviu-p: I updated the dismiss notification system to be based on user ID which is stored in dashboard options.

js/src/components/class-counter.js
Liviu-p: I updated a bit the `FrmCounter` functionality, it will have the `timeToFinish` option property as a reference for setting up the counters' speed. It's set to default to 1.4s which means that counters will finish at the same time( 1.4s )

classes/models/FrmEntry.php
Liviu-p: I updated it to `x.x`

css/frm_admin.css
Liviu-p: I removed the CSS duplicates

classes/views/dashboard/templates/inbox.php
Liviu-p: I removed the duplicate block by moving into a larger `foreach` loop.

classes/controllers/FrmFormsController.php
Liviu-p: It should be updated in the most recent commit.

stephywells: Thank you!

classes/controllers/FrmDashboardController.php
stephywells: Looks great!

classes/views/dashboard/templates/license-management.php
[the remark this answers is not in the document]
stephywells: Actually... I'm still seeing issues. For example, when pro is installed and I don't have a license active, I'm seeing this:
Screenshot 2023-12-21 at 1 27 16 PM

It should be like this since a license is required:
Screenshot 2023-12-21 at 3 38 38 PM

I'm seeing this for Elite:
Screenshot 2023-12-21 at 1 25 58 PM

I pushed an update to use the existing code instead of entirely rewriting it and running it in two different ways.

stephywells (changes requested):

formidable.php
```diff
@@ -88,6 +88,8 @@ function frm_class_autoloader( $class_name, $filepath ) {
 			$filepath .= 'helpers/';
 		} elseif ( preg_match( '/^.+Controller$/', $class_name ) ) {
 			$filepath .= 'controllers/';
+		} elseif ( preg_match( '/^.+View$/', $class_name ) ) {
```
stephywells, line 91: Looks like this should be removed now.

js/src/form-templates/events/categoryListener.js
```diff
@@ -4,8 +4,9 @@
 import { getElements } from '../elements';
 import { PREFIX, CURRENT_CLASS, getAppState, setAppState } from '../shared';
 import { showSelectedCategory } from '../ui';
-import { fadeIn, onClickPreventDefault } from '../utils';
+import { onClickPreventDefault } from '../utils';
 import { resetSearchInput } from './';
+import { FrmAnimate } from '../../common/utilities/animation';
```
stephywells, line 9: Let's keep this consistent. Looks like we're using this in another spot:
import { FrmAnimate } from '../utils';

So it should be:
import { FrmAnimate, onClickPreventDefault } from '../utils';

It would be best if the naming is consistent too with a lowercase f (frmAnimate)

classes/views/FrmDashboardView.php
stephywells: This works great. Thank you!

js/src/dashboard.js
stephywells: Thanks so much! This will be much better.

classes/views/dashboard/templates/youtube-video.php
```diff
@@ -0,0 +1,10 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+?>
+<iframe style="width: 100%; height: 224px; margin-bottom:-6px"
+	src="https://www.youtube.com/embed/aLJLDtPHXY8?si=7_1qoP9fPiaiUt0f" 
+	title="YouTube video player" frameborder="0" 
+	allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen>
```
stephywells, line 9: I'm seeing a console error:
> Unrecognized feature: 'web-share'.

classes/controllers/FrmInboxController.php
```diff
@@ -49,16 +48,30 @@ public static function dismiss_all_button( $atts ) {
...
 		self::add_tracking_request();
 		self::add_free_template_message();
 
-		$inbox    = new FrmInbox();
-		$messages = $inbox->get_messages( 'filter' );
-		$messages = array_reverse( $messages );
-		$user     = wp_get_current_user();
+		$inbox              = new FrmInbox();
+		$unread_messages    = $inbox->get_messages();
+		$dismissed_messages = $inbox->get_messages();
```
stephywells, line 65: Let's reuse the messages already fetched instead of calling the same code again.
```suggestion
		$dismissed_messages = $unread_messages;
```

classes/controllers/FrmInboxController.php
```diff
@@ -49,16 +48,30 @@ public static function dismiss_all_button( $atts ) {
...
+
+		$inbox->filter_messages( $unread_messages, 'filter' );
+		$inbox->filter_messages( $dismissed_messages, 'dismissed' );
 
-		include( FrmAppHelper::plugin_path() . '/classes/views/inbox/list.php' );
+		return array(
+			'unread'    => array_reverse( $unread_messages ),
+			'dismissed' => array_reverse( $dismissed_messages ),
+			'user'      => wp_get_current_user(),
```
stephywells, line 73: Is this being used?

classes/controllers/FrmAppController.php
```diff
@@ -174,6 +170,46 @@ private static function is_white_page() {
...
+	}
+
+	/**
+	 * Check if Views is an active page.
+	 *
+	 * @since x.x
+	 *
+	 * @return bool
+	 */
+	private static function is_views_page() {
```
stephywells, line 197: We've got FrmAppHelper::is_view_builder_page().

Let's move this new function there where it's more accessible.

js/src/form-templates/ui/showHeaderCancelButton.js
```diff
@@ -2,7 +2,7 @@
  * Internal dependencies
  */
 import { getElements } from '../elements';
-import { fadeIn } from '../utils';
+import { FrmAnimate } from '../../common/utilities/animation';
```
stephywells, line 5: Looks like this is a different title case that everything else. Looks like we need to update all "FrmAnimate" to "frmAnimate"

js/src/form-templates/events/categoryListener.js
```diff
@@ -4,8 +4,9 @@
 import { getElements } from '../elements';
 import { PREFIX, CURRENT_CLASS, getAppState, setAppState } from '../shared';
 import { showSelectedCategory } from '../ui';
-import { fadeIn, onClickPreventDefault } from '../utils';
+import { onClickPreventDefault } from '../utils';
 import { resetSearchInput } from './';
+import { FrmAnimate } from '../utils';
```
stephywells, line 9: This has a different source file than all the other FrmAnimate occurances.

classes/helpers/FrmDashboardHelper.php
```diff
@@ -0,0 +1,330 @@
...
+	 *                @type string $payments['counters'][]['items'][]['counter_label']
+	 *                @type int    $payments['counters'][]['items'][]['counter']
+	 *    @type array $video Array of video args
+	 *        @type string $video['id'] YouTube video ID
+	 * }
+	 *
+	 * @return void
+	 */
+	public function __construct( $data ) {
+		if ( isset( $data['counters'] ) ) {
```
stephywells, line 91: Looks like this can be simplified quite a bit.
```
$sections = array( 'counters', 'license', 'payments', ... );
foreach ( $sections as $section ) {
  if ( isset( $data[ $section ] ) ) {
    $this->view[ $section ] = $data[ $section ];
  }
}
```

css/admin/dashboard.css
```diff
@@ -0,0 +1,506 @@
...
+	margin: 0px;
+	padding: 0px;
+}
+
+.frm-counter-card-wrapper.frm-payments-placeholder .frm-counter-card h3,
+.frm-dashboard-container .frm-dashboard-widget h2.frm-widget-heading {
+	font-size: var(--text-md);
+	font-weight: 600;
+	position: relative;
+	z-index: 22;
```
stephywells, line 45: Are these two lines doing anything? I'm seeing this in a couple spots, but I'm not seeing anything different when I remove them.

```
position: relative;
z-index: 22;
```

stephywells (approved):
We're so close! This is my final review and then we can get some extra QA and a final review from Mike.

Thanks so much for hanging in there. I've learned from this PR and Sherv's form templates PR that we need to figure out how we can be working on the whole plugin with easy maintenance and global consistency in mind. Part of this is knowing what elements are available to reuse, so it may require a big refactor of the admin css to make it easier to know what we have.

We can add more notes in ClickUp on this, but I'd love to hear your ideas on how we can do this. Thanks so much for all the work you put into creating this dashboard!!!

classes/controllers/FrmAppController.php
```diff
@@ -174,6 +170,34 @@ private static function is_white_page() {
...
+	 * @return bool
+	 */
+	private static function is_grey_page() {
+		$grey_pages = array(
+			'formidable-applications',
+			'formidable-dashboard',
+		);
+
+		$is_grey_page = self::is_page_in_list( $grey_pages );
+		return apply_filters( 'frm_is_grey_page', $is_grey_page );
```
Crabcyborg, line 187: It's nice to add block comments for new filters. We usually include a `@since` comment and `@param` comments for the variables that are passed to the filter.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,621 @@
...
+	 * @var string
+	 */
+	private static $option_meta_name = 'frm-dashboard-options';
+
+	/**
+	 * Option name used to store the time when auto redirected for welcome.
+	 *
+	 * @var string
+	 */
+	public static $redirect_meta_name = 'frm_activation_redirect';
```
Crabcyborg, line 34: Can this be made into a constant?

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,621 @@
...
+				),
+				'license'  => array(),
+				'inbox'    => self::view_args_inbox(),
+				'entries'  => array(
+					'widget-heading'   => __( 'Latest Entries', 'formidable' ),
+					'cta'              => array(
+						'label' => __( 'View All Entries', 'formidable' ),
+						'link'  => admin_url( 'admin.php?page=formidable-entries' ),
+					),
+					'show-placeholder' => 0 < (int) $counters_value['entries'] ? false : true,
```
Crabcyborg, line 163: `? false : true` feels really backwards. Why not just flip the comparison?

```
'show-placeholder' => 0 >= (int) $counters_value['entries'],
```

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,621 @@
...
+		return false;
+	}
+
+	/**
+	 * Check if is dashboard page.
+	 *
+	 * @return boolean
+	 */
+	public static function is_dashboard_page() {
+		if ( FrmAppHelper::is_admin_page( 'formidable-dashboard' ) ) {
```
Crabcyborg, line 445: This function is probably fine just reduced to `return FrmAppHelper::is_admin_page( 'formidable-dashboard' );` I don't imagine this needs to be split up like this. It likely won't change often or at all.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,621 @@
...
+
+	/**
+	 * Detect if the logged user's email is subscribed. Used for inbox email subscribe.
+	 *
+	 * @param string $email The logged user's email.
+	 * @return boolean
+	 */
+	public static function email_is_subscribed( $email ) {
+		$subscribed_emails = self::get_subscribed_emails();
+		return false !== array_search( $email, $subscribed_emails, true ) ? true : false;
```
Crabcyborg, line 459: `? true : false` is really just redundant here. This line would be the same as just `return false !== array_search( $email, $subscribed_emails, true )`.

Crabcyborg: We probably don't need `array_search` either though. `in_array` is a little more readable.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,621 @@
...
+	}
+
+	/**
+	 * Enqueue controller assets.
+	 *
+	 * @return void
+	 */
+	public static function enqueue_assets() {
+
+		if ( false === self::is_dashboard_page() ) {
```
Crabcyborg, line 479: I don't think we need to worry about strict checks with simple `bool` functions that don't return any other type. `if ( ! self::is_dashboard_page() ) {` a bit more natural.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,621 @@
...
+		return self::inbox_prepare_messages( FrmInboxController::get_inbox_messages() );
+	}
+
+	/**
+	 * Prepare inbox messages data.
+	 *
+	 * @return array
+	 */
+	private static function inbox_prepare_messages( $data ) {
+		foreach ( $data as $key => &$messages ) {
```
Crabcyborg, line 502: Can we do this without using the `&messages` and `&$message` passing by reference in a loop? This can get messy.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,621 @@
...
+		// remove dismiss button
+		$pattern = '/<a[^>]*class="[^"]*frm_inbox_dismiss[^"]*"[^>]*>.*?<\/a>/is';
+		return preg_replace( $pattern, ' ', $cta );
+	}
+
+	/**
+	 * Get the embed YouTube video from YouTube feed api. If there are 0 entries we show the welcome video otherwise latest video from FF YouTube channel is displayed.
+	 *
+	 * @param int $entries_count The total entries available.
+	 * @return string The YouTube video ID.
```
Crabcyborg, line 523: This returns `string|null`. We should make sure the type is correct.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,621 @@
...
+	 *
+	 * @param string $email The user email address.
+	 * @return void
+	 */
+	private static function save_subscribed_email( $email ) {
+		$subscribed_emails = self::get_subscribed_emails();
+		if ( false === array_search( $email, $subscribed_emails, true ) ) {
+			$subscribed_emails[] = $email;
+			self::update_dashboard_options( $subscribed_emails, 'inbox-subscribed-emails' );
+			return;
```
Crabcyborg, line 555: This return is unnecessary for now at least.

classes/helpers/FrmAppHelper.php
```diff
@@ -4117,4 +4117,26 @@ public static function dismiss_warning_message( $option = '' ) {
...
+
+		if ( is_callable( 'FrmProAddonsController::get_readable_license_type' ) && ! class_exists( 'FrmProDashboardController' ) ) {
+			// Manage PRO versions without PRO dashboard functionality.
+			$license_type = FrmProAddonsController::get_readable_license_type();
+			if ( 'lite' !== strtolower( $license_type ) ) {
+				$message = 'Formidable Pro ' . $license_type;
+			}
+		}
+
+		return apply_filters( 'frm_license_type_text', $message );
```
Crabcyborg, line 4140: It looks like in Pro this is used like an action. `$message` is never returned, so this function returns `null`.

I think We shouldn't be returning a string in Pro instead of just outputting the HTML immediately.

classes/helpers/FrmDashboardHelper.php
```diff
@@ -0,0 +1,363 @@
...
+	 *
+	 * @return void
+	 */
+	public function get_youtube_video( $classes ) {
+		$template = $this->view['video'];
+		if ( null === $template['id'] ) {
+			return;
+		}
+
+		echo '<div class="' . esc_attr( $classes ) . '">' .
```
Crabcyborg, line 276: Can we do this with a view or something? We generally try to avoid echoing HTML strings like this.

classes/helpers/FrmDashboardHelper.php
```diff
@@ -0,0 +1,363 @@
...
+	 *
+	 * @return void
+	 */
+	public static function load_entries_template( $template ) {
+		if ( true === $template['show-placeholder'] ) {
+			self::load_placeholder_template( $template );
+			return;
+		}
+
+		echo '<div class="frm-flex-box frm-justify-between"><h2 class="frm-widget-heading">' . esc_html( $template['widget-heading'] ) . '</h2><a class="frm-widget-cta" href="' . esc_url( $template['cta']['link'] ) . '">' . esc_html( $template['cta']['label'] ) . '</a></div>';
```
Crabcyborg, line 332: Here as well. It would be nice not to echo an HTML string.

classes/models/FrmInbox.php
```diff
@@ -174,11 +174,11 @@ private function clean_messages() {
 	/**
 	 * @return void
 	 */
-	private function filter_messages( &$messages ) {
+	public function filter_messages( &$messages, $type = 'unread' ) {
 		$user_id = get_current_user_id();
 		foreach ( $messages as $k => $message ) {
 			$dismissed = isset( $message['dismissed'] ) && isset( $message['dismissed'][ $user_id ] );
-			if ( empty( $k ) || $this->is_expired( $message ) || $dismissed ) {
+			if ( empty( $k ) || $this->is_expired( $message ) || ( $dismissed && 'dismissed' !== $type ) || ( 'dismissed' === $type && ! $dismissed ) ) {
```
Crabcyborg, line 181: This check is kind of funny. Can it be reduced to something like `$type === 'dismissed' === $dismissed``?

deprecated/FrmWelcomeController.php
```diff
@@ -0,0 +1,123 @@
...
+	 */
+	public static function load_hooks() {
+		_deprecated_function( __METHOD__, '6.7' );
+	}
+
+	/**
+	 * Performs a safe (local) redirect to the welcome screen
+	 * when the plugin is activated
+	 *
+	 * @deprecated 6.7
```
Crabcyborg, line 27: We should change these `6.7` versions to either `x.x` so they get changed on release, or `6.8` since they'll be in that version anyway.

classes/controllers/FrmDashboardController.php
```diff
@@ -0,0 +1,615 @@
...
+	 * Option name used to store the time when auto redirected for welcome.
+	 */
+	const REDIRECT_META_NAME = 'frm_activation_redirect';
+
+	/**
+	 * Handle name used for registering controller scripts and style.
+	 *
+	 * @var string Handle name used for wp_register_script|wp_register_style
+	 */
+	public static $page_slug = 'formidable-dashboard';
```
Crabcyborg, line 18: We could make more of these variable constants. It's nice to use a constant if we don't expect the value to ever change. If it isn't constant, then you really have no way of knowing a variable's value.

stripe/models/FrmTransLitePayment.php
```diff
@@ -78,21 +78,23 @@ public function get_defaults() {
...
 	 * @return array            Contains `count` and `total`.
 	 */
-	public function get_payments_stats( $from_date, $to_date ) {
+	public function get_payments_stats( $from_date = null, $to_date = null ) {
 		$data = array(
 			'count' => 0,
 			'total' => array(),
 		);
 
+		$where = array();
```
Crabcyborg, line 87: The `$where` was getting defined after the check for `if ( ! FrmTransLiteAppHelper::payments_table_exists() ) {`.

We don't need `$where` if we're exiting early there, so it makes sense to keep this logic after the guard clause.

js/src/components/class-tabs-navigator.js
```diff
@@ -0,0 +1,55 @@
...
+		this.navs             = this.wrapper.querySelectorAll( '.frm-tabs-navs ul > li' );
+		this.slideTrackLine   = this.wrapper.querySelector( '.frm-tabs-active-underline' );
+		this.slideTrack       = this.wrapper.querySelector( '.frm-tabs-slide-track' );
+		this.slides           = this.wrapper.querySelectorAll( '.frm-tabs-slide-track > div' );
+
+		this.init();
+	}
+
+	init() {
+		if ( null === this.wrapper || null === this.navs || null === this.trackLine || null === this.slideTrack || null === this.slides ) {
```
Crabcyborg, line 18: Since `this.slides` uses `querySelectorAll`, it won't ever be `null`. It would be an empty array.

Crabcyborg, line 18: `this.navs` also uses `querySelectorAll` and will never be null.

js/src/common/utilities/animation.js
```diff
@@ -0,0 +1,103 @@
...
+		}
+		frmAnimate.init = true;
+		this.loadCssFile();
+	}
+
+	getCssFileUrl() {
+		if ( ! window.frmGlobal ) {
+			return '';
+		}
+		return window.frmGlobal.url + '/css/' + this.cssFilePath  ;
```
Crabcyborg, line 65: There seems to be some weird spacing here between `this.cssFilePath` and the semi-colon.

js/formidable_admin.js
```diff
@@ -10174,6 +10234,15 @@ function frmAdminBuildJS() {
 						return true;
 					}
 					fadeOut( message, function() {
+						if ( null !== dismissedMessagesWrapper ) {
+							dismissedMessage.classList.remove( 'frm-fade' );
+							dismissedMessage.querySelector( '.frm-inbox-message-heading' ).removeChild( dismissedMessage.querySelector( '.frm-inbox-message-heading .frm_inbox_dismiss' ) );
+							dismissedMessagesWrapper.append( dismissedMessage );
+						}
+						if ( 0 === message.parentNode.querySelectorAll( '.frm-inbox-message-container' ).length - 1 ) {
```
Crabcyborg, line 10242: This check is weird. `if ( 0 === length - 1 )` seems more complicated to read than just `if ( 1 === length )`.

js/formidable_admin.js
```diff
@@ -202,33 +223,63 @@ var FrmFormsConnect = window.FrmFormsConnect || ( function( document, window, $
...
+		},
+
+		/**
+		 * Use the data-success element to replace the element content.
+		 */
+		showInlineSuccess: function() {
+			const successElement = document.querySelectorAll( '.frm-confirm-msg [data-success]' );
+			if ( successElement.length ) {
+				successElement.forEach( function( element ) {
+					element.innerHTML = element.getAttribute( 'data-success' );
```
Crabcyborg, line 278: Do we need to use `innerHTML` here? Can we run this through the `purifyHtml` function we have this in file, or use `textContent` if we don't need to support HTML here?

classes/controllers/FrmInboxController.php
Crabcyborg: @Liviu-p Is this comment resolved?

deprecated/FrmWelcomeController.php
```diff
@@ -0,0 +1,123 @@
...
+	 * @return void
+	 */
+	public static function redirect() {
+		_deprecated_function( __METHOD__, 'x.x' );
+	}
+
+	/**
+	 * Add a submenu welcome screen for the formidable parent menu
+	 *
+	 * @deprecated 6.7
```
Crabcyborg, line 37: It looks like there's still a lot of these saying 6.7. Can we update all of these to `x.x` or `6.8`?

css/admin/animations.css
```diff
@@ -0,0 +1,97 @@
+/* File is Loaded via js only - FrmAnimate src: js/src/common/utilities/animation.js */
```
Crabcyborg, line 1: Do we need to load this via JS? It might be nicer on the page load to just include this CSS in `frm_admin.css`? Is there a reason to keep this separate?

Crabcyborg (approved):
Thanks @Liviu-p!

I think this is finally ready to merge.

🥳

PR #1357, opened 2023-10-30
Force truncate for strings way over the limit with no spaces

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3811

I chose some hard coded values here.

If the length is 50 characters over what it should be, and I can't detect any spaces, then I truncate at the limit + 10.

**Before**
Screen Shot 2023-10-30 at 1 35 25 PM

**After**
Screen Shot 2023-10-30 at 1 35 16 PM

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -2408,7 +2408,33 @@ class FrmAppHelper {
 			unset( $total_len, $word );
 		}
 
+		$sub = self::maybe_force_truncate_on_string_with_no_spaces( $sub, $length );
+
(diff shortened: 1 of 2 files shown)


stephywells (commented):
This looks great! I just have one question. "img src" includes a space, so I'm a little confused by how this is catching it.

Crabcyborg:
@stephywells The `strpos` check has a `$offset = $length` in its check, so it only checks for spaces after the limit.

I think the truncate works fine before that space, but it gets too long after the space.

stephywells (approved):

stephywells:
Thanks! I missed that! Great solution!

garretlaxton (approved):
This looks great!

PR #1358, opened 2023-11-01
Support saving blank options for radio fields and checkbox field

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/3385

Since I noticed there were other proposed solutions to this like the one in https://github.com/Strategy11/formidable-forms/pull/1171, I'm just trying a different approach of allowing blank options to be saved/retrieved without being trimmed.

I'm not quite sure if this is worth consider or not but it could be used to brainstorm on it.

Screencast shared by Mike on how to replicate it:
HUMIJ1WBdr


AbdiTolesa:
@Crabcyborg I have drafted this PR to know what you think about it, I'm thinking of an alternative approach to support saving blank options in radio, radio based and checkbox fields. This at least has solved the issue for me. The blank options won't be selected by default. What's your thoughts on it?

Crabcyborg:
I'm going to close this for now since it's so old.

I'd like to find a cleaner fix to the issue if possible.

PR #1366, opened 2023-11-08
Ensure Proper Zebra Striping with Blank Fields

shervElmi:
This PR ensures that zebra striping remains consistent in repeater fields, even when some fields are left blank. The visual integrity of forms will be preserved.

## Related Issue:
[Issue #3688](https://github.com/Strategy11/formidable-pro/issues/3688)

## Testing Instruction:
1. Navigate to `WP Admin > Formidable > Forms`.
2. Create an entry for a form ensuring some fields are left blank.
3. Verify the zebra striping remains correct as shown in the 'After' output.

### Before
image

---

image

### After
image

---

image

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -9819,10 +9819,34 @@ function frmAdminBuildJS() {
 		footerLinks.classList.remove( 'frm_hidden' );
 	}
 
+	/**
+	 * Apply zebra striping to a table while ignoring empty rows.
+	 *
+	 * @param {string} tableSelector The CSS selector for the table.
+	 * @param {string} emptyRowClass The class name used to identify empty rows.
+	 */
+	function applyZebraStriping( tableSelector, emptyRowClass ) {
+		// Get all non-empty table rows within the specified table
+		const rows = document.querySelectorAll( `${tableSelector} tr:not(.${emptyRowClass})` );
+		if ( rows?.length < 1 ) {
+			return;
+		}
+
+		let isOdd = true;
+		rows.forEach( row => {
+			row.classList.add( isOdd ? 'frm-odd' : 'frm-even' );
+			isOdd = !isOdd;
+		});
+
+		const tables = document.querySelectorAll( tableSelector );
+		tables.forEach( table => table.classList.add( 'frm-zebra-striping' ) );
+	};
+
 	return {
 		init: function() {
 			initAddMyEmailAddress();
 			addAdminFooterLinks();
+			applyZebraStriping( '.frm-alt-table', 'frm-empty-row' );
 
 			s = {};
 
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -4142,18 +4142,35 @@ label input[type="checkbox"], label input[type="radio"] {
 }
 
 .frm-alt-table {
+	position: relative;
 	width:100%;
 	border-collapse:collapse;
 	margin-top:0.5em;
 	font-size:15px;
 }
 
+/* Before `applyZebraStriping` is executed, an empty space is displayed */
+.frm-alt-table:not(.frm-zebra-striping)::before {
+	content: '';
+	position: absolute;
+	top: 0;
+	right: 0;
+	bottom: 0;
+	left: 0;
+	background-color: #fff;
+}
+
 .frm-alt-table th {
 	width:200px;
 }
 
-.frm-alt-table tr {
+.frm-alt-table tr,
+.frm-alt-table tr.frm-odd {
 	background-color:transparent;
+}
+
+.frm-alt-table tr.frm-even {
+	background-color: var(--sidebar-color);
 }
 
 .frm-alt-table th,
@@ -4162,10 +4179,6 @@ label input[type="checkbox"], label input[type="radio"] {
 	vertical-align:top;
 	text-align:left;
 	padding:20px;
-}
-
-.frm-alt-table tr:nth-child(even) {
-	background-color: var(--sidebar-color);
 }
 
 .frm-alt-table h3 {


garretlaxton:
This looks good, but if I click the `show empty fields` button the striping isn't correct. Not sure if that matters though?

shervElmi:
@Crabcyborg, what do you think about Garret's input? You can use the QA URL in the description for testing.

Crabcyborg:
@shervElmi Is it possible to just remove the new classes and then run this function again to add them back again when that button is clicked?

js/formidable_admin.js
```diff
@@ -9819,10 +9819,34 @@ function frmAdminBuildJS() {
...
+	/**
+	 * Apply zebra striping to a table while ignoring empty rows.
+	 *
+	 * @param {string} tableSelector The CSS selector for the table.
+	 * @param {string} emptyRowClass The class name used to identify empty rows.
+	 */
+	function applyZebraStriping( tableSelector, emptyRowClass ) {
+		// Get all non-empty table rows within the specified table
+		const rows = document.querySelectorAll( `${tableSelector} tr:not(.${emptyRowClass})` );
+		if ( rows?.length < 1 ) {
```
Crabcyborg, line 9831: This `?` is totally pointless. `querySelectorAll` will always return an array.

js/formidable_admin.js
```diff
@@ -9819,10 +9819,34 @@ function frmAdminBuildJS() {
...
+
+		const tables = document.querySelectorAll( tableSelector );
+		tables.forEach( table => table.classList.add( 'frm-zebra-striping' ) );
+	};
+
 	return {
 		init: function() {
 			initAddMyEmailAddress();
 			addAdminFooterLinks();
+			applyZebraStriping( '.frm-alt-table', 'frm-empty-row' );
```
Crabcyborg, line 9849: It might be nice to keep this out of the global scope and add a new `frmAdminBuild.entriesInit();` like we have for other pages.

Or we could take this moment to add a new `entries.js` file. We have one in Pro already. If we went in that direction, we could help get code out of this file. It has nearly 11,000 lines so I'd love to get it down rather than add more code to this file.

shervElmi:
@garretlaxton, I've refactored all the code. Could you please review this PR again? Thanks! :)

garretlaxton (approved):
This looks great!
image
image

js/admin/entries.js
```diff
@@ -0,0 +1,48 @@
...
+	showEmptyFieldsButton.dataset.show = showEmptyFieldsButton.dataset.show || 'false';
+
+	onClickPreventDefault( showEmptyFieldsButton, () => {
+		// Change button state and update table striping
+		const newShowState = showEmptyFieldsButton.dataset.show === 'true' ? 'false' : 'true';
+		const isShowingEmptyFields = newShowState === 'true';
+
+		showEmptyFieldsButton.dataset.show = newShowState;
+
+		setTimeout(() => {
```
truongwp, line 44: Why do we need this `setTimeout()`. Why is it `200`?

shervElmi: Here's the code where toggleItem uses the jQuery `show` and `hide` methods. The `hide` method defaults to a `200ms` duration for hiding an element. To prevent a flashing design and avoid creating a bug, I've set a setTimeout. After the hiding is complete, I call `applyZebraStriping` again.

image

truongwp: That's ok. Thank you.

truongwp (approved):
This looks good. Thanks @shervElmi.

js/admin/entries.js
```diff
@@ -0,0 +1,48 @@
...
+	applyZebraStriping( '.frm-alt-table', 'frm-empty-row' );
+
+	/**
+	 * Manages the 'Show Empty Fields' button.
+	 *
+	 * Clicking the button switches its state between showing and hiding empty table fields.
+	 * It also toggles zebra striping on the table to reflect this change, making it easy
+	 * for users to see or hide empty fields.
+	 */
+	const showEmptyFieldsButton = document.getElementById( 'frm-entry-show-empty-fields' );
```
Crabcyborg, line 32: Since this script is also loaded on the entries list page, a JS error is triggered when this element doesn't exist.

Screen Shot 2023-11-15 at 10 05 47 AM

js/admin/entries.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * entries.js File
+ *
+ * Contains scripts for the Entries admin page.
+ */
+
+/**
+ * WordPress dependencies
+ */
+const { domReady } = wp;
```
Crabcyborg, line 10: Can we declare these inside of the `domReady` function? We shouldn't add anything to the global scope to avoid conflicts.

js/formidable_admin.js
```diff
@@ -10655,6 +10681,7 @@ function frmAdminBuildJS() {
 			}
 		},
 
+		applyZebraStriping: applyZebraStriping,
```
Crabcyborg, line 10684: We could simplify all of these and just show `applyZebraStriping` instead of `applyZebraStriping: applyZebraStriping,`, `infoModal` instead of `infoModal: infoModal,`, etc.

Crabcyborg: The syntax only works in browsers that support ES6, so we don't do it on our front end scripts, but updating our admin JS to make it more modern is always welcome.

Crabcyborg (approved):
This looks good!

I made a couple tiny commits to add a space and improve readability a bit.

I'll merge this for today's release.

Thanks @shervElmi!

Thank you @truongwp and @garretlaxton for your help reviewing this!

🚀

shervElmi:
Thanks, @Crabcyborg! 🚀

PR #1369, opened 2023-11-10
Fix directory shows license error

Crabcyborg:
This fixes an issue I'm seeing in our directory add on.

But we should probably fix this in the API.


Crabcyborg:
I don't think we'll actually merge this so I'm going to close it.

PR #1371, opened 2023-11-11
Compliance with White Labeling in Backend

shervElmi:
This PR addresses an oversight in the backend white labeling, particularly in the import menu. It ensures that when white labeling is enabled, the user-chosen brand name replaces 'Formidable' in all backend interfaces, including the import page.

## Related Issue:
[White labeling needs to comply in the backend](https://github.com/Strategy11/formidable-pro/issues/4058)

## QA URL
https://qa.formidableforms.com/sherv11/wp-admin/admin.php?page=formidable-import

## Testing Instruction:
1. Navigate to `WP Admin > Formidable > Global Settings > White Labeling`.
2. Change the 'Plugin Label' to a value different from 'Formidable'.
3. Go to `WP Admin > Formidable > Import/Export`.
4. Confirm that 'Formidable' is replaced with the new brand name in the page descriptions and the XML file title section.
5. > Important: After changing the brand name, ensure you `disable` the` Formidable PRO` plugin. Note that there is a separate PR for the PRO version.
https://github.com/Strategy11/formidable-pro/pull/4559

### Before
<kbd>
image
</kbd>

---

<kbd>
image
</kbd>

### After
<kbd>
image
</kbd>

---

<kbd>
image
</kbd>


garretlaxton (approved):
This works great!

truongwp:
@Crabcyborg This looks good. But users will need to translate those strings again. Do you think we should check if the white label is `Formidable`, keep the old strings?

Crabcyborg:
Thanks Truong. That sounds like a good idea.

shervElmi:
@Crabcyborg and @truongwp, I'd like to hear your thoughts on how I can implement your suggestion with less repetition. Here's an example if we want to separate them using a simple if condition:

image

truongwp:
@shervElmi Yes, but the condition should be `! FrmAppHelper::is_formidable_branding()`.

truongwp (approved):

Crabcyborg (approved):
This looks great!

Thank you @shervElmi @truongwp and @garretlaxton!

🚀

PR #1372, opened 2023-11-13
Stop adding aria-required to role="group" and add aria-required to th…

Crabcyborg:
…e first checkbox instead

Fixes https://github.com/Strategy11/formidable-pro/issues/4450

In the more recent ticket, they're using axe. The same error appears there as it does in Pagespeed.

Screen Shot 2023-11-13 at 10 28 20 AM

This update fixes both of those.

The issue is that `role="group"` doesn't support `aria-required`. There is no good role for a group of checkboxes that works like `radiogroup`. A `radiogroup` with `aria-required` is totally valid, and passes in both accessibility tests, but `group` doesn't.

What I've done now, is I add the `aria-required="true"` attribute to the **first** checkbox in a group only. This way it isn't read for each option.

### DIFF
--- a/classes/models/FrmFieldFormHtml.php
+++ b/classes/models/FrmFieldFormHtml.php
@@ -501,30 +501,22 @@ class FrmFieldFormHtml {
 			return;
 		}
 
-		$field                       = (array) $this->field_obj->get_field();
-		$attributes                  = array();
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This solves the issue on pagespeed for me.

PR #1383, opened 2023-11-21
Fix how 'other' option checked state is determined

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4578 together with https://github.com/Strategy11/formidable-pro/pull/4584

I believe this fix would ideally be in Pro but since most of the stuff is built into Lite, I'm just updating lite.

Fixes in this update:
- Checked state of other option on front end page load.

### Screencast
https://github.com/Strategy11/formidable-forms/assets/41271840/e652528c-94ab-45dd-adbe-0922763d0ffc

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -1236,6 +1236,10 @@ class FrmFieldsHelper {
 			$checked = 'checked="checked" ';
 		}
 
+		if ( ! $checked && is_array( $args['field']['value'] ) && in_array( $args['field_val'], $args['field']['value'], true ) ) {
+			$checked = 'checked="checked" ';
+		}
+
 		return $other_args;
 	}
 
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -1236,7 +1236,13 @@ class FrmFieldsHelper {
 			$checked = 'checked="checked" ';
 		}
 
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This looks great!

classes/helpers/FrmFieldsHelper.php
```diff
@@ -1236,6 +1236,10 @@ public static function prepare_other_input( $args, &$other_opt, &$checked ) {
 			$checked = 'checked="checked" ';
 		}
 
+		if ( ! $checked && is_array( $args['field']['value'] ) && in_array( $args['field_val'], $args['field']['value'], true ) ) {
```
truongwp, line 1239: We should check if `$args['field_val']` exists because this method is called in some other places.

AbdiTolesa: Nice catch, thanks!

truongwp (approved):

classes/helpers/FrmFieldsHelper.php
```diff
@@ -1236,6 +1236,10 @@ public static function prepare_other_input( $args, &$other_opt, &$checked ) {
 			$checked = 'checked="checked" ';
 		}
 
+		if ( ! $checked && is_array( $args['field']['value'] ) && isset( $args['field_val'] ) && in_array( $args['field_val'], $args['field']['value'], true ) ) {
```
Crabcyborg, line 1239: @AbdiTolesa What do you think about adding a field type check here for checkboxes? It helps add more context to this. It may also benefit to have a comment since it's a fair bit of logic without any description.

Crabcyborg (approved):
Thanks @AbdiTolesa!

And thank you @truongwp and @garretlaxton for reviewing it!

🚀

PR #1385, opened 2023-11-21
Set a no_update value on update plugins transient so auto update is a…

Crabcyborg:
…vailable when up to date

Related issue https://github.com/Strategy11/formidable-pro/issues/2624

Prior to this update, the Enabled/Disable auto-updates options were only visible when a form was not up to date (when it shows an option to Update Now).

I looked at https://github.com/Strategy11/business-directory-premium/blob/main/models/class-wpbdp-addons.php#L107-L111 and saw it had this `else` condition so I added something similar. Now it always shows the Disable/Enable options.

Screen Shot 2023-11-21 at 4 07 24 PM

I don't know if this is all that needs to change though. I didn't have any luck getting it to auto-update on my local environment.

According to my Site Health it doesn't try to auto-update when you're using Git.
Screen Shot 2023-11-21 at 4 09 08 PM

@stephywells Is this all that we should need to do? What's the best way to test this?

### DIFF
--- a/classes/controllers/FrmAddonsController.php
+++ b/classes/controllers/FrmAddonsController.php
@@ -352,12 +352,11 @@ class FrmAddonsController {
 				continue;
 			}
 
-			$wp_plugin  = isset( $wp_plugins[ $folder ] ) ? $wp_plugins[ $folder ] : array();
-			$wp_version = isset( $wp_plugin['Version'] ) ? $wp_plugin['Version'] : '1.0';
+			$wp_plugin    = isset( $wp_plugins[ $folder ] ) ? $wp_plugins[ $folder ] : array();
+			$wp_version   = isset( $wp_plugin['Version'] ) ? $wp_plugin['Version'] : '1.0';
+			$plugin->slug = explode( '/', $folder )[0];
 
 			if ( version_compare( $wp_version, $plugin->new_version, '<' ) ) {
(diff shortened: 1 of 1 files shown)


stephywells (commented):
This has worked for me in the past. That's cool that WP doesn't auto update git plugins.

We might also need to move these two lines before the condition:
$slug                           = explode( '/', $folder );
$plugin->slug                   = $slug[0];

This worked for me when I tested with those two lines moved up, and I added this in formidable.php to trigger the auto updates:
```
require ABSPATH . WPINC . '/pluggable.php';
require_once ABSPATH . 'wp-admin/includes/admin.php';
	require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';

	$upgrader = new WP_Automatic_Updater();
	$upgrader->run();
```

Crabcyborg:
Thanks Steph.

I see what you mean about the slug. I moved that up now.

I'm still having no luck getting auto update to work though. I tried a folder that wasn't a git directory, setting my upgrade folder to 777. I tried that script before, and tried it again. No luck.

stephywells:
I tested with turning on auto updates for a plugin that isn't ours, and I changed the folder name for one of my git plugins, so I could download a non-git version from the Formidable -> add-ons page. It updated fine for me.

PR #1387, opened 2023-11-22
Add composite indexes for entries tables

Crabcyborg:
Related update https://github.com/Strategy11/formidable-pro/pull/4540#pullrequestreview-1730324322

This moves the indexes into Lite so database queries are optimized for all sites.

I run this after it creates the tables so it always runs. I tried getting it to work in the `dbDelta` call but I don't believe that's supported for this kind of index.


garretlaxton (approved):
this looks good for me!

Crabcyborg:
Thanks Garret!

🚀

PR #1390, opened 2023-11-23
Reset 'max' option when switching paragraph field type to others

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/3768

When the 'Field Type' for a paragraph field is changed to other field types, it used to be limiting the number of characters allowed since it is assigned to the `maxlength` attribute. It is proposed to clear the 'max' value as a solution instead and just fixed that in this update.

### Screenshot
image


garretlaxton:
If you switch this back to a paragraph field the Rows option is blank, but still use the previous value.

AbdiTolesa:
> If you switch this back to a paragraph field the Rows option is blank, but still use the previous value.

@garretlaxton I have deployed and activated this on my site. It looks going back to paragraph removes the option and doesn't use the previous value. Here is the preview:
https://qa.formidableforms.com/wp-admin/admin-ajax.php?action=frm_forms_preview&form=issue3768
The form: https://qa.formidableforms.com/wp-admin/admin.php?page=formidable&frm_action=edit&id=286

Do you see any issues with that?

garretlaxton (approved):
Thanks Abdi, I think I was seeing old cache on my end. This looks great to me.

classes/models/FrmForm.php
```diff
@@ -377,6 +377,7 @@ public static function update_fields( $id, $values ) {
 			}
 
 			self::prepare_field_update_values( $field, $values, $new_field );
+			self::maybe_update_max_option( $new_field, $field, $values );
```
truongwp, line 380: I think we should change the order of these params to the same as the `prepare_field_update_values()` method above.

truongwp (approved):
This is good now. Thanks @AbdiTolesa!

Crabcyborg (approved):
Thanks @AbdiTolesa! This looks good.

Thank you @truongwp and @garretlaxton for reviewing this!

🚀

PR #1391, opened 2023-11-28
Add a filter that allows updating form ids for fields in Stripe action

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-stripe/issues/98


stripe/models/FrmTransLiteAction.php
```diff
@@ -103,13 +103,22 @@ private function default_currency() {
 	 * @return array
 	 */
 	public function get_field_options( $form_id ) {
+
+		/**
+		 * @since x.x
+		 *
+		 * @param mixed $form_id
+		 */
+		$form_ids = apply_filters( 'frm_get_field_options_form_id', absint( $form_id ) );
```
truongwp, line 112: I think we need the original form ID as the third param. Because after running [this](https://github.com/Strategy11/formidable-pro/pull/4596/files#diff-cb53198069014818d2202e56c5be8ba6bdd1ed222f31ecceadae456c60f2dec0R493), the result is an array of form IDs. If I add a function to this filter after that, I won't know what the original form ID is.

stripe/models/FrmTransLiteAction.php
```diff
@@ -103,13 +103,26 @@ private function default_currency() {
...
 	 */
 	public function get_field_options( $form_id ) {
+
+		$form_id  = absint( $form_id );
+		$form_ids = $form_id;
+
+		/**
+		 * @since x.x
+		 *
+		 * @param mixed $form_ids
```
truongwp, line 113: Instead of `mixed`, we can use `int|int[]` here for more detail.
Is `$form_id` always `int`?

truongwp (approved):

stripe/models/FrmTransLiteAction.php
```diff
@@ -103,13 +103,26 @@ private function default_currency() {
 	 * @return array
 	 */
 	public function get_field_options( $form_id ) {
+
+		$form_id  = absint( $form_id );
+		$form_ids = $form_id;
+
+		/**
+		 * @since x.x
```
Crabcyborg, line 111: A comment why this exists may be useful.

stripe/models/FrmTransLiteAction.php
```diff
@@ -103,13 +103,26 @@ private function default_currency() {
...
+		 *
+		 * @param int|int[] $form_ids
+		 * @param int $form_id
+		 */
+		$form_ids = apply_filters( 'frm_get_field_options_form_id', $form_ids, $form_id );
```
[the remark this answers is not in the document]
AbdiTolesa: Thanks for suggesting a better name! I have updating this and reflected the change in pro as well (https://github.com/Strategy11/formidable-pro/pull/4596/files#diff-cb53198069014818d2202e56c5be8ba6bdd1ed222f31ecceadae456c60f2dec0R493)

Crabcyborg:
Is it possible to keep the JS logic in Lite the same and add the new query logic in the Pro `frmstrp.js` file instead?

In that file, I add a couple of functions to the `window.frmProForm` global object.

We could add something like `window.frmProForm.queryForNameField` and call that if it exists, or fallback to the Lite logic when it isn't callable.

AbdiTolesa:
> Is it possible to keep the JS logic in Lite the same and add the new query logic in the Pro `frmstrp.js` file instead?
> 
> In that file, I add a couple of functions to the `window.frmProForm` global object.
> 
> We could add something like `window.frmProForm.queryForNameField` and call that if it exists, or fallback to the Lite logic when it isn't callable.

@Crabcyborg I have added the function to pro and then tried to call it if available from the lite and stripe scripts or fallback to the old query otherwise.

stripe/js/frmstrp.js
```diff
@@ -229,7 +230,9 @@
 			settings = frm_stripe_vars.settings,
 			firstNameID = '',
 			lastNameID = '',
-			getNameFieldValue;
+			getNameFieldValue,
+			queryForNameFieldIsFound,
```
Crabcyborg, line 234: Is there a reason why this variable is in this scope?

It looks like it's only used in `getNameFieldItem`. It would make sense to declare it there instead to reduce the scope of it.

js/admin/dom.js
```diff
@@ -516,6 +516,7 @@
 				if ( height ) {
 					settings.height = height;
 				}
+				console.log( settings );
```
Crabcyborg, line 519: There's a console log here.

Crabcyborg (approved):
This looks good to go.

🚀

PR #1396, opened 2023-11-30
Hide journey if it is collected by User Flow add-on

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4597


Crabcyborg (approved):
Thanks @AbdiTolesa!

This fix seems good.

PR #1399, opened 2023-11-30
Add left styling rule when auto fold class is not set on body

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4464

To replicate Kristine's issue, you can use this snippet:
```
global $_updated_user_settings;
$_updated_user_settings['unfold'] = true;
```

Basically she has the "unfold" user setting on. I'm not totally sure how this gets set though.

This update adds a new `frm-unfold` class when this `unfold` setting is on (so `auto-fold` is never added) and that the `mfold` user preference is not set to `f` which adds the `folded`.

Maybe this is complicating it though. May we just want to not check for the `.auto-fold` class and always set `left: 160px;` by default?

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -76,6 +76,10 @@ class FrmAppController {
 			$classes .= ' frm-lite ';
 		}
 
+		if ( get_user_setting( 'unfold' ) && 'f' !== get_user_setting( 'mfold' ) ) {
+			$classes .= ' frm-unfold ';
+		}
+
 		return $classes;
 	}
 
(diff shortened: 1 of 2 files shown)


stephywells (approved):
I couldn't find how to add the setting either. I also had to switch my post settings away from full screen mode, and then add this unfold code you included. 

I think this solution looks good though, despite the weirdness of it. This was originally a report from Syed and Kristine was the only one who could see it.

garretlaxton (approved):
I couldn't replicate this, but I don't have any issues with this fix.

Crabcyborg:
Thanks Steph and Garret!

🚀

PR #1401, opened 2023-12-01
Avoid deprecation warnings when calling template functions

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4478

We can know whether calling functions `get_header` and `get_footer` would cause deprecation warnings ahead of time and avoid it if so. We can instead copy pieces of code from core to call relevant functions and output html fragments required.

### Testing
Preview a form in theme and check the error log for deprecation warnings.

> PHP Deprecated: File Theme without header.php is deprecated since version 3.0.0 with no alternative available. Please include a header.php template in your theme. in /var/www/src/wp-includes/functions.php on line 5646
> PHP Deprecated: str_replace(): Passing null to parameter https://github.com/Strategy11/formidable-pro/issues/3 ($subject) of type array|string is deprecated in /var/www/src/wp-includes/post-template.php on line 257
> PHP Deprecated: File Theme without footer.php is deprecated since version 3.0.0 with no alternative available. Please include a footer.php template in your theme. in /var/www/src/wp-includes/functions.php on line 5646

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -467,26 +467,7 @@ class FrmFormsController {
 			<link rel="stylesheet" href="<?php bloginfo( 'stylesheet_url' ); ?>" type="text/css" media="screen" />
 			<link rel="pingback" href="<?php bloginfo( 'pingback_url' ); ?>" />
 
-			<?php if ( file_exists( get_stylesheet_directory() . '/images/kubrickbgwide.jpg' ) ) { ?>
-			<style type="text/css" media="screen">
-
-				<?php
-				// Checks to see whether it needs a sidebar.
-				/** @psalm-suppress UndefinedVariable */
-				if ( empty( $withcomments ) && ! is_single() ) { // phpstan-ignore-line
-					?>
-				#page { background: url("<?php bloginfo( 'stylesheet_directory' ); ?>/images/kubrickbg-<?php bloginfo( 'text_direction' ); ?>.jpg") repeat-y top; border: none; }
-			<?php } else { // No sidebar. ?>
-				#page { background: url("<?php bloginfo( 'stylesheet_directory' ); ?>/images/kubrickbgwide.jpg") repeat-y top; border: none; }
-			<?php } ?>
-
-			</style>
-			<?php } ?>
-
 			<?php
-			if ( is_singular() ) {
-				wp_enqueue_script( 'comment-reply' );
-			}
 			wp_head();
 			?>
 			</head>
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -460,7 +460,7 @@ class FrmFormsController {
 			<div id="page">
 				<div id="header" role="banner">
 					<div id="headerimg">
-						<h1><a href="<?php echo home_url(); ?>/"><?php bloginfo( 'name' ); ?></a></h1>
+						<h1><a href="<?php echo esc_url( home_url() ); ?>/"><?php bloginfo( 'name' ); ?></a></h1>
 						<div class="description"><?php bloginfo( 'description' ); ?></div>
 					</div>
 				</div>
@@ -478,8 +478,8 @@ class FrmFormsController {
 						<?php
 						printf(
 							/* translators: 1: Site name, 2: WordPress */
-							__( '%1$s is proudly powered by %2$s' ),
-							get_bloginfo( 'name' ),
+							esc_html__( '%1$s is proudly powered by %2$s', 'formidable' ),
+							esc_html( get_bloginfo( 'name' ) ),
 							'<a href="https://wordpress.org/">WordPress</a>'
 						);
 						?>
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -472,7 +472,8 @@ class FrmFormsController {
 
 				<?php
 				// Checks to see whether it needs a sidebar.
-				if ( empty( $withcomments ) && ! is_single() ) {
+				/** @psalm-suppress UndefinedVariable */
+				if ( empty( $withcomments ) && ! is_single() ) { // phpstan-ignore-line
 					?>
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
This fix works great!

classes/controllers/FrmFormsController.php
```diff
@@ -425,15 +425,88 @@ public static function preview_block_theme_body_classnames( $classes ) {
...
+	}
+
+	/**
+	 * Returns true if calling a template function doesn't trigger deprecation warnings.
+	 *
+	 * @since x.x
+	 * @param string $template
+	 * @return bool
+	 */
+	private static function should_try_getting_template( $template ) {
```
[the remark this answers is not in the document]
truongwp: Oh sorry about that. It's https://developer.wordpress.org/reference/functions/get_theme_file_path/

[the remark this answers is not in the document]
AbdiTolesa: @truongwp That function returns string instead of boolean. It requires some extra processing to make it useable as this function. I think this is better as is.

truongwp (approved):
Thanks @AbdiTolesa. This looks good!

classes/controllers/FrmFormsController.php
```diff
@@ -425,15 +425,88 @@ public static function preview_block_theme_body_classnames( $classes ) {
...
+		$stylesheet_path = get_stylesheet_directory();
+		$template_path   = get_template_directory();
+		$is_child_theme  = $stylesheet_path !== $template_path;
+		$template_name   = $template . '.php';
+
+		return file_exists( $stylesheet_path . '/' . $template_name ) || $is_child_theme && file_exists( $template_path . '/' . $template_name );
```
Crabcyborg, line 509: Can we add braces around the second check?

Right now the way this works is that it evaluates `file_exists( $stylesheet_path . '/' . $template_name ) || $is_child_theme` and then checks that against `file_exists( $template_path . '/' . $template_name )`.

So if `file_exists( $stylesheet_path . '/' . $template_name )` is true but `$is_child_theme` is false, it will still check if `file_exists( $template_path . '/' . $template_name )` because it's checking` if ( true && file_exists( $template_path . '/' . $template_name ) )`.

classes/controllers/FrmFormsController.php
```diff
@@ -425,17 +427,90 @@ public static function preview_block_theme_body_classnames( $classes ) {
...
+	private static function get_template( $template ) {
+		if ( self::should_try_getting_template( $template ) ) {
+			call_user_func( 'get_' . $template );
+			return;
+		}
+
+		if ( 'header' === $template ) {
+			wp_head();
+			?>
+			</head>
```
Crabcyborg, line 460: It doesn't look like `wp_head` adds a `<head>` tag here.

I checked my page source and it looks like there's a `</head>` close tag but no `<head>` tag opening it.

AbdiTolesa: @Crabcyborg I inspected my page and noticed the `<head>` tag is also there as in this screenshot:
image

Crabcyborg: @AbdiTolesa That's just because your browser tries to correct it. If you check the actual HTML output, it isn't there.

Try right clicking and going to "View Page Source",
Screen Shot 2023-12-29 at 10 00 15 AM

classes/controllers/FrmFormsController.php
```diff
@@ -455,6 +455,37 @@ private static function get_template( $template ) {
...
+			<head>
+			<link rel="profile" href="https://gmpg.org/xfn/11" />
+			<meta http-equiv="Content-Type" content="<?php bloginfo( 'html_type' ); ?>; charset=<?php bloginfo( 'charset' ); ?>" />
+
+			<title><?php echo wp_get_document_title(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></title>
+
+			<link rel="stylesheet" href="<?php bloginfo( 'stylesheet_url' ); ?>" type="text/css" media="screen" />
+			<link rel="pingback" href="<?php bloginfo( 'pingback_url' ); ?>" />
+
+			<?php if ( file_exists( get_stylesheet_directory() . '/images/kubrickbgwide.jpg' ) ) { ?>
```
Crabcyborg, line 470: We can probably simplify this. I don't think we need this part.

classes/controllers/FrmFormsController.php
```diff
@@ -455,6 +455,37 @@ private static function get_template( $template ) {
...
+			<?php } else { // No sidebar. ?>
+				#page { background: url("<?php bloginfo( 'stylesheet_directory' ); ?>/images/kubrickbgwide.jpg") repeat-y top; border: none; }
+			<?php } ?>
+
+			</style>
+			<?php } ?>
+
+			<?php
+			if ( is_singular() ) {
+				wp_enqueue_script( 'comment-reply' );
```
Crabcyborg, line 487: I think we should leave this part off too. We don't want any comments on our page.

classes/controllers/FrmFormsController.php
```diff
@@ -425,17 +427,103 @@ public static function preview_block_theme_body_classnames( $classes ) {
...
+
+			<link rel="stylesheet" href="<?php bloginfo( 'stylesheet_url' ); ?>" type="text/css" media="screen" />
+			<link rel="pingback" href="<?php bloginfo( 'pingback_url' ); ?>" />
+
+			<?php
+			wp_head();
+			?>
+			</head>
+			<body <?php body_class(); ?>>
+			<div id="page">
```
Crabcyborg, line 475: @AbdiTolesa Since this is so much HTML, it would be nice to move this into new header/footer view files.

Crabcyborg: It would also be useful to have a comment that this code is mostly copied/pasted from the WordPress `theme-compat.php` file.

Crabcyborg (approved):
Thanks @AbdiTolesa! I think this is good to go now.

Thank you @truongwp and @garretlaxton for your help with review!

🚀

PR #1409, opened 2023-12-08
Change the HTML order of previous buttons, add flex styling to submit button container and keep it left using order: -1

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3790

**New forms only** will now have new submit button HTML with flex styling.

This issue isn't fixed for old forms. You will need to clear your old submit button HTML and save to reset it to the default to fix it in an old form.

_The gap between buttons is larger now, but that's intentional._

@stephywells I wouldn't normally request your review on this one but now this requires 2 approvals to merge?


garretlaxton (approved):
This looks great!

PR #1413, opened 2023-12-11
Pro issue 3967

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/3967

Here's a basic snippet to simulate something similar to the issue from this ticket https://github.com/Strategy11/formidable-pro/issues/3967#issuecomment-1850615714

When any `$_POST` data includes the word "select", this will force a redirect to a 404 page.

```
foreach ( $_POST as $key => $value ) {
	if ( false !== strpos( $value, 'select' ) ) {
		wp_redirect( site_url() . '/404-page' );
		die();
	}
}
```

I tested this fix out as a beta on the customer's website, and it fixes the issue with their security plugin blocking saving a dropdown "select" field.

@stephywells What do you think? Is there anything else I'd need to do here to cover this?

### DIFF
--- a/classes/models/FrmField.php
+++ b/classes/models/FrmField.php
@@ -444,6 +444,8 @@ class FrmField {
 
 		if ( isset( $values['type'] ) ) {
 			if ( 'dropdown' === $values['type'] ) {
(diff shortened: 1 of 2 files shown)


stephywells (approved):
Nice solution!

PR #1415, opened 2023-12-11
Update Phpstan

stephywells:
Load phpstan in composer
Use Pro and Views functions directory instead of stubs since phpstan thinks they are always defined. This might need more work for the phpstan workflow in GH.
Fix and stop excluding PHPDoc tag @param rule
Fix and stop excluding is not nullable for static values
Exclude less commenting phpcs and specify which rules to exclude for now.
Fix the comment rules that can be fixed automatically.


classes/controllers/FrmAddonsController.php
```diff
@@ -1246,13 +1246,15 @@ public static function install_addon_api() {
...
 	 * @since 4.09
+	 *
+	 * @param string       $plugin
+	 * @param array|string $upgrade_link_args
+	 * @return void
 	 */
 	public static function conditional_action_button( $plugin, $upgrade_link_args ) {
 		if ( is_callable( 'FrmProAddonsController::conditional_action_button' ) ) {
-			return FrmProAddonsController::conditional_action_button( $plugin, $upgrade_link_args );
+			FrmProAddonsController::conditional_action_button( $plugin, $upgrade_link_args );
```
Crabcyborg, line 1257: @stephywells Getting rid of the return here would mean a bunch of other code runs after this line.

I think we still need a `return`, just on its own line.

classes/controllers/FrmAddonsController.php
```diff
@@ -1275,14 +1277,19 @@ public static function conditional_action_button( $plugin, $upgrade_link_args )
...
+	 *     @type string|false $license_type
+	 *     @type string $plan_required
+	 *     @type string $upgrade_link
+	 * }
+	 * @return void
 	 */
 	public static function show_conditional_action_button( $atts ) {
 		if ( is_callable( 'FrmProAddonsController::show_conditional_action_button' ) ) {
-			return FrmProAddonsController::show_conditional_action_button( $atts );
+			FrmProAddonsController::show_conditional_action_button( $atts );
```
Crabcyborg, line 1292: Same here. We should call the function, then return.

classes/controllers/FrmXMLController.php
```diff
@@ -363,7 +363,7 @@ public static function import_xml() {
 			return;
 		}
 
-		//add_filter('upload_mimes', 'FrmXMLController::allow_mime');
+		// add_filter('upload_mimes', 'FrmXMLController::allow_mime');
```
Crabcyborg, line 366: Should we get rid of this line? We shouldn't have commented out code.

tests/forms/test_FrmFormsControllerAjax.php
```diff
@@ -47,7 +47,7 @@ private function _setup_post_values( $form_id ) {
 			'id' => $form_id,
 			'action' => 'update',
 			'frm_save_form' => wp_create_nonce( 'frm_save_form_nonce' ),
-			//'_wp_http_referer' =>
+			// '_wp_http_referer' =>
```
Crabcyborg, line 50: Should we just remove this line?

classes/controllers/FrmAddonsController.php
stephywells: Good catch! I just reviewed every file again to make sure I didn't miss anything else like this.

classes/controllers/FrmXMLController.php
stephywells: Done.

Crabcyborg (approved):
Thanks @stephywells!

🚀

PR #1418, opened 2023-12-12
Form builder permission settings scripts

omidakhavans:
### Summary
This update would remove the unnecessary script for permission settings tab based on new updates.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -10278,7 +10278,7 @@ function frmAdminBuildJS() {
 		settingsInit: function() {
 			const $formActions = jQuery( document.getElementById( 'frm_notification_settings' ) );
 
-			let formSettings, $loggedIn, $cookieExp, $editable;
+			let formSettings, $loggedIn, singleEntryType, $editable;
 
 			// BCC, CC, and Reply To button functionality
 			$formActions.on( 'click', '.frm_email_buttons', showEmailRow );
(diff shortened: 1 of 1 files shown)


truongwp:
Does this cause any problem if Pro version hasn't been updated?

omidakhavans:
@truongwp Some fields wouldn't show hide properly. Do you have some suggestion?

truongwp:
I think we should check to remove these scripts only when those settings are not on the Settings page.

omidakhavans:
@truongwp Since its effects on only a dropdown and others would work fine. I guess it's fine. if we want to make a condition and keep the old codes the point about cleaning would be lost. Can you please test it to see if we are fine or not? Also maybe @Crabcyborg could have some points about this?

js/formidable_admin.js
```diff
@@ -10453,6 +10396,49 @@ function frmAdminBuildJS() {
...
+					return;
+				}
+
+				Array.prototype.forEach.call( singleEntryType, ( selector ) => {
+					selector.classList.remove( 'frm_hidden' );
+				});
+			});
+
+			// jQuery event trigger for backward compability.
+			triggerChange( jQuery( document.getElementById( 'save_draft' ) ) );
```
Crabcyborg, line 10434: This line is causing issues. As soon as I load the settings page I get prompted with the unsaved changes pop up because this triggers change on load.

Crabcyborg:
@omidakhavans Maybe we can pull the old logic into a temporary backward-compatibility function? It doesn't seem ideal that this update breaks Pro when it isn't up to date.

We generally avoid breaking changes like that when possible.

Is it much more work to make these listeners work with the old HTML as well?

Crabcyborg:
Also, @omidakhavans, is it possible to add the new logic in Pro instead? Since these settings are for Pro features it would be nice to leave future changes out of Lite where possible.

We have a `settings.js` now in Pro. It used to be that we didn't have a dedicated JS file yet in Pro when these were added.

Crabcyborg:
I'm closing this as it is no longer required.

PR #1422, opened 2023-12-15
Add new JS filter for getFieldList()

truongwp:
This is required for a Surveys fix.

I also fixed some variables aren't declared before using.


Crabcyborg (approved):
Thanks @truongwp!

This looks good to me.

🚀

PR #1425, opened 2023-12-20
Stop requiring CC field if stripe intent is in request

Crabcyborg:
This fixes the issue with required CC fields and Stripe Lite actions.

I hadn't copied over a similar function from the Stripe add on, but clearly it's important.

This is a simplified version that covers the smaller Stripe Lite scope.


jonathanenlared (approved):
It works great 👍

PR #1428, opened 2023-12-21
Add a sent mothership request flag to FrmAddon to limit requests to o…

Crabcyborg:
…ur API

Related Slack conversation https://strategy11.slack.com/archives/G01KDLFJTJS/p1703078624583319?thread_ts=1702904869.778139&cid=G01KDLFJTJS

Now instead of all ~20 requests happening at once, it makes a single request and spreads the 20 requests across 20 HTTP requests.

It took me about 12 seconds to load a page with the 20 requests as well, so this update helps take a lot of the load off a single user's request that may make their site seem really slow.

I left some of the requests alone. This is just when calling `is_license_revoked` which can get called a lot in a single request because it's part of the `FrmAddon` constructor.

### DIFF
--- a/classes/models/FrmAddon.php
+++ b/classes/models/FrmAddon.php
@@ -20,6 +20,16 @@ class FrmAddon {
 	public $license;
 	protected $get_beta = false;
 
+	/**
+	 * This is used to flag other add ons not to send a request.
+	 * We only want to send a single API request per HTTP request.
+	 *
+	 * @since x.x
+	 *
+	 * @var bool
+	 */
+	private static $sent_mothership_request = false;
+
 	public function __construct() {
 
 		if ( empty( $this->plugin_slug ) ) {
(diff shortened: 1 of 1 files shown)


classes/models/FrmAddon.php
```diff
@@ -472,6 +486,7 @@ private function is_license_revoked() {
 			return;
 		}
 
+		self::$sent_mothership_request = true;
```
stephywells, line 489: Do you think we could set this as an array and add $this->license to the array? I want to be sure we don't check the license for an add-on and never check it for the main plugin.

Crabcyborg: @stephywells I don't think we need to worry about an addon happening first.

Immediately when Pro is loaded it calls `FrmProAppHelper::get_updater();` which triggers the `FrmProEddController` constructor right away, so the first mothership request is always going to be from Pro.

Screen Shot 2023-12-28 at 11 23 10 AM

stephywells: That sounds great then. I don't think that would be an issue.

stephywells (approved):

PR #1431, opened 2023-12-27
Implement Deactivate and Uninstall Buttons for Add-Ons on Formidable Add-Ons Page

shervElmi:
This PR addresses the feature request in [Issue #4349](https://github.com/Strategy11/formidable-pro/issues/4349), which involves enhancing the user interface of the Formidable Add-Ons page (`/wp-admin/admin.php?page=formidable-addons`). The update adds "Deactivate" and "Uninstall" buttons for add-ons, providing clearer options for managing their states.

## Related Issue:
[Add Deactivate and Uninstall buttons to the Formidable Add-Ons page #4349](https://github.com/Strategy11/formidable-pro/issues/4349)

## Testing Instructions:
1. Navigate to `WP Admin > Formidable > Add-Ons`.
2. For each add-on, verify the presence of the appropriate button(s) based on its current state:
   - "Deactivate" when the add-on is active.
   - "Uninstall" when the add-on is deactivated.
3. Test the functionality of the new buttons to ensure they correctly deactivate and uninstall the add-ons.

## Formidable PRO Plugin PR:
https://github.com/Strategy11/formidable-pro/pull/4676


shervElmi:
Hey @jonathanenlared!

I wasn't able to create a PR URL for my PR, and I need to change my branch name. This change will close this PR, and I'll have to create a new one. Please try to test this on your side, but if you need a QA URL, let me know.

**Note:** Please note that this PR is related to another PR in the pro plugin.
https://github.com/Strategy11/formidable-pro/pull/4676

garretlaxton (approved):
This worked great for me!

classes/controllers/FrmAddonsController.php
```diff
@@ -934,6 +934,52 @@ protected static function activate_plugin( $plugin, $redirect = '', $network_wid
...
+	 *
+	 * @param string $plugin
+	 * @param bool   $silent
+	 * @return null|WP_Error Null on success, WP_Error on invalid file.
+	 */
+	protected static function deactivate_plugin( $plugin, $silent = false ) {
+		if ( ! function_exists( 'deactivate_plugins' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/plugin.php';
+		}
+		return deactivate_plugins( $plugin, $silent );
```
Crabcyborg, line 950: It looks like `deactivate_plugins` never returns anything. It's a `void` return function so the return result would always be `null`.

classes/controllers/FrmAddonsController.php
```diff
@@ -1055,23 +1101,129 @@ protected static function install_addon() {
...
+			case 'uninstall':
+				$result = self::uninstall_plugin( $installed );
+				break;
+		}
+
+		if ( is_wp_error( $result ) ) {
+			// Ignore the invalid header message that shows with nested plugins.
+			if ( $result->get_error_code() !== 'no_plugin_header' ) {
+				if ( wp_doing_ajax() ) {
+					echo json_encode( array( 'error' => $result->get_error_message() ) );
```
Crabcyborg, line 1214: Can we use `wp_send_json_error` here instead so we're doing this the WordPress way?

classes/controllers/FrmAddonsController.php
```diff
@@ -1055,23 +1101,129 @@ protected static function install_addon() {
...
+			return;
+		}
+
+		$result = null;
+		switch ( $action ) {
+			case 'activate':
+				$result = self::activate_plugin( $installed );
+				break;
+			case 'deactivate':
+				$result = self::deactivate_plugin( $installed );
```
Crabcyborg, line 1203: You would also want to remove `$result =` here.

js/formidable_admin.js
```diff
@@ -8607,7 +8606,16 @@ function frmAdminBuildJS() {
 		jQuery( '#frm_upgrade_modal .frm_lock_icon use' ).attr( 'xlink:href', '#frm_lock_open_icon' );
 
 		// Proceed with CSS changes
-		el.parent().removeClass( 'frm-addon-not-installed frm-addon-installed' ).addClass( 'frm-addon-active' );
+		const frmAdminJs = frm_admin_js; // eslint-disable-line camelcase
+		const actionMap = {
+			'frm_activate_addon': { class: 'frm-addon-active', message: frmAdminJs.active },
```
Crabcyborg, line 8611: We don't need the quotes around these keys here.

js/formidable_admin.js
```diff
@@ -8607,7 +8606,16 @@ function frmAdminBuildJS() {
...
+			'frm_deactivate_addon': { class: 'frm-addon-installed', message: frmAdminJs.installed },
+			'frm_uninstall_addon': { class: 'frm-addon-not-installed', message: frmAdminJs.not_installed }
+		};
+		actionMap.frm_install_addon = actionMap.frm_activate_addon;
+
+		message.text( actionMap[action].message );
+		el.parent().removeClass( 'frm-addon-not-installed frm-addon-installed frm-addon-active' ).addClass( actionMap[action].class );
```
Crabcyborg, line 8618: Could we do this with less jQuery? It would be nice to remove some jQuery with this update.

js/formidable_admin.js
```diff
@@ -8607,7 +8606,16 @@ function frmAdminBuildJS() {
 		jQuery( '#frm_upgrade_modal .frm_lock_icon use' ).attr( 'xlink:href', '#frm_lock_open_icon' );
 
 		// Proceed with CSS changes
-		el.parent().removeClass( 'frm-addon-not-installed frm-addon-installed' ).addClass( 'frm-addon-active' );
+		const frmAdminJs = frm_admin_js; // eslint-disable-line camelcase
```
shervElmi, line 8609: Thank you for your review, @Crabcyborg! What do you think about relocating this line to the top of the file? This could help us avoid the necessity of using `// eslint-disable-line camelcase` in future implementations.

Crabcyborg: Sure @shervElmi, that sounds like a good workaround.

I would try to leave it out of the global scope at least though.

If we put it inside of the top of `frmAdminBuildJS`, nearby where the `frmDom` functions are delcared, it should be usable in most places as least.

classes/controllers/FrmAddonsController.php
```diff
@@ -934,6 +934,52 @@ protected static function activate_plugin( $plugin, $redirect = '', $network_wid
...
+		if ( ! current_user_can( 'delete_plugins' ) ) {
+			return new WP_Error( 'uninstall_failed', __( 'Current user cannot delete plugins.', 'formidable' ) );
+		}
+
+		if ( ! function_exists( 'delete_plugins' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/plugin.php';
+		}
+
+		self::deactivate_plugin( $plugin, true );
+		// Deactivate plugin before uninstalling.
```
Crabcyborg, line 971: This comment is a bit weird. It's above the `delete_plugins` call, but it's describing the call above. The function is also really just the name of the function, so I think we're better off just removing it.

classes/controllers/FrmAddonsController.php
```diff
@@ -934,6 +934,52 @@ protected static function activate_plugin( $plugin, $redirect = '', $network_wid
...
+		$result = delete_plugins( array( $plugin ) );
+
+		if ( is_wp_error( $result ) ) {
+			return $result;
+			// Return the error from delete_plugins.
```
Crabcyborg, line 976: This comment is in an odd place as well. Comments should be above stuff, not under.

classes/controllers/FrmAddonsController.php
```diff
@@ -934,6 +934,52 @@ protected static function activate_plugin( $plugin, $redirect = '', $network_wid
...
+		deactivate_plugins( $plugin, $silent );
+	}
+
+	/**
+	 * Uninstall a specified plugin.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $plugin
+	 * @return null|WP_Error Null on success, WP_Error on invalid file.
```
Crabcyborg, line 959: `null` on success seems like a weird return type.

Maybe we could return `true` instead?

classes/controllers/FrmAddonsController.php
```diff
@@ -934,6 +934,52 @@ protected static function activate_plugin( $plugin, $redirect = '', $network_wid
...
+		// Deactivate plugin before uninstalling.
+		$result = delete_plugins( array( $plugin ) );
+
+		if ( is_wp_error( $result ) ) {
+			return $result;
+			// Return the error from delete_plugins.
+		}
+
+		return null;
+		// Success.
```
Crabcyborg, line 980: This comment seems unnecessary as well.

classes/controllers/FrmAddonsController.php
```diff
@@ -925,7 +925,7 @@ public static function connect_pro() {
 	 * @param string $redirect
 	 * @param bool   $network_wide
 	 * @param bool   $silent
-	 * @return null|WP_Error Null on success, WP_Error on invalid file.
+	 * @return true|WP_Error True on success, WP_Error on invalid file.
```
Crabcyborg, line 928: It looks like the WordPress `active_plugin` function actually does return `@return null|WP_Error Null on success` so we should leave this one.

classes/controllers/FrmAddonsController.php
```diff
@@ -1055,23 +1098,128 @@ protected static function install_addon() {
...
 		FrmAppHelper::set_current_screen_and_hook_suffix();
 
 		$plugin = FrmAppHelper::get_param( 'plugin', '', 'post', 'sanitize_text_field' );
-		self::maybe_activate_addon( $plugin );
+		call_user_func( $action_callback, $plugin );
+
+		if ( is_callable( $response_callback ) ) {
+			echo json_encode( call_user_func( $response_callback ) );
+		} else {
+			echo json_encode( array( 'success' => true ) );
```
Crabcyborg, line 1174: Can we use `wp_send_json_success()` here?

classes/controllers/FrmAddonsController.php
shervElmi: Could you please check this link: https://developer.wordpress.org/reference/functions/activate_plugins/
It seems it returns true.

image

classes/controllers/FrmAddonsController.php
shervElmi: Done!

classes/controllers/FrmAddonsController.php
Crabcyborg: That's another function though. This is activate_plugin getting called here. Not activate_plugins.

I didn't see documentation for activate_plugin but I looked at the source code and pulled they Null return comment from there.

Crabcyborg:
@shervElmi It looks like there is a failing unit test.

Screen Shot 2024-01-08 at 1 10 34 PM

shervElmi:
Hey @Crabcyborg,

I've just commented out a line of code that was causing the PR to fail, and it appears to be unrelated to the changes in this PR. Could you please confirm this? Once you've confirmed, I'll revert my last commit, and we can create a new issue to address this separately.

Crabcyborg:
Thanks @shervElmi.

@truongwp Do you have any idea why the unit test would break when deleting the email summary helper option? https://github.com/Strategy11/formidable-forms/pull/1431/commits/b06574088bc4227b46f4e32739279e6c6c2a32c9

truongwp:
@Crabcyborg I don't know why but I created a new PR and uncommented that line, there is no failing tests: https://github.com/Strategy11/formidable-forms/pull/1452

shervElmi:
Thank you @truongwp and @Crabcyborg, for reviewing this issue. I've merged the master PR into the current branch, and now all checks have passed.

js/formidable_admin.js
```diff
@@ -8616,7 +8616,24 @@ function frmAdminBuildJS() {
...
+		const messageElement = message[0];
+		messageElement.textContent = actionMap[action].message;
+
+		const buttonElement = button[0];
+		const parentElement = buttonElement.parentElement;
+		parentElement.classList.remove( 'frm-addon-not-installed', 'frm-addon-installed', 'frm-addon-active' );
+		parentElement.classList.add( actionMap[action].class );
+		buttonElement.classList.remove( 'frm_loading_button' );
+
+		message.text( actionMap[action].message );
```
Crabcyborg, line 8635: You're already updating this with `messageElement.textContent = actionMap[action].message;` a couple lines up.

I don't think we need this line.

js/formidable_admin.js
```diff
@@ -8616,7 +8616,24 @@ function frmAdminBuildJS() {
...
+		el.parent().removeClass( 'frm-addon-not-installed frm-addon-installed frm-addon-active' ).addClass( actionMap[action].class );
```
Crabcyborg, line 8636: Can we do this part without jQuery?

css/frm_admin.css
```diff
@@ -2404,8 +2409,20 @@ h2.frm-h2 + .howto {
...
+.frm-addon-installed .frm-deactivate-addon,
+.frm-addon-active  .frm-activate-addon,
+.frm-addon-active  .frm-install-addon,
+.frm-addon-active .frm-uninstall-addon,
+.plugin-card-pro.frm-addon-active .button {
+	display: none !important;
+}
+
+#frm-addons-page .frm-addon-button {
+	opacity: 1 !important;
```
Crabcyborg, line 2425: What style is getting in the way here?

It looks like there is a `frm-opacity-100` class now. Does that work? Does it need to be important?

shervElmi: We are using these buttons on other pages, so I've decided not to change the JavaScript (which sets this button's opacity to 0) and HTML, and instead, I used CSS, which is a more consistent approach for me. If you think I should do it differently, let me know. Yes, the `important` here is mandatory.

Crabcyborg:
Thanks for those updates @shervElmi!

@garretlaxton This has been updated a bit since you last tested it.

There's one change where a lot of JavaScript string variables were updated, pointing to a reference instead of the actual variable for code-styling purposes. It would probably be good to test for that as well - basically making sure that all of the strings used on admin pages are working as expected.

Thank you!

garretlaxton (approved):
I don't have any issues with this!

Crabcyborg (approved):
Thanks @shervElmi! I think this is good to go now.

Thank you @truongwp and @garretlaxton for your help!

🚀

PR #1436, opened 2024-01-02
Avoid a db error when getting payments stats

Crabcyborg:
If you haven't connected Stripe Connect, and you're not using PayPal, the table may not exist at all.

I missed that we weren't checking that during review.

Related ticket https://wordpress.org/support/topic/database-errors-from-scheduled-frmemailsummaryhelper/


truongwp (approved):

PR #1442, opened 2024-01-04
Support fields shortcode modal in html field

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4672

image


garretlaxton:
My popup is appearing off the page for me. I cleared my browser cache, but that didn't help.
image

AbdiTolesa:
> My popup is appearing off the page for me. I cleared my browser cache, but that didn't help. image

I have deployed this to the QA (https://qa.formidableforms.com/abdi/wp-admin/admin.php?page=formidable&frm_action=edit&id=79), do you see any issues with it yet?

garretlaxton:
@AbdiTolesa Yes, I see this on the QA site as well. I'm wondering if this is a browser issue? I'm using Firefox.

AbdiTolesa:
> @AbdiTolesa Yes, I see this on the QA site as well. I'm wondering if this is a browser issue? I'm using Firefox.

@garretlaxton I noticed a different style issue in firefox and fixed it. But I couldn't see the issue you are facing even with Firefox browser. Does my latest commit probably fix it for you? Here is my Firefox screenshot:
image

garretlaxton (approved):
Yes! This looks great on Firefox now.

Crabcyborg:
It looks like this branch is breaking drag-and-drop.

When I start dragging my field, it gets stuck on my cursor. I can't drop it. It just stays in drag mode.

ds6zYl1jG9

AbdiTolesa:
> It looks like this branch is breaking drag-and-drop.
> 
> When I start dragging my field, it gets stuck on my cursor. I can't drop it. It just stays in drag mode while my mouse is up.

@Crabcyborg I don't see this issue on this branch, here is a screen record:

https://github.com/Strategy11/formidable-forms/assets/41271840/fd90a990-c0f2-46c7-bdaa-d05a6c1fc2bd

Do you see issues with drag and dropping fields with this branch checked out @garretlaxton?

js/formidable_admin.js
```diff
@@ -9691,6 +9753,8 @@ function frmAdminBuildJS() {
 		},
 
 		buildInit: function() {
+			jQuery( '#frm_builder_page' ).on( 'mouseup', '*:not(.frm-show-box)', maybeHideShortcodes );
```
Crabcyborg, line 9756: It's this line. Can we do this without listening for `mouseup`, or at least making sure we aren't targeting anything that is draggable.

garretlaxton:
@AbdiTolesa Yes, I see this on Firefox.

js/formidable_admin.js
AbdiTolesa: @Crabcyborg It looks preventing `e.stopPropagation()` on builder page fixes this issue.

AbdiTolesa:
> @AbdiTolesa Yes, I see this on Firefox.

Does the latest commit fix it for you @garretlaxton?

garretlaxton:
Yes, this works great for me now!

js/formidable_admin.js
```diff
@@ -9511,6 +9550,31 @@ function frmAdminBuildJS() {
...
+		if ( e.target.classList.contains( 'frm-show-box' )  || e.target.parentElement.classList.contains( 'frm-show-box' ) ) {
+			return;
+		}
+
+		const sidebar = document.getElementById( 'frm_adv_info' );
+		if ( ! sidebar ) {
+			return;
+		}
+
+		if ( sidebar.getAttribute( 'data-fills' ) === e.target.id && typeof e.target.id !== 'undefined' ) {
```
Crabcyborg, line 9567: Does `sidebar.dataset.fills` work here instead? Using `getAttribute` for `data` attributes feels kind of messy.

js/formidable_admin.js
```diff
@@ -9511,6 +9550,31 @@ function frmAdminBuildJS() {
 		tables.forEach( table => table.classList.add( 'frm-zebra-striping' ) );
 	};
 
+	function maybeHideShortcodes( e ) {
+		if ( ! document.getElementById( 'frm_builder_page' ) ) {
+			e.stopPropagation();
+		}
+
+		if ( e.target.classList.contains( 'frm-show-box' )  || e.target.parentElement.classList.contains( 'frm-show-box' ) ) {
```
Crabcyborg, line 9558: It looks like there are too many spaces here between `( 'frm-show-box' )` and the `||`.

Crabcyborg: It may also be good to confirm `e.target` actually has a `parentElement`. I'm not sure if there's a real risk of that, but it may be good just in case.

js/formidable_admin.js
```diff
@@ -9511,6 +9550,31 @@ function frmAdminBuildJS() {
...
+		const sidebar = document.getElementById( 'frm_adv_info' );
+		if ( ! sidebar ) {
+			return;
+		}
+
+		if ( sidebar.getAttribute( 'data-fills' ) === e.target.id && typeof e.target.id !== 'undefined' ) {
+			return;
+		}
+
+		const isChild = jQuery( e.target ).closest( '#frm_adv_info' ).length > 0;
```
Crabcyborg, line 9571: Can we do this line without jQuery?

js/formidable_admin.js
```diff
@@ -9511,6 +9550,31 @@ function frmAdminBuildJS() {
...
+
+		if ( ! isChild && sidebar.display !== 'none' ) {
```
Crabcyborg, line 9573: I think this line is wrong.

I think `sidebar.display` is always undefined. You're likely trying to access `sidebar.style.display` here instead.

js/formidable_admin.js
```diff
@@ -6945,6 +6944,46 @@ function frmAdminBuildJS() {
...
+		const wrapTextareaWithIconContainer = fieldSettingsContainer => {
+			const textarea = document.querySelector( fieldSettingsContainer + ' textarea' );
+			const wrapperSpan = span({ className: 'frm-with-right-icon' });
+			textarea.parentNode.insertBefore( wrapperSpan, textarea );
+			wrapperSpan.appendChild( textarea );
+			wrapperSpan.insertBefore( createModalTriggerIcon(), textarea );
+		};
+
+		const createModalTriggerIcon = () => {
+			const modalTriggerIcon = document.createElementNS( 'http://www.w3.org/2000/svg', 'svg' );
```
Crabcyborg, line 6971: You should be able to use `frmDom.svg({ href: '#frm_more_horiz_solid_icon' })` to remove a lot of this code.

js/formidable_admin.js
```diff
@@ -9511,6 +9550,31 @@ function frmAdminBuildJS() {
...
+		const sidebar = document.getElementById( 'frm_adv_info' );
+		if ( ! sidebar ) {
+			return;
+		}
+
+		if ( sidebar.dataset.fills === e.target.id && typeof e.target.id !== 'undefined' ) {
+			return;
+		}
+
+		const isChild = e.target .closest( '#frm_adv_info' );
```
Crabcyborg, line 9571: There's a weird space here between `e.target` and `.closest`.

truongwp:
@AbdiTolesa I still get the drag issue like Mike got. It happens when I click on the HTML field to add that field:

https://github.com/Strategy11/formidable-forms/assets/19748318/2ac47f15-5f5d-4179-ae51-91d6c2783646

I'm using Chrome.

js/formidable_admin.js
AbdiTolesa: @crabcyborg I have tried using that before deciding to write a new function and I think it was missing something something important in the trigger icon.

AbdiTolesa:
> @AbdiTolesa I still get the drag issue like Mike got. It happens when I click on the HTML field to add that field:

@truongwp I tried this again but can't replicate it anymore with different browsers. @garretlaxton Do you see issues interacting with an HTML field with this branch?

js/formidable_admin.js
```diff
@@ -6968,13 +6968,11 @@ function frmAdminBuildJS() {
 		};
 
 		const createModalTriggerIcon = () => {
-			const modalTriggerIcon = document.createElementNS( 'http://www.w3.org/2000/svg', 'svg' );
-			modalTriggerIcon.className.baseVal = 'frmsvg frm-show-box';
-			const useElement = document.createElementNS( 'http://www.w3.org/2000/svg', 'use' );
-			useElement.setAttributeNS( 'http://www.w3.org/1999/xlink', 'href', '#frm_more_horiz_solid_icon' );
-			modalTriggerIcon.appendChild( useElement );
+			const moreIcon = frmDom.svg({ href: '#frm_more_horiz_solid_icon', classList: [ 'frm', 'frm-show-box' ] });
```
Crabcyborg, line 6971: What is going on with `frm` here as a class name? The `frmsvg` gets added automatically so I imagine we don't need this `frm` string here at all.

js/formidable_admin.js
```diff
@@ -6968,13 +6968,11 @@ function frmAdminBuildJS() {
...
+			moreIcon.firstElementChild.removeAttribute( 'href' );
+			moreIcon.firstElementChild.setAttributeNS( 'http://www.w3.org/1999/xlink', 'xlink:href', '#frm_more_horiz_solid_icon' );
```
Crabcyborg, line 6973: So the issue has to do with needing to use the deprecated `xlink:href` attribute?

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href

Where is it breaking? We may need to update other code to expect more modern attribute values?

Crabcyborg: @AbdiTolesa What about fixing it like this? https://github.com/Strategy11/formidable-forms/pull/1489

garretlaxton:
@AbdiTolesa No, I don't have any issues with this anymore.

js/formidable_admin.js
AbdiTolesa: Thanks, that looks better!

truongwp:
Let me try replicating that issue on a live site.

truongwp:
I can't replicate it anymore. Sorry for the delay.

truongwp (approved):

classes/views/frm-fields/back-end/html-content.php
```diff
@@ -3,25 +3,16 @@
 	die( 'You are not allowed to call this page directly.' );
 }
 ?>
-<p class="frm-has-modal">
```
Crabcyborg, line 6: What is the reason for removing this class?

js/formidable_admin.js
```diff
@@ -783,7 +782,7 @@ function frmAdminBuildJS() {
 		$link.closest( 'div' ).children( '.tabs-panel' ).not( t ).not( c ).hide();
 		document.getElementById( t.replace( '#', '' ) ).style.display = 'block';
 
-		if ( this.id === 'frm_insert_fields_tab' ) {
+		if ( this.id === 'frm_insert_fields_tab' && ! document.getElementById( 'frm_builder_page' ) ) {
```
Crabcyborg, line 785: Why is it that we do not want to clear the settings box on the builder page?

A comment may be useful here.

It isn't clear why this check is here.

js/formidable_admin.js
```diff
@@ -6945,6 +6944,40 @@ function frmAdminBuildJS() {
...
+			return;
+		}
+
+		const fieldSettingsContainer = '#frm-single-settings-' + fieldId;
+		if ( document.querySelector( fieldSettingsContainer + ' .frm-show-box' ) ) {
+			return;
+		}
+		singleField.querySelector( '.wp-editor-container' ).classList.add( 'frm_has_shortcodes' );
+
+		const wrapTextareaWithIconContainer = fieldSettingsContainer => {
```
Crabcyborg, line 6962: There is no need to pass this as a param in its current scope. It just makes this more confusing to follow, especially since it uses the same variable name.

Also, `fieldSettingsContainer` is not a container. I would reserve that name for the actual element, not the selector string. I would make that more clear in the name of the variable, probably like `fieldSettingsSelector` to keep it from getting too long. The `Container` details is less important in this case.

js/formidable_admin.js
```diff
@@ -6945,6 +6944,40 @@ function frmAdminBuildJS() {
...
+			const textarea = document.querySelector( fieldSettingsContainer + ' textarea' );
+			const wrapperSpan = span({ className: 'frm-with-right-icon' });
+			textarea.parentNode.insertBefore( wrapperSpan, textarea );
+			wrapperSpan.appendChild( textarea );
+			wrapperSpan.insertBefore( createModalTriggerIcon(), textarea );
```
Crabcyborg, line 6967: Can you just append the trigger icon first?

Like
```
wrapperSpan.appendChild( createModalTriggerIcon() );
wrapperSpan.appendChild( textarea );
```

This way you don't need to use `.insertBefore` which is a bit more awkward to work with.

js/formidable_admin.js
```diff
@@ -6945,6 +6944,40 @@ function frmAdminBuildJS() {
...
+		};
+
+		const createModalTriggerIcon = () => {
+			return frmDom.svg({ href: '#frm_more_horiz_solid_icon', classList: [ 'frm-show-box' ] });
+		};
+
+		wrapTextareaWithIconContainer( fieldSettingsContainer );
+		document.getElementById( 'frm-html-tags' ).style.display = 'none';
```
Crabcyborg, line 6975: What is exactly going on on this line? Why do we need to hide `#frm-html-tags`? Will `frm-html-tags` stay hidden when it shouldn't? Is there a possibility of a bug there?

Can we add a comment here explaining all of that?

js/formidable_admin.js
```diff
@@ -9511,6 +9550,31 @@ function frmAdminBuildJS() {
 		tables.forEach( table => table.classList.add( 'frm-zebra-striping' ) );
 	};
 
+	function maybeHideShortcodes( e ) {
+		if ( ! document.getElementById( 'frm_builder_page' ) ) {
```
Crabcyborg, line 9554: Can we maybe set this to a variable on load so we aren't calling `document.getElementById` all over the place?

classes/views/frm-fields/back-end/html-content.php
AbdiTolesa: I think it was used in the old broken modal we replaced but I don't see its purpose anymore now.

js/formidable_admin.js
AbdiTolesa: I have added a comment about this line since it would cause the field settings to hide on popup open.

js/formidable_admin.js
AbdiTolesa: That div was being displayed under the 'Fields' tab for some reason like in the screenshot below while it shouldn't and I had to add this line to make it consistent with how it works in other places. Removing this line isn't causing the issue for me anymore though and I'm removing it.
image

js/formidable_admin.js
```diff
@@ -783,7 +783,8 @@ function frmAdminBuildJS() {
 		$link.closest( 'div' ).children( '.tabs-panel' ).not( t ).not( c ).hide();
 		document.getElementById( t.replace( '#', '' ) ).style.display = 'block';
 
-		if ( this.id === 'frm_insert_fields_tab' ) {
+		// clearSettingsBox would hide field settings when opening the fields modal and we want to skip it there.
+		if ( this.id === 'frm_insert_fields_tab' && ! builderPage ) {
```
Crabcyborg, line 787: So, I'm thinking this doesn't really make sense still.

`clearSettingsBox();` is really a function for the builder page. If we don't want this on the builder page, we'd want to remove it.

The problem is really that we're triggering `clickNewTab` when `#frm-nav-tabs a` is clicked.


And the problem here is that there are two `#frm-nav-tabs` elements.

We need to figure out how to make this unique.

AbdiTolesa: Thanks for the insight @Crabcyborg! I have updated this to selectively avoid calling the function for the fields shortcode. I also found that the previous update of switching to `href` from `xlink:href` on the `use` tags broke the function that toggles the icon and I have pushed a small update to fix that.

Crabcyborg (approved):
I think this is good now.

🥳

PR #1450, opened 2024-01-08
Switch a few options to not autoload for performance

stephywells:
I'm going through a list of autoload options, and removing what isn't needed on every page.

Also, the settings are duplicated in the autoloading with both the transient and the option. This PR simplifies and optimizes getting the global settings. I'm not seeing a good reason to muddy this up with an extra transient when the option is being auto loaded.

@Crabcyborg do you see any possible issues with this?


Crabcyborg (approved):
Thanks @stephywells. This looks good to me.

I've requested some QA from Garret just in case.

garretlaxton (approved):
This looks good to me.

PR #1457, opened 2024-01-10
Summary email enhancement

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/4711

### DIFF
--- a/classes/controllers/FrmEmailSummaryController.php
+++ b/classes/controllers/FrmEmailSummaryController.php
@@ -20,6 +20,8 @@ class FrmEmailSummaryController {
 	 */
 	public static function maybe_send_emails() {
 		$emails = FrmEmailSummaryHelper::should_send_emails();
+		self::maybe_skip_email( $emails );
+
 		if ( ! $emails ) {
 			return;
 		}
@@ -28,6 +30,27 @@ class FrmEmailSummaryController {
 			if ( method_exists( 'FrmEmailSummaryHelper', 'send_' . $email ) ) {
 				call_user_func( array( 'FrmEmailSummaryHelper', 'send_' . $email ) );
 			}
+		}
+	}
+
+	/**
+	 * Skips yearly email if there is no entry in that period.
+	 *
+	 * @since x.x
+	 *
+	 * @param array $emails Summary emails should be sent.
+	 */
+	private static function maybe_skip_email( &$emails ) {
+		$index = array_search( FrmEmailSummaryHelper::YEARLY, $emails );
+		if ( false === $index ) {
+			return;
+		}
+
+		$to_date       = FrmEmailSummaryHelper::get_date_from_today();
+		$from_date     = gmdate( 'Y-m-d', strtotime( $to_date . '-364 days' ) );
+		$entries_count = FrmEmailSummaryHelper::get_entries_count( $from_date, $to_date );
+		if ( ! $entries_count ) {
+			unset( $emails[ $index ] );
 		}
 	}
 }
--- a/classes/controllers/FrmEmailSummaryController.php
+++ b/classes/controllers/FrmEmailSummaryController.php
@@ -20,7 +20,6 @@ class FrmEmailSummaryController {
 	 */
 	public static function maybe_send_emails() {
 		$emails = FrmEmailSummaryHelper::should_send_emails();
-
 		if ( ! $emails ) {
 			return;
 		}
--- a/classes/controllers/FrmEmailSummaryController.php
+++ b/classes/controllers/FrmEmailSummaryController.php
@@ -20,7 +20,6 @@ class FrmEmailSummaryController {
 	 */
 	public static function maybe_send_emails() {
 		$emails = FrmEmailSummaryHelper::should_send_emails();
-		self::maybe_skip_email( $emails );
 
 		if ( ! $emails ) {
 			return;
@@ -30,27 +29,6 @@ class FrmEmailSummaryController {
 			if ( method_exists( 'FrmEmailSummaryHelper', 'send_' . $email ) ) {
 				call_user_func( array( 'FrmEmailSummaryHelper', 'send_' . $email ) );
 			}
-		}
-	}
-
-	/**
-	 * Skips yearly email if there is no entry in that period.
-	 *
-	 * @since x.x
-	 *
-	 * @param array $emails Summary emails should be sent.
-	 */
-	private static function maybe_skip_email( &$emails ) {
-		$index = array_search( FrmEmailSummaryHelper::YEARLY, $emails );
-		if ( false === $index ) {
-			return;
-		}
-
-		$to_date       = FrmEmailSummaryHelper::get_date_from_today();
-		$from_date     = gmdate( 'Y-m-d', strtotime( $to_date . '-364 days' ) );
-		$entries_count = FrmEmailSummaryHelper::get_entries_count( $from_date, $to_date );
-		if ( ! $entries_count ) {
-			unset( $emails[ $index ] );
 		}
 	}
 }
--- a/classes/helpers/FrmEmailSummaryHelper.php
+++ b/classes/helpers/FrmEmailSummaryHelper.php
@@ -540,4 +540,15 @@ class FrmEmailSummaryHelper {
 		$skip_emails = is_string( $addons['no_emails'] ) ? explode( ',', $addons['no_emails'] ) : (array) $addons['no_emails'];
 		$recipients  = array_diff( $recipients, $skip_emails );
 	}
+
+	/**
+	 * Echos string in plain text email.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $string string.
+	 */
+	public static function plain_text_echo( $string ) {
+		echo wp_strip_all_tags( $string ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+	}
 }
--- a/classes/helpers/FrmEmailSummaryHelper.php
+++ b/classes/helpers/FrmEmailSummaryHelper.php
@@ -530,7 +530,7 @@ class FrmEmailSummaryHelper {
 			return;
 		}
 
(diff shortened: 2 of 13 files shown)


classes/views/summary-emails/stats.php
```diff
@@ -70,7 +70,7 @@
 				foreach ( $args['top_forms'] as $index => $top_form ) {
 					?>
 					<tr>
-						<td align="left" style="padding: 1em 0.33em 0.28em 0"><?php echo intval( $index + 1 ); ?>. <?php echo esc_html( $top_form->form_name ); ?></td>
+						<td align="left" style="padding: 1em 0.33em 0.28em 0"><?php echo intval( $index + 1 ); ?>. <?php echo esc_html( wp_unslash( $top_form->form_name ) ); ?></td>
```
stephywells, line 73: Can we strip slashes when get get the data in get_top_forms()? I think that would be more reliable.

classes/views/summary-emails/stats-plain.php
```diff
@@ -12,38 +12,39 @@
 	die( 'You are not allowed to call this page directly.' );
 }
 
-echo esc_html( $args['subject'] );
+//phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped,WordPress.Security.EscapeOutput.UnsafePrintingFunction -- don't need to escape for plain-text email.
```
stephywells, line 15: Can you tell me what to add this exception? Is it causing a problem when the output is escaped? It's best if we can avoid making exceptions unless there's a really good reason for it.

classes/controllers/FrmEmailSummaryController.php
```diff
@@ -20,6 +20,8 @@ class FrmEmailSummaryController {
 	 */
 	public static function maybe_send_emails() {
 		$emails = FrmEmailSummaryHelper::should_send_emails();
+		self::maybe_skip_email( $emails );
```
stephywells, line 23: This looks like it might be running a DB query a lot more often than we need and creates some duplication.

Would it make sense to return right after getting the $entries_count in FrmEmailStats instead? And check again if it's empty in FrmEmailSummary::send?

classes/helpers/FrmEmailSummaryHelper.php
```diff
@@ -515,4 +515,22 @@ public static function get_date_from_today( $date_diff = '' ) {
...
+	 * @param array $recipients Recipients.
+	 */
+	public static function maybe_remove_recipients_from_api( &$recipients ) {
+		$api    = new FrmFormApi();
+		$addons = $api->get_api_info();
+		if ( empty( $addons['no_emails'] ) ) {
+			return;
+		}
+
+		$skip_emails = explode( ',', $addons['no_emails'] );
```
stephywells, line 533: Would you prefer the api sends an array or comma separated list? We can do either.

classes/models/FrmEmailSummary.php
```diff
@@ -130,9 +131,14 @@ protected function get_content_args() {
 	 */
 	public function send() {
 		$recipients = $this->get_recipients();
-		$content    = $this->get_content();
-		$subject    = $this->get_subject();
-		$headers    = $this->get_headers();
+		if ( ! $recipients ) {
+			// Return true to not try to send this email on the next day.
+			return true;
```
stephywells, line 136: Is it possible to kill the email before we run all the queries?

stephywells (commented):
Thanks so much for the quick update @truongwp !

I have a couple questions.

classes/views/summary-emails/stats-plain.php
truongwp: In plain-text email, `esc_html()` will encode special characters, like `'`.

classes/controllers/FrmEmailSummaryController.php
truongwp: Once `yearly` is in the emails list, there is always at least one `get_entries_count()` call. I use `FrmDb` so the result should be cached. I think this is fine.

But your solution will reduce some duplicate code, so I will go that way.

classes/helpers/FrmEmailSummaryHelper.php
truongwp: Both are ok for me. We should go with the format that takes less bandwidth.
I'll update this code to support both array and string.

classes/models/FrmEmailSummary.php
truongwp: All queries are run in the `$this->get_content()`, which is after this line. So no heavy queries run if no recipient. Is that what you want?

classes/views/summary-emails/stats-plain.php
stephywells: Would wp_strip_all_tags work instead?

classes/helpers/FrmEmailSummaryHelper.php
stephywells: Sounds good. Thank you!

classes/models/FrmEmailSummary.php
stephywells: Perfect! Thank you!

classes/views/summary-emails/stats-plain.php
truongwp: The problem is with quotes. Example, if the string we want to output is `We're`, the output of `esc_html( $str )` will be `We&quot;re`.

We can use `wp_specialchars_decode( esc_html( $str ) )` or `wp_specialchars_decode( wp_strip_all_tags( $content ), ENT_QUOTES )`, but this way still needs the PHPCS exception.

classes/views/summary-emails/stats.php
truongwp: I did it.

classes/controllers/FrmEmailSummaryController.php
truongwp: I updated this to skip sending the yearly email in the `send()` method. Please check it.

stephywells: Very nice!

classes/views/summary-emails/stats-plain.php
stephywells: @Crabcyborg do you have suggestions on escaping the HTML in the plain text emails?

Crabcyborg: @truongwp Are you sure `wp_strip_all_tags` has the same issues with quotes?

I played with a small example,
```
$var = "\"Mike's string\"";
var_dump( $var === wp_strip_all_tags( $var ) );
```

The result is true.

`esc_html` calls `_wp_specialchars`, but `wp_strip_all_tags` doesn't.

Removing some unimportant code from the source code of `wp_strip_all_tags`, it doesn't encode anything. It just removes stuff:

```
$text = preg_replace( '@<(script|style)[^>]*?>.*?</\\1>@si', '', $text );
$text = strip_tags( $text );

if ( $remove_breaks ) {
	$text = preg_replace( '/[\r\n\t ]+/', ' ', $text );
}

return trim( $text );
```

truongwp: @Crabcyborg `wp_strip_all_tags()` will work. We also used it in one of the summary email code. But we still need phpcs:ignore comment if using it.

Crabcyborg: That's okay.

I think we should remove the `phpcs:disable` for the whole file.

Do we actually need to change from `esc_html__` to `__` in this file?

truongwp: @Crabcyborg Yes, we should change them to `__`. So will we have multiple `php:ignore`?

Crabcyborg: @truongwp What if we move all of the echo logic to a single line of code so we can just add one `phpcs:ignore` comment, reduce the number of calls to `wp_strip_all_tags`, and also have more flexibility in the future in case we want to make changes.

I mocked up what I'm thinking in just one of the files for now (`stats-plain.php`). I haven't actually tested it either.

truongwp: Thanks @Crabcyborg. That's a great idea. I'll update your PR and merge it into this.

truongwp: @stephywells I updated to use Mike's solution. Please check it.

classes/models/FrmEmailSummary.php
```diff
@@ -44,7 +44,7 @@ public function __construct() {
 		 *
 		 * @param bool $enable Set to `true` to enable.
 		 */
-		$this->is_html = apply_filters( 'frm_html_summary_emails', true );
+		$this->is_html = apply_filters( 'frm_html_summary_emails', false );
```
stephywells, line 47: Was this line changed intentionally?

stephywells (approved):

classes/views/summary-emails/stats-plain.php
stephywells: Thanks for the update! This looks great. I didn't test though, so this could still use a round of testing.

classes/models/FrmEmailSummary.php
truongwp: Sorry. That was the test code. I fixed it.

classes/helpers/FrmEmailSummaryHelper.php
```diff
@@ -324,6 +324,13 @@ public static function get_top_forms( $from_date, $to_date, $limit = 5 ) {
 				intval( $limit )
 			)
 		);
+
+		// Remove slashes from form name.
+		foreach ( $result as &$value ) {
```
Crabcyborg, line 329: Can we possibly do this loop without passing by reference?

garretlaxton:
@truongwp @Crabcyborg I'm not able to find the `allow people with an Elite license to disable these emails from their profile page` option. Where can I find this?

truongwp:
@garretlaxton You can skip it. Steph told me that she would add that setting when this is released.

garretlaxton (approved):
This looks great!

PR #1458, opened 2024-01-10
Refactor code to reflect change in pro script

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-stripe/issues/98


stripe/js/frmstrp.js
```diff
@@ -262,17 +262,26 @@
...
 				return queryForNameFieldIsFound ?
 				window.frmProForm.queryForNameField( fieldID, 'container' ) :
 				document.querySelector( '#frm_field_' + fieldID + '_container, .frm_field_' + fieldID + '_container' );
 			}
+
 			return queryForNameFieldIsFound ?
-			window.frmProForm.queryForNameField( fieldID, 'field', $form ) :
-			$form.find( '#frm_field_' + fieldID + '_container input, input[name="item_meta[' + fieldID + ']"], .frm_field_' + fieldID + '_container input' );
+			window.frmProForm.queryForNameField( fieldID, 'field', $form[0]) :
+			$form[0].querySelector( '#frm_field_' + fieldID + '_container input, input[name="item_meta[' + fieldID + ']"], .frm_field_' + fieldID + '_container input' );
```
Crabcyborg, line 284: Are we guaranteed that `$form[0]` would never be undefined or that `$form` is never going to be `null`? If it is, this could cause errors.

AbdiTolesa: When `$form` is `null`, we are exiting early in the `if` block above. Otherwise, I can see  it is defined like `$form = jQuery( thisForm );` and used without a check in multiple places so it looks safe to assume that this line doesn't potentially cause an error.

Crabcyborg (approved):
Thanks @AbdiTolesa!

This looks good.

🚀

PR #1460, opened 2024-01-11
Update frm-new-pill to match tags

stephywells:
I missed this one the first time around. The "New" tag is also green now.

Before:
Screenshot 2024-01-11 at 9 18 06 AM

After:
Screenshot 2024-01-11 at 9 17 53 AM
Screenshot 2024-01-11 at 9 19 28 AM
Screenshot 2024-01-11 at 9 52 21 AM

I had to make a few changes to the application list so this would fit. We'll need to circle back to this later to match the template page better since this isn't ideal.
Screenshot 2024-01-11 at 10 16 18 AM

### DIFF
--- a/css/admin/applications.css
+++ b/css/admin/applications.css
@@ -67,8 +67,8 @@
 
 /* Card title */
 .frm-application-card h3 {
-	display: inline-flex;
-	max-width: 100%;
+	display: flex;
+	width: 100%;
 	align-items: center;
 	flex-wrap: nowrap;
 	gap: var(--gap-xs);
--- a/css/admin/applications.css
+++ b/css/admin/applications.css
@@ -61,27 +61,36 @@
 	box-shadow: var(--box-shadow-lg);
 }
 
-.frm-application-template-card:hover h4 {
+.frm-application-template-card:hover h3 {
 	color: var(--primary-500);
 }
 
 /* Card title */
-.frm-application-card h4 {
-	display: inline-block;
+.frm-application-card h3 {
+	display: inline-flex;
 	max-width: 100%;
+	align-items: center;
+	flex-wrap: nowrap;
+	gap: var(--gap-xs);
+	margin: 0;
+	font-size: var(--text-sm) !important;
+	font-weight: 500;
+}
+
+.frm-application-card h3 .frmsvg {
+	width: 11px;
+	height: 13px;
+}
+
+.frm-application-card h3 .frm-inner-text {
+	max-width: calc( 100% - 50px );
 	overflow: hidden;
 	white-space: nowrap;
 	text-overflow: ellipsis;
+}
+
+.frm-application-card h3 .frm-meta-tag {
 	margin: 0;
-}
-
-.frm-locked-application-template h4 {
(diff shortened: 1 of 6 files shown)


css/admin/applications.css
```diff
@@ -61,27 +61,36 @@
...
-.frm-application-template-card:hover h4 {
+.frm-application-template-card:hover h3 {
 	color: var(--primary-500);
 }
 
 /* Card title */
-.frm-application-card h4 {
-	display: inline-block;
+.frm-application-card h3 {
+	display: inline-flex;
```
shervElmi, line 70: I think we should use `display: flex;` here because with this approach, we could avoid bugs like the one below:
Bug Example

shervElmi (approved):
Thanks so much @stephywells, for your work. I really see improvements after your PRs on matching tags!

css/admin/applications.css
stephywells: Oh wow... That looks bad. Not sure how I missed that. I must have had a width set in my console and didn't make it to the page, and didn't check without the new tags last thing.

This is fixed now.

Crabcyborg (approved):
This looks good to me!

🚀

css/admin/applications.css
shervElmi: Thanks for your fix @stephywells, it's great now! 🚀

---

I've also found a way to reduce CSS, but I wasn't sure about it. Here's my idea 😄:

I searched across all repositories under org:Strategy11 for `frm-new-pill` and found that we only use it in the places you worked on in this PR:
frm-new-pill usage

Therefore, I think we can remove the current styles from `frm-new-pill`, even if they are inside the `frm_admin_global.css` file. Here are my changes:

Proposed CSS changes 1

Proposed CSS changes 2

stephywells: Yeah, I checked that too. We had it in the global css since we added the "new" pill on the "applications" menu...

In fact... Instead of removing it, we should add it to the "Form Templates" nav with a date for when it should stop showing. Would you mind adding this @shervElmi ? Here's an example:
https://github.com/Strategy11/formidable-forms/commit/cae476c0100b4c7e159110cbb6a470a990056ae1

Let's show that "New" pill until the end of February.

shervElmi: Sure, that's a great idea.

shervElmi: I have two questions for you, @stephywells:

1. **Menu Display Issue:** After adding the new pill to the menu, I noticed that the "Form Templates New" label splits into two lines, as depicted in the screenshot. Does this look good to you?

   Menu Display Issue

2. **Time-Related Example:** I'm struggling to find an example of how to add time in the context mentioned in the recent commit. Could you point me in the right direction?

stephywells: 1. It looks like we need to remove the left margin in the global css and move that to our admin css. That looks like it fits for me.
2. Would could do it like this:
`$new_pill = gmdate( 'Y-m-d', ) < '2023-02-28' ? '<span class="frm-new-pill">NEW</span>' : '';`

PR #1461, opened 2024-01-11
Fix: Correct Placement of Support Footer Links on Lite-Reports Page

shervElmi:
This PR addresses the issue where support footer links are not correctly positioned on the "lite-reports" page (#4714). Currently, these links appear at the top of the page, above the content, and move up and down when scrolling, rather than staying in a fixed position.

## QA URL:
https://qa.formidableforms.com/sherv/wp-admin/admin.php?page=formidable&frm_action=lite-reports&form=1162

## Related Issue:
[Support footer links don't appear in the right place on the "lite-reports" page](https://github.com/Strategy11/formidable-pro/issues/4714)

## Accessing the Issue:
To reproduce this issue:
1. Disable Pro.
2. Go to a form with entries.
3. Click on the greyed-out "Reports" tab.
   - Example URL: `https://example.com/wp-admin/admin.php?page=formidable&frm_action=lite-reports&form=1162`

## Changes Made:
This PR includes adjustments to ensure the support footer links appear in their correct position on the "lite-reports" page, enhancing the user interface and usability.

### Before
image

### After
image

### DIFF
--- a/classes/views/shared/reports-info.php
+++ b/classes/views/shared/reports-info.php
@@ -4,34 +4,32 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 ?>
 <div class="frm_wrap">
-	<div class="frm_page_container">
-		<?php
-		FrmAppHelper::get_admin_header(
-			array(
-				'label'       => __( 'Reports', 'formidable' ),
-				'form'        => $form,
-				'close'       => $form ? admin_url( 'admin.php?page=formidable&frm_action=reports&form=' . $form->id ) : '',
-			)
-		);
-		?>
-		<div class="frmcenter" style="margin-top:10vh">
-			<h2><?php esc_html_e( 'Get Live Graphs and Reports', 'formidable' ); ?></h2>
-			<p style="max-width:400px;margin:20px auto">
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This looks great!

truongwp (approved):

Crabcyborg:
@shervElmi I wonder if we actually just want to update the HTML of the `lite-reports` page?

It looks like, if my browser window height is fairly small, that I'm unable to actually scroll to the bottom of the page.

Here's a recording. I scroll, but only the sidebar ever moves.
HFGfwkYFyl

classes/views/shared/reports-info.php
```diff
@@ -4,34 +4,32 @@
 }
 ?>
 <div class="frm_wrap">
-	<div class="frm_page_container">
```
shervElmi, line 7: The `#frm_page_container` makes our top menu sticky. However, on this page, if we retain it and address the problem you mentioned, it creates a rather amusing issue, as shown below. Therefore, I decided to remove this element from this page.

Let's also keep the JavaScript updated in this PR, as it offers a better method for managing where floating links should be.

Image Description

Crabcyborg (approved):
Thanks @shervElmi!

The scrolling looks better now.

Thank you @truongwp and @garretlaxton for your help with review!

🚀

PR #1462, opened 2024-01-11
Pro issue 4713 / Add a draft flag to new fields and hide them from the front end except for on preview

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4713

I cannot reliably delete the draft fields. The only event listener (that isn't deprecated) that gets triggered when leaving the page is `beforeunload` which is cancellable (with no way of knowing if it was cancelled or not) which doesn't make it a reliable event to try deleting drafts from as they would get deleted when you cancel because of unsaved changes.

And if we delete on form builder load there is a risk of deleting another user's draft fields that they are working on, so I'm not deleting fields due to the possible bugs that could occur from deleting fields that shouldn't be deleted.

This update adds a new `draft_fields` input to the page that stores a CSV of all draft field IDs. This allows us to cross-reference this before making a database check, and it guarantees that the draft fields getting marked as active are actually part of the form getting saved rather than part of another incomplete session.

This update also adds a `draft` flag to new fields. When a new field is added it will not appear in the front end until the save/update button is clicked. There is an exception though - a draft field is shown on the "preview" page for a user that can edit forms. If the user visiting the preview page cannot edit forms, or the form is embedded in a page, the draft fields will not be shown.

With this update, the unsaved changes pop up will trigger now whenever a new field is added.

**TODO**
- [x] Add unit tests.
- [x] Add additional unsaved changes messages.
- [x] Steph's comments https://github.com/Strategy11/formidable-pro/issues/4713#issuecomment-1889704905

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -213,6 +213,24 @@ class FrmFormsController {
 		if ( count( $errors ) > 0 ) {
 			return self::get_edit_vars( $id, $errors );
 		} else {
+			$draft_field_rows = FrmDb::get_results(
+				'frm_fields',
+				array(
+					'form_id'            => $id,
+					'field_options LIKE' => 's:5:"draft";i:1;',
+				),
+				'id, field_options'
+			);
+			foreach ( $draft_field_rows as $row ) {
+				FrmAppHelper::unserialize_or_decode( $row->field_options );
+				if ( ! is_array( $row->field_options ) || empty( $row->field_options['draft'] ) ) {
+					continue;
+				}
+
+				$row->field_options['draft'] = 0;
+				FrmField::update( $row->id, array( 'field_options' => $row->field_options ) );
+			}
+
 			FrmForm::update( $id, $values );
 			$message = __( 'Form was successfully updated.', 'formidable' );
 
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -213,23 +213,7 @@ class FrmFormsController {
 		if ( count( $errors ) > 0 ) {
 			return self::get_edit_vars( $id, $errors );
 		} else {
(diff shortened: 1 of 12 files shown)


classes/controllers/FrmFormsController.php
```diff
@@ -233,6 +235,47 @@ public static function update( $values = array() ) {
...
+		if ( ! $draft_field_ids ) {
+			// Exit early if the draft fields input is invalid. It should be a CSV of integer values.
+			return;
+		}
+
+		$draft_field_rows = FrmDb::get_results(
+			'frm_fields',
+			array(
+				'id'                 => $draft_field_ids,
+				'form_id'            => $form_id,
```
stephywells, line 263: Does this cover fields inside a repeater?

stephywells (commented):
This is looking good and is a nice, simple approach.

Does this work right with sections since it auto adds the end divider field?

classes/controllers/FrmFormsController.php
Crabcyborg: Oh, good point. It probably doesn't. I'll look into that.

Crabcyborg:
@stephywells I made some adjustments based on your comments.

- I now check the `msg` variable after a field is added and query for all of the fields so I can catch all of the new ones.
- I added a query that gets all children form ids and includes that in the field query when "Pro is installed".
- I made a new `get_draft_field_results` function so I can re-use it in both places without needing to repeat logic. I added some new unit tests to cover this.

stephywells (approved):
Thanks for the updates! This looks great. I didn't test it, but the approach looks good.

garretlaxton (approved):
This fix works great! Only new fields are added to the preview page, and any page that the form is embedded on isn't changed.

Crabcyborg:
Thank you Garret and Steph!

🥳

PR #1463, opened 2024-01-11
Fix: Hide 'Entries' for XML Export in Lite Mode (#4709)

shervElmi:
This PR resolves the issue where the 'Entries' checkbox appeared for XML exports in Formidable Lite, despite being non-functional. Now, the 'Entries' option is hidden when XML is selected in Lite mode, aligning with the expected behavior.

## QA URL:
https://qa.formidableforms.com/sherv2/wp-admin/admin.php?page=formidable-import

## Related Issue:
[#4709 - Stop showing "Entries" as an XML export option when Pro is inactive](https://github.com/Strategy11/formidable-pro/issues/4709)

## Testing Instructions:
1. Disable Pro.
2. Go to `WP Admin > Formidable > Import/Export`.
3. Verify that the 'Entries' checkbox is not visible for XML export.

### Before
image

### After
image

### DIFF
--- a/classes/views/xml/import_form.php
+++ b/classes/views/xml/import_form.php
@@ -109,7 +109,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 			<p id="frm_csv_data_export" class="xml_opts">
 				<label><?php esc_html_e( 'Include the following in the export file', 'formidable' ); ?></label>
 				<?php foreach ( $export_types as $t => $type ) { ?>
-					<label class="frm_inline_label frm-<?php echo esc_attr( sanitize_title( $type ) ); ?>">
+					<label class="frm_inline_label frm-<?php echo esc_attr( sanitize_title( $t ) ); ?>">
 						<input type="checkbox" name="type[]" value="<?php echo esc_attr( $t ); ?>"/>
 						<?php echo esc_html( $type ); ?>
 					</label> &nbsp;
--- a/classes/views/xml/import_form.php
+++ b/classes/views/xml/import_form.php
@@ -109,7 +109,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 			<p id="frm_csv_data_export" class="xml_opts">
 				<label><?php esc_html_e( 'Include the following in the export file', 'formidable' ); ?></label>
 				<?php foreach ( $export_types as $t => $type ) { ?>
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This looks great!

classes/views/xml/import_form.php
```diff
@@ -109,7 +109,7 @@
 			<p id="frm_csv_data_export" class="xml_opts">
 				<label><?php esc_html_e( 'Include the following in the export file', 'formidable' ); ?></label>
 				<?php foreach ( $export_types as $t => $type ) { ?>
-					<label class="frm_inline_label">
+					<label class="frm_inline_label frm-<?php echo esc_attr( sanitize_title( $type ) ); ?>">
```
truongwp, line 112: We should use `$t` because `$type` could be translated.

classes/views/xml/import_form.php
```diff
@@ -109,7 +109,7 @@
 			<p id="frm_csv_data_export" class="xml_opts">
 				<label><?php esc_html_e( 'Include the following in the export file', 'formidable' ); ?></label>
 				<?php foreach ( $export_types as $t => $type ) { ?>
-					<label class="frm_inline_label">
+					<label class="frm_inline_label frm-<?php echo esc_attr( sanitize_title( $t ) ); ?>">
```
truongwp, line 112: I don't think we need `sanitize_title()` anymore.

truongwp (approved):

classes/views/xml/import_form.php
```diff
@@ -109,7 +109,7 @@
 			<p id="frm_csv_data_export" class="xml_opts">
 				<label><?php esc_html_e( 'Include the following in the export file', 'formidable' ); ?></label>
 				<?php foreach ( $export_types as $t => $type ) { ?>
-					<label class="frm_inline_label">
+					<label class="frm_inline_label frm-<?php echo esc_attr( $t ); ?>">
```
Crabcyborg, line 112: I'm concerned that `frm-$t` is too short. It could pretty easily lead to a styling conflict since `frm-items` isn't a very unlikely class name.

What about `frm-export-xml-$t` instead?

Crabcyborg (approved):
Thanks for updating that Sherv!

And thank you @truongwp and @garretlaxton for reviewing this!

🚀

PR #1465, opened 2024-01-12
Do not strip tags used in separators on display

Crabcyborg:
Related Slack comment https://strategy11.slack.com/archives/CJFQ599V5/p1705078375154979?thread_ts=1705053738.151109&cid=CJFQ599V5

This update checks the `sep` option on display. When set, if it includes HTML tags, those tags are not removed.

I'm not sure if it's worth checking the tags for attributes as well? Right now, those would still get stripped.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -634,13 +634,22 @@ class FrmAppHelper {
 		return $value;
 	}
 
-	public static function sanitize_value( $sanitize, &$value ) {
+	/**
+	 * @since x.x The args param was added.
+	 *
+	 * @param callable     $sanitize The sanitize function to call.
+	 * @param array|string $value    The value to sanitize.
+	 * @param array        $args     Arguments to pass to the sanitize function (if any).
+	 */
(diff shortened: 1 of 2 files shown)


stephywells (commented):
Just wondering... Is it possible to sanitize before the sep is added? Or would that be really complicated?

Crabcyborg:
@stephywells The main problem is that it's happening in Pro, so it's not even happening here in this plugin.

It's imploding on the separator here in the middle of `FrmProContent::replace_single_shortcode` https://github.com/Strategy11/formidable-pro/blob/31f07a51a9186f65149bf42882472fca426737dd/classes/models/FrmProContent.php#L140

stephywells (approved):
That does make it complicated... This looks like a great solution. Unit tests would be really helpful too.

Crabcyborg:
@stephywells It looks like this is only an issue in Pro. When I disable Pro this works just fine.

i think it's fine to add this logic as part of `FrmProContent::replace_single_shortcode` instead.

So I made a new PR. I added unit tests there https://github.com/Strategy11/formidable-pro/pull/4723

This PR is no longer needed so I'm going to close this.

PR #1468, opened 2024-01-12
Fix pro issue 4717 / Check svg has a use child when cleaning HTML

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4717

I think this is because the cleaning happens now more than once, to clean nested HTML as well.

The error is because there are `svg` elements here with no `use` children. The svg has no child at all.

The cause is because the svg used on the option itself is getting sanitized multiple times and uses an inconsistent `xlink:href`/`href` value.

Screen Shot 2024-01-12 at 4 17 14 PM

Now I check for both values just in case one fails.

This code worked on the assumption an `svg` always had this child. I've updated it to not break in that case either.

### DIFF
--- a/js/admin/dom.js
+++ b/js/admin/dom.js
@@ -813,10 +813,14 @@
 		const tagType = node.tagName.toLowerCase();
 
 		if ( 'svg' === tagType ) {
-			return svg({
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This works great!

PR #1469, opened 2024-01-15
Stop sanitizing on display if updated by user is privileged

Crabcyborg:
Related comment https://github.com/Strategy11/formidable-pro/issues/4647#issuecomment-1892513008

I'm bumping up the `updated_by` check. If it's set, the permission check only checks the `updated_by` user's permissions.

This means that old entry data submitted by untrusted users that wasn't sanitized in the past (in older versions prior to v6.7.1) won't be sanitized with the additional strict filtering on display if the entry was updated by a privileged user that may not have noticed an unsafe anchor tag or img, but that's pretty low risk.

### DIFF
--- a/classes/models/fields/FrmFieldType.php
+++ b/classes/models/fields/FrmFieldType.php
@@ -1395,7 +1395,7 @@ DEFAULT_HTML;
 	protected function should_strip_most_html( $entry ) {
 		if ( $entry->updated_by ) {
 			// Stop stripping HTML on display when the entry has been updated by a privileged user.
-			return ! $this->user_id_is_privileged( $entry->updated_by, 'frm_edit_entries' );
+			return ! $this->user_id_is_privileged( $entry->updated_by );
 		}
 		if ( ! $entry->user_id || ! $this->user_id_is_privileged( $entry->user_id ) ) {
 			return true;
--- a/classes/models/fields/FrmFieldType.php
+++ b/classes/models/fields/FrmFieldType.php
@@ -1387,20 +1387,21 @@ DEFAULT_HTML;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


classes/models/fields/FrmFieldType.php
```diff
@@ -1393,15 +1393,30 @@ protected function should_strip_most_html_before_preparing_display_value( $atts
 	 * @return bool
 	 */
 	protected function should_strip_most_html( $entry ) {
-		if ( ! $entry->user_id || ( ! user_can( $entry->user_id, 'frm_edit_entries' ) && ! user_can( $entry->user_id, 'administrator' ) ) ) {
-			return true;
-		}
 		if ( $entry->updated_by ) {
-			return ! user_can( $entry->updated_by, 'frm_edit_entries' ) && ! user_can( $entry->updated_by, 'administrator' );
+			// Stop stripping HTML on display when the entry has been updated by a privileged user.
+			return ! $this->user_id_is_privileged( $entry->updated_by );
```
stephywells, line 1398: Can we only return here if the user has permission? If no permission, it would be nice to check the created user too.

stephywells (commented):
Thanks for the quick update!

classes/models/fields/FrmFieldType.php
Crabcyborg: I think that's fine. Since the sanitizing also happens on save for unprivileged I don't see a big risk if we stopped sanitizing on display when the entry was created by a privileged user but then edited by an untrusted one.

I modified the function to just return false whenever a user id check is made that is privileged https://github.com/Strategy11/formidable-forms/pull/1469/commits/52fef2218524ea0528ab25e866e8a230adf8f82a

stephywells (approved):
This looks great! Thanks so much!

PR #1471, opened 2024-01-15
Load 15 fields at a time in builder fix

Crabcyborg:
I noticed that all of my `frm_load_field` calls were for 1 or 2 fields at a time.

This code here is capped at 15.

But because of my side-by-side fields update, the `nextElementSibling` isn't always the next field.

I missed this until now.


Crabcyborg:
Thanks @garretlaxton!

🚀

PR #1475, opened 2024-01-17
Prevent null from being passed to string functions

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/3453

This update adds a check for a variable value against null and then convert it to empty string before passing it to string manipulation functions.

### Test steps
1. Make sure your PHP version is `>=8.1`
2. There could be different methods to test this but I just used `FrmAppHelper::recursive_function_map( array(), 'trim' );` in a code snippet.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1948,7 +1948,24 @@ class FrmAppHelper {
 				}
 			}
 		} else {
+			$value = self::maybe_update_value_if_null( $value, $function );
 			$value = call_user_func( $function, $value );
+		}
+
+		return $value;
+	}
+
+	/**
+	 * Updates value to empty string if it is null and being passed to a string function.
+	 *
+	 * @since x.x
+	 * @param mixed $value
+	 * @param string $function
+	 * @return mixed
+	 */
+	private static function maybe_update_value_if_null( $value, $function ) {
+		if ( null === $value && in_array( $function, array( 'trim', 'strlen' ), true ) ) {
+			$value = '';
 		}
 
 		return $value;
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1959,7 +1959,7 @@ class FrmAppHelper {
 	 * Updates value to empty string if it is null and being passed to a string function.
 	 *
 	 * @since x.x
-	 * @param mixed $value
+	 * @param mixed  $value
 	 * @param string $function
 	 * @return mixed
 	 */
--- a/tests/misc/test_FrmAppHelper.php
+++ b/tests/misc/test_FrmAppHelper.php
@@ -753,4 +753,31 @@ class test_FrmAppHelper extends FrmUnitTest {
 			$this->assertEquals( $assertion['expected'], $result );
 		}
 	}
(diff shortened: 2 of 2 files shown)


AbdiTolesa:
Note: There is a codestyle issue that isn't related to this update and I couldn't find on the line mentioned in the error.

Crabcyborg:
> Note: There is a codestyle issue that isn't related to this update and I couldn't find on the line mentioned in the error.

Thanks @AbdiTolesa. That was my fault with an update I made quickly yesterday.

I just fixed it and merged master here to get it passing.

garretlaxton:
Do you have an example of the snippet you used?

classes/helpers/FrmAppHelper.php
```diff
@@ -1951,12 +1951,29 @@ public static function recursive_function_map( $value, $function ) {
 				}
 			}
 		} else {
+			$value = self::maybe_update_value_if_null( $value, $function );
```
Crabcyborg, line 1954: @AbdiTolesa I think a fix like this should probably have some unit test coverage as well.

AbdiTolesa:
> 2\. FrmAppHelper::recursive_function_map( array(), 'trim' );

I have put that in the test steps for you (2). I just called `FrmAppHelper::recursive_function_map( array(), 'trim' );` and checked there is no deprecation warning in PHP 8.1.

classes/helpers/FrmAppHelper.php
AbdiTolesa: @Crabcyborg Sounds good but I see an error when trying to run tests in lite, do you know how to work around that?
image

Crabcyborg: @AbdiTolesa It seems like your auto-loading is possibly broken.

Does it work after running `composer install`?

AbdiTolesa: @Crabcyborg I have tried that but it didn't help fix the issue.

garretlaxton (approved):
This fix works great!

truongwp (approved):

classes/helpers/FrmAppHelper.php
Crabcyborg: @AbdiTolesa Maybe try adding `cacheResult="false"` to your PHPUnit XML, or deleting the `.phpunit.result.cache` file. It may be caching related.

The auto-loading should all be fine. Our workflow works fine, etc. The issue seems to be on your end somewhere.

truongwp: Maybe the plugin folder name is wrong. I see `formidable-forms`. I think it should be `formidable`.

Crabcyborg: @AbdiTolesa It looks like this might have gotten lost a bit. Do any of these suggested fixes solve your issue with running tests? I'd like to merge this, but I think we should have tests.

AbdiTolesa: Thanks @truongwp! That indeed was my mistake. @Crabcyborg I have added a unit test to cover that function now.

Crabcyborg (approved):
I think this looks good.

🚀

PR #1484, opened 2024-01-22
Resolve Display Issue of Install Add-On Buttons in Formidable Forms (Lite Version PR)

shervElmi:
### Main PR: https://github.com/Strategy11/formidable-pro/pull/4726 ###

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -8596,11 +8596,12 @@ function frmAdminBuildJS() {
 			success: function( response ) {
 				let saveAndReload;
 
-				if ( 'string' !== typeof response && 'string' === typeof response.message ) {
-					if ( 'undefined' !== typeof response.saveAndReload ) {
-						saveAndReload = response.saveAndReload;
+				if ( 'string' !== typeof response && 'string' === typeof response?.data.message ) {
+					const { message: dataMessage, saveAndReload: dataSaveAndReload } = response.data;
+					if ( 'undefined' !== typeof dataSaveAndReload ) {
+						saveAndReload = dataSaveAndReload;
 					}
-					response = response.message;
+					response = dataMessage;
 				}
 
 				const error = extractErrorFromAddOnResponse( response );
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This looks great!

Something I do notice is that if the plugin is installed, but not activated it shows the same message that the plugin is not installed. Maybe we could improve that in the future?
image

shervElmi:
Thanks Garret for your comment. @Crabcyborg, what are your thoughts on Garret's comment here? It's outside of this PR, but I would like to know if we need to create a new issue for that or not.

js/formidable_admin.js
```diff
@@ -8596,11 +8596,12 @@ function frmAdminBuildJS() {
 			success: function( response ) {
 				let saveAndReload;
 
-				if ( 'string' !== typeof response && 'string' === typeof response.message ) {
-					if ( 'undefined' !== typeof response.saveAndReload ) {
-						saveAndReload = response.saveAndReload;
+				if ( 'string' !== typeof response && 'string' === typeof response?.data.message ) {
```
Crabcyborg, line 8599: @shervElmi This is causing issues when I activate an add-on from the builder page.

My response is now in this new `response.data` format. It is at `response.message`.

Screen Shot 2024-01-23 at 9 27 34 AM

This results in this TypeError,
Screen Shot 2024-01-23 at 9 28 05 AM

shervElmi:
@Crabcyborg, did you use the `fix-addon-buttons-display-issue` branch for the Formidable Pro plugin?

Crabcyborg:
Thanks @garretlaxton and @shervElmi.

@stephywells Do you think it's worth trying to conditionally change the pop up title and message content based on state of the button? If it says "Activate", we could change it to say something like "This plugin is not activated. Would you like to activate it now?".

image

Crabcyborg:
@shervElmi Yes I did. That update has no changes in it though that would change the response to include a `data` key though.

Crabcyborg:
@shervElmi Can we make it work as well if Pro isn't up to date? It would be nice not to have these errors.

shervElmi:
@Crabcyborg, should this PR be retested by Garret? Made some refactoring in my last commit.

stephywells:
> Do you think it's worth trying to conditionally change the pop up title and message content based on state of the button? If it says "Activate", we could change it to say something like "This plugin is not activated. Would you like to activate it now?".

It would be much better if the text is all applicable for the situation (install vs activate). I know this has come up several times in the past, but I'm not sure if we created an issue for it.

Crabcyborg:
Thanks Steph!

@shervElmi Can you make an issue for it?

shervElmi:
@Crabcyborg, here is the issue URL: https://github.com/Strategy11/formidable-pro/issues/4745

Crabcyborg:
Thank you Sherv!

Crabcyborg (approved):
Thanks @shervElmi! This looks good.

🚀

PR #1486, opened 2024-01-23
Fix action limit message for actions with more than one limit

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4556

This could also be fixed in PHP by adding a new string for "99" action limit but still requires js update to check which action is clicked on and then showing the appropriate corresponding message but I believe we can limit the fix to js by manipulating the message as in this PR, which sounds less complex.

### Steps to test
1. Create a new form and add 99 'Confirmation' actions to the form. I used the following snippet to automate it:
```
$action_control = FrmFormActionsController::get_form_actions( 'on_submit' );
$i = 0;
while ( $i++ < 10 ) {
    $action_control->create( $form_id );
}
```
2. The icon for Confirmation action would be disabled when the limit is hit and clicking on it should show a message like in this screenshot:
image


js/formidable_admin.js
```diff
@@ -7149,7 +7149,12 @@ function frmAdminBuildJS() {
 	}
 
 	function onlyOneActionMessage() {
-		infoModal( frmAdminJs.only_one_action );
+		let message = frmAdminJs.only_one_action;
+		if ( this.classList.contains( 'frm_on_submit_action' ) ) {
+			message = message.replace( 'one', 99 );
```
Crabcyborg, line 7154: The 99 is filterable and could be any number. I wouldn't hard code it. I also wouldn't only do this for on submit actions, but any action at its limit where `data-limit` is not 1.

The action anchor element has a `data-limit=99` on it.

Screen Shot 2024-01-23 at 9 40 25 AM

-

Crabcyborg (commented):
I wonder if we want to conditionally display the second sentence.

"Please edit the existing form action" really only makes sense when there is one action.

garretlaxton (approved):
This looks good!

js/formidable_admin.js
```diff
@@ -7149,7 +7149,17 @@ function frmAdminBuildJS() {
...
 
 	function onlyOneActionMessage() {
-		infoModal( frmAdminJs.only_one_action );
+		let message = frmAdminJs.only_one_action;
+		let limit   = this.dataset.limit;
+
+		if ( 'undefined' !== typeof limit ) {
+			limit = parseInt( limit );
+			if ( limit > 1 ) {
+				message = message.replace( __( 'one', 'formidable' ), limit ).replace( __( 'Please edit the existing form action.', 'formidable' ), '', limit ).trim();
```
Crabcyborg, line 7158: I'm still not in love with this solution.

We don't actually have a translation for "'Please edit the existing form action." so someone would need to translate it only for this purpose, and it would need to be an exact match for their other translation.

It might make more sense to break the translation in PHP, wrap this specific text in a `span` element and then conditionally show or hide it based on the number here.

AbdiTolesa: Ok, that makes more sense, thanks! I have updated this except wrapping the text in a `span` as I didn't get why that is needed here yet.

js/formidable_admin.js
```diff
@@ -7148,8 +7148,20 @@ function frmAdminBuildJS() {
...
-	function onlyOneActionMessage() {
-		infoModal( frmAdminJs.only_one_action );
+	function actionLimitMessage() {
+		let message = frmAdminJs.only_one_action;
+		let limit   = this.dataset.limit;
+
+		if ( 'undefined' !== typeof limit ) {
+			limit = parseInt( limit );
+			if ( limit > 1 ) {
+				message  = message.replace( __( 'one', 'formidable' ), limit ).trim();
```
Crabcyborg, line 7158: We have the same translation issue here with the `one` translation. I'd like to avoid this breaking if this string isn't translated, or there's a translation mismatch.

Maybe instead of `one` we can just use `%d` in the translation and replace it with an actual `1`?

Crabcyborg (approved):

I think this looks good now.

🥳

PR #1488, opened 2024-01-24
Duplicate unsaved field opts

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/3726

This PR extends to making sure field options are synced when a field is duplicated for Checkbox, Radio and Dropdown fields.

### Test steps
1. Create a form and add a Checkbox, Radio and Dropdown fields
2. Add/modify/remove options from each of those fields as needed.
3. Duplicate the fields
4. Check that the duplicate field options are in sync with the original field.


garretlaxton (approved):
This works great!

js/formidable_admin.js
```diff
@@ -2143,13 +2143,114 @@ function frmAdminBuildJS() {
...
+	function copyUnsavedDeleteOptions( origOptsContainer, newOptsContainer, args ) {
+		const originalOpts = origOptsContainer.querySelectorAll( 'li' );
+
+		if ( ! originalOpts ) {
+			return;
+		}
+
+		const { originalFieldId, newFieldId } = args;
+
+		for ( li of originalOpts ) {
```
truongwp, line 2198: Look like we're missing `let` for `li` variable.

js/formidable_admin.js
```diff
@@ -2143,13 +2143,114 @@ function frmAdminBuildJS() {
...
+	}
+
+	function replaceElementAttribute( element, args ) {
+		const { originalFieldId, originalFieldKey, newFieldId, newFieldKey } = args;
+
+		let regex         = new RegExp( originalFieldId, 'g' );
+		let elementString = element.outerHTML.replace( regex, newFieldId );
+		regex             = new RegExp( originalFieldKey, 'g' );
+		elementString     = elementString.replace( regex, newFieldKey );
+		const tempDiv     = document.createElement( 'div' );
```
truongwp, line 2248: Can we use a function in `js/admin/dom.js` to create an element?

js/formidable_admin.js
```diff
@@ -2143,13 +2143,114 @@ function frmAdminBuildJS() {
...
+			return;
+		}
+		newOptsContainer.innerHTML = '';
+
+		const newFieldKey      = newOptsContainer.dataset.key;
+		const originalFieldKey = origOptsContainer.dataset.key;
+
+		const args = { originalFieldId, originalFieldKey, newFieldId, newFieldKey };
+
+		copyUnsavedDeleteOptions( origOptsContainer, newOptsContainer, args );
```
truongwp, line 2176: Why don't we move `origOptsContainer` and `newOptsContainer` to the `args`?

AbdiTolesa: @truongwp Because I didn't want to pass objects we don't need to `copyUnsavedOptions` which also receives `args`.

truongwp (approved):

js/formidable_admin.js
```diff
@@ -2163,13 +2163,114 @@ function frmAdminBuildJS() {
...
+	 * Syncs the options in the form preview area.
+	 *
+	 * @param {object} args
+	 * @returns {void}
+	 */
+	function copyUnsavedOptions( args ) {
+		const { originalFieldId, newFieldId } = args;
+
+		const originalFieldOpts = document.getElementById( `field_${originalFieldId}_inner_container` ).querySelector( '.frm_opt_container' );
+		const newFieldOpts      = document.getElementById( `field_${newFieldId}_inner_container` ).querySelector( '.frm_opt_container' );
```
Crabcyborg, line 2248: Should we use `?` nullable checks here in case the inner containers happened to not exist?

js/formidable_admin.js
```diff
@@ -2163,13 +2163,114 @@ function frmAdminBuildJS() {
...
+			const newOpt = replaceElementAttribute( child, args );
+			newFieldOpts.append( newOpt );
+		}
+	}
+
+	function replaceElementAttribute( element, args ) {
+		const { originalFieldId, originalFieldKey, newFieldId, newFieldKey } = args;
+
+		let regex         = new RegExp( originalFieldId, 'g' );
+		let elementString = element.outerHTML.replace( regex, newFieldId );
```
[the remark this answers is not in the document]
AbdiTolesa: Yes I added `_` prefix and `"`s around the id/keys used in the pattern. That sounds safer to me now too, thank you @Crabcyborg!

js/formidable_admin.js
```diff
@@ -2163,13 +2163,116 @@ function frmAdminBuildJS() {
...
+		for ( const child of originalFieldOpts.children ) {
+			const newOpt = replaceElementAttribute( child, args );
+			newFieldOpts.append( newOpt );
+		}
+	}
+
+	function replaceElementAttribute( element, args ) {
+		const { originalFieldId, originalFieldKey, newFieldId, newFieldKey } = args;
+
+		let regex         = new RegExp( `_${originalFieldId}`, 'g' );
```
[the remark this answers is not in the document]
AbdiTolesa: @Crabcyborg We have string structures like those below in the field HTML, where `3930` is my old field id I want to replace:
```
	_3930-
	_3930"
	_3930]
	_3930_
```
They all start with `_old_field_id` but followed by different characters. I have now created a function that loops over the possible combinations and do regex replacements to avoid errors.

Crabcyborg:
I think I'd rather fix this by passing the target field's live data to the AJAX endpoint and rendering it already.

I don't think the current implementation would work very well with change events or any other logic where the UI changes based on a value.

PR #1494, opened 2024-01-29
Add aria-expanded to accordion h3 elements

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4611

### Test steps
1. Go to Visual Styler page and turn on your Assistive software.
2. Try tabbing through the accordion headers and confirm that the state of the currently tabbed accordion element is announced as you do so.

### DIFF
--- a/js/admin/style.js
+++ b/js/admin/style.js
@@ -1249,7 +1249,7 @@
 			el.setAttribute( 'role', 'button' );
 			el.addEventListener( 'click', ( event ) => {
 				event.stopPropagation();
-				frmAdminBuild.maybeCollapseSettings.call( el, 'open', true );
+				frmAdminBuild.maybeCollapseSettings.call( el, event, 'open', true );
 			});
 		});
 	}
--- a/js/admin/style.js
+++ b/js/admin/style.js
@@ -1248,21 +1248,23 @@
 			el.setAttribute( 'aria-expanded', el.parentElement.classList.contains( 'open' ) );
 			el.setAttribute( 'role', 'button' );
 			el.addEventListener( 'click', event => {
-				event.stopPropagation();
-				maybeCollapseSettings( el );
+				// event.stopPropagation();
+				updateAriaExpandedAtt( el );
 			});
 			el.addEventListener( 'keydown', event => {
 				if ( event.key === ' ' ) {
 					event.preventDefault();
-					maybeCollapseSettings( el );
+					updateAriaExpandedAtt( el );
 				}
 			});
 		});
 	}
 
-	function maybeCollapseSettings( target ) {
-		const expanded = target.parentElement.classList.toggle( 'open' );
-		target.setAttribute( 'aria-expanded', expanded );
+	/**
+	 * @param {HTMLElement} target
+	 */
+	function updateAriaExpandedAtt( target ) {
+		target.setAttribute( 'aria-expanded', target.parentElement.classList.contains( 'open' ) );
 	}
 
 	/**
--- a/js/admin/style.js
+++ b/js/admin/style.js
@@ -1260,7 +1260,7 @@
 	}
 
 	/**
-	 * @param {HTMLElement} target
+	 * @param {Event} event
 	 */
 	function maybeCollapseSettings( event ) {
 		let expanded;
--- a/js/admin/style.js
+++ b/js/admin/style.js
@@ -1248,13 +1248,12 @@
 			el.setAttribute( 'aria-expanded', el.parentElement.classList.contains( 'open' ) );
 			el.setAttribute( 'role', 'button' );
 			el.addEventListener( 'click', event => {
-				// event.stopPropagation();
-				updateAriaExpandedAtt( el );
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This looks and sounds great!

truongwp (approved):

js/formidable_admin.js
```diff
@@ -10346,6 +10346,14 @@ function frmAdminBuildJS() {
 			$previewWrapper.on( 'click', '.frm_remove_image_option', removeImageFromOption );
 
 			wp.hooks.doAction( 'frm_style_editor_init' );
+
+			document.querySelectorAll( '.styling_settings h3.accordion-section-title' ).forEach( el => {
```
Crabcyborg, line 10350: @AbdiTolesa This can go in the `style.js` script instead, right?

The `initEditPage` function is triggered on the `frm_style_editor_init` hook that is called here.

That way we don't need to add more code to this big `formidable_admin.js` file wither over 10k lines of code.

js/admin/style.js
```diff
@@ -1243,6 +1243,14 @@
 
 			radio.closest( '.dropdown-item' ).classList.add( 'active' );
 		});
+
+		document.querySelectorAll( '.styling_settings h3.accordion-section-title' ).forEach( el => {
+			el.setAttribute( 'aria-expanded', el.parentElement.classList.contains( 'open' ) );
+			el.setAttribute( 'role', 'button' );
+			el.addEventListener( 'click', ( event ) => {
+				el.setAttribute( 'aria-expanded', event.target.parentElement.classList.contains( 'open' ) );
```
Crabcyborg, line 1251: Can we refactor our `formidable_admin.js` code a bit to avoid duplicate code?

We have this code for the builder,
```
$builderForm.on( 'click', '.frm-single-settings h3', maybeCollapseSettings );
		$builderForm.on( 'keydown', '.frm-single-settings h3', function( event ) {
		// If so, only proceed if the key pressed was 'Enter' or 'Space'
		if ( event.key === 'Enter' || event.key === ' ' ) {
			event.preventDefault();
			maybeCollapseSettings.call( this, event );
		}
});
```

It would be nice to make a shared function that both of these used, so we are consistent and aren't rewriting the same thing everywhere.

js/formidable_admin.js
```diff
@@ -5238,9 +5238,10 @@ function frmAdminBuildJS() {
...
 
-	function maybeCollapseSettings() {
+	function maybeCollapseSettings( event, toggleClass = 'frm-collapsed', toggleParentClass = false ) {
+		const toggleElement = toggleParentClass ? this.parentElement : this;
 		/*jshint validthis:true */
-		this.classList.toggle( 'frm-collapsed' );
+		toggleElement.classList.toggle( toggleClass );
 
 		// Toggles the "aria-expanded" attribute
 		let expanded = this.getAttribute( 'aria-expanded' ) === 'true' || false;
```
truongwp, line 5247: @AbdiTolesa As in a similar PR, we can declare `expanded` variable based on the `classList.toggle()` instead of `getAttribute()`.

truongwp (approved):
This looks good again. Thanks @AbdiTolesa!

js/admin/style.js
Crabcyborg: This example code here uses the `event.preventDefault();` on the space key down event to avoid a scroll event on the page.

Right now, if I open/close a styler section using the spacebar, the list scrolls past the section I'm trying to open, which isn't very accessible.

Crabcyborg: @AbdiTolesa It doesn't look like this has been addressed yet.

AbdiTolesa: @Crabcyborg I'm sorry I didn't notice that but already updated my PR to add the spacebar event.

Crabcyborg (commented):
This seems to be backwards. When my section is open, it's says it's collapsed and when it's closed it's saying it's expanded.

The value of `aria-expanded` is always wrong at the moment.

js/formidable_admin.js
```diff
@@ -5238,13 +5238,13 @@ function frmAdminBuildJS() {
 		parentCont.classList.toggle( 'frm-section-collapsed' );
 	}
 
-	function maybeCollapseSettings() {
+	function maybeCollapseSettings( event, toggleClass = 'frm-collapsed', toggleParentClass = false ) {
```
Crabcyborg, line 5241: I'm not sure how much it's worth trying to re-use `maybeCollapseSettings` at this point. It looks like there's too much modification that needs to be done here. I'd rather not complicate this function. Having it as a 0 param function and now changing it accepting 3 params seems like way too much complexity added.

I think we'd be better off not making this function global and just making a new function in the styler, or updating the markup so it's the same as other collapsible sections.

This just seems too messy.

Crabcyborg (approved):

I think this could probably still be a bit more simple, but this is good enough.

🚀

Thank you @truongwp and @garretlaxton for your help reviewing this!

PR #1496, opened 2024-01-31
Add message with Email summary settings

stephywells:
- Show a message with link to disable summary emails in profile for Elite and Business users (licenses that could have lots of sites)
- Changed setting name from "Recipients" to "Summary email recipients" for clarity
- Check white labeling based on icon instead of name. For a while, the menu default was "Forms" so this is excluding too many sites. Also, don't check if pro is inactive (Part of https://github.com/Strategy11/formidable-pro/issues/4687 )

Screenshot 2024-01-31 at 2 06 30 PM

### DIFF
--- a/classes/helpers/FrmEmailSummaryHelper.php
+++ b/classes/helpers/FrmEmailSummaryHelper.php
@@ -531,28 +531,14 @@ class FrmEmailSummaryHelper {
 	 * @param array $recipients Recipients.
 	 */
 	public static function maybe_remove_recipients_from_api( &$recipients ) {
-		$skip_emails = self::api_emails_to_skip();
-		if ( empty( $skip_emails ) ) {
-			return;
-		}
-		$recipients = array_diff( $recipients, $skip_emails );
-	}
-
-	/**
-	 * Checks if API email summary is on.
-	 *
-	 * @since x.x
-	 *
-	 * @return array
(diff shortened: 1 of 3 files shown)


classes/helpers/FrmEmailSummaryHelper.php
```diff
@@ -531,14 +531,28 @@ public static function get_date_from_today( $date_diff = '' ) {
...
+	}
+
+	/**
+	 * Checks if API email summary is on.
+	 *
+	 * @since x.x
+	 *
+	 * @return array
+	 */
+	public static function api_emails_to_skip() {
```
truongwp, line 548: This method should be private since I don't see it's used outside this class.

stephywells: Good catch! I was using it at first, but didn't think about changing it when I stopped using this. Guess I didn't need to make this change at all....

stephywells: I went ahead and reverted that change since I wasn't using the new function.

truongwp (approved):
This looks good! Thanks @stephywells for working on this.

Crabcyborg (approved):
This looks good.

Thanks @stephywells and @truongwp for working on this!

🚀

PR #1499, opened 2024-02-07
Check if id is defined before invoking a method on it

AbdiTolesa:
I updated Quiz score global settings page to use frm_font_icon instead of the +/- <i> tags in https://github.com/Strategy11/formidable-quizzes/pull/194 and this error came up when clicking the delete row icon.

image

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -707,7 +707,7 @@ function frmAdminBuildJS() {
 				show = '';
 				hide = '';
 			}
-		} else if ( id.indexOf( 'frm_postmeta_' ) === 0 ) {
+		} else if ( id && id.indexOf( 'frm_postmeta_' ) === 0 ) {
 			if ( jQuery( '#frm_postmeta_rows .frm_postmeta_row' ).length < 2 ) {
 				show = '.frm_add_postmeta_row.button';
 			}


Crabcyborg:
@AbdiTolesa I don't think we should fix this by adding this check.

I think we should try to update quizzes so it works more consistently with other remove tag triggers.

I think we should define a `data-removeid` attribute so it isn't null.

And we should make sure that the value of this attribute is the id of the row that we're removing. It should fade out like the other rows do.

We should be able to remove the `el.parent().remove();` code from the quizzes add-on and just let Lite handle it.

AbdiTolesa:
> I think we should define a `data-removeid` attribute so it isn't null.

@crabcyborg I don't we can do that since we are referencing a pseudo-element here and it can't have a data attribute.

Crabcyborg:
@AbdiTolesa We need to fix that too. No elements should ever have the same ID. That's a bug.

AbdiTolesa:
> @AbdiTolesa We need to fix that too. No elements should ever have the same ID. That's a bug.

Noted, @Crabcyborg! I have created a new issue to track that and working on it, https://github.com/Strategy11/formidable-pro/issues/4796.

Crabcyborg:
I'm going to close this PR as updating Lite as part of this would cause backward incompatibility issues as well as inconsistency issues. If we aren't using these attributes the way they're meant to be we shouldn't be using them.

PR #1501, opened 2024-02-07
Check entry->updated_by is set fix

Crabcyborg:
Fixes https://secure.helpscout.net/conversation/2503809609/189379/

> Warning: undefined property stdClass::$updated_by

I'm not sure how the entry doesn't have an `updated_by`, but adding a check doesn't hurt.

### DIFF
--- a/classes/models/fields/FrmFieldType.php
+++ b/classes/models/fields/FrmFieldType.php
@@ -1420,7 +1420,8 @@ DEFAULT_HTML;
 	 * @return bool
 	 */
 	protected function should_strip_most_html( $entry ) {
-		if ( $entry->updated_by && $this->user_id_is_privileged( $entry->updated_by ) ) {
+		// $entry->updated_by may not always be set.
+		if ( ! empty( $entry->updated_by ) && $this->user_id_is_privileged( $entry->updated_by ) ) {
 			return false;
 		}
 		if ( $entry->user_id && $this->user_id_is_privileged( $entry->user_id ) ) {


classes/models/fields/FrmFieldType.php
```diff
@@ -1420,7 +1420,8 @@ protected function should_strip_most_html_before_preparing_display_value( $atts
 	 * @return bool
 	 */
 	protected function should_strip_most_html( $entry ) {
-		if ( $entry->updated_by && $this->user_id_is_privileged( $entry->updated_by ) ) {
+		// $entry->updated_by may not always be set.
+		if ( ! empty( $entry->updated_by ) && $this->user_id_is_privileged( $entry->updated_by ) ) {
```
stephywells, line 1424: @Crabcyborg do you think it would be helpful to set this in FrmEntry::getOne?

Crabcyborg: Thanks @stephywells.

That prompted me to look further into this.

The issue isn't `FrmEntry::getOne` but `FrmProSummaryValues::base_entry`. `user_id` isn't set either.

I made a PR in Pro to fix this https://github.com/Strategy11/formidable-pro/pull/4789
As well as another PR in Lite to add the other check. This way it's fixed on both sides https://github.com/Strategy11/formidable-forms/pull/1503

stephywells: Thanks for digging into this to get to the source! Awesome!

PR #1504, opened 2024-02-09
Make form API error handling support string errors

Crabcyborg:
**Related tickets**
https://secure.helpscout.net/conversation/2506175920/189635
https://secure.helpscout.net/conversation/2506108172/189628/

It looks like an error for a nulled license works differently.

Instead of the expected `$addons['error']` array, it's a string like "Your license has been blocked!".

The "Y" shows up in PHP 7.4.
image

Add more power to your forms, and bring

In PHP 8 it throws a fatal error instead,
> Uncaught TypeError: Cannot access offset of type string on string in .../formidable/classes/models/FrmFormApi.php:261
> Stack trace:
> #0 .../formidable/classes/models/FrmFormApi.php(245): FrmFormApi->get_error_from_response()
> #1 .../formidable/classes/models/FrmAddon.php(327): FrmFormApi->error_for_license()
> #2 .../formidable/classes/models/FrmAddon.php(317): FrmAddon->show_license_message()
> #3 .../class-wp-hook.php(326): FrmAddon->maybe_show_license_message()

The error seems to be causing issues with disconnecting a nulled license. I couldn't figure out where this is being defined though. When I searched the error message, I couldn't find where this message was actually coming from.

Pre-release,
[formidable-6.8.1b.zip](https://github.com/Strategy11/formidable-forms/files/14225789/formidable-6.8.1b.zip)

### DIFF
--- a/classes/models/FrmAddon.php
+++ b/classes/models/FrmAddon.php
@@ -230,10 +230,10 @@ class FrmAddon {
 	/**
 	 * @since 3.04.03
 	 *
-	 * @param array $error
+	 * @param array|string $error
 	 */
 	public function maybe_clear_license( $error ) {
-		if ( $error['code'] === 'disabled' && $error['license'] === $this->license ) {
+		if ( is_array( $error ) && $error['code'] === 'disabled' && $error['license'] === $this->license ) {
 			$this->clear_license();
 		}
 	}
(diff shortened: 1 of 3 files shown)


stephywells (approved):
This should be fixed from the Cloudflare end, but we could still merge this if you'd like. It's a nice simple fix.

PR #1505, opened 2024-02-13
Add Missing Animation to Field Group Message Element

shervElmi:
This PR addresses the omission of animation for the Field Group Message Element. Instead of using JavaScript for enqueuing the necessary CSS for animations, we have transitioned to a PHP-based approach for improved performance and consistency.

## Related Issue:
[Missing Animation for Field Group Message Element After Recent Dashboard UI Update #4758](https://github.com/Strategy11/formidable-pro/issues/4758)

## Testing Instructions:
To ensure the animation has been successfully integrated and functions as expected, please follow these steps:

1. **Form Builder:**
   - Open the form builder.
   - Select any field to trigger the bottom banner.
   - Confirm that the bottom banner appears with the intended animation.

2. **General Animation Check:**
   - Navigate to the Dashboard and Form Templates pages, and also check the Overlay element.
   - Observe the UI to confirm that animations are present and functioning correctly.


shervElmi:
@Crabcyborg, I noticed that the "🚀 Deploy website on push / 🎉 Deploy (pull_request)" action has failed. Just wanted to bring it to your attention in case you have a solution for this issue.

Crabcyborg:
@shervElmi I don't. Our hosting has changed so all of the deploy workflows have broken.

@stephywells Do we have a new replacement for these breaking deploy workflows?

stephywells:
Yep! The deploy and delete workflows here are updated, and can be copied into other plugins:
https://github.com/Strategy11/formidable-abandonment/blob/main/.github/workflows/qadeploy.yml

garretlaxton (approved):
This looks great!

js/form-templates.js
truongwp, line 1: This file isn't minified. Can you fix it?

shervElmi: Thanks for the review, @truongwp. Actually, it's not minified because of the related `.map.js` file that Webpack is generating.

truongwp: I tried running `npm run build` then these files are minified. Did you use that script to build these files?

shervElmi: I commit the development version of auto-generated files, and when we create a package, they are minified by default, similar to `npm run build`. In pull requests, I commit the `npm run dev` generated files so that we can debug them.

truongwp: The dev version of generated files should be used when we are debugging. When the code is ready, we should minify it, as what we deliver to users, so we can discover any issues with the minified files if have any.

shervElmi: Thank you for sharing your thoughts on this, @truongwp. I get your point! 👍

@Crabcyborg, I believe committing the dev version of generated files could be beneficial for error checking on the QA sites. However, Truong holds a differing view on this subject. I'm not entirely certain, but it seems to me that Steph has previously committed the dev version of generated files. What are your thoughts on this matter?

shervElmi:
> Yep! The deploy and delete workflows here are updated, and can be copied into other plugins: https://github.com/Strategy11/formidable-abandonment/blob/main/.github/workflows/qadeploy.yml

@Crabcyborg, could you please check the abandonment `qadeploy.yml` file and let me know what needs to be updated for the Onboarding Wizard feature?

shervElmi:
Yes, I do. Thank you so much, @Crabcyborg!

js/form-templates.js
Crabcyborg: @shervElmi It does make the diff really big. It would be nicer for packing a release if this wasn't such a big diff.

Since Garret had already approved it I think it makes sense to use the minified version. We probably only need to use the dev version _**after**_ someone reports a JS console error and we can't trace it back. This usually shouldn't be necessary.

Crabcyborg (approved):
Thanks for updating this @shervElmi!

🚀

PR #1507, opened 2024-02-14
Update logic row html for element ids and names if duplicate id is found

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4796

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -7238,6 +7238,12 @@ function frmAdminBuildJS() {
 				nonce: frmGlobal.nonce
 			},
 			success: function( html ) {
+				let logicRowContainerID = jQuery( html ).attr( 'id' );
+				while ( document.getElementById( logicRowContainerID ) ) {
+					html = updateLogicRowHTML( html, id, logicRowContainerID );
+					logicRowContainerID = jQuery( html ).attr( 'id' )
+				}
+
 				jQuery( document.getElementById( 'logic_link_' + id ) ).fadeOut( 'slow', function() {
 					var $logicRow = jQuery( document.getElementById( 'frm_logic_row_' + id ) );
 					$logicRow.append( html );
@@ -7246,6 +7252,18 @@ function frmAdminBuildJS() {
 			}
 		});
 		return false;
+	}
+
+	function updateLogicRowHTML( html, id, newLogicRowID ) {
+		const logicRowNumberToReplace = parseInt( newLogicRowID.split( '_' ).pop() );
+		const newLogicRowNumber  = logicRowNumberToReplace + 1;
+
+		html = html
+		.replaceAll( `_${id}_${logicRowNumberToReplace}`, `_${id}_${newLogicRowNumber}` )
+		.replaceAll( `[${logicRowNumberToReplace}]`, `[${newLogicRowNumber}]`)
+		.replaceAll( ",'" + logicRowNumberToReplace + "',", ",'" + newLogicRowNumber + "',");
+
+		return html;
 	}
 
 	function toggleSubmitLogic() {


garretlaxton (approved):
This fix works great!

js/formidable_admin.js
```diff
@@ -7248,6 +7254,28 @@ function frmAdminBuildJS() {
...
+	 * @param {String} newLogicRowID
+	 * @returns {String}
+	 */
+	function updateLogicRowHTML( html, id, newLogicRowID ) {
+		const logicRowNumberToReplace = parseInt( newLogicRowID.split( '_' ).pop() );
+		const newLogicRowNumber       = logicRowNumberToReplace + 1;
+
+		html = html
+		.replaceAll( `_${id}_${logicRowNumberToReplace}`, `_${id}_${newLogicRowNumber}` )
+		.replaceAll( `[${logicRowNumberToReplace}]`, `[${newLogicRowNumber}]` )
```
truongwp, line 7273: This replacement could be wrong if the action ID is a small number like `frm_email_action[3][post_content][conditions][3][hide_opt]`

js/formidable_admin.js
```diff
@@ -7248,6 +7254,28 @@ function frmAdminBuildJS() {
...
+		.replaceAll( ',\'' + logicRowNumberToReplace + ',\'', ',\''  + newLogicRowNumber + ',\'' );
```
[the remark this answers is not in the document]
AbdiTolesa: It would trigger eslint error if we do that, @truongwp.

js/formidable_admin.js
AbdiTolesa: Thanks, @truongwp! That is addressed now.

js/formidable_admin.js
```diff
@@ -7248,6 +7254,28 @@ function frmAdminBuildJS() {
...
+	 * @param {String} newLogicRowID
+	 * @returns {String}
+	 */
+	function updateLogicRowHTML( html, id, newLogicRowID ) {
+		const logicRowNumberToReplace = parseInt( newLogicRowID.split( '_' ).pop() );
+		const newLogicRowNumber       = logicRowNumberToReplace + 1;
+
+		html = html
+		.replaceAll( `_${id}_${logicRowNumberToReplace}`, `_${id}_${newLogicRowNumber}` )
+		.replaceAll( `[${logicRowNumberToReplace}][hide_opt]`, `[${newLogicRowNumber}][hide_opt]` )
```
truongwp, line 7273: Not only `hide_opt`, we have `frm_email_action[585][post_content][conditions][0][hide_field_cond]` too. I think we should replace `[conditions][${number}]`.

js/formidable_admin.js
```diff
@@ -7248,6 +7254,28 @@ function frmAdminBuildJS() {
...
+		.replaceAll( ',\'' + logicRowNumberToReplace + ',\'', ',\''  + newLogicRowNumber + ',\'' );
```
truongwp, line 7274: This replacement could be wrong with a low ID, like the above replacement.

truongwp (approved):

Crabcyborg (commented):
@AbdiTolesa Is it possible to get more of this logic into Pro?

Ideally we wouldn't add more Pro logic into Lite.

We could add a new `frm_after_add_form_logic_row` action and then listen for that hook in `builder.js` in Pro.

AbdiTolesa:
> @AbdiTolesa Is it possible to get more of this logic into Pro?
> 
> Ideally we wouldn't add more Pro logic into Lite.
> 
> We could add a new `frm_after_add_form_logic_row` action and then listen for that hook in `builder.js` in Pro.

@Crabcyborg I believe you meant `settings.js` in pro as builder isn't loaded in the form settings page and I have moved the fix there. https://github.com/Strategy11/formidable-pro/pull/4816

js/formidable_admin.js
```diff
@@ -7238,6 +7238,8 @@ function frmAdminBuildJS() {
 				nonce: frmGlobal.nonce
 			},
 			success: function( html ) {
+				html = wp.hooks.applyFilters( 'frm_after_add_form_logic_row', html, id );
```
Crabcyborg, line 7241: If this is a filter, I think the hook could use a more descriptive name.

Mabye `frm_form_logic_row_html` instead?

AbdiTolesa: Makes sense!

PR #1510, opened 2024-02-14
Fix: parameters getting stripped from connect flow

stephywells:
Remove the old v1 connect js since it was unused

Firefox is removing all paramters from the URL if any utms are included. This removed the utms so the link will work.

### DIFF
--- a/classes/controllers/FrmAddonsController.php
+++ b/classes/controllers/FrmAddonsController.php
@@ -1284,7 +1284,7 @@ class FrmAddonsController {
 			$auth = hash( 'sha512', wp_rand() );
 			update_option( 'frm_connect_token', $auth, 'no' );
 		}
-		$link = FrmAppHelper::admin_upgrade_link( 'connect', 'api-connect' );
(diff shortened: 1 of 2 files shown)


shervElmi (approved):
Thank you for your work on this issue, @stephywells! I've tested it on my QA site, and it works.

Crabcyborg (approved):
I just uploaded a Formidable Lite beta to QA and tested it there, and it worked for me, for the first time ever 😅.

I'm used to only seeing error messages on this page.

Screen Shot 2024-02-14 at 4 54 22 PM

It's also exciting to see a ton of code getting removed 🥳.

🚀

stephywells:
Yay!! That's great news!!

Crabcyborg:
@stephywells Since this didn't make it into today's release, do you think it's worth making another Lite release tomorrow so this gets out too?

stephywells:
Well... Looks like another issue came up, so this would be great to include in the release.

PR #1511, opened 2024-02-15
Include term_parent in exported xml

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/799

### Test steps
1. Create a form with and configure a 'Create Post' action in such a way that Category taxonomy is mapped to a field in the form.
2. Preview the form and select Category that has a Parent category.
3. Export the form xml along with entries.
4. Open the export xml file and check that the parent category is found like `<term_parent>parent_id</term_parent>`

### DIFF
--- a/classes/controllers/FrmXMLController.php
+++ b/classes/controllers/FrmXMLController.php
@@ -560,7 +560,7 @@ class FrmXMLController {
 	 * @return array
 	 */
 	public static function get_parent_terms_slugs( $terms ) {
-		$parent_term_ids = array_filter( wp_list_pluck( $terms, 'parent' ) );
+		$parent_term_ids = array_filter( array_unique( wp_list_pluck( $terms, 'parent' ) ) );
 		$parent_slugs    = array();
 
 		if ( ! $parent_term_ids ) {
--- a/classes/controllers/FrmXMLController.php
+++ b/classes/controllers/FrmXMLController.php
@@ -553,6 +553,28 @@ class FrmXMLController {
 	}
 
 	/**
+	 * Returns an array that has parent term slugs for the terms provided.
+	 *
+	 * @since x.x
+	 * @param array $terms
+	 * @return array
+	 */
+	public function get_parent_terms_slugs( $terms ) {
+		$parent_term_ids = implode( ',', array_filter( wp_list_pluck( $terms, 'parent' ) ) );
+		$parent_slugs    = array();
+
+		if ( $parent_term_ids ) {
+			global $wpdb;
+			$results = $wpdb->get_results( $wpdb->prepare( "SELECT term_id, slug FROM {$wpdb->prefix}terms WHERE term_id IN (%s)", $parent_term_ids ) );
+			foreach ( $results as $result ) {
+				$parent_slugs[ $result->term_id ] = $result->slug;
+			}
+		}
+
+		return $parent_slugs;
+	}
+
+	/**
 	 * @return void
 	 */
 	private static function prepare_types_array( &$type ) {
--- a/classes/controllers/FrmXMLController.php
+++ b/classes/controllers/FrmXMLController.php
@@ -567,7 +567,7 @@ class FrmXMLController {
 			return $parent_slugs;
 		}
 
-		$results = FrmDb::get_results( 'terms', array( 'term_id' => $parent_term_ids  ), 'term_id, slug' );
+		$results = FrmDb::get_results( 'terms', array( 'term_id' => $parent_term_ids ), 'term_id, slug' );
 		foreach ( $results as $result ) {
 			$parent_slugs[ $result->term_id ] = $result->slug;
 		}
--- a/classes/controllers/FrmXMLController.php
+++ b/classes/controllers/FrmXMLController.php
@@ -563,12 +563,14 @@ class FrmXMLController {
 		$parent_term_ids = implode( ',', array_filter( wp_list_pluck( $terms, 'parent' ) ) );
 		$parent_slugs    = array();
 
-		if ( $parent_term_ids ) {
-			global $wpdb;
-			$results = $wpdb->get_results( $wpdb->prepare( "SELECT term_id, slug FROM {$wpdb->prefix}terms WHERE term_id IN (%s)", $parent_term_ids ) );
-			foreach ( $results as $result ) {
-				$parent_slugs[ $result->term_id ] = $result->slug;
-			}
+		if ( ! $parent_term_ids ) {
+			return $parent_slugs;
+		}
+
+		global $wpdb;
+		$results = $wpdb->get_results( $wpdb->prepare( "SELECT term_id, slug FROM {$wpdb->prefix}terms WHERE term_id IN (%s)", $parent_term_ids ) );
+		foreach ( $results as $result ) {
+			$parent_slugs[ $result->term_id ] = $result->slug;
 		}
 
 		return $parent_slugs;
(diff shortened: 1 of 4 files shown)


classes/views/xml/posts_xml.php
```diff
@@ -109,8 +109,10 @@
 
 	$frm_inc_tax[] = $term->term_id;
 	$label = ( 'category' === $term->taxonomy || 'tag' === $term->taxonomy ) ? $term->taxonomy : 'term';
+	$term_parent = ! empty( $term->parent ) ? '<term_parent>' . esc_html( $term->parent ) . '</term_parent>' : '';
```
truongwp, line 112: I think we should use `if` to check then print the `<term_parent>` tag, like we do with the term name and description below. So we can remove a `phpcs:ignore`.

AbdiTolesa: @truongwp Both of those lines have `phpcs:ignore` as well, I think you haven't noticed that.

truongwp: Term parent is int here, we don't need to use `CDATA`. It's fine to use `esc_html`, so we can remove the `phpcs:ignore`.

AbdiTolesa: I see, sorry for being confused!

truongwp (approved):
This looks good! Thanks @AbdiTolesa.

There is only one small thing. Probably we want no space after `</term_taxonomy>`. After exporting, each term will be on one line. I'm not sure if this is what we want. @Crabcyborg Please take a look at this.

AbdiTolesa:
> Thanks @truongwp. Yeah, I think we should try to keep it to a single line like it was before.
> 
> @AbdiTolesa We'll need to move the `<?php` tag back to where it was. If it's on the next line, it adds spaces to the file. It looks better if keep it where it was.

@Crabcyborg That is moved back now.

Crabcyborg:
Thanks @AbdiTolesa!

Does this import back as expected? Have you tried that?

AbdiTolesa:
> Thanks @AbdiTolesa!
> 
> Does this import back as expected? Have you tried that?
That part was overlooked, I was just positive it won't have an issue looking at the code but small casting to integer commit fixed it easily.

AbdiTolesa:
> > Thanks @AbdiTolesa!
> > Does this import back as expected? Have you tried that?
@Crabcyborg That part was overlooked, I only checked the imported posts are assigned the respective taxonomy terms and I was just positive it won't have another issue looking at the code but small casting to integer commit fixed it easily.

classes/helpers/FrmXMLHelper.php
```diff
@@ -192,7 +192,7 @@ public static function import_xml_terms( $terms, $imported ) {
 	private static function get_term_parent_id( $t ) {
 		$parent = (string) $t->term_parent;
 		if ( ! empty( $parent ) ) {
-			$parent = term_exists( (string) $t->term_parent, (string) $t->term_taxonomy );
+			$parent = term_exists( (int) $t->term_parent, (string) $t->term_taxonomy );
```
Crabcyborg, line 195: @AbdiTolesa It actually looks like we want to pass a string.

Steph's example shows "parent-slug".
```
<term_parent>parent-slug</term_parent>
```

Let's try to show the term parent slug instead of the ID.

AbdiTolesa: @Crabcyborg Yeah but getting `parent_slug` requires a db query and I thought we can avoid that using the parent ID instead. I believe this could be significant improvement particularly when there are many terms being exported. What do you think?

Crabcyborg: I think it's worth getting it.

Maybe we could loop through the posts ahead of time and get all of the term parent IDs so we could get them all with one database request?

classes/controllers/FrmXMLController.php
```diff
@@ -552,6 +552,30 @@ public static function generate_xml( $type, $args = array() ) {
...
+	public static function get_parent_terms_slugs( $terms ) {
+		$parent_term_ids = implode( ',', array_filter( wp_list_pluck( $terms, 'parent' ) ) );
+		$parent_slugs    = array();
+
+		if ( ! $parent_term_ids ) {
+			return $parent_slugs;
+		}
+
+		global $wpdb;
+		$results = $wpdb->get_results( $wpdb->prepare( "SELECT term_id, slug FROM {$wpdb->prefix}terms WHERE term_id IN (%s)", $parent_term_ids ) );
```
Crabcyborg, line 571: Can this be done using `FrmDb` methods?

Like
```
$results = FrmDb::get_results( 'terms', array(.'term_id' => $parent_term_ids  ), 'term_id, slug' );
```

Where `$parent_iterm_ids` is an array and not a CSV?

classes/controllers/FrmXMLController.php
```diff
@@ -552,6 +552,29 @@ public static function generate_xml( $type, $args = array() ) {
...
 
+	/**
+	 * Returns an array that has parent term slugs for the terms provided.
+	 *
+	 * @since x.x
+	 * @param array $terms
+	 * @return array
+	 */
+	public static function get_parent_terms_slugs( $terms ) {
+		$parent_term_ids = implode( ',', array_filter( wp_list_pluck( $terms, 'parent' ) ) );
```
Crabcyborg, line 563: It looks like you're still imploding. We want to pass `$parent_term_ids` as an array.

classes/controllers/FrmXMLController.php
```diff
@@ -552,6 +552,29 @@ public static function generate_xml( $type, $args = array() ) {
...
+		$parent_slugs    = array();
+
+		if ( ! $parent_term_ids ) {
+			return $parent_slugs;
+		}
+
+		$results = FrmDb::get_results( 'terms', array( 'term_id' => $parent_term_ids ), 'term_id, slug' );
+		foreach ( $results as $result ) {
```
Crabcyborg, line 571: Instead of this loop we could just use `wp_list_pluck`.

https://developer.wordpress.org/reference/functions/wp_list_pluck/

phpcs.xml
```diff
@@ -25,7 +25,7 @@
 		<exclude name="Squiz.Commenting.FunctionComment.MissingParamTag" />
 		<exclude name="Squiz.Commenting.InlineComment.InvalidEndChar" />
 		<exclude name="Generic.Commenting.DocComment.MissingShort" />
-
+		<exclude name="Squiz.Commenting.LongConditionClosingComment.Missing" />
```
Crabcyborg, line 28: @AbdiTolesa We've been adding these. Instead of adding an exception, you should add an `//end if` to the end of any really long if condition.

classes/controllers/FrmXMLController.php
```diff
@@ -552,6 +552,27 @@ public static function generate_xml( $type, $args = array() ) {
...
 
+	/**
+	 * Returns an array that has parent term slugs for the terms provided.
+	 *
+	 * @since x.x
+	 * @param array $terms
+	 * @return array
+	 */
+	public static function get_parent_terms_slugs( $terms ) {
+		$parent_term_ids = array_filter( wp_list_pluck( $terms, 'parent' ) );
```
Crabcyborg, line 563: Maybe we should use `array_unique` here as well since there could be duplicate IDs.

Crabcyborg (approved):
Thanks @AbdiTolesa for updating that!

This looks good now.

@garretlaxton Would you be able to test this once more? It's gone through some changes. It would export the ID before, but now it exported the slug. This is required for the impot to work as it was intended.

Can you test with exporting and importing back?

Thank you!

garretlaxton (approved):
This looks good still!

Crabcyborg:
Thanks Garret!

PR #1513, opened 2024-02-15
Pro issue 4776 / Add support for a new value arg in show search box f…

Crabcyborg:
…unction

Fixes https://github.com/Strategy11/formidable-pro/issues/4776

This update adds support for a `value` arg. If it is a string (rather than the `false` default value), it is used as the `value` param for the input.

Before it always used `_admin_search_query`.

The body of that function is simple:
```
echo isset( $_REQUEST['s'] ) ? esc_attr( wp_unslash( $_REQUEST['s'] ) ) : '';
```

I just copied that as the fallback value when there is no value arg set.

I changed this to use `FrmAppHelper::array_to_html_params` to echo out all of the input params as well.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -1338,6 +1338,7 @@ class FrmAppHelper {
 			'tosearch'    => '',
 			'text'        => __( 'Search', 'formidable' ),
 			'input_id'    => '',
+			'value'       => false,
 		);
 		$atts = array_merge( $defaults, $atts );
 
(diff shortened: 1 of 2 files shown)


classes/helpers/FrmAppHelper.php
```diff
@@ -1352,19 +1353,26 @@ public static function show_search_box( $atts ) {
 
 		$input_id = $atts['input_id'] . '-search-input';
 
+		$input_atts = array(
+			'type'          => 'search',
+			'id'            => $input_id,
+			'name'          => 's',
+			'value'         => is_string( $atts['value'] ) ? $atts['value'] : ( isset( $_REQUEST['s'] ) ? wp_unslash( $_REQUEST['s'] ) : '' ),
```
stephywells, line 1360: Would it be worth using the FrmAppHelper::get_simple_request function here and sanitizing the value?

stephywells (commented):
Great solution!

classes/helpers/FrmAppHelper.php
Crabcyborg: I considered that but it uses sanitize functions. This doesn't. I'm concerned if we sanitized it we would break the search results as well.

stephywells: Gotcha. Sounds good.

stephywells (approved):

PR #1514, opened 2024-02-16
Stop showing "All Selected" text in Bootstrap multiselect dropdowns

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4810

Bootstrap multiselect by default shows "All selected" when all items are selected.

This isn't very useful when there are only a couple options. With this update, I'm removing the "All selected text", so it never shows.

**Before**
Screen Shot 2024-02-16 at 9 40 01 AM

**After**
Screen Shot 2024-02-16 at 9 39 41 AM

If you select a lot of options, it starts to count them. IF you select all of them, after this update, instead of "All selected", it will just continue to show a count.
Screen Shot 2024-02-16 at 9 41 47 AM

**I'm not sure if we want to make this more dynamic?**

It starts to count items elected when the number of options selected exceeds `numberDisplayed`. The default value for `numberDisplayed` is 3.

We could consider bumping this number as well. For example, if I bump it to 10, it looks like this,
Screen Shot 2024-02-16 at 9 44 23 AM

It might be that we want to still show "All selected" if there are more than 3 options, since "All selected" may be nicer than "11 selected".

### DIFF
--- a/js/admin/dom.js
+++ b/js/admin/dom.js
@@ -157,6 +157,8 @@
 				nonSelectedText: '',
 				// Prevent the dropdown from showing "All Selected" when every option is checked.
 				allSelectedText: '',
+				// This is 3 by default. We want to show more options before it starts showing a count.
+				numberDisplayed: 8,
 				onDropdownShown: function( event ) {
 					const action = jQuery( event.currentTarget.closest( '.frm_form_action_settings, #frm-show-fields' ) );
 					if ( action.length ) {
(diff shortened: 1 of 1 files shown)


stephywells (approved):
Thank you! This has been bugging my, but I hadn't looked into it.

I like the look of the truncated, comma-separated list in your last screenshot.

Crabcyborg:
Thanks Steph!

> I like the look of the truncated, comma-separated list in your last screenshot.

I went ahead and set a `numberDisplayed` value of 8 then. So now it only starts showing a count at 9+ items selected.

PR #1515, opened 2024-02-16
Use please select placeholder for bootstrap multiselect

Crabcyborg:
Quoting Steph from https://github.com/Strategy11/formidable-pro/pull/4630#pullrequestreview-1883987426
> It would also help to show -- Select -- in the dropdown if nothing is selected.

Before this was always an empty string so there were no placeholders.

I figure this is fine to add to all multiselect dropdowns and not just the one in the linked PR.

Screen Shot 2024-02-16 at 2 11 10 PM

### DIFF
--- a/js/admin/dom.js
+++ b/js/admin/dom.js
@@ -154,7 +154,7 @@
 					button: '<button type="button" class="multiselect dropdown-toggle btn" data-toggle="dropdown" ' + labelledBy + '><span class="multiselect-selected-text"></span> <b class="caret"></b></button>'
 				},
 				buttonContainer: '<div class="btn-group frm-btn-group dropdown" />',
-				nonSelectedText: '',
+				nonSelectedText: __( '— Please Select —', 'formidable' ),
 				onDropdownShown: function( event ) {
 					const action = jQuery( event.currentTarget.closest( '.frm_form_action_settings, #frm-show-fields' ) );
 					if ( action.length ) {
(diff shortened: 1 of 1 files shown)


stephywells:
@Crabcyborg do you see a way to remove line breaks from the title? Or remove the title?
Screenshot 2024-02-16 at 11 32 42 AM

js/admin/dom.js
```diff
@@ -154,7 +154,7 @@
 					button: '<button type="button" class="multiselect dropdown-toggle btn" data-toggle="dropdown" ' + labelledBy + '><span class="multiselect-selected-text"></span> <b class="caret"></b></button>'
 				},
 				buttonContainer: '<div class="btn-group frm-btn-group dropdown" />',
-				nonSelectedText: '',
+				nonSelectedText: __( '— Please Select —', 'formidable' ),
```
stephywells, line 157: Can we remove the "please" here? It would be nice if the existing strings would cover this for translations.

stephywells (changes requested):
Thanks for the update!

js/admin/dom.js
Crabcyborg: They won't because it needs to use the actual `—` character and not `&mdash;` like the others. I tried `&mdash;` but it shows that literally in the dropdown option instead.

Would you prefer to translate using `&mdash;`and then use a string replace to swap them to `—`

stephywells (approved):

PR #1516, opened 2024-02-16
Set empty option title attributes for bootstrap multiselect

Crabcyborg:
Related comment https://github.com/Strategy11/formidable-forms/pull/1515#issuecomment-1949084973

By defining a `" "` title value on my option elements, bootstrap-multiselect uses that as the `title` attribute for the options. When it's just a space character, the title never shows (tested in Firefox and Chrome).


stephywells (approved):
Yay!!

PR #1517, opened 2024-02-17
Reduce server load and page load times

stephywells:
- Set a 2 minute transient to avoid duplicate requests
- Increase the time between checking for new templates
- Use site options for multisite update checks so every site in the network is not checking if they use the same license
- Don't check if the site url changed and needs to be activated if the last try was not successful
- Don't cache a rate limited response for the normal time. Allow the next check sooner.
- Use utc times for api calls instead of adjusting for utc
- Fix expired banner and php errors on form templates page when site is rate limited

Part of https://github.com/Strategy11/formidable-pro/issues/4813 (covers checked boxes in linked issue)

How to test:
- Force a rate limit response. This will add a line to the error log each time an API request to ff.com is triggered. When rate limited, the site will check every 5 minutes.
```
add_filter(
	'pre_http_request',
	function( $pre, $parsed_args, $url ) {
		if ( strpos( $url, 'formidableforms.com' ) === false ) {
			return $pre;
		}
		error_log( 'http_request: ' . $url );

		return array(
			'response' => array( 'code' => 429, 'message' => 'Limited' ),
			'body'     => '{"error":{"type":"basic","code":"valid","expires":"","message":"Your site has been rate limited for 1 minute. Your forms will continue to function properly. If this message persists, please click the link to Check for a recent purchase on the Settings page","license":"xxx"},"expires":""}',
		);
	},
	10,
	3
);
```
- Click "check for recent purchases" and watch the log with this snippet to see how often api calls are being triggered:
```
add_filter(
	'pre_http_request',
	function( $pre, $parsed_args, $url ) {
		if ( strpos( $url, 'formidableforms.com' ) === false ) {
			return $pre;
		}
		error_log( 'http_request: ' . $url );

		return $pre;
	},
	10,
	3
);
```


classes/views/form-templates/template.php
```diff
@@ -9,6 +9,22 @@
...
+		<?php
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		echo FrmAppHelper::kses( $template['message'], array( 'a', 'b', 'strong', 'br' ) );
+		?>
+	</li>
+	<?php
+	return;
+}
+
+if ( ! is_array( $template ) ) {
```
Crabcyborg, line 24: Would it make sense to check this first rather than checking `
if ( ! empty( $template['message'] ) ) {` first?

It seems to make more sense that way since we're trying to access `$template` as an array before checking it is an array.

classes/models/FrmAddon.php
```diff
@@ -500,35 +491,57 @@ private function is_license_revoked() {
...
 			$last_checked = get_site_option( $this->transient_key() );
 		} else {
 			$last_checked = get_option( $this->transient_key() );
 		}
-		return $last_checked;
+		if ( $last_checked && ! is_array( $last_checked ) ) {
+			// Get string into array for existing values.
+			$last_checked = array( 'time' => $last_checked );
+		}
+		return (array) $last_checked;
```
Crabcyborg, line 533: If `$last_checked` is false, this will convert it to `array( false )` which would then validate as `true`, which could lead to issues.

I wonder if `$last_checked` is not an array, if we should just return an empty array instead?

Crabcyborg (approved):
I only left a few minor comments.

I think this looks good!

Would you want to try doing a release ahead of scheduled today to get this out sooner?

🥳

classes/models/FrmFormApi.php
```diff
@@ -185,26 +233,51 @@ public function get_pro_updater() {
...
 		}
 
-		$version     = FrmAppHelper::plugin_version();
-		$for_current = isset( $cache['version'] ) && $cache['version'] == $version;
-		if ( ! $for_current ) {
-			// Force a new check.
-			return false;
+		$values = json_decode( $cache['value'], true );
+		if ( isset( $values['response_code'] ) ) {
+			unset( $values['response_code'] );
```
Crabcyborg, line 262: @stephywells What is the reason for unsetting `$values['response_code']` from the cached value?

This causes issues for me.

I tried working on https://github.com/Strategy11/formidable-pro/issues/4825.

I think it makes sense to just not ever return "expired" for a rate limited response.

But if we're unsetting the response code, we have no way of knowing if the response was a 429 or not.

classes/models/FrmAddon.php
```diff
@@ -207,22 +216,6 @@ public function set_license( $license ) {
...
-	 * @since 2.04
-	 */
-	private function is_time_to_auto_activate() {
-		$last_try = get_option( $this->option_name . 'last_activate' );
-
-		return ( ! $last_try || $last_try < strtotime( '-1 day' ) );
-	}
-
-	private function set_auto_activate_time() {
-		update_option( $this->option_name . 'last_activate', time() );
```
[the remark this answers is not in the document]
stephywells: I was thinking about that, but didn't see a good way of finding them to delete them without a frequent check. Do you have an idea on how to check without sacrificing performance?

classes/models/FrmFormApi.php
stephywells: I wasn't sure if keeping it there would cause issues elsewhere with a new, unexpected value.

Your PR looks great though. I would prefer to only allow 429 now. I switched valid sites to get a 429.

classes/models/FrmAddon.php
stephywells: Good catch. Done.

classes/views/form-templates/template.php
stephywells: Definitely. Done.

classes/models/FrmFormApi.php
Crabcyborg: Cool. I had it at 429, but noticed that email summary issue mentions 403, so I added that as well. I can change it back to 429s only.

stephywells: Thank you!!

classes/models/FrmAddon.php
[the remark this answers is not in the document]
Crabcyborg: It looks like these are only used when you have a defined license constant.

So most people likely don't have these options defined.

If we add the constant check first, it won't have a performance impact of most sites at least.

I don't think it would hurt to just delete the options after the request happens, which isn't too often. That way we don't have to try looping every plugin all at once.

Something like,
```
$response = $this->activate_license( $license );

if ( $this->get_defined_license() ) {
    delete_option( $this->option_name . 'active' );
    delete_option( $this->option_name . 'key' );
}
```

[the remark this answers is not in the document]
stephywells: What do you think of this?
https://github.com/Strategy11/formidable-forms/pull/1517/commits/9925cdff6a2b5822700145594a019301e012e515

stephywells:
@Crabcyborg I just had a thought... Maybe we could prep a beta version and upload it on formidableforms.com for testing?

classes/models/FrmAddon.php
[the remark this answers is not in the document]
Crabcyborg: That would clear if when you delete your license, but otherwise it would never get cleaned up. I suspect most people don't usually disconnect their license.

classes/models/FrmAddon.php
```diff
@@ -785,4 +809,44 @@ public function manually_queue_update() {
...
+		return get_transient( $this->lock_key() );
+	}
+
+	/**
+	 * @since x.x
+	 *
+	 * @return void
+	 */
+	protected function set_running() {
+		set_transient( $this->lock_key(), true, 2 * MINUTE_IN_SECONDS );
```
Crabcyborg, line 841: Would it make sense to lock at a multisite level?

We might want to use `get_site_transient`/`set_site_transient` instead on multisite?

https://developer.wordpress.org/reference/functions/get_site_transient/
https://developer.wordpress.org/reference/functions/set_site_transient/

garretlaxton:
Where is `check for recent purchases` located?

Crabcyborg:
@garretlaxton That's a link in the bottom-right of the license settings on the Global Settings page.

Screen Shot 2024-02-26 at 12 11 09 PM

garretlaxton:
For the second one, I only see the api calls each time I click the button. Is that how it should work?

Crabcyborg:
Attached is a pre-release of this branch.

[formidable-beta.zip](https://github.com/Strategy11/formidable-forms/files/14419397/formidable-beta.zip)

Crabcyborg (approved):
This looks great! 🥳

classes/models/FrmAddon.php
stephywells: I just made a couple more updates to this. Now it will use multisite transient if there's no license key used for the request.

PR #1518, opened 2024-02-19
Stop stripping most HTML from HTML field display value fix

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2511714226/190280/

When showing a quiz score's correct answers, `FrmProEntriesController::show_entry_shortcode` is called.

When a user is not logged in, `FrmAppHelper::strip_most_html` is removing `img` tags from HTML field display.

This was introduced with https://github.com/Strategy11/formidable-forms/pull/1439

An HTML field doesn't really need any of the logic from `FrmFieldType::get_display_value`.

This skips some code, so `fill_default_atts`, `FrmAppHelper::strip_most_html`, `prepare_display_value`, and some checks for if `$value` is an array with checks for `$atts['show']`, are all skipped entirely.

HTML fields really don't need any of that. They should just display as is, since there isn't any user content directly defining its output.

### DIFF
--- a/classes/models/fields/FrmFieldHTML.php
+++ b/classes/models/fields/FrmFieldHTML.php
@@ -70,17 +70,16 @@ class FrmFieldHTML extends FrmFieldType {
 	}
 
 	/**
-	 * Overwrite FrmFieldType::get_display_value.
+	 * Overwrite FrmFieldType::should_strip_most_html_before_preparing_display_value.
 	 * An HTML field is not populated from user input.
 	 * It does not need to be passed through FrmAppHelper::strip_most_html.
 	 *
 	 * @since x.x
 	 *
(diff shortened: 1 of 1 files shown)


stephywells (commented):
Would it be better to set it back to how it was working before be adding the should_strip_most_html_before_preparing_display_value function here instead of overwriting get_display_value?

Crabcyborg:
Thanks @stephywells.

My thought was that none of the checks in `get_display_value` really applied to HTML fields and were just adding overhead.

I made that change though. I hadn't considered that and it works pretty well that way.

stephywells (approved):
Gotcha. I didn't look in detail into what the get_display_value does, so I'm good with whichever way you prefer.

PR #1520, opened 2024-02-20
Ranking Field

Liviu-p:
Related PR: 
https://github.com/Strategy11/formidable-surveys/pull/251


classes/views/frm-fields/single-option.php
```diff
@@ -12,7 +12,9 @@
 	}
 
 	?>
-	<input type="<?php echo esc_attr( $default_type ); ?>" name="<?php echo esc_attr( $field_name ); ?>" <?php echo ( isset( $checked ) && $checked ? 'checked="checked"' : '' ); ?> value="<?php echo esc_attr( $field_val ); ?>"/>
+	<?php if ( 'ranking' !== $default_type ) : ?>
```
Crabcyborg, line 15: What is the reason for not showing this input for ranking fields? It would be nice not to have the hard coded type check for another add-on here. Without any comment either, there's really no obvious explanation for his change.

css/_single_theme.css.php
```diff
@@ -234,6 +234,14 @@
 	box-shadow:0 1px 1px rgba(0, 0, 0, 0.075) inset, 0 0 8px rgba(<?php echo esc_html( FrmStylesHelper::hex2rgb( $border_color_active ) ); ?>, 0.6);
 	<?php } ?>
 }
+.<?php echo esc_html( $style_class ); ?> .frm-ranking-field-option:hover,
```
Crabcyborg, line 237: CSS for ranking fields should be in `frm-surveys.css`. I don't believe this should be in Lite.

css/admin/style.css
```diff
@@ -164,9 +164,10 @@ This file is added to the visual styler admin page, accessed from the style tab.
 
 	/*
 	When a background image is set the children in the fieldset have a z-index of 1.\
```
Crabcyborg, line 166: It looks like this comment has a stray `\` backslash. It might be good to remove that.

css/frm_admin.css
```diff
@@ -3436,7 +3438,9 @@ a.frm_option_icon:hover::before {
 .progress-bars-style .accordion-section-title:before{
 	content: "\e939" !important;
 }
-
+.ranking-fields-style .accordion-section-title:before{
```
Crabcyborg, line 3441: We have admin CSS for the surveys add-on. We should be able to add any admin styles for ranking fields there instead.

fonts/s11-fp-1.svg
Crabcyborg, line 1: Can we remove any font icons changes? Can this be done without using font icons?

js/formidable_admin.js
```diff
@@ -5662,16 +5662,29 @@ function frmAdminBuildJS() {
...
 
+		if ( 'ranking' === type ) {
+			inputType = 'hidden';
+		}
+
 		other = '<input type="text" id="field_' + fieldKey + '-' + opt.key + '-otext" class="frm_other_input frm_pos_none" name="item_meta[other][' + fieldId + '][' + opt.key + ']" value="" />';
 
-		single = '<div class="frm_' + type + ' ' + type + ' ' + classes + '" id="frm_' + type + '_' + fieldId + '-' + opt.key + '"><label for="' + id +
+		this.getSingle = function() {
+			if ( 'ranking' === type ) {
```
Crabcyborg, line 5672: Can we use a filter to move this JS into the surveys add-on?

js/formidable_admin.js
```diff
@@ -7205,20 +7225,8 @@ function frmAdminBuildJS() {
...
-			limit = parseInt( limit );
-			if ( limit > 1 ) {
-				message  = message.replace( 1, limit ).trim();
-			} else {
-				message += ' ' + frmAdminJs.edit_action_text;
-			}
-		}
-
-		infoModal( message );
+	function onlyOneActionMessage() {
```
Crabcyborg, line 7228: It looks like there was a mrege conflict here where an update got lost. We want `actionLimitMessage` not `onlyOneActionMessage`.

classes/models/FrmStyle.php
```diff
@@ -750,6 +750,9 @@ public function get_defaults() {
 			'progress_border_size'     => '2px',
 			'progress_size'            => '24px',
 
+			'ranking_bg_color_active'     => 'rgba(128, 187, 254, 0.04)',
```
Crabcyborg, line 753: Do we need these in Lite? It looks like these are getting added in the Surveys add-on using the filter below.

classes/models/FrmField.php
```diff
@@ -104,6 +104,13 @@ public static function pro_field_selection() {
 				'icon' => 'frm_icon_font frm_upload_icon',
 				'message' => 'Add file uploads to save time and cut down on back-and-forth. Upgrade to Pro to get Upload fields and more.',
 			),
+			'ranking'  => array(
+				'name'            => __( 'Ranking', 'formidable' ),
+				'icon'            => 'frm_icon_font frm_chart_bar_icon frm_show_upgrade',
+				'dynamic-message' => '<span class="frm-pb-sm frm-block">The Surveys add-on is not installed. Do you want to install the add-on now?</span>
```
Crabcyborg, line 110: We don't need this `dynamic-message` data.

Just passing `'addon' => 'surveys',` should be enough. The rest shouldn't be handled here.

classes/helpers/FrmFieldsHelper.php
```diff
@@ -1889,9 +1889,10 @@ public static function convert_field_object_to_flat_array( $field ) {
 	public static function show_add_field_buttons( $args ) {
 		$field_key    = $args['field_key'];
 		$field_type   = $args['field_type'];
+		$field_badge  = isset( $field_type['badge'] ) ? '<span class="frm-meta-tag frm-text-xs frm-lt-green-tag">' . $field_type['badge'] . '</span>' : '';
```
Crabcyborg, line 1892: Can we use `frm-new-pill` here? We have a function `FrmAppHelper::show_pill_text` which would be nice to use here.

fonts/s11-fp-1.svg
Liviu-p: Hey @Crabcyborg, I updated the icons to use SVGs.  I ended up working on multiple solutions on this which none of them worked well because of `do_accordion_sections`: I tried to insert the SVG icons in `global $wp_meta_boxes` but title sanitizing from `do_accordion_sections` wouldn't allow to display svg. Tried to extract each svg icon from font and use it separately in CSS which worked but we would end up in icon duplicates and bunch of new SVGs files in images. 
I couldn't find a better way to make it work with our icon integration than by copying the `do_accordion_sections` function into `FrmStylesHelper` and adapting it a bit to fit our needs.

classes/helpers/FrmStylesHelper.php
```diff
@@ -715,6 +715,82 @@ function( $value ) {
...
+
+		if ( empty( $screen ) ) {
+			$screen = get_current_screen();
+		} elseif ( is_string( $screen ) ) {
+			$screen = convert_to_screen( $screen );
+		}
+
+		$page = $screen->id;
+
+		$hidden = get_hidden_meta_boxes( $screen );
```
Crabcyborg, line 744: Maybe we can reduce this function a bit. I don't believe we're using any hidden meta boxes - this part of the code could probably go.

classes/helpers/FrmStylesHelper.php
```diff
@@ -715,6 +715,82 @@ function( $value ) {
...
+
+						$open_class = '';
+						if ( ! $first_open && empty( $hidden_class ) ) {
+							$first_open = true;
+							$open_class = 'open';
+						}
+						?>
+						<li class="control-section accordion-section <?php echo esc_attr( $hidden_class ); ?> <?php echo esc_attr( $open_class ); ?> <?php echo esc_attr( $box['id'] ); ?>" id="<?php echo esc_attr( $box['id'] ); ?>">
+							<h3 class="accordion-section-title hndle" tabindex="0">
+								<?php FrmAppHelper::icon_by_class( 'frmfont ' . $icon_id ); ?>
```
Crabcyborg, line 772: These 3 lines all open and close PHP tags. It would make sense to just open/close once.

classes/helpers/FrmStylesHelper.php
```diff
@@ -715,6 +715,82 @@ function( $value ) {
...
 	}
 
+	/**
+	 * Duplicate of WordPress do_accordion_section function, it adds an additional svg icon support.
+	 *
+	 * @since x.x
+	 *
+	 * @return int
+	 */
+	public static function do_accordion_sections( $screen, $context, $data_object ) {
```
Crabcyborg, line 725: Instead of a new function, we could probably just use `FrmStylesController` since we already have that function there. This way we don't need another public function for the same thing.

classes/views/frm-fields/single-option.php
Crabcyborg: I'm seeing this is to avoid the checkmark/radio for selecting a default value.

Maybe we can just hide this with CSS in the surveys add-on so we don't need to make a change in Lite?

classes/models/fields/FrmFieldType.php
```diff
@@ -486,9 +486,12 @@ public function show_default_value_field( $field, $default_name, $default_value
 
 	/**
 	 * @since 4.04
+	 *
+	 * @param array $args
+	 * @return bool
 	 */
 	protected function should_continue_to_field_options( $args ) {
-		return in_array( $args['field']['type'], array( 'select', 'radio', 'checkbox' ) );
+		return in_array( $args['field']['type'], array( 'select', 'radio', 'checkbox' ), true );
```
Crabcyborg, line 494: I removed ranking here. There's a new `should_continue_to_field_options` function for the ranking field that returns true in the surveys add-on instead now.

classes/helpers/FrmStylesHelper.php
Crabcyborg: I made a small commit to fix this.

classes/helpers/FrmFieldsHelper.php
Crabcyborg: I updated this.

classes/models/FrmField.php
Crabcyborg: I've updated this. I dropped `dynamic-message` for just the regular `message` attribute.

And added an `upsell_image` option instead.

css/admin/style.css
Crabcyborg: I removed this.

js/formidable_admin.js
```diff
@@ -5575,13 +5575,13 @@ function frmAdminBuildJS() {
...
 		} else {
 			opts = getMultipleOpts( fieldId );
-			type = input.attr( 'type' );
 			jQuery( '#field_' + fieldId + '_inner_container > .frm_form_fields' ).html( '' );
 			fieldInfo = getFieldKeyFromOpt( jQuery( '#frm_delete_field_' + fieldId + '-000_container' ) );
 
 			var container = jQuery( '#field_' + fieldId + '_inner_container > .frm_form_fields' ),
 				hasImageOptions = imagesAsOptions( fieldId ),
 				imageSize = hasImageOptions ? getImageOptionSize( fieldId ) : '',
+				type = ( 'hidden' === input.attr( 'type' ) ? input.data( 'field-type' ) : input.attr( 'type' ) ),
```
Crabcyborg, line 5584: `type` is already defined earlier up. We shouldn't redeclare a variable.

js/formidable.js
```diff
@@ -1642,6 +1642,8 @@ function frmFrontFormJS() {
 				}
 			}
 
+			jQuery( document ).trigger( 'frmOnFormSubmit', [ object, e ]);
```
Crabcyborg, line 1645: I left a comment about this in the Surveys add-on, but can we do this without a new event? This might not catch every case. It would be great if we could just handle this with PHP if possible.

classes/views/frm-fields/single-option.php
Crabcyborg: I see what happens when you try to make it visible. The `<input type="ranking" />` field is pretty not ideal.

I changed this to a `radio`/`checkbox` check instead so we don't see any unexpected types here and we can avoid hard coding `ranking`.

css/_single_theme.css.php
Liviu-p: I removed this @Crabcyborg, it was already in added in `frm-surveys.css`

css/custom_theme.css.php
```diff
@@ -264,6 +264,29 @@
 	font-weight:var(--field-weight);
 	box-shadow:var(--box-shadow)<?php echo esc_html( $important ); ?>;
 }
+.with_frm_style .frm-ranking-field-container .frm-ranking-field-option {
```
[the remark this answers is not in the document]
Liviu-p: It has been removed with SVG icons update.

js/formidable_admin.js
Liviu-p: I updated it @Crabcyborg as requested

js/formidable_admin.js
Liviu-p: I noticed you took care of this, thank you @Crabcyborg!

classes/models/fields/FrmFieldType.php
```diff
@@ -488,7 +488,7 @@ public function show_default_value_field( $field, $default_name, $default_value
 	 * @since 4.04
 	 */
 	protected function should_continue_to_field_options( $args ) {
-		return in_array( $args['field']['type'], array( 'select', 'radio', 'checkbox' ) );
+		return in_array( $args['field']['type'], array( 'select', 'radio', 'checkbox', 'ranking' ) );
```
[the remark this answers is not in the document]
Liviu-p: I can see it has been updated by you, thanks!

classes/models/FrmStyle.php
Liviu-p: nope, these were handled via a filter in surveys add-on, I removed it.

classes/models/FrmField.php
Liviu-p: Thank you @Crabcyborg!

classes/helpers/FrmStylesHelper.php
Liviu-p: Yep, seems to be working fine without that. I removed it.

classes/helpers/FrmStylesHelper.php
Liviu-p: I updated it.

js/formidable_admin.js
Liviu-p: I updated it

css/custom_theme.css.php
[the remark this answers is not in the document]
Crabcyborg: Hm, this appears to still be here.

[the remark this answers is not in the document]
Liviu-p: I've just pushed a commit to address this

js/formidable_admin.js
```diff
@@ -5664,14 +5664,31 @@ function frmAdminBuildJS() {
...
+
+			/**
+			 * Get single option template.
+			 * @param {Object} option  Object containing the option data.
+			 * @param {string} type    The field type.
+			 * @param {string} fieldId The field id.
+			 * @param {string} classes The option clasnames.
+			 * @param {string} id      The input id attribute.
+			 */
+			single = wp.hooks.applyFilters( 'frm_admin.build_single_option_template', { opt, type, fieldId, classes, id });
```
Crabcyborg, line 5677: This filter isn't quite being used the right way.

Here's a recording. You can see the live preview of the radio button options all disappear when I add a radio button option (with surveys on master).

Like with the PHP filters, the first variable sent should be the value we are passing.

We should **not** check if `single` is `undefined`. This should never be possible, if the filter is being used correctly.

We could pass an empty string `''` and then check if single is not an empty string instead, then return it. Right now it tries to return the `{ opt, type, fieldId, classes, id }` object instead which isn't right.

In the surveys add-on, you do not want to do a void return. We'll need to change this line in surveys:
```
if ( 'ranking' !== type ) {
	return;
}
```

TdRE7YIFiP

js/formidable.js
Liviu-p: I removed this line, as the validation will be handled in backend only from now.

css/admin/style.css
```diff
@@ -237,7 +238,7 @@ Time fields use a div instead of a label, so don't target by the primary label b
 	line-height: var(--line-height);
 }
 
-#frm_style_preview .with_frm_style :not(.ui-datepicker-title) > select {
+#frm_style_preview .with_frm_style :not(.ui-datepicker-title) > select:not(.frm-ranking-position) {
```
Crabcyborg, line 241: @Liviu-p Do we need these `:not(.frm-ranking-position)` rules in the CSS? It would be nice if we could remove these.

These just normalize the dropdowns a bit. I don't imagine they should be causing too many issues?

I tried looking at a styler preview of a ranking field without these modifications, and it seems like the dropdown looks a bit better with dropdown normalizing.

The dropdown arrow looks like it's in the wrong place right now,
Screen Shot 2024-03-05 at 10 06 35 AM

Liviu-p: I've just removed it. It was later updated in the surveys add-on. 
Thank you @Crabcyborg for double-checking on this!

css/admin/style.css
```diff
@@ -247,7 +248,7 @@ Time fields use a div instead of a label, so don't target by the primary label b
 	appearance: none; /* Hide the default icon */
 }
 
-#frm_style_preview select:not(.ui-datepicker-month):not(.ui-datepicker-year) {
+#frm_style_preview select:not(.ui-datepicker-month):not(.ui-datepicker-year):not(.frm-ranking-position) {
```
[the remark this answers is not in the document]
Liviu-p: Hey @Crabcyborg, I find it required as it fixes the the dropdown icon position in the "Style" tab.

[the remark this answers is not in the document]
Crabcyborg: It seems fine to me without it, but I'm fine with keeping it.

Crabcyborg (approved):
Thanks @Liviu-p for working on this!

I think we're good to merge this for today's core release.

🚀

PR #1522, opened 2024-02-20
Add field name shortcode for error messages

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2505693625/189582/

> if a field is left blank, the error message needs to include what field , such as "The name field cannot be left blank", "the message field cannot be left blank", etc.

I initially brought this up in Slack (https://strategy11.slack.com/archives/G01KDLFJTJS/p1707484587610489).

Quoting Steph's reply,
> We could use "[label] cannot be blank", but I think it's fine if we replace "this field" so we don't have to add extra support, and fix this for existing forms.

I'm using `[field_name]` (rather than `[label]`) which is already a supported in a few places. I believe it works with JavaScript validation without making further changes because it gets replaced in field HTML already.

**Some issues with blank error messages require https://github.com/Strategy11/formidable-pro/pull/4822 or the "This field"/ text will still be used when Pro is active.**

### General Description
- Some substrings are now replaced in error messages.
    - "This field" and "This value" will be replaced by a field name dynamically. This is for backward compatibility so old fields are also supported. This can be disabled using a new `frm_error_substrings_to_replace_with_field_name` filter.
    -  On settings pages, "This field" and "This value" will be changed to `[field_name]` automatically in the builder.
- I try to use `[field_name]` everywhere as much as possible. **This works better with translations** than trying to only use the "This field"/"This value" substrings.

I left some comments in the PR with more details as to why certain changes were made.

I added a new `frm_error_substrings_to_replace_with_field_name` filter so this can be tweaked. If someone wants to keep their "This field" or "This value" substrings from getting converted to `[field_name]` then they can use those filters to remove the strings from the array.

**The hook can also be used to add additional translated strings, if you wanted to convert more substrings universally in a single place.**

Example usage to only replace `[field_name]` shortcodes,
```
add_filter( 'frm_error_substrings_to_replace_with_field_name', function() {
	return array( '[field_name]' );
} );
```

There is however no way to avoid the `str_replace` happening when a new field loads its default value from the global settings. I always replace "This field" and "This value" from the global setting to `[field_name]` when a new field is inserted.

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -352,32 +352,34 @@ class FrmFieldsHelper {
 		$msg = empty( $msg ) ? $defaults[ $error ]['part'] : $msg;
 		$msg = do_shortcode( $msg );
 
-		$substrings_to_replace_with_field_name = array(
-			'[field_name]',
-			'This value',
-			'This field',
-		);
-		/**
-		 * @since x.x
-		 *
-		 * @param array<string> $substrings_to_replace_with_field_name
-		 * @param array         $args {
-		 *     @type string       $msg   The current error message before the substrings are replaced.
-		 *     @type string       $error A key including 'unique_msg', 'invalid', 'blank', or 'conf_msg'.
-		 *     @type array|object $field The field with the error.
-		 * }
-		 */
-		$filtered_substrings = apply_filters( 'frm_error_substrings_to_replace_with_field_name', $substrings_to_replace_with_field_name, compact( 'msg', 'error', 'field' ) );
+		if ( trim( $field_name ) ) {
+			$substrings_to_replace_with_field_name = array(
+				'[field_name]',
+				'This value',
+				'This field',
+			);
+			/**
+			 * @since x.x
+			 *
+			 * @param array<string> $substrings_to_replace_with_field_name
+			 * @param array         $args {
+			 *     @type string       $msg   The current error message before the substrings are replaced.
+			 *     @type string       $error A key including 'unique_msg', 'invalid', 'blank', or 'conf_msg'.
+			 *     @type array|object $field The field with the error.
+			 * }
+			 */
+			$filtered_substrings = apply_filters( 'frm_error_substrings_to_replace_with_field_name', $substrings_to_replace_with_field_name, compact( 'msg', 'error', 'field' ) );
 
-		if ( is_array( $filtered_substrings ) ) {
-			$substrings_to_replace_with_field_name = $filtered_substrings;
-		} else {
-			_doing_it_wrong( __FUNCTION__, 'Only arrays should be returned when using the frm_error_substrings_to_replace_with_field_name filter.', 'x.x' );
-		}
+			if ( is_array( $filtered_substrings ) ) {
+				$substrings_to_replace_with_field_name = $filtered_substrings;
+			} else {
+				_doing_it_wrong( __FUNCTION__, 'Only arrays should be returned when using the frm_error_substrings_to_replace_with_field_name filter.', 'x.x' );
+			}
 
-		foreach ( $substrings_to_replace_with_field_name as $substring ) {
-			if ( false !== strpos( $msg, $substring ) ) {
-				$msg = str_replace( $substring, FrmAppHelper::maybe_kses( $field_name ), $msg );
+			foreach ( $substrings_to_replace_with_field_name as $substring ) {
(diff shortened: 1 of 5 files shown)


classes/models/fields/FrmFieldType.php
```diff
@@ -640,9 +640,6 @@ public function get_new_field_defaults() {
 			'options'       => '',
 			'default_value' => '',
 			'required'      => false,
-			'blank'         => $frm_settings->blank_msg,
```
Crabcyborg, line 643: I removed these because it looks like these were being defined in the wrong array. `blank`, `unique_msg`, and `invalid` are all "field options", not part of the `$field` object itself.

classes/models/fields/FrmFieldType.php
```diff
@@ -651,27 +648,11 @@ public function get_new_field_defaults() {
 		return array_merge( $field, $field_options );
 	}
 
-	protected function default_unique_msg() {
```
Crabcyborg, line 654: This is deprecated now because it wasn't really being used.

This is only called when a new field is created, where the field would never be unique.

Then the `unique_msg` is filtered through `frm_default_field_options`. There, Pro sets it to `$frm_settings->unique_msg`.

I don't think we need this logic in Lite since `unique` is a Pro-only setting anyway.

classes/models/fields/FrmFieldType.php
```diff
@@ -708,13 +689,15 @@ protected function new_field_settings() {
...
 	 */
 	public function get_default_field_options() {
-		$opts       = array(
+		$frm_settings = FrmAppHelper::get_settings();
+		$opts         = array(
 			'size'               => '',
 			'max'                => '',
 			'label'              => '',
-			'blank'              => '',
+			'blank'              => FrmFieldsHelper::default_blank_msg(),
```
Crabcyborg, line 697: A new field would always have a blank required message because there was no default defined.

This line fixes that so the default blank message is there right away for the new field instead of after it is saved as blank.

classes/models/fields/FrmFieldType.php
```diff
@@ -1592,4 +1567,26 @@ public function maybe_decode_value( $value ) {
...
+		$message      = $frm_settings->unique_msg;
+		return $message;
+	}
+
+	/**
+	 * @deprecated x.x
+	 *
+	 * @return string
+	 */
+	protected function default_invalid_msg() {
```
Crabcyborg, line 1588: I deprecated this because it doesn't appear to be necessary.

This was only being used somewhere where I don't think it was ever working (see https://github.com/Strategy11/formidable-forms/pull/1522/files#r1496412221).

Removing it doesn't appear to have any issue. I see an invalid message as soon as I add a new Email or Phone field.

classes/models/fields/FrmFieldType.php
```diff
@@ -708,13 +682,15 @@ protected function new_field_settings() {
 	 * @return array
 	 */
 	public function get_default_field_options() {
-		$opts       = array(
+		$frm_settings = FrmAppHelper::get_settings();
```
stephywells, line 685: This line looks like it might be unused.

classes/helpers/FrmFieldsHelper.php
```diff
@@ -178,6 +176,48 @@ private static function fill_cleared_strings( $field, array &$field_array ) {
...
+	/**
+	 * @since x.x
+	 *
+	 * @return string
+	 */
+	public static function default_unique_msg() {
+		$frm_settings   = FrmAppHelper::get_settings();
+		$unique_message = $frm_settings->unique_msg;
+
+		if ( false !== strpos( $unique_message, 'This value' ) ) {
```
stephywells, line 198: Is it helpful to have this check before replacing the value?

stephywells (approved):
This looks like a great approach! I haven't tested this, but the code looks great.

classes/models/fields/FrmFieldType.php
Crabcyborg: Good catch! I must have added that and then changed it to use ` FrmFieldsHelper::default_blank_msg` instead.

classes/helpers/FrmFieldsHelper.php
Crabcyborg: I don't think there's really any need for the extra checks. I removed them wherever it's just adding extra code.

Thanks!

garretlaxton (approved):
This looks good!

Crabcyborg:
Thanks Garret!

PR #1523, opened 2024-02-21
Add shortcodes popup to ai field

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-ai/issues/51
This update in lite is required to make shortcode popups work for an AI field in https://github.com/Strategy11/formidable-ai/pull/50 work

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -6985,7 +6985,8 @@ function frmAdminBuildJS() {
 	}
 
 	function shouldAddShortcodesModalTriggerIcon( fieldType ) {
-		const fieldsWithShortcodesBox = [ 'html', 'ai' ];
+		const fieldsWithShortcodesBox = wp.hooks.applyFilters( 'frm_fields_with_shortcode_popup', [ 'html' ] );
+
 		return fieldsWithShortcodesBox.includes( fieldType );
 	}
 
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):

js/formidable_admin.js
```diff
@@ -6985,7 +6985,7 @@ function frmAdminBuildJS() {
 	}
 
 	function shouldAddShortcodesModalTriggerIcon( fieldType ) {
-		const fieldsWithShortcodesBox = [ 'html' ];
+		const fieldsWithShortcodesBox = [ 'html', 'ai' ];
```
truongwp, line 6988: I think we should call an `applyFilters()` here instead of adding `ai`, so other fields can be added without changing this plugin.

AbdiTolesa: Thanks @truongwp! That is a good suggestion and I have added a new filter here and hooked into it in ai.

js/formidable_admin.js
```diff
@@ -6985,7 +6985,8 @@ function frmAdminBuildJS() {
 	}
 
 	function shouldAddShortcodesModalTriggerIcon( fieldType ) {
-		const fieldsWithShortcodesBox = [ 'html' ];
+		const fieldsWithShortcodesBox = wp.hooks.applyFilters( 'frm_fields_with_shortcode_popup', [ 'html' ]);
```
truongwp, line 6988: We should use camelCase for JS hook name. Please change it.

truongwp: @Crabcyborg Ok. I'm messed up with the event name. Thank you.

truongwp (approved):

Crabcyborg (approved):
This looks good to me.

🚀

PR #1527, opened 2024-02-22
improve table style for two cell rows

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-quizzes/issues/189

When any of the first column cells have a long content, it pushes the second column far to the right and squashes it.

It looks max-width is not supported for non-replaced inline elements (See https://developer.mozilla.org/en-US/docs/Web/CSS/max-width#formal_definition). I just set the first cell's width to `50%` and believe it is just enough.

**Before**:
image

**After**:
image

### Test steps
1. Create a form and add some fields to it.
2. Edit one of the fields to make the label be a very long text.
3. Configure 'Send email' action to send the default message to your email.
4. Submit an entry populating the fields with values.
5. Check the table in the email sent out and confirm that the maximum width of the labels is 50% of the total row width

### DIFF
--- a/classes/models/FrmTableHTMLGenerator.php
+++ b/classes/models/FrmTableHTMLGenerator.php
@@ -275,6 +275,20 @@ class FrmTableHTMLGenerator {
 	}
 
 	/**
+	 * Appends inline css props to existing style string of style="prop:value" format
+	 *
+	 * @since x.x
+	 *
+	 * @param string $props
+	 * @param string $style
+	 *
+	 * @return string
+	 */
+	private function add_props_to_inline_style( $props, $style ) {
+		return rtrim( $style, '"' ) . $props . '"';
+	}
+
+	/**
 	 * Generate a two cell row for an HTML table
 	 *
 	 * @since 2.04
@@ -285,12 +299,12 @@ class FrmTableHTMLGenerator {
 	 * @return string
 	 */
 	public function generate_two_cell_table_row( $label, $value ) {
-		$row = '<tr' . $this->tr_style();
+		$row = '<tr ' . $this->add_props_to_inline_style( 'display:flex;', $this->tr_style() );
 		$row .= $this->add_row_class( $value === '' );
 		$row .= '>';
 
-		$label = '<th' . rtrim( $this->td_style, '"' ) . ' width: 50%;"' . '>' . wp_kses_post( $label ) . '</th>';
-		$value = '<td' . $this->td_style . '>' . wp_kses_post( $value ) . '</td>';
+		$label = '<th ' . $this->add_props_to_inline_style( 'min-width:50%;max-width: 50%;', $this->td_style ) . ' >' . wp_kses_post( $label ) . '</th>';
+		$value = '<td ' . $this->add_props_to_inline_style( 'width: 100%;', $this->td_style ) . ' >' . wp_kses_post( $value ) . '</td>';
 
 		if ( 'rtl' == $this->direction ) {
 			$row .= $value;
--- a/classes/models/FrmTableHTMLGenerator.php
+++ b/classes/models/FrmTableHTMLGenerator.php
@@ -299,12 +299,12 @@ class FrmTableHTMLGenerator {
 	 * @return string
 	 */
 	public function generate_two_cell_table_row( $label, $value ) {
-		$row = '<tr ' . $this->add_props_to_inline_style( 'display:flex;', $this->tr_style() );
+		$row = '<tr' . $this->add_props_to_inline_style( 'display:flex;', $this->tr_style() );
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This looks good!

classes/models/FrmTableHTMLGenerator.php
```diff
@@ -285,12 +299,12 @@ public function generate_table_footer() {
 	 * @return string
 	 */
 	public function generate_two_cell_table_row( $label, $value ) {
-		$row = '<tr' . $this->tr_style();
+		$row = '<tr' . $this->add_props_to_inline_style( 'display:flex;', $this->tr_style() );
```
truongwp, line 302: `$this->tr_style()` and `$this->td_style` can be empty string. So your custom method might not work.

Can you change the styling in `tr_style()` and `init_td_style()` methods instead of creating a new one?

AbdiTolesa: > Can you change the styling in `tr_style()` and `init_td_style()` methods instead of creating a new one?

I don't want to affect other places those values are used and that's why I created a new function but I have updated it to consider empty strings.

truongwp:
@AbdiTolesa Sorry. I've just noticed that this PR fixes a Quizzes issue. In the Quizzes addon, we have `FrmQuizzesTableHTMLGenerator` class that extends this class. Can you make change in the Quizzes addon to fix the issue?

AbdiTolesa:
> @AbdiTolesa Sorry. I've just noticed that this PR fixes a Quizzes issue. In the Quizzes addon, we have `FrmQuizzesTableHTMLGenerator` class that extends this class. Can you make change in the Quizzes addon to fix the issue?

@truongwp I have discussed this in the issue page, on this comment (https://github.com/Strategy11/formidable-quizzes/issues/189#issuecomment-1946347021) and I think we are convinced this update is worth do it here to keep the emails consistent. In fact, it isn't caused by the Quiz add-on. Any email with very long entry values could have caused the same issue.

truongwp:
@AbdiTolesa Thank you. If we fix it here to keep consistency, I think we should move changes in `tr_style()` and `td_style` to fix this issue in other places. With this extra CSS, I don't think it will break the table in other places.

AbdiTolesa:
> @AbdiTolesa Thank you. If we fix it here to keep consistency, I think we should move changes in `tr_style()` and `td_style` to fix this issue in other places. With this extra CSS, I don't think it will break the table in other places.

@truongwp I couldn't figure out how to test how doing that might affect `generate_single_cell_table_row` since it uses those class members. Have you tried that? We should at least check that.

truongwp (approved):
@AbdiTolesa After checking this PR again, I found that you add different CSS to `<th>` and `<td>`. They used `$this->td_style` before. So making change directly to the `td_style` won't work.

This PR looks good. Thank you.

classes/models/FrmTableHTMLGenerator.php
```diff
@@ -285,12 +302,12 @@ public function generate_table_footer() {
...
 		$row .= $this->add_row_class( $value === '' );
 		$row .= '>';
 
-		$label = '<th' . $this->td_style . '>' . wp_kses_post( $label ) . '</th>';
-		$value = '<td' . $this->td_style . '>' . wp_kses_post( $value ) . '</td>';
+		$label = '<th' . $this->add_props_to_inline_style( 'min-width:50%;max-width: 50%;', $this->td_style ) . '>' . wp_kses_post( $label ) . '</th>';
```
Crabcyborg, line 309: It looks like this still hasn't really been figured out.

Steph mentions "Adding the 50% width is not a good solution.".

Adding a `min-width` isn't really a better solution than adding `width`.

AbdiTolesa: @Crabcyborg Yeah I remember she said to go with `max-width:50%` but if we apply that to the cells, they would get a different width and the table columns wouldn't be aligned at all. This is the best look I found being limited to inline styling. The table's width would auto shrink/grow based on the content with this at least.

Crabcyborg: Since you are using `display: flex` on the row, what about using `flex: 1` on the cells?

That seems to work fine https://jsfiddle.net/3t07n48q/

AbdiTolesa: @Crabcyborg Oh, one thing I didn't mention above is that I found that some css properties get removed from the table by the client, including `flex:1`. It isn't working for me at least with gmail.

Crabcyborg:
This update still really just looks like a messy way of setting width to 50%. I don't think this is the direction we wanted to go in.

I'm just going to close this for now. I think we can try to come up with a better solution to this issue.

PR #1532, opened 2024-02-23
Update connect flow

stephywells:
- Remove the ajax for v1 connect process. The JS for the 'frm_connect' ajax was removed in an earlier PR
- Pass a return page and installed flag to the connect progress. If it's already installed, the api won't return a file_url

The server-side updates are already in place, and most of the changes are there.

With this update:
- After connecting, we'll return to the same page we started from (settings, dashboard, add-ons)
- If the plugin is already installed, there shouldn't be any errors
- The connect process should be much smoother. This includes when pro is installed but disconnected and when pro is not installed.

### DIFF
--- a/classes/controllers/FrmAddonsController.php
+++ b/classes/controllers/FrmAddonsController.php
@@ -890,38 +890,6 @@ class FrmAddonsController {
 	}
 
 	/**
-	 * Install Pro after connection with Formidable.
-	 *
-	 * @since 4.02.05
-	 *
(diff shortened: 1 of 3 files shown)


Crabcyborg (approved):
This looks good to me.

Thanks @stephywells!

I haven't tested this but I've requested QA from Garret.

garretlaxton (approved):
This works fine for me!

PR #1533, opened 2024-02-26
Try using a placeholder element and replacing it with ajax response

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4796

This is an alternative fix to https://github.com/Strategy11/formidable-pro/pull/4816 suggested by Mike.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -7125,16 +7125,16 @@ function frmAdminBuildJS() {
 
 	function addFormLogicRow() {
 		/*jshint validthis:true */
-		var id = jQuery( this ).data( 'emailkey' ),
-			type = jQuery( this ).closest( '.frm_form_action_settings' ).find( '.frm_action_name' ).val(),
-			formId = document.getElementById( 'form_id' ).value,
-			logicRowsContainer = document.getElementById( 'frm_logic_row_' + id ),
-			logicRows = logicRowsContainer.querySelectorAll( '.frm_logic_row' ),
-			newRowID =  getNewRowId( logicRows, 'frm_logic_' + id + '_' ),
-			placeholder = div({
-				id: 'frm_logic_' + id + '_' + newRowID,
-				className: 'frm_logic_row frm_hidden'
-			});
+		const id                 = jQuery( this ).data( 'emailkey' );
+		const type               = jQuery( this ).closest( '.frm_form_action_settings' ).find( '.frm_action_name' ).val();
+		const formId             = document.getElementById( 'form_id' ).value;
+		const logicRowsContainer = document.getElementById( 'frm_logic_row_' + id );
+		const logicRows          = logicRowsContainer.querySelectorAll( '.frm_logic_row' );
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -7128,23 +7128,31 @@ function frmAdminBuildJS() {
 		var id = jQuery( this ).data( 'emailkey' ),
 			type = jQuery( this ).closest( '.frm_form_action_settings' ).find( '.frm_action_name' ).val(),
 			formId = document.getElementById( 'form_id' ).value,
-			logicRows = document.getElementById( 'frm_form_action_' + id ).querySelectorAll( '.frm_logic_row' );
+			logicRowsContainer = document.getElementById( 'frm_logic_row_' + id ),
+			logicRows = logicRowsContainer.querySelectorAll( '.frm_logic_row' ),
+			newRowID =  getNewRowId( logicRows, 'frm_logic_' + id + '_' ),
+			placeholder = div({
```
Crabcyborg, line 7134: We could use `const` here. It would be nice to stop using `var` in this function while we're working on it.

js/formidable_admin.js
```diff
@@ -7128,23 +7128,31 @@ function frmAdminBuildJS() {
...
 				nonce: frmGlobal.nonce
 			},
 			success: function( html ) {
-				html = wp.hooks.applyFilters( 'frm_action_logic_row_html', html, id );
-
 				jQuery( document.getElementById( 'logic_link_' + id ) ).fadeOut( 'slow', function() {
-					var $logicRow = jQuery( document.getElementById( 'frm_logic_row_' + id ) );
-					$logicRow.append( html );
+					var $logicRow = jQuery( logicRowsContainer );
+					var tempDiv   = document.createElement( 'div' );
```
Crabcyborg, line 7153: Let's use const for these.

Also, we can use `div()` here to make this line smaller.

js/formidable_admin.js
```diff
@@ -7128,23 +7128,31 @@ function frmAdminBuildJS() {
 		var id = jQuery( this ).data( 'emailkey' ),
 			type = jQuery( this ).closest( '.frm_form_action_settings' ).find( '.frm_action_name' ).val(),
 			formId = document.getElementById( 'form_id' ).value,
-			logicRows = document.getElementById( 'frm_form_action_' + id ).querySelectorAll( '.frm_logic_row' );
+			logicRowsContainer = document.getElementById( 'frm_logic_row_' + id ),
+			logicRows = logicRowsContainer.querySelectorAll( '.frm_logic_row' ),
+			newRowID =  getNewRowId( logicRows, 'frm_logic_' + id + '_' ),
```
Crabcyborg, line 7133: There's are too many spaces here after this `=` sign.

js/formidable_admin.js
Crabcyborg: Don't worry about making this change.

I made a PR branching off of this https://github.com/Strategy11/formidable-forms/pull/1539

I think we can do this without needing a tempDev.

PR #1535, opened 2024-02-26
Remove a lot of old CSS rules that are no longer required

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2518544642/191058/

This update removes a lot of the prefixed `-webkit-` and `-moz-` CSS rules. These are legacy from back when CSS3 was really new and are no longer required. Many of these now throw warnings when validating site CSS (like using https://jigsaw.w3.org/).

Many of these at least no longer work in Firefox, and documentation is fully removed.

I didn't fix everything. This is more of a phase 1 update. It helps drop a lot of errors from the CSS validation list so the remaining pieces are easier to digest. So far I was able to drop about 50 lines of CSS across various files.

I mostly target a few rules in this update including `-webkit-box-sizing`, `-moz-border-radius`, `-webkit-box-shadow`, and `animation` rules.

Once this is approved, I'll start working on a Pro update as well. For now, I haven't started working on reducing CSS in Pro.

### DIFF
--- a/css/_single_theme.css.php
+++ b/css/_single_theme.css.php
@@ -265,8 +265,6 @@ if ( '' === $field_height || 'auto' === $field_height ) {
 	color:<?php echo esc_html( $submit_text_color . $important ); ?>;
 	cursor:pointer;
 	font-weight:<?php echo esc_html( $submit_weight . $important ); ?>;
-	-moz-border-radius:<?php echo esc_html( $submit_border_radius . $important ); ?>;
-	-webkit-border-radius:<?php echo esc_html( $submit_border_radius . $important ); ?>;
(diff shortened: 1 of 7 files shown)


css/_single_theme.css.php
```diff
@@ -195,10 +195,6 @@
 <?php echo esc_html( $style_class ); ?> textarea:-ms-input-placeholder{
 	color: <?php echo esc_html( $text_color_disabled . $important ); ?>;
 }
-.<?php echo esc_html( $style_class ); ?> input:-moz-placeholder,
```
Crabcyborg, line 198: I dropped this because there's an identical style above.

For now I've left the placeholder styles alone.

stephywells (approved):
This looks good! We should be sure to mention this in the readme, along with minimum browser versions.

PR #1536, opened 2024-02-26
Make hex to rgb function work with rgb values

Crabcyborg:
I noticed some issues when testing my Formidable CSS against https://jigsaw.w3.org/css-validator/

Colour values when using RGB values are coming through wrong as `background: rgba(, , , 0.15) !important;`.

The problem is that `hex2rgb` is called unconditionally and doesn't work with rgb values.

frm_style_formidable-style-5 with frm style td ui-datepicker-current-day,

I also added a check if `$border_width` is empty to avoid a parsing error,
height auto;

Really, we should be checking if every setting is empty, but each check makes this look more messy.

### DIFF
--- a/classes/helpers/FrmStylesHelper.php
+++ b/classes/helpers/FrmStylesHelper.php
@@ -185,7 +185,23 @@ class FrmStylesHelper {
 		<?php
 	}
 
+	/**
+	 * Convert a hex color to RGB.
+	 *
+	 * @param string $hex Hex color value.
+	 * @return string RGB value without the rgb() wrapper.
+	 */
 	public static function hex2rgb( $hex ) {
+		if ( 0 === strpos( $hex, 'rgb' ) ) {
+			$rgb = str_replace( array( 'rgb(', 'rgba(', ')' ), '', $hex );
+			$rgb = explode( ',', $rgb );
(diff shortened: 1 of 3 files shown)


classes/helpers/FrmStylesHelper.php
```diff
@@ -185,7 +185,23 @@ public static function bs_icon_select( $style, $frm_style, $type = 'arrow' ) {
...
 	}
 
+	/**
+	 * Convert a hex color to RGB.
+	 *
+	 * @param string $hex Hex color value.
+	 * @return string RGB value without the rgb() wrapper.
+	 */
 	public static function hex2rgb( $hex ) {
+		if ( 0 === strpos( $hex, 'rgb' ) ) {
```
stephywells, line 195: Can we move this to a new function so it can have a description? The existing function description and name don't really make sense.

Crabcyborg: Thanks @stephywells!

I just broke this up now into 2 new functions - `get_rgb_array_from_rgb` and `get_rgb_array_from_rgb`.

stephywells (approved):
Thank you! That's much more clear now.

PR #1537, opened 2024-02-27
Add filter for limiting repeater entries in list page

AbdiTolesa:
Related issue: https://github.com/Strategy11/formidable-pro/issues/4128

### Test procedure
1. Create a form and add a repeater to it.
2. Add a text input the repeater.
3. Submit many long text values for the input field.
4. View the form entries list and there must be 5 repeated entries per an entry in the list which is the new default.
5. Use the new `frm_repeater_entries_limit` to increase the number of entries displayed in the list.
Example.
`add_filter( 'frm_repeater_entries_limit', function( $limit ) { return 10;} );`
7. Confirm that number of entries is increased accordingly.


garretlaxton (approved):
This looks great!

truongwp (approved):
This looks good. Maybe we should change the filter name for more specific to the entry listing page.

Thanks @AbdiTolesa!

classes/helpers/FrmEntriesHelper.php
```diff
@@ -216,8 +216,19 @@ public static function prepare_display_value( $entry, $field, $atts ) {
...
+			/**
+			 * This filter allows updating the limit of child entries in entry list page. The default is 5.
+			 *
+			 * @since x.x
+			 * @param array $args {
+			 *    @type object $field
+			 *    @type array  $atts
+			 * }
+			 */
+			$child_entries_limit = apply_filters( 'frm_repeater_entries_limit', 5, compact( 'field', 'atts' ) );
```
Crabcyborg, line 228: I agree with Truong this could use a more descriptive name.

I think we should mention that is for a display value, and that this is specific for the entries list.

Maybe `frm_entries_list_repeater_display_limit`?

Crabcyborg: We're also getting a missing `//end if` PHPCS error here.

When these come up, it's a good indicator that there's too much code in this this if condition.

If we move out some of the logic into a new function, like a `get_display_value_repeater_entries`, we wouldn't need to add the `//end if`,

Screen Shot 2024-03-06 at 12 56 07 PM

classes/helpers/FrmEntriesHelper.php
```diff
@@ -257,6 +246,41 @@ public static function prepare_display_value( $entry, $field, $atts ) {
...
+			 *    @type array  $atts
+			 * }
+			 */
+			$child_entries_limit = apply_filters( 'frm_entries_list_repeater_display_limit', 5, compact( 'field', 'atts' ) );
+
+			// This is a repeating section.
+			return FrmEntry::getAll( array( 'it.parent_item_id' => $entry->id ), '', $child_entries_limit, true );
+		}
+
+		// Get all values for this field.
```
Crabcyborg, line 274: It looks like we lost the comment `// This is an embeded form.`. It would be nice to make it more clear that this is code is for handling embedded forms. Right now that is not very clear.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -257,6 +246,41 @@ public static function prepare_display_value( $entry, $field, $atts ) {
 		return FrmAppHelper::kses( $val, 'all' );
 	}
 
+	/**
+	 * @since x.x
```
Crabcyborg, line 250: A function description would be helpful here since this code isn't very self descriptive.

Crabcyborg (commented):
Thanks @AbdiTolesa.

I think this looks great.

I just have one more thought here. I wonder if we could possibly show an ellipses or something to show that there are more entries? Right now it isn't clear at a glance that it's actually being truncated.

Screen Shot 2024-03-08 at 2 20 16 PM

classes/helpers/FrmEntriesHelper.php
```diff
@@ -254,9 +243,76 @@ public static function prepare_display_value( $entry, $field, $atts ) {
...
+	 *
+	 * @return string
+	 */
+	private static function maybe_append_ellipses( $val, $entry, $field, $atts ) {
+		if ( substr( $val, -3 ) === '...' ) {
+			return $val;
+		}
+
+		$child_entries_limit    = apply_filters( 'frm_entries_list_repeater_display_limit', 5, compact( 'field', 'atts' ) );
+		$total_repeater_entries = FrmEntry::getRecordCount( array( 'it.parent_item_id' => $entry->id ) );
```
Crabcyborg, line 270: @AbdiTolesa Instead of making two queries, can we just use a limit of `$child_entries_limit + 1`? We'll only return the `$child_entries_limit` amount, so we'll need to pop off the last result, but if there are `$child_entries_limit + 1` entries, then we know to show the ellipses.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -254,9 +252,77 @@ public static function prepare_display_value( $entry, $field, $atts ) {
...
+			/**
+			 * This filter allows updating the limit of child entries in entry list page. The default is 5.
+			 *
+			 * @since x.x
+			 * @param array $args {
+			 *    @type object $field
+			 *    @type array  $atts
+			 * }
+			 */
+			$child_entries_limit = apply_filters( 'frm_entries_list_repeater_display_limit', 5, compact( 'field', 'atts' ) );
```
Crabcyborg, line 304: The main reason I haven't merged this is that I'm not totally excited about having a repeater-specific filter in Lite.

Is there a better way we could do this where this new filter is in Pro instead?

AbdiTolesa: The logic looks tightly tied to the Lite plugin. The only way I can think of is creating a Pro function and calling it from Lite with a feature check. I don't know if that is any better.

Crabcyborg: @AbdiTolesa It looks like the code here in Lite should be in Pro. It would be nice to take this opportunity to remove Pro logic from Lite.

Everything after line 213 is Pro logic since it goes past a `if ( ! FrmAppHelper::pro_is_installed() ) {` check on that line.

The code after that calls Pro functions already (`FrmProEntryMetaHelper::get_post_or_meta_value`).

We could move this into the Pro version of the file `FrmProEntriesHelper`.

What we've done a lot in the past is check if `FrmProEntriesHelper::prepare_child_display_value` is callable and if it is, return the result of that function call. If it isn't, we fall back on the old logic here. After several versions, we remove the callable check and just always call the new function and remove the old code.

AbdiTolesa: @Crabcyborg Sure thing! I have created https://github.com/Strategy11/formidable-pro/pull/5232 and moved the pro logic there.

classes/helpers/FrmEntriesHelper.php
```diff
@@ -210,51 +210,10 @@ public static function prepare_display_value( $entry, $field, $atts ) {
 			return self::display_value( $field_value, $field, $atts );
 		}
 
-		if ( ! FrmAppHelper::pro_is_installed() ) {
+		if ( ! is_callable( 'FrmProEntriesHelper::prepare_child_display_value' ) ) {
 			return '';
 		}
-
```
Crabcyborg, line 216: @AbdiTolesa This isn't quite what I asked for.

We should flip this check here to `is_callable` and `return FrmProEntriesHelper::prepare_child_display_value( $entry, $field, $atts );` if it is.

Then keep all of this old code for backward compatibility until a later date.

We don't want to break this for users who have only updated Lite and not Pro.

AbdiTolesa: That's my bad, sorry for not considering backward compatibility!

classes/helpers/FrmEntriesHelper.php
```diff
@@ -210,8 +210,8 @@ public static function prepare_display_value( $entry, $field, $atts ) {
 			return self::display_value( $field_value, $field, $atts );
 		}
 
-		if ( ! FrmAppHelper::pro_is_installed() ) {
```
Crabcyborg, line 213: @AbdiTolesa I think we'll still need this pro check.

1. The fallback still shouldn't run if this is false.
2. We still shouldn't call `FrmProEntriesHelper::prepare_child_display_value` if there is no pro license set.

PR #1538, opened 2024-02-27
Do not allow drag after submit button using index

Crabcyborg:


truongwp (approved):
This works great! Thank you so much!

PR #1539, opened 2024-02-27
Use insertAdjacentHTML for 4796 fix

Crabcyborg:
Related PR https://github.com/Strategy11/formidable-forms/pull/1533

This is just a small attempt to clean this part of this update up a bit. I don't think we need to create a temporary element and risk getting only a single child when the HTML returns multiple.


AbdiTolesa (approved):
Thank you for cleaning up my code @Crabcyborg! This looks good to me.

PR #1540, opened 2024-02-27
Pro issue 4842 / Convert string recipients to array fix

Crabcyborg:
With https://github.com/Strategy11/formidable-forms/pull/1457/ the function `maybe_remove_recipients_from_api` was introduced.

The problem is that `$recipients` is a string, not an array, so the function doesn't really work as expected.

### DIFF
--- a/classes/helpers/FrmEmailSummaryHelper.php
+++ b/classes/helpers/FrmEmailSummaryHelper.php
@@ -528,7 +528,8 @@ class FrmEmailSummaryHelper {
 	 *
 	 * @since 6.8
 	 *
-	 * @param array $recipients Recipients.
+	 * @param string $recipients Recipients.
+	 * @return void
 	 */
 	public static function maybe_remove_recipients_from_api( &$recipients ) {
 		$api    = new FrmFormApi();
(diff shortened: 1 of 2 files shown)


classes/helpers/FrmEmailSummaryHelper.php
```diff
@@ -538,7 +539,10 @@ public static function maybe_remove_recipients_from_api( &$recipients ) {
 		}
 
 		$skip_emails = is_string( $addons['no_emails'] ) ? explode( ',', $addons['no_emails'] ) : (array) $addons['no_emails'];
-		$recipients  = array_diff( $recipients, $skip_emails );
+
+		$recipients = array_map( 'trim', explode( ',', $recipients ) );
```
truongwp, line 543: I think we should check if `$skip_emails` isn't empty before running extra code.

Do you think we should run `$recipients = array_map( 'trim', explode( ',', $recipients ) );` in `get_recipients()`, before this method?

Crabcyborg: @truongwp This function starts with:

```
if ( empty( $addons['no_emails'] ) ) {
	return;
}
```

I don't think `$skip_emails` would ever be empty after that check.

I'm fine with making sure that the recipients CSV has no spaces before hand, but since this function is public, if it were re-used more than once, you may end up needing to do that check in multiple places. If it's just handled here, it may be easier to manage.

truongwp: Oh sorry, I didn't notice that check.
I'm ok to leave `$recipients` as string for now.

truongwp (approved):
This looks good! Thanks @Crabcyborg!

PR #1541, opened 2024-02-28
Add new filter that allows updating display value

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-wpml/issues/4

### DIFF
--- a/classes/models/fields/FrmFieldType.php
+++ b/classes/models/fields/FrmFieldType.php
@@ -1439,7 +1439,16 @@ DEFAULT_HTML;
 	 * @return array|string
 	 */
 	protected function prepare_display_value( $value, $atts ) {
-		return $value;
+		/**
+		 * Allows updating display value for fields from classes not extending this class. Override this method in child classes otherwise.
+		 *
+		 * @since x.x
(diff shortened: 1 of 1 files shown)


classes/models/fields/FrmFieldType.php
```diff
@@ -1439,7 +1439,16 @@ protected function fill_default_atts( &$atts ) {
...
+		/**
+		 * Allows updating display value for fields from classes not extending this class. Override this method in child classes otherwise.
+		 *
+		 * @since x.x
+		 *
+		 * @param array|string $value
+		 * @param object $field
+		 * @param array $atts
+		 */
+		return apply_filters( 'frm_prepare_display_value', $value, $this->field, $atts );
```
Crabcyborg, line 1451: Is it possible to use a filter that already exists?

We have a lot already. `frm_display_value`, `frm_get_display_value`.

It would be nice to not add another.

AbdiTolesa: Oh, I couldn't realise I could use those ones indeed and we can definitely use them. We can close this PR then.

PR #1544, opened 2024-02-28
Filter import file options

Crabcyborg:
This update adds a `accept=".xml, .csv"` attribute when Pro is active, or a `accept=".xml"` attribute when only Lite is active.

I figure this is better for usability. It's more clear what can be uploaded since the wrong types are all faded out.

I got a little mixed up trying to upload a `.zip` of a `.xml` for a moment. This update helps make the types easier to tell apart, preventing a `.zip` upload from happening by accident.


stephywells (approved):
Great idea!

PR #1545, opened 2024-02-29
do frm_after_field_is_imported action even if form exists

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/1361

https://github.com/Strategy11/formidable-pro/pull/4518 alone doesn't cover it all. When you try to import a given form, while it already exist, the issue is still there and this PR triggers `frm_after_field_is_imported` action to make the pro fix apply there as well.

## Summary by CodeRabbit

- **New Features**
	- Enhanced field import functionality to automatically update field options with defaults and handle additional actions post-import.


Crabcyborg (commented):
Thanks @AbdiTolesa.

I think this looks good.

But I wonder if we want to use the same hook for this or if we'd want a new hook.

@stephywells Do you think we should use the same hook when a new field is imported and when an imported field triggers a field update? Right now it looks like there is only a hook getting called for a new imported field that didn't previously exist, so there's a gap when trying to use the hook.

stephywells:
@Crabcyborg I think it would be okay to use the same hook if we also include a parameter for the action. But a different hook may be easier, like we currently do in other places.

AbdiTolesa:
> But I wonder if we want to use the same hook for this or if we'd want a new hook.

@Crabcyborg I have added a new hook `frm_after_existing_field_is_imported` here and used it in Pro based on Steph's feedback.

classes/helpers/FrmXMLHelper.php
```diff
@@ -473,6 +477,43 @@ private static function import_xml_fields( $xml_fields, $form_id, $this_form, &$
...
 	}
 
+	/**
+	 * @since x.x
+	 *
+	 * @param array $field_array
+	 * @return array
+	 */
+	private static function update_field_options_with_defaults( $field_array ) {
+		$defaults               = self::default_field_options( $field_array['type'] );
```
Crabcyborg, line 487: @AbdiTolesa What is the reason for adding the defaults? Do we need this part?

AbdiTolesa: I just wanted to make the `$field_array` consistent with `frm_after_field_is_imported`  action as it is also called there.

Crabcyborg: I think we're good without it.

It isn't really part of the `frm_after_field_is_imported` hook. It's really just because that function is doing everything. It calls `FrmField::create( $f );` as well.

We don't need to change how the import logic works. I think we should just pass the variables we have.

classes/helpers/FrmXMLHelper.php
```diff
@@ -447,12 +447,17 @@ private static function import_xml_fields( $xml_fields, $form_id, $this_form, &$
...
 
+					$old_field_id = $f['id'];
+
 					// check for field to edit by field key
 					unset( $f['id'] );
 
 					FrmField::update( $form_fields[ $f['field_key'] ], $f );
 					$imported['updated']['fields'] ++;
 
+					$f['id'] = $old_field_id;
```
Crabcyborg, line 458: I think it probably makes more sense to use the new field ID for `$f['id']` and then pass the `old_field_id` as a filter argument.

The way it is right now feels a bit background. The "updated field" might not actually exist.

AbdiTolesa: @Crabcyborg If we add a third parameter to `frm_after_existing_field_is_imported` action, I think it would be inconsistent with `frm_after_field_is_imported`. We can't hook the same function to both hooks anymore as well and that requires another update. I'm not certain I have understand the risk you mentioned well when you say '"updated field" might not actually exist.' though. Do you mean we should do something like this where `$field_array` has the new id?
`do_action( 'frm_after_existing_field_is_imported', $field_array, $form_fields[ $field_array['field_key'] ], $old_field_id );`

Crabcyborg: Yeah. I think `$field_array['id']` should be the field that was updated, not the ID that was in the XML. The ID from the XML should be `$old_field_id`.

Which I mean by the field might not exist, is that there could be no field with `$old_field_id` as its ID, any we don't want people mistaking that for the new ID of the field.

If you really want to re-use one function, you could just set a `$old_field_id = 0` parameter and not use it when it's `0`.

classes/helpers/FrmXMLHelper.php
```diff
@@ -684,7 +725,7 @@ private static function create_imported_field( $f, &$imported ) {
 		$new_id = FrmField::create( $f );
 		if ( $new_id != false ) {
 			$imported['imported']['fields'] ++;
-			do_action( 'frm_after_field_is_imported', $f, $new_id );
+			do_action( 'frm_after_field_is_imported', $f, $new_id, 0 );
```
Crabcyborg, line 728: Do we need this `0` here?

I was thinking we would keep this to 2 parameters, but in the Pro code, you can use `2` as the number of arguments when defining the hook, and then using `$old_field_id = 0` as the default 3rd parameter in that function on Pro.

Crabcyborg (approved):
This looks good!

Thanks @AbdiTolesa!

🚀

PR #1546, opened 2024-02-29
Support multiple shortcode popups per field

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-ai/pull/50

We want to support multiple shortcodes popups per field as per Mike's comment in Guide prompt for AI field here: https://github.com/Strategy11/formidable-ai/pull/50#issuecomment-1969627465

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -6958,7 +6958,7 @@ function frmAdminBuildJS() {
 				textarea.parentNode.insertBefore( wrapperSpan, textarea );
 				wrapperSpan.appendChild( createModalTriggerIcon() );
 				wrapperSpan.appendChild( textarea );
-			})
+			});
 		};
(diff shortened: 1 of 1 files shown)


Crabcyborg:
@AbdiTolesa Is there anything we can do about the shortcode pop up not staying in one place when you scroll? It's pretty easy for it to be off-screen.

Here's a recording. The rich text field is using the "Smart default values" right now, and that pop up follows the input box on scroll.
zjL98Qv7gg

Also, can we use this filter in Pro so rich text fields also use this pop up instead of the one that's being used right now? The pop up doesn't have very useful shortcode options.

Crabcyborg (approved):
This looks good.

But my comment above also apply. I'll make new issues for those.

Thanks @AbdiTolesa!

PR #1543, opened 2024-02-28
Set form options to array if string when setting on submit migrated f…

Crabcyborg:
…lag fix

Related ticket https://secure.helpscout.net/conversation/2518729092/191090/

> Fatal error: Uncaught TypeError: Cannot access offset of type string on string in .../formidable/classes/helpers/FrmOnSubmitHelper.php:400 Stack trace: #0 .../formidable/classes/helpers/FrmOnSubmitHelper.php(305): FrmOnSubmitHelper::save_migrated_success_actions() #1 .../formidable/classes/controllers/FrmFormsController.php(153): FrmOnSubmitHelper::maybe_migrate_submit_settings_to_action() #2 .../formidable/classes/controllers/FrmFormsController.php(1754): FrmFormsController::settings() #3 .../wp-includes/class-wp-hook.php(324): FrmFormsController::route() #4 .../wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters() #5 .../wp-includes/plugin.php(517): WP_Hook->do_action() #6 .../wp-admin/admin.php(259): do_action() #7 {main} thrown in .../formidable/classes/helpers/FrmOnSubmitHelper.php on line 400

Pre-release,
[formidable-6.8.3b.zip](https://github.com/Strategy11/formidable-forms/files/14438017/formidable-6.8.3b.zip)

@truongwp I requested your review since I'm updating `save_migrated_success_actions`. Do you think this fix makes sense?

PR #1547, opened 2024-02-29
New function `FrmForm::prepare_form_row_data`

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2518729092/191090/

A lot of code works on the assumption that `$form->options` is an array.

I tried adding some checks in some updates yesterday (in https://github.com/Strategy11/formidable-forms/pull/1543, https://github.com/Strategy11/formidable-pro/pull/4845 and https://github.com/Strategy11/formidable-pro/pull/4844) but I think we're better off just forcing `$form->options` as an array right away. Trying to fix this in every place where we assume `$form->options` is an array is turning into too much work and too many undetected bugs.

**This update now forces all results from `FrmForm::getOne` to have an array `$form->options` value when `$form` is not null. I used the lite form defaults so some basic keys are defined, to help fix an issue with a PHP warning.**

**Questions**
1. Would we want to try calling `FrmFormsHelper::get_default_opts()` as well to define some Pro defaults?
2. Would we want to start only filtering `frm_form_object` when it's an object and not null? That might help with stability as well.

> Fatal error: Uncaught TypeError: Cannot access offset of type string on string in .../wp-content/plugins/formidable/classes/models/FrmForm.php:605 Stack trace: #0 .../wp-content/plugins/formidable/classes/models/FrmForm.php(559): FrmForm::trash() #1 .../wp-content/plugins/formidable/classes/controllers/FrmFormsController.php(658): FrmForm::set_status() #2 .../wp-content/plugins/formidable/classes/controllers/FrmFormsController.php(626): FrmFormsController::change_form_status() #3 .../wp-content/plugins/formidable/classes/controllers/FrmFormsController.php(1754): FrmFormsController::trash() #4 .../wp-includes/class-wp-hook.php(324): FrmFormsController::route() #5 .../wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters() #6 .../wp-includes/plugin.php(517): WP_Hook->do_action() #7 .../wp-admin/admin.php(259): do_action() #8 {main} thrown in .../wp-content/plugins/formidable/classes/models/FrmForm.php on line 605

Coming from this line of code:
```
$options['trash_time'] = time();
```

Since `$options` is a string, trying to set `$options['trash_time']` triggers a fatal error.

### DIFF
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -817,17 +817,20 @@ class FrmForm {
 	 */
 	private static function prepare_form_row_data( $row ) {
 		$row = wp_unslash( $row );
-		if ( is_object( $row ) && ! is_array( $row->options ) ) {
+		if ( ! is_object( $row ) ) {
+			return $row;
+		}
+
+		if ( ! is_array( $row->options ) ) {
 			$row->options = FrmFormsHelper::get_default_opts();
 		}
-		if ( is_object( $row ) ) {
-			/**
-			 * @since 4.03.02
-			 *
-			 * @param stdClass $row
-			 */
-			$row = apply_filters( 'frm_form_object', $row );
-		}
+
+		/**
+		 * @since 4.03.02
+		 *
+		 * @param stdClass $row
+		 */
+		$row = apply_filters( 'frm_form_object', $row );
 		return $row;
 	}
 
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -818,7 +818,7 @@ class FrmForm {
 	private static function prepare_form_row_data( $row ) {
 		$row = wp_unslash( $row );
 		if ( is_object( $row ) && ! is_array( $row->options ) ) {
-			$row->options = array();
+			$row->options = FrmFormsHelper::get_default_opts();
 		}
 		return apply_filters( 'frm_form_object', $row );
 	}
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -787,8 +787,7 @@ class FrmForm {
 				if ( isset( $cache->options ) ) {
 					FrmAppHelper::unserialize_or_decode( $cache->options );
 				}
-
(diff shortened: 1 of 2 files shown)


joaovictorangeline:
@Crabcyborg Hi! 
I installed the provided FormidableForms version, and I tested the following:

- [x]Moving the form to the Trash.
- [x]Clicking on Undo when moving the form to the trash.
- [x]Build, Settings, Entries, and Views tabs work fine.
- [ ]Style tab is returning this warning 
```
"Warning: Undefined array key "custom_style" in
/home/u103776375/domains/protecontech.com/public_html/wp-content/plugins/formidable/classes/views/styles/_styles-list.php
on line
11"
```
-[x]Added field and updated the form successfully
-[x]Form looks good when previewing it.
-[x]Form submitted fine.

Crabcyborg:
Thanks @joaovictorangeline.

I'm thinking we may want to set the default form options instead of an empty array to solve that issue.

joaovictorangeline:
> Thanks @joaovictorangeline.
> 
> I'm thinking we may want to set the default form options instead of an empty array to solve that issue.

Yes, when I enabled the default style, it worked fine! 
Should I wait for a new patch or should I answer the customer with what we already have? The warning will disappear when WP_DEBUG is false.

Crabcyborg:
@joaovictorangeline 

Here's a more up to date pre-release you can share that should fix that issue as well,
[formidable-6.8.3b.zip](https://github.com/Strategy11/formidable-forms/files/14450961/formidable-6.8.3b.zip)

joaovictorangeline:
I tested and it worked @Crabcyborg ! 
Thanks for putting your efforts on this.

stephywells (commented):
This looks like a great approach.

1. I think it's fine to not set the defaults for the pro options. The code should handle that already since many forms would start as lite forms before an upgrade.
2. That sounds like a good adjustment to only run the hook if it's an object.

PR #1548, opened 2024-02-29
Drop old deprecated frontend JS functions

Crabcyborg:
I don't think we need these functions anymore.

These functions were deprecated back in September 2015, nearly 9 years ago.

In this commit https://github.com/Strategy11/formidable-forms/commit/5695ba18b4806ae6e4d0838d0d443f779df43673

This helps simplify and reduce the size of our front end JS, dropping 45 lines of old deprecated code.


stephywells (approved):
Awesome! This looks great!

PR #1552, opened 2024-03-01
use button for draft trigger

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4022
### Test steps
1. Create a form, add some fields to it.
2. Enable saving drafts in the form permission settings
3. Preview the form and confirm that the draft trigger is no longer an anchor link but a button element
4. Edit the form to make it multi page.
5. Preview the form and click around the buttons and confirm that a loader is only added to the currently clicked element and everything is intuitive.

### Screen record
CleanShot 2024-03-04 at 10 07 20

## Summary by CodeRabbit

- **New Features**
	- Introduced a new button element for saving drafts, enhancing user interaction.
- **Style**
	- Updated CSS for loading indicators and elements, improving visual feedback during loading states.
- **Enhancements**
	- Enhanced form submission by disabling the save draft button during processing.
	- Improved error handling for form submissions, ensuring better feedback for users.

### DIFF
--- a/classes/helpers/FrmFormsHelper.php
+++ b/classes/helpers/FrmFormsHelper.php
@@ -449,7 +449,7 @@ BEFORE_HTML;
 	}
 
 	public static function get_draft_link() {
-		$link = '[if save_draft]<a href="#" tabindex="0" class="frm_save_draft" [draft_hook]>[draft_label]</a>[/if save_draft]';
+		$link = '[if save_draft]<button class="frm_save_draft frm_button_submit" [draft_hook]>[draft_label]</button>[/if save_draft]';
 
 		return $link;
 	}
--- a/classes/helpers/FrmFormsHelper.php
+++ b/classes/helpers/FrmFormsHelper.php
@@ -449,7 +449,7 @@ BEFORE_HTML;
 	}
 
 	public static function get_draft_link() {
-		$link = '[if save_draft]<button class="frm_save_draft frm_button_submit" [draft_hook]>[draft_label]</button>[/if save_draft]';
+		$link = '[if save_draft]<button class="frm_save_draft" [draft_hook]>[draft_label]</button>[/if save_draft]';
 
 		return $link;
 	}
--- a/css/custom_theme.css.php
+++ b/css/custom_theme.css.php
@@ -730,28 +730,28 @@ legend.frm_hidden{
 	visibility:visible !important;
 }
 
-.with_frm_style .frm_loading_prev .frm_prev_page,
-.with_frm_style .frm_loading_form .frm_button_submit {
+.with_frm_style .frm_loading_prev .frm_show_loader.frm_prev_page,
+.with_frm_style .frm_loading_form .frm_show_loader.frm_button_submit {
 	position: relative;
 	opacity: .8;
 	color: transparent !important;
 	text-shadow: none !important;
 }
 
-.with_frm_style .frm_loading_prev .frm_prev_page:hover,
-.with_frm_style .frm_loading_prev .frm_prev_page:active,
-.with_frm_style .frm_loading_prev .frm_prev_page:focus,
-.with_frm_style .frm_loading_form .frm_button_submit:hover,
-.with_frm_style .frm_loading_form .frm_button_submit:active,
-.with_frm_style .frm_loading_form .frm_button_submit:focus {
+.with_frm_style .frm_loading_prev .frm_show_loader.frm_prev_page:hover,
+.with_frm_style .frm_loading_prev .frm_show_loader.frm_prev_page:active,
+.with_frm_style .frm_loading_prev .frm_show_loader.frm_prev_page:focus,
+.with_frm_style .frm_loading_form .frm_show_loader.frm_button_submit:hover,
+.with_frm_style .frm_loading_form .frm_show_loader.frm_button_submit:active,
+.with_frm_style .frm_loading_form .frm_show_loader.frm_button_submit:focus {
 	cursor: not-allowed;
 	color: transparent;
 	outline: none !important;
 	box-shadow: none;
 }
 
-.with_frm_style .frm_loading_prev .frm_prev_page::before,
-.with_frm_style .frm_loading_form .frm_button_submit:before {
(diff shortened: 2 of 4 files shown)


garretlaxton (approved):
This looks great. But if the form is already created, it doesn't use a button and still uses just a link.

garretlaxton (approved):
Understood. This looks great then.

js/formidable.js
```diff
@@ -1571,6 +1571,13 @@ function frmFrontFormJS() {
 			jQuery( document ).on( 'elementor/popup/show', frmRecaptcha );
 
 			enableSubmitButtonOnBackButtonPress();
+			documentOn(
+				'click',
+				'input[type="submit"], .frm_prev_page, .frm_save_draft, .frm_button_submit',
+				function( event ) {
+					event.target.className += ' frm_show_loader';
```
truongwp, line 1578: We need to remove `frm_show_loaded` class from other buttons. You can set a field to required, enable validate and submit form via AJAX, you will see that class isn't removed.

AbdiTolesa: I didn't get it when you say "other buttons", I tried this with ajax submission and validation turned on and see no issues with it. Could you please share your screen record?
CleanShot 2024-03-07 at 12 40 19

truongwp: @AbdiTolesa Can you make a field required then leave it empty? When a validation error shows, the form isn't reloaded, that class isn't removed.
Sorry, I'm on my phone so I can't record a screencast.

AbdiTolesa: Sorry I accidentally cut off the part where I tried saving empty value as well but that also didn't cause the issue you said. I tried that again now as well. Can you try that from your end when you get the chance?

truongwp: Sorry, my explanation wasn't clear enough. You need to submit the form first, then save draft, you will see both buttons have loading icon.


https://github.com/Strategy11/formidable-forms/assets/19748318/27c33d7f-c6b8-497e-8d28-dd3f67b410aa

AbdiTolesa: Thanks that is clear now @truongwp! I have tried to remove that class from the submit button on ajax validation error and I think that should suffice it here. Let me know if you have any further concerns!

truongwp (approved):
This looks great now. Thanks @AbdiTolesa!

js/formidable.js
Crabcyborg, line 1578: Do we really need this new class?

Is this just to avoid some styles on save draft buttons? Can we just avoid styles on those buttons instead?

It would be nice if this update didn't require any JavaScript updates.

AbdiTolesa: @Crabcyborg I was thinking a bit of enhancement in this update in addition to making the draft trigger a button. All buttons would show loader when you click draft previously and I felt that is wrong and confusing to the user and I wanted to fix that too in this PR.
CleanShot 2024-03-12 at 10 08 23

Crabcyborg: @AbdiTolesa I get why it's there, but it's needlessly complicated.

I made a PR off of this branch to just handle it all in CSS like I asked.

https://github.com/Strategy11/formidable-forms/pull/1565

css/custom_theme.css.php
```diff
@@ -771,6 +771,10 @@
 	animation: spin 2s linear infinite;
 }
 
+.with_frm_style .frm_prev_page[disabled] {
+	opacity: 0.5;
```
Crabcyborg, line 775: We have the opacity rule for our next button in Pro css. It's in `single-style.css.php` because we need to make sure we're checking the `$settings['submit_style']` setting first.

Can we move this style into Pro instead?

Crabcyborg:
The styling part of this update is incomplete.

Razvan's image shows the draft button as an inverted button, where the background and the text are swapped.

I've made a PR in Pro to support this https://github.com/Strategy11/formidable-pro/pull/4867

js/formidable.js
AbdiTolesa: @Crabcyborg I tried your branch but this is how it looks with that, should we remove my effort to make it look consistent? I clicked the draft button but the styles all look inconsistent, loader shows on the other button etc. Sorry if I'm getting confused on this.

CleanShot 2024-03-12 at 16 03 35

Crabcyborg: @AbdiTolesa That's how it's always worked. That's not an issue.

AbdiTolesa: @Crabcyborg I understand it has been working like that but I was thinking it would be great if we could enhance that a bit (so that the loader reflects the button clicked while other buttons are only disabled) and make it more consistent.

css/custom_theme.css.php
AbdiTolesa: Sure, I have created https://github.com/Strategy11/formidable-pro/pull/4870

js/formidable.js
Crabcyborg: Thanks. I'm running this change by Steph in check-in.

I think this reason makes sense.

I think we'll probably want to come up with a cleaner solution still though. I think there's a better way to handle this.

Crabcyborg: Steph responded back that we're fine going with the way it's been working. I merged my PR into this.

> for the spinner question, I think it would be fine to only show the spinner on the submit button when save draft is clicked.

classes/helpers/FrmFormsHelper.php
```diff
@@ -449,7 +449,7 @@ public static function get_default_html( $loc ) {
 	}
 
 	public static function get_draft_link() {
-		$link = '[if save_draft]<a href="#" tabindex="0" class="frm_save_draft" [draft_hook]>[draft_label]</a>[/if save_draft]';
+		$link = '[if save_draft]<button class="frm_save_draft frm_button_submit" [draft_hook]>[draft_label]</button>[/if save_draft]';
```
Crabcyborg, line 452: Can we do this without `frm_button_submit`? What is the reasoning for using the submit button class?

I fear this could cause weird conflicts.

Crabcyborg: I've gone ahead and removed them.

I think this is good to merge now. I think we can handle the rest in Pro.

PR #1556, opened 2024-03-06
Reset hcaptcha when navigating pages and switch to explicit rendering

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4838

hCaptcha wouldn't render if the global hcaptcha is already set and that is the case when navigating between pages. I have attempted to nullify `window.hcaptcha` when navigating between pages to solve that issue.

### Test steps
1. Select hcaptcha from the global captcha settings.
2. Create a form, add a field then a summary field and captcha field.
3. Preview the form.
4. Fill the form and then on the summary page, click edit to edit the first field value.
5. Edit the field and then click next to the summary page again.
6. Confirm that hcaptcha is always rendered on the summary page.
7. Try the same by replacing the summary field with any other field and navigating around the pages.

## Summary by CodeRabbit

- **New Features**
	- Enhanced captcha support by appending specific parameters to the captcha API URL for improved loading and rendering.
	- Introduced additional element attributes for reCAPTCHA to enhance front-end integration.

- **Bug Fixes**
	- Adjusted the behavior of reCAPTCHA in scenarios where the captcha size is 'invisible' and multiple captchas are not enabled, by not setting the `data-callback` attribute.

- **Refactor**
	- Updated the captcha handling logic in JavaScript to support both Google reCAPTCHA and hCaptcha, including renaming functions and adding destruction handling for hCaptcha.


garretlaxton (approved):
This looks good! I navigated back and forth with many different fields on the last page and the Captcha works fine.

truongwp (approved):

js/formidable.js
```diff
@@ -1895,11 +1911,15 @@ jQuery( document ).ready( function() {
 	frmFrontForm.init();
 });
 
-function frmRecaptcha() {
```
Crabcyborg, line 1898: Since `frmRecaptcha` is a global function we shouldn't just remove it without deprecating it.

js/formidable.js
```diff
@@ -1568,24 +1578,30 @@ function frmFrontFormJS() {
...
+			var formID, captchaID,
 				size = captcha.getAttribute( 'data-size' ),
 				rendered = captcha.getAttribute( 'data-rid' ) !== null,
 				params = {
 					'sitekey': captcha.getAttribute( 'data-sitekey' ),
 					'size': size,
 					'theme': captcha.getAttribute( 'data-theme' )
-				};
+				},
+				activeCaptcha = reCaptchaIsSelected() ? grecaptcha : hcaptcha;
```
Crabcyborg, line 1604: There's a lot of if/else stuff here. We have a Cloudflare Turnstile captcha now too so it would be good to make this work better for multiple types.

I did some testing and it looks like Turnstile also has this bug at the moment. It would be good to try fixing that as well at the same time.

js/formidable.js
```diff
@@ -1568,24 +1578,30 @@ function frmFrontFormJS() {
...
 
 			// Elementor popup show event. Fix Elementor Popup && FF Captcha field conflicts
-			jQuery( document ).on( 'elementor/popup/show', frmRecaptcha );
+			jQuery( document ).on( 'elementor/popup/show', frmCaptcha );
 
 			enableSubmitButtonOnBackButtonPress();
+			jQuery( document ).on(
+				'click',
+				'.frm-edit-page-btn.frm_page_back, .frm_prev_page, .frm_button_submit:not(.frm_final_submit)',
+				destroyhCaptcha
```
Crabcyborg, line 1587: It would be nice to have a more reliable way here than a click handler.

Do we have other options on how to trigger this?

classes/models/fields/FrmFieldCaptcha.php
```diff
@@ -160,7 +160,7 @@ protected function recaptcha_api_url( $frm_settings ) {
 	}
 
 	protected function hcaptcha_api_url() {
-		$api_js_url = 'https://js.hcaptcha.com/1/api.js';
+		$api_js_url = 'https://js.hcaptcha.com/1/api.js?&onload=frmCaptcha&render=explicit';
```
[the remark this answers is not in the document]
AbdiTolesa: hCaptcha's documentation says:

> To specify a custom onload callback function, you must pass the function name as an onload query parameter in the hCaptcha javascript tag. In addition, you must set the render query parameter to explicit.

https://docs.hcaptcha.com/configuration/#explicitly-render-hcaptcha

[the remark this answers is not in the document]
Crabcyborg: They also have documentation for a data-callback. Did you try it?

All of these captcha scripts use the same interfaces. They all support the same options. We should be as consistent as possible.

js/formidable.js
AbdiTolesa: I have fixed the issue with Turnstile too now.

classes/models/fields/FrmFieldCaptcha.php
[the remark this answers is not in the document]
AbdiTolesa: Yes I have tried that before and again now, it is not working for me. I can output `data-callback` to the captcha element overriding `add_front_end_element_attributes` in `FrmHcaptchaSettings` but doesn't help.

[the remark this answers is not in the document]
AbdiTolesa: I don't think `data-callback` can be used as a replacement for `onload` param.

js/formidable.js
```diff
@@ -1897,13 +1914,31 @@ jQuery( document ).ready( function() {
 });
 
 function frmRecaptcha() {
+	console.warn( 'DEPRECATED: function frmRecaptcha in v6.x use frmCaptcha' );
```
Crabcyborg, line 1917: We could still call `frmCaptcha` here to avoid breaking anything.

js/formidable.js
```diff
@@ -1897,13 +1914,31 @@ jQuery( document ).ready( function() {
...
+		return grecaptcha;
+	}
+	if ( document.querySelector( '.h-captcha' ) ) {
+		return hcaptcha;
+	}
+	return turnstile;
+}
+
+function getSelectedCaptchaElements() {
+	return jQuery( '.frm-g-recaptcha, .h-captcha, .cf-turnstile' );
```
Crabcyborg, line 1939: Can we stop using jQuery here?

js/formidable.js
```diff
@@ -1897,13 +1914,31 @@ jQuery( document ).ready( function() {
...
-		frmFrontForm.renderRecaptcha( captchas[c]);
+		frmFrontForm.renderCaptcha( captchas[c]);
 	}
 }
 
+function getSelectedCaptcha() {
+	if ( document.querySelector( '.frm-g-recaptcha' ) ) {
+		return grecaptcha;
+	}
+	if ( document.querySelector( '.h-captcha' ) ) {
```
Crabcyborg, line 1932: Maybe we should target a form as well. This could be a false positive if there was hcaptcha for something else but not in a Formidable form.

js/formidable.js
```diff
@@ -1522,6 +1522,16 @@ function frmFrontFormJS() {
 		});
 	}
 
+	/**
+	 * Destroys the global hcaptcha object since it wouldn't otherwise render.
+	 */
+	function destroyhCaptcha() {
+		if ( ! window.hasOwnProperty( 'hcaptcha' ) ) {
```
Crabcyborg, line 1529: Maybe we should have another check here to confirm that there is a Formidable hcaptcha field before removing the global object?

It's possible that this could break a non-Formidable hCaptcha implementation.

js/formidable.js
```diff
@@ -1897,13 +1914,32 @@ jQuery( document ).ready( function() {
 });
 
 function frmRecaptcha() {
+	console.warn( 'DEPRECATED: function frmRecaptcha in v6.x use frmCaptcha' );
+	frmRecaptcha();
```
Crabcyborg, line 1918: This is calling itself. This would cause an infinite loop.

js/formidable.js
```diff
@@ -1897,13 +1914,32 @@ jQuery( document ).ready( function() {
...
 	var c, cl,
-		captchas = jQuery( '.frm-g-recaptcha' );
+		captchas = getSelectedCaptchaElements();
 	for ( c = 0, cl = captchas.length; c < cl; c++ ) {
-		frmFrontForm.renderRecaptcha( captchas[c]);
+		frmFrontForm.renderCaptcha( captchas[c]);
 	}
 }
 
+function getSelectedCaptcha() {
```
Crabcyborg, line 1929: I feel like we'd be better off just passing the captcha type and not having to figure this out with querying for elements which can slow down an already slow page.

We could keep the `frmRecaptcha` function and have it call like `frmCaptcha( '.frm-g-recaptcha' )`.

Then add an `frmHcaptcha`, `frmTurnstile`.

It's safer this way as it has a more direct path.

js/formidable.js
```diff
@@ -1897,13 +1914,32 @@ jQuery( document ).ready( function() {
 });
 
 function frmRecaptcha() {
+	console.warn( 'DEPRECATED: function frmRecaptcha in v6.x use frmCaptcha' );
```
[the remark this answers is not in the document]
AbdiTolesa: I think we are keeping this function based on the comment before this and we probably don't need to deprecate it.

js/formidable.js
```diff
@@ -1522,6 +1522,16 @@ function frmFrontFormJS() {
 		});
 	}
 
+	/**
+	 * Destroys the formidable generated global hcaptcha object since it wouldn't otherwise render.
+	 */
+	function destroyhCaptcha() {
+		if ( ! window.hasOwnProperty( 'hcaptcha' ) || ! document.querySelector( '.h-captcha' ) ) {
```
Crabcyborg, line 1529: `h-captcha` is still too generic. I'd check if it's in `frm-show-form` here.

js/formidable.js
```diff
@@ -1571,21 +1581,27 @@ function frmFrontFormJS() {
 			jQuery( document ).on( 'elementor/popup/show', frmRecaptcha );
 
 			enableSubmitButtonOnBackButtonPress();
+			jQuery( document ).on(
```
Crabcyborg, line 1584: We have a function `documentOn`. Does it work in-place of `jQuery( document ).on(`?

AbdiTolesa: I tried this with that but that is not working, probably because `frmPageChanged` is a jQuery event?

Crabcyborg (approved):
Thanks @AbdiTolesa!

Thank you @truongwp and @garretlaxton for reviewing this!

🚀

PR #1562, opened 2024-03-07
Fix missing views arrow icons

Crabcyborg:
This broke with the ranking field update.

These styles that were removed are still required for the collapsible sections in the visual views editor.

Screen Shot 2024-03-07 at 11 08 22 AM

On master, these icons are missing now.
Screen Shot 2024-03-07 at 11 08 49 AM

I also dropped the new chart SVG from the `s11-fp.svg` file as I believe we don't need this anymore.

This update adds back some the styles that were removed, and adds some new styles in the `style.css` file to only make these changes on the styler rather than everywhere.


Liviu-p (commented):
Thank you @Crabcyborg for fixing this. Regarding "Advanced Settings" & "No Entries Message" navs, may I have the page path where these can be found?

Crabcyborg:
Thanks Liviu!

PR #1564, opened 2024-03-11
Update position of shortcodes popup on scroll

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4846

The new shortcodes popup position was fixed and not updated when scroll and this update fixes that.
Followup conversation https://github.com/Strategy11/formidable-pro/pull/4856#discussion_r1516262084


Crabcyborg (approved):
Thanks @AbdiTolesa!

This looks good to me.

🚀

PR #1566, opened 2024-03-12
add new filter that will support enabling the Bulk Edit Options, fix …

Liviu-p:
This adds a new filter that will allow to enable the "Bulk Edit Options" for other fields rather than radio, checkbox or select.
Left margin fix and "save" icon position fix when "Separate Values" is activated.


css/frm_admin.css
```diff
@@ -4347,7 +4347,7 @@ li.ui-state-default.edit_field_type_divider .frm-show-hover {
 .frm_image_preview_wrapper,
 #new_fields .frm_single_option select,
 #new_fields .frm_single_option input[type=text] {
-	width: calc(100% - 73px);
+	width: calc(100% - 83px);
```
Crabcyborg, line 4350: I don't think we need to reduce the width here for other field types, like radio buttons.

Also, why is it that the ranking options are 10px less wide? Can we fix this in the surveys add-on instead?

Liviu-p: The weird thing is that I was able to replicate the same issue related to "save" icon in "Separate Value"  on radio buttons and checkboxes  in older version with surveys add-on disabled. I believe the way how "save" icon is position via an absolute position inside a `span` which has an inline display and no `position:relative` makes the "save" icon position and inputs width inconsistent across different browsers & screen resolution. 
I see this more as a separate issue and non related to ranking feature.

Crabcyborg: It works fine for other fields. There was briefly a bug in master, but it was never released.

I made a change in the surveys add-on to make the inputs larger. I also fixed the icon not being visible issue there.

Crabcyborg: I've removed this change so we can merge the bulk edit update on its own.

Crabcyborg (approved):
This looks good. Thanks @Liviu-p.

I made some changes to the variable names the filter name.

This branch was out of date. Make sure you're using the latest version of master. I think you were seeing some buggy CSS in your branch.

PR #1568, opened 2024-03-14
Return empty string instead of 'noform'

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4314

### Test steps
1. Install and enable classic non-block themes like Twenty Twenty or something
2. Install "Classic Widgets" plugin for compatibility
3. Go to Appearance > Widgets and drag the 'Formidable Forms' widget to an active widget area.
4. Try changing the widget title and form selected
5. Confirm that the 'Save' button is visible and you can save it.
6. View a page on which this widget area is displayed and confirm everything is as expected.

## Summary by CodeRabbit

- **Bug Fixes**
	- Adjusted the form method to return an empty string for consistency, enhancing the user experience.

### DIFF
--- a/classes/widgets/FrmShowForm.php
+++ b/classes/widgets/FrmShowForm.php
@@ -92,6 +92,6 @@ class FrmShowForm extends WP_Widget {
 			</label>
 		</p>
 		<?php
-		return 'noform';
+		return '';
 	}
 }


garretlaxton (approved):
This works great!
image

truongwp (approved):
This looks great. Thanks @AbdiTolesa!

classes/widgets/FrmShowForm.php
```diff
@@ -92,6 +92,6 @@ public function form( $instance ) {
 			</label>
 		</p>
 		<?php
-		return 'noform';
+		return '';
```
Crabcyborg, line 95: @stephywells It looks like this bug was introduced in the v6.0 styling update with this commit https://github.com/Strategy11/formidable-forms/commit/72d6c392ab142a72c1448e57548d53782327d727

I'm not too familiar with this code. The change here looks simple.

Do you see any issue changing this to an empty string instead?

stephywells: That should be just fine. That was added for the phpcs requirement to have a return value. But an empty string would cover that too.

Crabcyborg: Thanks Steph!

Crabcyborg (approved):

🚀

PR #1570, opened 2024-03-15
Hide bulk edit only for image buttons

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-surveys/issues/211

## Summary by CodeRabbit

- **Refactor**
	- Enhanced readability of displaying the bulk edit link in image options settings by refactoring the logic.

### DIFF
--- a/classes/views/frm-fields/back-end/field-options.php
+++ b/classes/views/frm-fields/back-end/field-options.php
@@ -6,7 +6,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 $field_option_count = is_array( $args['field']['options'] ) ? count( $args['field']['options'] ) : 0;
 $display_format     = FrmField::get_option( $args['field'], 'image_options' );
 ?>
-<span class="frm-bulk-edit-link <?php echo $display_format === '1' ? 'frm_hidden' : ''; ?> ">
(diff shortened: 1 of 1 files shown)


classes/views/frm-fields/back-end/field-options.php
```diff
@@ -4,8 +4,9 @@
 }
 
 $field_option_count = is_array( $args['field']['options'] ) ? count( $args['field']['options'] ) : 0;
+$display_format     = FrmField::get_option( $args['field'], 'image_options' );
```
Crabcyborg, line 7: This variable doesn't seem very well named. The option is called `image_options`, the variable is called `display_format`. That doesn't really match up. It should be more like `uses_images` or something that describes the setting?

AbdiTolesa: If we use `uses_images`, it would imply that `image_options === 1` but that is yet to be checked in the line below this. I used `display_format` because we labeled that setting 'Display Format' in form builder. I can change it if that still doesn't make sense.
image

Crabcyborg: Oh, thanks. Maybe we can add a comment instead if the issue is the option name?

Crabcyborg (approved):

Thanks Abdi!

PR #1577, opened 2024-03-18
Skip sending monthly summary email if there is no entries

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/4881

We did the same thing for the yearly email before.

## Summary by CodeRabbit

- **New Features**
	- Enhanced monthly email functionality to check for entries before sending, ensuring emails are only sent when necessary.


Crabcyborg (approved):
Thanks @truongwp! This looks good.

🚀

PR #1587, opened 2024-03-21
Use a less generic id in floating links icon fix

Crabcyborg:
I noticed that with the Cloudflare Turnstile update, the floating links icon got all messed up.

The issue would go away immediately when I deleted the Cloudflare SVG.

It turns out that it's using a clipPath with `#a`. So is the Cloudflare SVG. This caused conflicting clip path issues.

This update updates both of the `#a` instances to something less conflict prone.

Screen Shot 2024-03-21 at 10 59 41 AM

## Summary by CodeRabbit

- **Refactor**
	- Updated SVG elements and styling in the CAPTCHA feature for improved visual consistency.


Crabcyborg:
Thanks @shervElmi!

PR #1588, opened 2024-03-22
Use card styling for review request

Crabcyborg:
Related comment https://github.com/Strategy11/formidable-pro/issues/4862#issuecomment-2013820862

> Can the x icon placement, margins, shadows, and border styling be the same on the review request?

This update aims to stop using the `notice` and `notice-info` classes everywhere.

I've added the `frm-card-item` class to the review request, and added a new `frm-compact-card-item` class to try sharing the padding between both the inbox slide-in and the review request.

X7asXyVj7g

I also modified the floating success message so it not longer uses the `frm-review-notice` class which isn't a very good fit. The fixed positioning is now handled in its class.

Screen Shot 2024-03-22 at 9 18 12 AM

## Summary by CodeRabbit

- **Style**
	- Updated CSS classes for review notices, enhancing their visual presentation.
	- Introduced a new CSS variable for box shadows, improving consistency across the application.
	- Adjusted padding and positioning for better layout and user experience.
- **Refactor**
	- Updated class names in JavaScript files for consistency with CSS changes.


stephywells:
Thanks @Crabcyborg ! It looks like the X in this one is still not positioned quite the same as the new inbox slide in.

Crabcyborg:
@stephywells I don't think we can position further to the right if that's what you mean. The review request is still right at the `right: 0`/`bottom: 0` position so the scroll bar gets in the way if you try to move it over more.

Is it wrong vertically as well? Or am I missing something?

stephywells:
Here's an example of the x in another spot. Looks like the color is a bit lighter. mMaybe the problem isn't the position. I was just looking at the screenshots.
Screenshot 2024-03-22 at 3 21 00 PM

Crabcyborg:
Thanks @stephywells.

I'm not sure why your icon is so faded out in that screenshot.

I'm using `var(--grey-800)`. The design also shows the X as a darker shade.

Screen Shot 2024-03-25 at 9 18 27 AM

I tried opening the new application modal in Firefox and Chrome, and in both, I'm seeing the darker icon.

Screen Shot 2024-03-25 at 9 22 52 AM

Any chance your dismiss icon is light because of a plugin conflict?

stephywells:
Weird... I'm not seeing the lighter color now either. I'm not sure what happened. Looks like we're all set!

PR #1590, opened 2024-03-25
Enhanced Formatting Options for Phone Field

shervElmi:
This PR introduces enhanced formatting options for the Phone Field, providing greater flexibility and international compatibility. The new options include 'None', 'International', and 'Custom' formats. The 'None' option applies a default pattern to the input, while the 'International' option integrates the [International Telephone Input](https://intl-tel-input.com/) JS library on the front end and employs a corresponding regex validation on the backend. The 'Custom' option allows users to define their patterns, preserving the functionality of previous versions.

### QA URL:
https://qa.formidableforms.com/sherv2/wp-admin/admin.php?page=formidable&frm_action=edit&id=29

### Required Pro Branch URL:
https://github.com/Strategy11/formidable-pro/tree/new-feature-country-code

### Testing Instructions:
1. Navigate to "WP Admin > Formidable > Forms" and create a new form with a phone field.
2. Access the Phone Field Settings and switch the Format between "International", "None", and "Custom".
3. Ensure that each format operates as expected, as demonstrated in the introductory videos.

### Introduction Video:
https://www.loom.com/share/9d385af44beb4d4d9b2a0511e48d7d26?sid=d79b3c19-00be-4571-a7f9-d0ee25660a4f

### Checklist:
- [x] Implement a migration step to set the `phone_type` value to `custom` for existing users who have previously utilized custom patterns.

## Summary by CodeRabbit

- **New Features**
	- Introduced an 'International' phone type option in form fields, allowing users to select from predefined or custom phone number formats.
	- Enhanced dropdown functionality with dependency management, improving user interaction by dynamically displaying relevant options based on selections.

- **Enhancements**
	- Improved phone number validation by adding support for international formats and sanitization processes.

- **Bug Fixes**
	- Adjusted form field styling for better visual consistency and usability.

- **Refactor**
	- Optimized event handling in JavaScript for form elements, ensuring smoother interactions and performance improvements.


js/formidable_admin.js
```diff
@@ -9662,6 +9675,53 @@ function frmAdminBuildJS() {
...
+		/**
+		 * Toggles the visibility of dependent elements associated with a select element based on its current selection.
+		 *
+		 * @since x.x
+		 *
+		 * @param {HTMLElement} select The select element whose dependencies need to be managed.
+		 * @return {void}
+		 */
+		function toggleDependencyVisibility( select ) {
+			const selectedOption = select.options[select.selectedIndex] || null;
```
Crabcyborg, line 9699: I don't know how helpful `|| null` is. We can probably just let it fallback to `undefined`.

Also, the spacing is off here. There should be spaces in the `[]` braces.

js/formidable_admin.js
```diff
@@ -9662,6 +9675,53 @@ function frmAdminBuildJS() {
...
+			});
+		}
+
+		// Initial setup: Show dependencies based on the current selection in each dropdown
+		selects.forEach( select => {
+			toggleDependencyVisibility( select );
+		});
+
+		// Update dependencies visibility on dropdown change
+		document.addEventListener( 'change', event => {
```
Crabcyborg, line 9716: We have a function `frmDom.util.documentOn` that you could use here instead that is basically doing what you're doing. It would make this a bit smaller.

More like,
`frmDom.util.documentOn( 'change', 'select.frm_select_with_dependency',( event ) => toggleDependencyVisibility( event.target ) );`

classes/views/frm-fields/back-end/phone/phone-type.php
```diff
@@ -0,0 +1,39 @@
...
+?>
+<p class="frm6 frm6_followed frm_form_field frm-phone-type">
+	<label for="phone_type_<?php echo esc_attr( $field_id ); ?>">
+		<?php esc_html_e( 'Format', 'formidable' ); ?>
+	</label>
+
+	<select
+		name="field_options[phone_type_<?php echo esc_attr( $field_id ); ?>]"
+		id="phone_type_<?php echo esc_attr( $field_id ); ?>"
+		class="frm_phone_type_dropdown frm_select_with_premium frm_select_with_dependency"
```
Crabcyborg, line 28: I'm thinking `frm_select_with_premium` might be a bit inconsistent of a name.

Maybe `frm_select_with_upgrade`?

garretlaxton:
If I put the phone field in a repeater, the country selector doesn't show for any additional fields added.
image

classes/controllers/FrmFieldsController.php
```diff
@@ -484,6 +484,10 @@ private static function add_input_classes( $field, array &$class ) {
 		if ( isset( $field['size'] ) && $field['size'] > 0 ) {
 			$class[] = 'auto_width';
 		}
+
+		if ( 'phone' === $field['type'] && isset( $field['phone_type'] ) && 'international' === $field['phone_type'] ) {
```
Crabcyborg, line 488: Is it possible to put this in Pro instead? Ideally the pro features don't get added in Lite.

classes/models/FrmEntryValidate.php
```diff
@@ -232,10 +237,22 @@ public static function validate_phone_field( &$errors, $field, $value, $args ) {
...
 	public static function phone_format( $field ) {
-		if ( FrmField::is_option_empty( $field, 'format' ) ) {
-			$pattern = self::default_phone_format();
-		} else {
-			$pattern = FrmField::get_option( $field, 'format' );
+		$phone_type = FrmField::get_option( $field, 'phone_type' );
+		$pattern    = '';
+
+		switch ( $phone_type ) {
+			case 'international':
```
Crabcyborg, line 244: Is it possible to use the `frm_phone_pattern` function and move this code into Pro?

classes/models/fields/FrmFieldPhone.php
```diff
@@ -38,10 +38,57 @@ protected function field_settings_for_type() {
 			'size'           => true,
 			'clear_on_focus' => true,
 			'invalid'        => true,
-			'format'         => true,
```
Crabcyborg, line 41: I think there's a huge backward compatibility bug here that we need to fix.

If I have an old phone field with a "Format" setting,

Screen Shot 2024-03-27 at 11 04 05 AM

But when I check out this branch, you removed the format setting and it just shows "None" when in reality, it probably needs to update to "Custom" and fill in.

Screen Shot 2024-03-27 at 11 04 49 AM

classes/models/FrmEntryValidate.php
```diff
@@ -257,6 +274,17 @@ private static function default_phone_format() {
...
 	}
 
+	/**
+	 * Returns regex for validating international phone numbers.
+	 *
+	 * @since x.x
+	 *
+	 * @return string Regex pattern.
+	 */
+	private static function get_international_phone_regex() {
```
Crabcyborg, line 284: Can we get this function into Pro?

js/formidable_admin.js
shervElmi: Addressed in the [Remove null fallback; Fix bracket spacing](https://github.com/Strategy11/formidable-forms/pull/1590/commits/602a7ef33724c61e12b28baefb495d6626592527) commit.

js/formidable_admin.js
shervElmi: Addressed in the [Refactor event listener with util](https://github.com/Strategy11/formidable-forms/pull/1590/commits/e5fec17fd15aa367b92d28c9ffad656ea9c1f4f2) commit.

classes/views/frm-fields/back-end/phone/phone-type.php
```diff
@@ -0,0 +1,39 @@
...
+		data-field-id="<?php echo intval( $field_id ); ?>"
+	>
+		<option value="none" <?php selected( $phone_type, 'none' ); ?>>
+			<?php esc_html_e( 'None', 'formidable' ); ?>
+		</option>
+		<?php $this->get_international_option(); ?>
+		<option value="custom" data-dependency="<?php echo esc_attr( '#frm-phone-field-custom-format-' . $field_id ); ?>" <?php selected( $phone_type, 'custom' ); ?>>
```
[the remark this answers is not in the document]
shervElmi: Addressed in the [Use dynamic echoing for phone field custom format dependency](https://github.com/Strategy11/formidable-forms/pull/1590/commits/ce0d2bd354cd8cd8ee7da8e75c3148d4049b5dbf) commit.

shervElmi:
Hi @garretlaxton, I've addressed the issue with the country selector in the repeater fields. Appreciate your feedback.

Regarding the inclusion of the country selector in the backend, it might lead to complications with pre-existing phone numbers. For instance, older numbers such as 2015550123 lack a country code, making the integration of this feature problematic. Additionally, there's the consideration of users with custom number formats.

classes/models/fields/FrmFieldPhone.php
shervElmi: @Crabcyborg, you've already suggested that if a custom "format" input is specified, we should use it and automatically switch the setting to "Custom". Before I proceed with this approach, could you also take into account a scenario where a **new user** enters a value in the custom format field and then selects "International"? In such cases, should we revert the format back to "Custom"?

Crabcyborg: @shervElmi

What do you think about using the `format` setting still and not introducing a new `phone_type` at all.

- If `format` equals `international` then we use international.
- If `format` is empty, use none.
- If `format` equals anything else, it is custom and the open input will show the raw format value (the custom format).

This way we don't have to migrate anything. We're really just adding support for a new `international` value and that's it. We just need to have a flexible UI and some logic on save.

shervElmi: Thank you, Mike!

Introducing a new `phone_type` is part of the Project Requirements, as illustrated in the image below:

CleanShot 2024-04-09 at 21 32 33@2x


ClickUp Link: https://app.clickup.com/9006050445/v/dc/8ccuv4d-20191/8ccuv4d-16591?block=block-0c90a4c3-36ea-4ae3-9b79-dda72f036caf

Crabcyborg: @shervElmi Yes, as part of the UI. I'm not suggesting interface changes.

Just `phone_type` will never get saved in the database. It's just determined based on the value of `format`.

shervElmi: I've updated my code, incorporating the excellent suggestions you provided! Below are the links to the commits:

- For Formidable Forms: [Refactor phone number formatting and preservation logic](https://github.com/Strategy11/formidable-forms/pull/1590/commits/b72e25da9a86856f467cc3a61e941e73128e65e5)
- For Formidable Pro: [Refactor phone number formatting and preservation logic](https://github.com/Strategy11/formidable-pro/pull/4896/commits/058398604081bc852bef2e16c108ed53b0152906)

Now, I need to implement some logic when saving the form. I'm planning to make these changes in the Pro plugin. Could you advise where I might do this? I've looked in the `builder.js` file but haven't found a suitable location yet.

shervElmi: @Crabcyborg, Before I proceed with implementing the save form functionality, I'd like to gather your thoughts on a couple of points:

1. For the scenario where a user selects the "None" option, I'm considering removing the format value when the form is saved. Do you think this is a suitable approach? My concern is that this might affect existing users who have a custom format in place and are merely experimenting with the "None" option. Without removing the format value, the "Custom" option would reappear, rendering the "None" option ineffective. I'd appreciate your insights on this.

2. When a user chooses the "International" option, I plan to set the format to "International." Consequently, if the user later selects the "Custom" option, the format input field will display "International" instead of being empty.

classes/models/FrmEntryValidate.php
shervElmi: Addressed in the following commits:
For Formidable Forms: [Refactor phone number formatting and preservation logic](https://github.com/Strategy11/formidable-forms/pull/1590/commits/b72e25da9a86856f467cc3a61e941e73128e65e5)
For Formidable Pro: [Refactor phone number formatting and preservation logic](https://github.com/Strategy11/formidable-pro/pull/4896/commits/058398604081bc852bef2e16c108ed53b0152906)

classes/models/FrmEntryValidate.php
shervElmi: Addressed in the following commits:
For Formidable Forms: [Refactor phone number formatting and preservation logic](https://github.com/Strategy11/formidable-forms/pull/1590/commits/b72e25da9a86856f467cc3a61e941e73128e65e5)
For Formidable Pro: [Refactor phone number formatting and preservation logic](https://github.com/Strategy11/formidable-pro/pull/4896/commits/058398604081bc852bef2e16c108ed53b0152906)

classes/models/fields/FrmFieldPhone.php
Crabcyborg: @shervElmi

1. I think it's okay if the format is lost after it is set to none.
2. I would check if it is "International" and make it empty when changing it Custom.

shervElmi:
@Crabcyborg and @garrettlaxton, I have addressed all the feedback provided. Thank you both for your thorough reviews.

classes/models/fields/FrmFieldPhone.php
```diff
@@ -38,10 +38,46 @@ protected function field_settings_for_type() {
...
+	}
+
+	/**
+	 * Retrieves the HTML for an 'International' option in a dropdown.
+	 *
+	 * @since x.x
+	 *
+	 * @return void Outputs the HTML option tag directly.
+	 */
+	protected function get_international_option() {
```
Crabcyborg, line 67: Since this function doesn't get anything, `print` or `show` is a better verb here.

js/formidable_admin.js
```diff
@@ -6501,6 +6534,7 @@ function frmAdminBuildJS() {
 		}
 
 		document.addEventListener( 'click', handleUpgradeClick );
+		frmDom.util.documentOn( 'change', 'select.frm_select_with_upgrade', ( event ) => handleUpgradeClick( event ) );
```
Crabcyborg, line 6537: Is there a reason for the extra function here?

I would just pass the function
```
frmDom.util.documentOn( 'change', 'select.frm_select_with_upgrade', handleUpgradeClick );
```

js/formidable_admin.js
```diff
@@ -9775,6 +9817,47 @@ function frmAdminBuildJS() {
...
+				if ( ! dependencyElement ) {
+					return;
+				}
+
+				dependencyElement.classList.toggle( 'frm_hidden', selectedOption !== option );
+			});
+		}
+
+		// Initial setup: Show dependencies based on the current selection in each dropdown
+		selects.forEach( select => {
```
Crabcyborg, line 9853: I don't think we need the extra function here either.

We could just do `selects.forEach( toggleDependencyVisibility );`/

css/frm_admin.css
```diff
@@ -304,6 +304,16 @@ body.frm-white-body ul#adminmenu > li.current > a.current:after {
...
 }
 
+.frm_form_field.frm6_followed + p,
+.frm_form_field.frm6_followed + div {
+	grid-column: span 6 / span 6;
+}
+
+.frm_form_field.frm-phone-type + p > label,
+.frm_form_field.frm-phone-type + div > label {
+	opacity: 0;
```
Crabcyborg, line 314: What is the reason for this CSS? Could you add a CSS comment here?

classes/views/frm-fields/back-end/value-format.php
```diff
@@ -2,8 +2,15 @@
 if ( ! defined( 'ABSPATH' ) ) {
 	die( 'You are not allowed to call this page directly.' );
 }
+
+$attributes          = array();
```
Crabcyborg, line 6: It might be more fit to put this logic in the controller rather than in the view.

js/formidable_admin.js
```diff
@@ -5331,6 +5331,16 @@ function frmAdminBuildJS() {
 		clickAction( this );
 	}
 
+	function maybeUpdatePhoneFormatInput( event ) {
```
Crabcyborg, line 5334: A function description here might help explain why this exists. If it could explain the process that would be helpful for anyone who doesn't know how this works.

It would also be good to add a `@since` comment here.

js/formidable_admin.js
```diff
@@ -6478,6 +6488,29 @@ function frmAdminBuildJS() {
 			b.classList.add( 'frm_loading_button' );
 		}
 		b.setAttribute( 'aria-busy', 'true' );
+
+		const phoneTypes = document.querySelectorAll( '.frm_phone_type_dropdown' );
```
Crabcyborg, line 6492: This should probably be in a new function with a description of what is happening here.

js/formidable_admin.js
```diff
@@ -9775,6 +9817,47 @@ function frmAdminBuildJS() {
...
+		 *
+		 * @param {HTMLElement} select The select element whose dependencies need to be managed.
+		 * @return {void}
+		 */
+		function toggleDependencyVisibility( select ) {
+			const selectedOption = select.options[ select.selectedIndex ];
+			select.querySelectorAll( 'option[data-dependency]' ).forEach( option => {
+				const dependencyElement = document.querySelector( option.dataset.dependency );
+				if ( ! dependencyElement ) {
+					return;
```
Crabcyborg, line 9845: Since this is just a one-liner it's fine to make this a bit smaller and just check the happy path.

```
if ( dependencyElement ) {
    dependencyElement.classList.toggle( 'frm_hidden', selectedOption !== option );
}
```

Since there is no `else` we don't really need the `return`.

garretlaxton (approved):
Thanks Sherv! This looks good!

classes/models/fields/FrmFieldPhone.php
shervElmi: Addressed in the [Change the `get_international_option` to `print_international_option`](https://github.com/Strategy11/formidable-forms/pull/1590/commits/313c59709b9c183c84ebb0fc1d7eb7b80f1b804a), and [Change the `get_international_option` to `print_international_option`](https://github.com/Strategy11/formidable-pro/pull/4896/commits/a9af1e0438a3ad34cba25fbd5919500bf1257449) commits.

js/formidable_admin.js
shervElmi: Addressed in the [Remove extra function for handleUpgradeClick](https://github.com/Strategy11/formidable-forms/pull/1590/commits/37af0cce4dc6e883d727b26a7b5e8c125b528f96) commit.

> Is there a reason for the extra function here?
No, it was just a simple mistake!

js/formidable_admin.js
shervElmi: Addressed in the [Remove extra function for `toggleDependencyVisibility`](https://github.com/Strategy11/formidable-forms/pull/1590/commits/56eb0bde639b54fefd88a0df5840121d8cb209c0) commit.

classes/views/frm-fields/back-end/value-format.php
shervElmi: @Crabcyborg, this is a general view and we use it in several places, so I've decided to add those logics to the view file.

shervElmi: Addressed in the [Move logic from view file to controller](https://github.com/Strategy11/formidable-forms/pull/1590/commits/49982fb39f3898ea38ddaa58f97448a931c96156) commit.

js/formidable_admin.js
shervElmi: Addressed in the [Refactor code into new function with detailed description](https://github.com/Strategy11/formidable-forms/pull/1590/commits/ba09b6e6b619a392bbd8826a21f76e2e1837b79a) commit.

js/formidable_admin.js
```diff
@@ -9976,6 +10081,7 @@ function frmAdminBuildJS() {
 
 			jQuery( '.frm_submit_ajax' ).on( 'click', submitBuild );
 			jQuery( '.frm_submit_no_ajax' ).on( 'click', submitNoAjax );
+			wp.hooks.addAction( 'frmAdmin.preFormSave', 'formidable', adjustFormatInputBeforeSave );
```
Crabcyborg, line 10084: Is this hook only to call a function in the same file?

We could just call `adjustFormatInputBeforeSave` directly and avoid the new hook. Ideally if we don't need a hook we shouldn't be adding one.

shervElmi: Addressed in the [Call `adjustFormatInputBeforeSave` directly](https://github.com/Strategy11/formidable-forms/pull/1590/commits/d8ad2d165af4c1023712e4a58b22f174a9fdfc9b) commit.

Crabcyborg (approved):
I think this all looks good now.

Thanks @shervElmi!

🚀

PR #1593, opened 2024-03-27
Fix field shortcode items long text style

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4659
### Test steps

1. Create a form with 2 fields, make the first field's label very long.
2. In the other field, switch Default value to calculation. Confirm that the `[field_id]` shortcode text is visible for the first field despite the long label it has.

### Screenshots
**Before**:
image

**After:**
image

## Summary by CodeRabbit

- **Style**
	- Adjusted text alignment to left and margin properties for improved layout in the admin form.
	- Updated max-width and display properties for enhanced styling.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -7486,7 +7486,6 @@ ul.frm_two_col {
 
 .frm_code_list.frm-full-hover a span {
 	max-width: 83px;
-	text-align: left;
 	margin-right: var(--gap-xs);
 }
 
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -7490,10 +7490,13 @@ ul.frm_two_col {
 	margin-left: auto;
 }
 
+.frm_code_list.frm-full-hover a {
+	overflow: hidden;
+}
+
 .frm_code_list.frm-full-hover a,
 .frm_code_list.frm-full-hover a span {
 	white-space: nowrap;
-	overflow: hidden;
 	text-overflow: ellipsis;
 }
 
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -8492,6 +8492,10 @@ button.frm_choose_image_box,
 	background-position: left 11px top 55%;
 }
 
+.rtl .frm_code_list.frm-full-hover a span {
+	margin-left: var(--gap-xs);
+}
+
 /* ---------------------------------------------------------------
 Clearfix
 --------------------------------------------------------------- */
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -7491,7 +7491,6 @@ ul.frm_two_col {
 }
 
 .frm_code_list.frm-full-hover a {
-	max-width: 95%;
     display: block;
 }
 
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -7492,7 +7492,7 @@ ul.frm_two_col {
 
 .frm_code_list.frm-full-hover a {
 	max-width: 95%;
-    display: inline-block;
+    display: block;
 }
 
 .frm_code_list.frm-full-hover a,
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This looks great. But I think this would look better if they were all aligned to the left.
image

Also, would it make sense to add 3 dots to the end of the label where it is cut off? Ex.
image


Just some suggestions though.

garretlaxton:
Awesome, this looks great!

truongwp:
@Crabcyborg What do you think about using `display` `block` vs `inline-block`?

`inline-block`:

https://github.com/Strategy11/formidable-forms/assets/19748318/e60e542b-9adf-4eea-ab1b-e694ec49f3f4

`block`:

https://github.com/Strategy11/formidable-forms/assets/19748318/6deea452-c9af-4f66-9cf0-b6422e08f2df

css/frm_admin.css
```diff
@@ -7486,8 +7486,13 @@ ul.frm_two_col {
...
 .frm_code_list.frm-full-hover a span {
 	max-width: 83px;
-	text-align: right;
-	margin-left: auto;
+	text-align: left;
+	margin-right: var(--gap-xs);
+}
+
+.frm_code_list.frm-full-hover a {
+	max-width: 95%;
```
truongwp, line 7494: Look like after changing to `block`, we don't need this `max-width` anymore. Can you check and confirm this?

AbdiTolesa: Yes indeed, thanks @truongwp!

css/frm_admin.css
```diff
@@ -7486,8 +7486,12 @@ ul.frm_two_col {
 
 .frm_code_list.frm-full-hover a span {
 	max-width: 83px;
-	text-align: right;
-	margin-left: auto;
+	text-align: left;
```
truongwp, line 7489: This seems to be unnecessary too.

truongwp (approved):

css/frm_admin.css
```diff
@@ -7486,8 +7486,11 @@ ul.frm_two_col {
 
 .frm_code_list.frm-full-hover a span {
 	max-width: 83px;
-	text-align: right;
-	margin-left: auto;
+	margin-right: var(--gap-xs);
```
Crabcyborg, line 7489: How does this look in rtl?

Would we want margin-left instead?

Crabcyborg:
@tuguirazvan I could use your input here on a design change we're doing.

There are before and after images included at the top, in the PR description.

Right now, it seems that the list of options isn't really indented the same as the other UI above and below it. I feel like we may need to reduce indentation so it aligns.

What do you think?

image

tuguirazvan:
@Crabcyborg 

I don't mind the indentation because simply removing it might not always solve the issue, especially considering the varying lengths of labels.

In this context, our approach should prioritize displaying the label ID first, followed by the label text. If the text exceeds one line, we can truncate it and include an ellipsis. Let me know what do you think!

css/frm_admin.css
AbdiTolesa: Yes indeed, I have overridden that now for rtl, thank you!

Crabcyborg:
@AbdiTolesa It looks like this update extends to more than just the code pop up.

Can you update this so the customization sidebar in the visual views editor looks the same as before?

**Before**
Screen Shot 2024-04-08 at 2 31 24 PM

**After**
Screen Shot 2024-04-08 at 2 31 06 PM

AbdiTolesa:
> Can you update this so the customization sidebar in the visual views editor looks the same as before?

@Crabcyborg Here is a quick update:
https://github.com/Strategy11/formidable-forms/pull/1612

PR #1597, opened 2024-03-28
Add new filter for validation messages

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4819

### Test steps
1. Create a multi-page form.
2. Add fields with "Required" turned on different pages to make sure `data-reqmsg` is included for JS validation purpose.
3. Preview the form and inspect the HTML for fields NOT included on the current page.
4. Confirm that `data-invmsg` and `data-reqmsg` are not there.
5. Inspect the HTML for fields on the current page.
6. Confirm that both data attributes are there.

## Summary by CodeRabbit

- **Enhancements**
	- Improved field validation by selectively including specific messages based on field type and conditions, enhancing user feedback during form submissions.
	- Added a new test method to ensure the `add_validation_messages` method functions correctly for different field configurations.

### DIFF
--- a/classes/controllers/FrmFieldsController.php
+++ b/classes/controllers/FrmFieldsController.php
@@ -689,27 +689,15 @@ class FrmFieldsController {
 	}
 
 	private static function add_validation_messages( $field, array &$add_html ) {
-		if ( $field['type'] === 'hidden' ) {
-			$field_messages_status = array(
-				'data-invmsg' => false,
-				'data-reqmsg' => false,
-			);
-		} else {
-			$field_messages_status = array(
-				'data-invmsg' => true,
-				'data-reqmsg' => true,
-			);
-		}
+		$field_validation_messages = self::get_field_validation_messages_status( $field );
 
-		$field_messages_status = apply_filters( 'frm_field_messages_status', $field_messages_status, $field );
-
-		if ( FrmField::is_required( $field ) && $field_messages_status['data-reqmsg'] ) {
+		if ( FrmField::is_required( $field ) && ! empty( $field_validation_messages['data-reqmsg'] ) ) {
 			$required_message        = FrmFieldsHelper::get_error_msg( $field, 'blank' );
 			$add_html['data-reqmsg'] = 'data-reqmsg="' . esc_attr( $required_message ) . '"';
 			self::maybe_add_html_required( $field, $add_html );
 		}
 
-		if ( ! FrmField::is_option_empty( $field, 'invalid' ) && $field_messages_status['data-invmsg'] ) {
+		if ( ! FrmField::is_option_empty( $field, 'invalid' ) && ! empty( $field_validation_messages['data-invmsg'] ) ) {
 			$invalid_message         = FrmFieldsHelper::get_error_msg( $field, 'invalid' );
 			$add_html['data-invmsg'] = 'data-invmsg="' . esc_attr( $invalid_message ) . '"';
 		}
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This looks great!

classes/controllers/FrmFieldsController.php
```diff
@@ -705,6 +707,38 @@ private static function add_validation_messages( $field, array &$add_html ) {
...
+	/**
+	 * Returns an array that contains field validation messages status.
+	 *
+	 * @since x.x
+	 *
+	 * @param array $field_array
+	 * @return array
+	 */
+	private static function get_field_validation_messages_status( $field_array ) {
+		if ( $field_array['type'] === 'hidden' ) {
```
truongwp, line 719: We should use `FrmField::get_field_type()` to prevent error if `$field` is an object.

AbdiTolesa: I don't we need that since this is called from `add_validation_messages` which is private function and all functions called before that are assuming `$field` to be an array.

truongwp: This method is in the `input_html()` method, which is public. 
In `FrmDatesField.php`, we call `FrmFieldsController::input_html( $this->field, false );`. In the past, I got a problem with `$this->field` sometimes is `object`, and sometimes is `array`. So I think we should use the `get_field_type()` here.

AbdiTolesa: Oh, that makes sense then. Thanks!

classes/controllers/FrmFieldsController.php
```diff
@@ -705,6 +707,38 @@ private static function add_validation_messages( $field, array &$add_html ) {
...
 
+	/**
+	 * Returns an array that contains field validation messages status.
+	 *
+	 * @since x.x
+	 *
+	 * @param array|object $field
+	 * @return array
+	 */
+	private static function get_field_validation_messages_status( $field) {
```
truongwp, line 718: A space is missing after `$field`.

classes/controllers/FrmFieldsController.php
```diff
@@ -705,6 +707,38 @@ private static function add_validation_messages( $field, array &$add_html ) {
...
+
+		/**
+		 * Allows controlling which field validation messages would be included in the field html.
+		 *
+		 * @since x.x
+		 *
+		 * @param array $field_validation_messages_status
+		 * @param array $field_array
+		 */
+		return apply_filters( 'frm_field_validation_messages_status', $field_validation_messages_status, $field_array );
```
truongwp, line 739: You didn't change the name of `$field_array`. And please update the `@param` comment of this filter.

classes/controllers/FrmFieldsController.php
```diff
@@ -705,6 +707,38 @@ private static function add_validation_messages( $field, array &$add_html ) {
...
+			);
+		}
+
+		/**
+		 * Allows controlling which field validation messages would be included in the field html.
+		 *
+		 * @since x.x
+		 *
+		 * @param array $field_validation_messages_status
+		 * @param array $field
```
truongwp, line 737: This param type should be `array|object`.

truongwp (approved):

classes/controllers/FrmFieldsController.php
```diff
@@ -705,6 +707,38 @@ private static function add_validation_messages( $field, array &$add_html ) {
...
+
+		/**
+		 * Allows controlling which field validation messages would be included in the field html.
+		 *
+		 * @since x.x
+		 *
+		 * @param array $field_validation_messages_status
+		 * @param array|object $field
+		 */
+		return apply_filters( 'frm_field_validation_messages_status', $field_validation_messages_status, $field );
```
Crabcyborg, line 739: I'm not in love with this filter name.

I don't feel "status" is the best term when this is really just to toggle whether or not something is included. "status" to me is generally more than just 2 states.

I tried running a prompt by Chat GPT and I got back `frm_field_validation_include_data_attributes`. I think it's a bit more clear what the filter is for when it mentions "data_attributes" in the name as well.

What do you think about changing this?

classes/controllers/FrmFieldsController.php
```diff
@@ -705,6 +707,38 @@ private static function add_validation_messages( $field, array &$add_html ) {
...
 
+	/**
+	 * Returns an array that contains field validation messages status.
+	 *
+	 * @since x.x
+	 *
+	 * @param array|object $field
+	 * @return array
+	 */
+	private static function get_field_validation_messages_status( $field ) {
```
Crabcyborg, line 718: Similar to how I'd rename the filter, I think we could call this function something else.

We're already in the "fields" controller so `field` isn't necessary in this name.

We could probably go with something like `get_validation_data_attribute_visibility_info`.

classes/controllers/FrmFieldsController.php
AbdiTolesa: Thank you, @Crabcyborg! That makes more sense indeed.

Crabcyborg (approved):

🚀

PR #1600, opened 2024-03-29
Do js validation for confirmation field even if its empty

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4882

### Test steps
1. Create a form and add fields that support confirmation like email and password field to it.
2. Enable JS validation in the form
3. Preview the form and fill only the first fields and leaving confirmation fields empty
4. Confirm that a JS validation error is triggered and you cannot submit the form.

image

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved form field validation to handle conditions and error messages more efficiently.

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -492,7 +492,11 @@ function frmFrontFormJS() {
 			confirmValue = confirmField.value;
 		}
 
-		if ( ! ( value === '' && confirmValue === ''  ) && value !== confirmValue ) {
+		if ( value === '' && confirmValue === '' ) {
+			return;
+		}
+
+		if ( value !== confirmValue ) {
 			errors[ 'conf_' + strippedFieldID ] = getFieldValidationMessage( confirmField, 'data-confmsg' );
 			return
 		}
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This works great!

js/formidable.js
```diff
@@ -488,7 +488,7 @@ function frmFrontFormJS() {
 			firstField = document.getElementById( strippedId );
 			value = firstField.value;
 			confirmValue = confirmField.value;
-			if ( '' !== value && '' !== confirmValue && value !== confirmValue ) {
+			if ( ! ( value === '' && confirmValue === '' ) && value !== confirmValue ) {
```
truongwp, line 491: Look like we just need to check `value !== confirmValue`. Because `value === '' && confirmValue === ''` means `value === confirmValue`.

AbdiTolesa: I think that helps avoiding extra call to `getFieldValidationMessage` when both fields are empty, what do you think?

truongwp: `value !== confirmValue` covers that case, because both fields can't be empty with that check.

AbdiTolesa: Sorry for being confused and confusing back 😓 This updated too now.

truongwp (approved):
Thanks @AbdiTolesa. This looks good!

PR #1604, opened 2024-04-03
Update sorting indicators

Crabcyborg:
**Related ticket** https://secure.helpscout.net/conversation/2558367506/194987/

In newer versions of WordPress, the markup for the sorting has been updated. It caused the sorting indicators on the form and entry list pages to disappear.

This update pulls some of the newer WordPress code that this is copied from to use the new sorting indicator markup. I also added a new `aria-sort` attribute to the active sorted column like WordPress added to improve accessibility.

Screen Shot 2024-04-03 at 3 58 57 PM
Screen Shot 2024-04-03 at 3 59 07 PM

I dropped a ton of `@access` comments as well which don't really contribute to the code base. It looks like these are mostly dropped on the WordPress file as well.

## Summary by CodeRabbit

- **Refactor**
    - Removed `@access public` annotation from the `load_admin_hooks` method in the `FrmDashboardController` class.
    - Adjusted visibility levels and made method adjustments in the dashboard and list helper functionalities in `FrmListHelper.php`.
- **New Features**
    - Added a conditional check in `FrmEntriesController.php` to include additional columns based on specific criteria.
- **Accessibility Improvements**
    - Added aria attributes and accessible text in various parts of the application for better usability.

### DIFF
--- a/classes/controllers/FrmDashboardController.php
+++ b/classes/controllers/FrmDashboardController.php
@@ -23,8 +23,6 @@ class FrmDashboardController {
 	/**
 	 * Register all of the hooks related to the welcome screen functionality
 	 *
-	 * @access public
-	 *
 	 * @return void
 	 */
 	public static function load_admin_hooks() {
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -289,6 +289,12 @@ class FrmEntriesController {
 			$form_id . '_is_draft'   => 'is_draft',
 		);
 
+		if ( ! $form_id ) {
+			$columns[ $form_id . '_user_id' ] = 'user_id';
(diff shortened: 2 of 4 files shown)


garretlaxton:
When I click a field to sort, there is a border around it. Should that be there?
image

Also, the form, entry name, and created by fields don't have a sort option on the entries page. Should they have sort options?
image

Crabcyborg:
Thanks @garretlaxton. The border is consistent with how the sorting works on the list of Pages, so I think that isn't an issue.

I went ahead and made those other columns sortable. I don't see a big reason why not.

Sorting by "Form" will sort by Form ID, not by Form Name alphabetically. Same with sorting by "Created By" will sort by User ID order.

garretlaxton (approved):
Thanks for clarifying that, this looks great!

PR #1606, opened 2024-04-04
Fix submit button PR breaks Pro unit tests

Crabcyborg:
This gets the Pro unit tests passing.

It fixes this error.
image

The hook gets loaded in `load_form_hooks` but that isn't triggered when the submit button is getting generated.

This fixes the issue, but I don't love it.

Maybe there's a better option?

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved form submission behavior by adding conditional checks before setting up field data.
- **New Features**
	- Added a new public static function to the `FrmField` class for setting default field settings.

### DIFF
--- a/classes/helpers/FrmSubmitHelper.php
+++ b/classes/helpers/FrmSubmitHelper.php
@@ -142,6 +142,10 @@ class FrmSubmitHelper {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


truongwp (approved):
Thank you. This works great!

Crabcyborg:
Thanks @truongwp!

I don't actually think we need to fix this in our code. It seems like this is really only an issue in the unit test.

I think https://github.com/Strategy11/formidable-forms/pull/1608 is a better fix.

I'm going to close this.

PR #1610, opened 2024-04-08
Make dashboard videos show conditionally, add js required for the new global setting

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4823
Works along https://github.com/Strategy11/formidable-pro/pull/4938

## Summary by CodeRabbit

- **New Features**
	- Enhanced the dashboard with conditional video display, showing videos only if specific conditions are met.
	- Improved user interface responsiveness by dynamically adjusting visibility of elements based on user input.

- **Bug Fixes**
	- Fixed logic flow in form field processing to ensure consistent behavior.

### DIFF
--- a/classes/controllers/FrmDashboardController.php
+++ b/classes/controllers/FrmDashboardController.php
@@ -172,7 +172,7 @@ class FrmDashboardController {
 
 		if ( FrmAppHelper::pro_is_installed() ) {
 			$frmpro_settings = new FrmProSettings();
-			$should_display_formidable_videos = ! isset( $frmpro_settings->hide_dashboard_videos ) || ! $frmpro_settings->hide_dashboard_videos || $frmpro_settings->menu_icon === '';
+			$should_display_formidable_videos = empty( $frmpro_settings->hide_dashboard_videos ) || $frmpro_settings->menu_icon === '';
 		} else {
 			$should_display_formidable_videos = true;
 		}
--- a/classes/controllers/FrmDashboardController.php
+++ b/classes/controllers/FrmDashboardController.php
@@ -170,7 +170,12 @@ class FrmDashboardController {
 			)
 		);
 
-		$frmpro_settings = new FrmProSettings();
+		if ( FrmAppHelper::pro_is_installed() ) {
+			$frmpro_settings = new FrmProSettings();
+			$should_display_formidable_videos = ! isset( $frmpro_settings->hide_dashboard_videos ) || ! $frmpro_settings->hide_dashboard_videos || $frmpro_settings->menu_icon === '';
+		} else {
+			$should_display_formidable_videos = true;
+		}
 
 		require FrmAppHelper::plugin_path() . '/classes/views/dashboard/dashboard.php';
 	}
--- a/classes/controllers/FrmDashboardController.php
+++ b/classes/controllers/FrmDashboardController.php
@@ -170,11 +170,10 @@ class FrmDashboardController {
 			)
 		);
 
-		if ( FrmAppHelper::pro_is_installed() ) {
-			$frmpro_settings = new FrmProSettings();
-			$should_display_formidable_videos = empty( $frmpro_settings->hide_dashboard_videos ) || $frmpro_settings->menu_icon === '';
+		if ( FrmAppHelper::pro_is_installed() && is_callable( 'FrmProDashboardHelper::should_display_videos' ) ) {
+			$should_display_videos = FrmProDashboardHelper::should_display_videos();
 		} else {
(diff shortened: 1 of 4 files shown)


garretlaxton (approved):

classes/views/dashboard/dashboard.php
```diff
@@ -32,7 +32,11 @@
 			<div class="frm-dashboard-widget frm-card-item frm-license-widget">
 				<?php $dashboard_view->get_license_management(); ?>
 			</div>
-			<?php $dashboard_view->get_youtube_video( 'frm-dashboard-widget frm-card-item frm-yt-widget' ); ?>
+			<?php
+			if ( ! $frmpro_settings->hide_dashboard_videos || $frmpro_settings->menu_icon === '' ) {
```
truongwp, line 36: Please add the comment block at the top of this file and add `@var` tags for global variables in this file.

truongwp: And I get `Undefined property: FrmProSettings::$hide_dashboard_videos` on this line after checking out this branch.

classes/controllers/FrmDashboardController.php
```diff
@@ -169,6 +169,14 @@ public static function route() {
 				'video'    => array( 'id' => self::get_youtube_embed_video( $counters_value['entries'] ) ),
 			)
 		);
+
+		if ( FrmAppHelper::pro_is_installed() ) {
```
truongwp, line 173: In my opinion, we should leave a filter here, and move this code to Pro. @Crabcyborg What do you think?

Crabcyborg: @truongwp Since this is white labelling specific, I think we may want to avoid a hook so we have better control over when this is hidden or not.

In some cases it's fine to just call a function directly.

I'd try something like:
```
$should_display_videos = is_callable( 'FrmProDashboardHelper::should_display_videos' ) ? FrmProDashboardHelper::should_display_videos() : false;
```

classes/controllers/FrmDashboardController.php
```diff
@@ -169,6 +169,14 @@ public static function route() {
...
+			$frmpro_settings = new FrmProSettings();
+			$should_display_formidable_videos = empty( $frmpro_settings->hide_dashboard_videos ) || $frmpro_settings->menu_icon === '';
```
Crabcyborg, line 175: @AbdiTolesa For a white label check, I think we should also check the menu name.

We should check if `'Formidable' !== FrmAppHelper::get_menu_name()`.

AbdiTolesa: @Crabcyborg I'm not sure if I have gotten it but is that we want to hide the videos regardless of the new setting? Like this? => https://github.com/Strategy11/formidable-pro/pull/4938/files#diff-79d2904eb1e07b16d771da86124a10170c07e6b1515357a9dfbe6e2200c5a2bdR242

classes/views/dashboard/dashboard.php
```diff
@@ -2,6 +2,9 @@
 if ( ! defined( 'ABSPATH' ) ) {
 	die( 'You are not allowed to call this page directly.' );
 }
+/**
+ * @var bool $should_display_formidable_videos
```
Crabcyborg, line 6: It looks like this comment uses `should_display_formidable_videos` but the variable is now `should_display_videos`.

js/formidable_admin.js
```diff
@@ -10508,6 +10508,10 @@ function frmAdminBuildJS() {
 
 			// Set fieldsUpdated to 0 to avoid the unsaved changes pop up.
 			frmDom.util.documentOn( 'submit', '.frm_settings_form', () => fieldsUpdated = 0 );
+
+			frmDom.util.documentOn( 'change', 'input[name="frm_menu_icon"]', ( e ) => {
```
Crabcyborg, line 10512: Is this only in Lite because we don't have a Pro JS file yet?

We have a `license.js` file in Pro.

I think we could re-purpose it to encompass everything in the Global settings, not just the license settings.

classes/models/FrmField.php
```diff
@@ -733,6 +733,7 @@ public static function get_all_types_in_form( $form_id, $type, $limit = '', $inc
 				'inc_repeat' => $inc_sub,
 			)
 		);
+		$results = false;
```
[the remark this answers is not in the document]
AbdiTolesa: Oh! Sorry about that!

Crabcyborg (approved):
Thanks @AbdiTolesa!

🚀

PR #1612, opened 2024-04-08
Undo changes to modal in views editor

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4659

Just reverting fields modal style in views editor.
**Before:**
image

image

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -7504,7 +7504,11 @@ ul.frm_two_col {
 	margin-right: var(--gap-xs);
 }
 
-.frm_code_list.frm-full-hover a {
+.frm-views-editor-body .frm_code_list.frm-full-hover a span {
+	margin-left: auto;
+}
+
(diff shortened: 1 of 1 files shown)


css/frm_admin.css
```diff
@@ -7504,7 +7504,11 @@ ul.frm_two_col {
 	margin-right: var(--gap-xs);
 }
 
-.frm_code_list.frm-full-hover a {
+.frm-views-editor-body .frm_code_list.frm-full-hover a span {
```
Crabcyborg, line 7507: I don't believe this class is used in Legacy views.

Could you try out version [v4.0.05](https://github.com/Strategy11/formidable-views/releases/tag/v4.0.05) of Views as well to make sure it looks the same as before?

Crabcyborg (approved):

I think this looks good now.

🚀

PR #1615, opened 2024-04-11
Add global setting for form submission message

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/3739

This PR adds a feature that announces a form submission message with screen reader when a form is submitted. There is also a new global setting added for updating the message.

image

### Test steps:
1. Go to the global settings and check that a new setting 'Submit Speak Text' is there.
2. Modify it as needed for your test.
3. Turn on a screen reader and try submitting a form.
4. Confirm that the message is announced by your screen reader.


garretlaxton (commented):
I am guessing this should get cut off when the form request is complete? I hear the text, but it stops after the form request is complete.

AbdiTolesa:
> I am guessing this should get cut off when the form request is complete? I hear the text, but it stops after the form request is complete.

@garretlaxton I think you mistyped some words in your comment probably? I didn't get it well.

Crabcyborg:
@AbdiTolesa It sounds like the page exits before the message is read.

I think we may be handling this from the wrong angle. I don't think we need to announce the submit event, but rather, the success event.

If the confirmation action is a message, I think we could probably just add `aria-live` to the success message, then that will be read out.

If there is a redirect action, we're probably fine to just not speak/add anything.

Crabcyborg:
Thanks @AbdiTolesa!

Maybe we can go with a shorter message. What you have is really long.

I think just "Form submitted" is probably enough.

AbdiTolesa:
> Thanks @AbdiTolesa!
> 
> Maybe we can go with a shorter message. What you have is really long.
> 
> I think just "Form submitted" is probably enough.

Should I keep the new global setting to customise that or we should hard code it? Or probably make it customisable with a new filter?

Crabcyborg:
@AbdiTolesa I think just a new filter is fine. This feature doesn't need a lot of visibility.

AbdiTolesa:
> @AbdiTolesa I think just a new filter is fine. This feature doesn't need a lot of visibility.

I have updated my code to just use a new filter.

js/formidable.js
```diff
@@ -1649,8 +1649,28 @@ function frmFrontFormJS() {
 			frmFrontForm.submitFormNow( object );
 		},
 
+		handleSubmit: function( e ) {
```
Crabcyborg, line 1652: Can we just call this in `frmFrontForm.submitForm` instead?

truongwp (approved):

Crabcyborg:
I'm still not sold on this change.

I think we need to come up with something better if we're going to make any changes at all.

PR #1617, opened 2024-04-11
Fix issue 4944 / Drag to index 0 when there are no fields

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/4944

This makes it work more like it did in v6.8.4. When there are no fields, the index is always 0, so the drag marker always shows at the top, above the no fields element.

Screen Shot 2024-04-11 at 2 51 09 PM

**Pre-release,**
[formidable-6.9b.zip](https://github.com/Strategy11/formidable-forms/files/14949646/formidable-6.9b.zip)

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1145,6 +1145,11 @@ function frmAdminBuildJS() {
 
 		let index, item, itemTop, returnIndex;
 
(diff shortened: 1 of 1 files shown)


truongwp (approved):
This works great. Thanks @Crabcyborg!

PR #1622, opened 2024-04-16
Add div with site-branding class when twenty nineteen is active

AbdiTolesa:
Related issue https://github.com/Strategy11/formidable-landing/issues/11

image

### Test steps
1. Install and activate Twenty Nineteen theme.
2. Create a menu and assign it to Primary menu (so that the js script being tested loads on the page)
3. Preview the form **On Blank page** and confirm that there is no JS error in the console.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved compatibility with the `twentynineteen` theme by conditionally adding specific HTML elements to ensure proper layout.
	- Enhanced conditional check for the existence of the function `twentynineteen_setup` in `form.php` and `direct.php` files.
	- Added a `<div>` with the class `site-branding frm_hidden` before displaying the form if the function is callable.
	- Deprecated the usage of `print_emoji_styles` by adding a `remove_action` call in the `load_direct_preview` method of `FrmFormsController.php`.

### DIFF
--- a/classes/views/frm-entries/form.php
+++ b/classes/views/frm-entries/form.php
@@ -27,6 +27,7 @@ FrmFormsController::maybe_load_css( $form, $values['custom_style'], $frm_vars['l
 
 // Get conditionally hidden fields
 $frm_hide_fields = FrmAppHelper::get_post_param( 'frm_hide_fields_' . $form->id, '', 'sanitize_text_field' );
+
 ?>
 <div class="frm_form_fields <?php echo esc_attr( apply_filters( 'frm_form_fields_class', '', $values ) ); ?>">
 <fieldset>
--- a/classes/views/frm-entries/form.php
+++ b/classes/views/frm-entries/form.php
@@ -28,6 +28,11 @@ FrmFormsController::maybe_load_css( $form, $values['custom_style'], $frm_vars['l
 // Get conditionally hidden fields
 $frm_hide_fields = FrmAppHelper::get_post_param( 'frm_hide_fields_' . $form->id, '', 'sanitize_text_field' );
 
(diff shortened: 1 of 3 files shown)


classes/views/frm-entries/form.php
```diff
@@ -28,6 +28,11 @@
 // Get conditionally hidden fields
 $frm_hide_fields = FrmAppHelper::get_post_param( 'frm_hide_fields_' . $form->id, '', 'sanitize_text_field' );
 
+if ( is_callable( 'twentynineteen_setup' ) ) {
+	?>
+<div class="site-branding frm_hidden"></div>
```
Crabcyborg, line 33: @AbdiTolesa It would be great if we could avoid adding this to this view.

Is this only an issue on preview pages? Maybe we can put it in `FrmFormsController::preview` instead?

AbdiTolesa: Thank you for the thought, @Crabcyborg! The issue is limited to when previewing on blank page and I have moved the fix to `direct.php`. I have created https://github.com/Strategy11/formidable-landing/pull/31 too for fixing the issue when previewing on landing page.

AbdiTolesa: Should we try reusing the fix in Landing plugin or are we okay duplicating it?

Crabcyborg: I think having it duplicated is fine.

Thanks for updating this!

Crabcyborg (approved):
Thanks @AbdiTolesa!

I think this is simple enough we can skip QA.

I added a small commit. I noticed I was seeing this deprecated message as well,
> PHP Deprecated:  Function print_emoji_styles is deprecated since version 6.4.0! Use wp_enqueue_emoji_styles instead. in /var/www/src/wp-includes/functions.php on line 6078

PR #1631, opened 2024-04-17
Exit early if call to mb_function fails

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/3321

I actually don't know how to replicate the issue so didn't test it, but I just believe this PR fixes the issue from points I collect from Mike's investigations on the issue.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the `truncate` method in `FrmAppHelper.php` for improved clarity and robustness in handling various input scenarios.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -2508,13 +2508,13 @@ class FrmAppHelper {
 		return $custom_style;
 	}
 
-	public static function truncate( $str, $length, $minword = 3, $continue = '...' ) {
-		if ( is_array( $str ) ) {
+	public static function truncate( $original_string, $length, $minword = 3, $continue = '...' ) {
+		if ( is_array( $original_string ) ) {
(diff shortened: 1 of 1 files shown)


classes/helpers/FrmAppHelper.php
```diff
@@ -2508,13 +2508,13 @@ public static function custom_style_value( $post_values ) {
 		return $custom_style;
 	}
 
-	public static function truncate( $str, $length, $minword = 3, $continue = '...' ) {
-		if ( is_array( $str ) ) {
+	public static function truncate( $original_string, $length, $minword = 3, $continue = '...' ) {
+		if ( is_array( $original_string ) ) {
```
Crabcyborg, line 2512: We should probably also change this to `! is_string`.

Crabcyborg (approved):
Thanks @AbdiTolesa!

I think this is good to go.

🚀

PR #1635, opened 2024-04-18
Issue 4799 logged out editing education

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4799

### DIFF
--- a/js/bootstrap-multiselect.js
+++ b/js/bootstrap-multiselect.js
@@ -563,10 +563,17 @@
             }
 
             this.$popupContainer.on("touchstart click", function (e) {
+                if ( this.optionIsDisabled( e.currentTarget ) ) {
+                    return;
+                }
                 e.stopPropagation();
             });
 
             this.$container.append(this.$popupContainer);
+        },
+
+        optionIsDisabled: function( el ) {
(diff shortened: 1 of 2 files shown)


js/bootstrap-multiselect.js
```diff
@@ -563,6 +563,9 @@
             }
 
             this.$popupContainer.on("touchstart click", function (e) {
+                if ( e.target.closest( '.frm_disabled_option' ) ) {
```
Crabcyborg, line 566: @AbdiTolesa This is a third party plugin. We can't update it directly - we just replace the full file when there's a new version.

garretlaxton (approved):
This looks good!

Crabcyborg (commented):
@AbdiTolesa

It looks like this icon is the closed version of `frm_lock_open_icon`.

For consistency, I think we should use a similar name and include its definition beside `frm_lock_open_icon`.

I think `frm_lock_closed_icon` is probably a good name.

Screen Shot 2024-04-24 at 9 36 29 AM

Crabcyborg (approved):
I think this is good.

🚀

PR #1645, opened 2024-04-19
Slevomat standard / Sort array keys in alphabetical order

Crabcyborg:


Crabcyborg:
I think this rule is likely not worth moving forward with. This would mess up the order of some things.

PR #1686, opened 2024-04-26
Feature/calendar views

Liviu-p:
Fixes https://github.com/Strategy11/formidable-pro/issues/4409

## Summary by CodeRabbit

- **New Features**
	- Added a method to check for specific classnames in inline styles.
- **Style**
	- Updated CSS to enable `flex-wrap` for radio control options in forms.


classes/helpers/FrmStylesHelper.php
```diff
@@ -755,6 +755,34 @@ function( $value ) {
...
 	}
 
+	/**
+	 * Check if a classname is already in the inline style of a stylesheet.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $classname The classname to check for.
+	 * @param string $style_handle The style handle to check.
+	 * @return boolean
```
Crabcyborg, line 765: @Liviu-p Can we do this without requiring updates in Lite?

Can we just add this function into Views?

css/admin/frm_admin_global.css
```diff
@@ -7,3 +7,34 @@
 	font-size: 9px;
 	vertical-align: middle;
 }
+button.frm-loading {
```
Crabcyborg, line 10: Is this style only for the block?

Maybe we should just add it there. This global file gets loaded on every admin page so it's nice to keep it really small.

images/blocks/views-block-placeholder.jpg
Crabcyborg, line 1: It looks like this file isn't optimized.

If I run it in https://imagecompressor.com/ it goes from 54KB to 12KB.

Let's optimize this so it takes less space and loads faster.

js/src/form/views.js
```diff
@@ -0,0 +1,97 @@
...
+		});
+
+		const activateViewsAddon = () => {
+			if ( true === addonActivateButton.isLoading ) {
+				return;
+			}
+			updateAddonActivateButton({ ...addonActivateButton, isLoading: true, classnames: addonActivateButton.defaultClassname + ' ' + addonActivateButton.loadingClassname });
+			if ( ! blockData.viewsAddon.installed && blockData.viewsAddon.hasAccess ) {
+				frmAddonAPI.toggleAddonState( 'frm_install_addon', blockData.viewsAddon.url ).then( () => {
+					window.location.reload();
```
Crabcyborg, line 40: Would reloading cause issues if there were unsaved changes in the Gutenberg editor?

classes/helpers/FrmStylesHelper.php
Liviu-p: Hey @Crabcyborg, this was moved to Views. My initial idea was that it could be used for more general cases.

css/admin/frm_admin_global.css
Liviu-p: of course @Crabcyborg, I moved it into a separate css file. It's loaded the block js file only from now.

images/blocks/views-block-placeholder.jpg
Liviu-p: good catch, I optimised it. Thank you!

js/src/form/views.js
Liviu-p: The reload issue is handled automatically by default in Gutenberg. For unsaved changes it will trigger a confirm modal before reloading.
Screenshot 2024-05-03 at 16 18 54

Crabcyborg (commented):
@Liviu-p I think something is off here. I only ever see the "Install" placeholder. I can't seem to get to the views block to pick a view.

I think our block shouldn't include `-placeholder` in the name. If someone installs Views, it should just replace the placeholder block with the real one.

Crabcyborg:
The other issue above might just be because I was changing stuff. But I do see two "Formidable Views" matches which we don't want.

Screen Shot 2024-05-03 at 5 02 14 PM

Crabcyborg:
I pushed updates to remove the `-placeholder.` I'm not sure what I was seeing earlier - this all seems to work well for me now.

Crabcyborg (approved):
I think this is good to go.

Thanks @Liviu-p!

🚀

PR #1689, opened 2024-04-29
Support conditional logic with stripe link

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4908

Reflects changes in Stripe add-on (https://github.com/Strategy11/formidable-stripe/pull/332) in Lite.

## Summary by CodeRabbit

- **New Features**
	- Enhanced Stripe payment forms with conditional logic to enable or disable the submit button based on payment element completion, improving user interaction and form handling.

### DIFF
--- a/stripe/js/frmstrp.js
+++ b/stripe/js/frmstrp.js
@@ -581,7 +581,7 @@
 	 * Check if the submit button is conditionally disabled.
 	 * This is required for Stripe link so the button does not get enabled at the wrong time after completing the Stripe elements.
 	 *
-	 * @since 3.0
+	 * @since x.x
 	 *
 	 * @param {String} formId
 	 * @returns {bool}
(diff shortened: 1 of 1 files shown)


Crabcyborg (commented):
@AbdiTolesa Is it possible to move more of this logic into the Pro `frmstrp.js` file instead?

Ideally this file only handles Lite requirements.

Crabcyborg:
@AbdiTolesa This update specifically should be in Pro since conditional logic is a Pro feature. Lite has no reason to include logic for Pro features (ie. conditional logic).

AbdiTolesa:
> @AbdiTolesa This update specifically should be in Pro since conditional logic is a Pro feature. Lite has no reason to include logic for Pro features (ie. conditional logic).

Understood but there is a bunch of new code to copy to Pro's `frmstrp.js`. I wonder if we want to reuse lite Stripe's function instead of that, probably using global variables but I'm not sure which global variable we should use if so. What do you think about that approach?

Crabcyborg:
@AbdiTolesa We can't use code from the Stripe add-on because Stripe Lite is only intended to work when there is no Stripe add-on.

The Stripe add-on fully replaces Stripe Lite.

PR #1691, opened 2024-04-29
Use decimal for a single decimal place fix

AbdiTolesa:
Reflects changes in https://github.com/Strategy11/formidable-payments/pull/81 which fixes https://github.com/Strategy11/formidable-pro/issues/4908

Quoting the original PR description:
Related ticket https://secure.helpscout.net/conversation/2493586368/188113/

> The payment amount is not accurate. It is setting a price of 1.115,00 € when it should be 111,50.

It would convert `111.50` to `111,50` but wasn't converting `111.5` to `111,5`, which causes the numbers to be off 10x more than expected.

I don't see harm in converting numbers with a single decimal place as well. I added a unit test to cover 3 "decimal places", which shouldn't be interpreted as a decimal.

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of decimal amounts in transactions to improve accuracy in financial calculations.
- **Tests**
	- Added new tests to ensure the reliability of decimal amount processing in transactions.


AbdiTolesa:
@garretlaxton I haven't tested this since my Stripe account is blocked now but I believe it should work for you.

garretlaxton (approved):
This looks great!

stripe/controllers/FrmTransLiteActionsController.php
```diff
@@ -275,8 +275,14 @@ private static function maybe_use_decimal( &$amount, $currency ) {
...
 
-		$amount_parts     = explode( '.', $amount );
-		$used_for_decimal = count( $amount_parts ) === 2 && strlen( $amount_parts[1] ) === 2;
+		$amount_parts = explode( '.', $amount );
+		if ( 2 !== count( $amount_parts ) ) {
+			return;
+		}
+
+		$strlen           = strlen( $amount_parts[1] );
+		$used_for_decimal = $strlen >= 1 && $strlen <= 2;
```
truongwp, line 284: Should this be `$strlen === 1 || `$strlen === 2`?

AbdiTolesa: I think so, thanks!

truongwp (approved):

Crabcyborg (approved):

🚀

PR #1695, opened 2024-04-30
Call pro intl phone validation function instead of checking regex pat…

Crabcyborg:
…tern for intl phones

Related PR https://github.com/Strategy11/formidable-pro/pull/5032


shervElmi (approved):
Great @Crabcyborg! :raised_hands::rocket:

PR #1707, opened 2024-05-05
Fix: [i18n] The strings in Applications are not displayed translated

DAnn2012:
Fix: [i18n] The strings in Applications are not displayed translated


Crabcyborg (approved):
Thanks @DAnn2012!

🚀

PR #1708, opened 2024-05-05
Fixed text domain in applications.js and style.js

DAnn2012:
Fixed text domain in applications.js and style.js


Crabcyborg (approved):
Thanks @DAnn2012!

🚀

PR #1710, opened 2024-05-06
Update quiz action type limit message

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-quizzes/issues/198

Requires https://github.com/Strategy11/formidable-quizzes/pull/201

### Screenshots
image

### Test steps
1. Add one type of a Quiz action (either Scored or Outcome)
2. Try adding the other Quiz type not already added
3. Confirm that a message appears that states only one Quiz action type could be included in a given form.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -7324,7 +7324,21 @@ function frmAdminBuildJS() {
 			}
 		}
 
+		if ( frmAdminJs.quiz_action_type_limit && shouldShowQuizActionLimitMessage( this ) ) {
+			message = frmAdminJs.quiz_action_type_limit;
+		}
+
 		infoModal( message );
+	}
+
+	function shouldShowQuizActionLimitMessage( actionLi ) {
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This looks great!

js/formidable_admin.js
```diff
@@ -7324,9 +7324,29 @@ function frmAdminBuildJS() {
 			}
 		}
 
+		if ( frmAdminJs.quiz_action_type_limit && shouldShowQuizTypeLimitMessage( this ) ) {
```
truongwp, line 7327: @Crabcyborg Do you think we should add a filter here and then modify the message in the Quizzes addon?

Crabcyborg: @truongwp I think we should avoid quiz code here. I think it's possible we could figure this out in a way that might not require a lite update at all though.

@AbdiTolesa We shouldn't have anything explicitly about quizzes here.

The logic for disabling quiz outcomes/scored quizzes is already entirely in the quizzes add-on. See the `frm-quizzes-form-action.js` file. We have functions there, `enableAction` and `disableAction`.

Maybe we can use different class names to avoid needing to touch Lite entirely?

PR #1715, opened 2024-05-06
Add subscription limit setting and a feature to auto-cancel a subscription after it reaches the payment limit

AbdiTolesa:
Syncs changes in https://github.com/Strategy11/formidable-stripe/pull/330

image

### Test steps
1. Go to form actions and add a Payment action
2. Set the Payment type to 'recurring'
3. Confirm that a new setting appears that allows configuring the payment limit
4. Configure all the required settings and save.
5. Confirm that the subscription is cancelled after the limit is hit


Crabcyborg:
@AbdiTolesa I wouldn't worry too much about getting unit tests running in Lite.

A lot of tests rely on the test secret, which we don't want to expose in a public repo.

tests/phpunit/stripe/FrmStrpLiteUnitTest.php
```diff
@@ -0,0 +1,544 @@
...
+	protected $use_test_credit_card_number;
+
+	public function setUp(): void {
+		parent::setUp();
+
+		// Reset settings so modified settings in one test don't break other tests.
+		$this->set_private_property( 'FrmStrpLiteAppHelper', 'settings', null );
+	}
+
+	protected function initialize_legacy_api() {
```
Crabcyborg, line 70: Anything "legacy" isn't part of the Stripe add-on. We don't want this.

tests/phpunit/stripe/FrmStrpLiteUnitTest.php
```diff
@@ -0,0 +1,544 @@
...
+	private function set_static_account_details_for_reuse( $initialized ) {
+		self::$shared_account_id        = $initialized->account_id;
+		self::$shared_connect_url       = $initialized->connect_url;
+		self::$shared_server_side_token = $initialized->password;
+		self::$shared_client_side_token = get_option( 'frm_strp_connect_client_password_test' );
+		$this->account_id               = $initialized->account_id;
+	}
+
+	private function get_secret_key() {
+		return getenv( 'FRM_STRP_API_TEST_SECRET' );
```
Crabcyborg, line 134: Let's not try to get this environment variable in Lite. if a test requires it, drop the test for now.

tests/phpunit/stripe/FrmStrpLiteUnitTest.php
```diff
@@ -0,0 +1,544 @@
...
+	}
+
+	protected function get_authenticated_stripe_client() {
+		return new \Stripe\StripeClient( $this->get_secret_key() );
```
Crabcyborg, line 138: We don't include the Stripe library in Lite. Any reference to `StripeClient` should get removed.

tests/phpunit/stripe/FrmStrpLiteUnitTest.php
```diff
@@ -0,0 +1,500 @@
...
+class FrmStrpLiteUnitTest extends FrmUnitTest {
+
+	public static function wpSetUpBeforeClass() {
+		parent::wpSetUpBeforeClass();
+		self::empty_tables();
+		self::frm_install();
+	}
+
+	/**
+	 * @var string $shared_account_id used to limit the number of accounts when testing: the static scope variables stick around from test to test.
```
Crabcyborg, line 12: @AbdiTolesa The issue you're seeing is that there's no benefit in including var names in var comment descriptions.

For example, on this line, you would remove `$shared_account_id`. We already know the name of the variable since this comment only covers `$shared_account_id`.

You can run php cs fixer by running `npm run php-cs-fixer`. If it doesn't work, make sure to run `composer install` first. It should be able to fix this automatically.

AbdiTolesa: Thanks!

AbdiTolesa:
@Crabcyborg Is there another approach to test the subscription cancellation aside from waiting for days and see the result that @garretlaxton can use to test this?

AbdiTolesa:
> @AbdiTolesa There's a full set of steps in the original PR.
> 
> Check out this comment [Strategy11/formidable-stripe#330 (comment)](https://github.com/Strategy11/formidable-stripe/pull/330#issuecomment-1959942373)

Thanks, it has passed the test for me!

garretlaxton (approved):
This works great!

stripe/controllers/FrmStrpLiteEventsController.php
```diff
@@ -238,12 +238,84 @@ private function prepare_from_invoice() {
...
+		}
+
+		if ( $this->get_payments_count( $sub->id ) < $payment_limit ) {
+			return;
+		}
+
+		// Flag to cancel subscription at period end.
+		// In this case, we do not want to cancel immediately.
+		$hook   = 'frm_stripe_cancel_subscription_at_period_end';
+		$filter = function () {
```
truongwp, line 282: We can use `'__return_true'` instead.

Crabcyborg: @truongwp The problem with using `__return_true` is if someone adds that themselves (as custom code), we remove the hook afterward - introducing side effects.

truongwp: Oh. Thank you. I didn't think about that case.

stripe/controllers/FrmStrpLiteEventsController.php
```diff
@@ -238,12 +238,84 @@ private function prepare_from_invoice() {
...
+
+	/**
+	 * Get the count of completed payments.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $sub_id Stripe subscriptino id prefixed with 'sub_'.
+	 * @return int
+	 */
+	private function get_payments_count( $sub_id ) {
```
truongwp, line 307: This has the same code https://github.com/Strategy11/formidable-forms/blob/2472158fec39d4b875a900b1864515b0fa02a623/stripe/helpers/FrmTransLiteListHelper.php#L463-L471

We can move this method and reuse it in the line above.

stripe/controllers/FrmStrpLiteEventsController.php
```diff
@@ -238,12 +238,100 @@ private function prepare_from_invoice() {
...
+	/**
+	 * Returns the count of completed payments.
+	 *
+	 * @since x.x
+	 *
+	 * @param array $payments
+	 *
+	 * @return int
+	 */
+	public static function count_completed_payments( $payments ) {
```
truongwp, line 324: I think this method should be in a helper, maybe `FrmTransLiteAppHelper`.

We can also add the second param for the status.

stripe/controllers/FrmStrpLiteEventsController.php
```diff
@@ -238,12 +238,80 @@ private function prepare_from_invoice() {
...
+	 *
+	 * @since x.x
+	 *
+	 * @param object $sub
+	 * @return void
+	 */
+	private function maybe_cancel_subscription( $sub ) {
+		$action = FrmFormAction::get_single_action_type( $sub->action_id, 'payment' );
+		// @phpstan-ignore-next-line
+		if ( ! ( $action instanceof WP_Post ) || empty( $action->post_content['payment_limit'] ) ) {
```
truongwp, line 260: I remember that I got a problem with checking if form action is instance of `WP_Post` in the past. The is a case the `$form_action` is an object like `WP_Post` but isn't a `WP_Post` object.

So I think instead of `! ( $action instanceof WP_Post )`, we should check `is_object( $action )`.

stripe/controllers/FrmStrpLiteEventsController.php
```diff
@@ -238,12 +238,80 @@ private function prepare_from_invoice() {
...
+			return;
+		}
+
+		$payment_limit = FrmStrpLiteSubscriptionHelper::prepare_payment_limit(
+			$action->post_content['payment_limit'],
+			// Form ID.
+			(int) $action->menu_order,
+			(int) $sub->item_id
+		);
+		if ( $payment_limit instanceof WP_Error ) {
```
truongwp, line 270: We can use `is_wp_error()` here.

truongwp (approved):

Crabcyborg (commented):
It looks like you're missing this update that was added in the payments submodule on this line https://github.com/Strategy11/formidable-payments/blob/e56bb154990e2bc120b97551d779b72fcbfbb22e/models/FrmTransAction.php#L41

```
if ( ! isset( $form_action->post_content['payment_limit'] ) ) {
	$form_action->post_content['payment_limit'] = '';
}
```

Without that I'm seeing a PHP Warning:

```
PHP Warning:  Undefined array key "payment_limit" in /var/www/src/wp-content/plugins/formidable/stripe/views/action-settings/payments-options.php on line 64
```

Crabcyborg (approved):

PR #1723, opened 2024-05-07
Use field keys for errors not ids

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/2590327517/198659

The inconsistency here is that we usually use the `html_id` for the error which is really `'field_' . $field['field_key']` and possibly more for an embedded/repeated form.

The `key` variable in this case is really an integer field ID, so it's inconsistent.

To replicate,
1. Add a field to a form. Make it required
2. Enable AJAX submit.
3. Submit the form.

**Expected result**
- The error should use the id `frm_error_field_29yf4d`.

**Actual result**
- The error is formatted like `frm_error_field_3` instead.

**Pre-release,**
[formidable-6.9.2b.zip](https://github.com/Strategy11/formidable-forms/files/15238502/formidable-6.9.2b.zip)

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -985,16 +985,11 @@ function frmFrontFormJS() {
 	 * @return {string}
 	 */
 	function getErrorElementId( key, input ) {
-		var split;
-
 		if ( isNaN( key ) || ! input.id ) {
 			// If key isn't a number, assume it's already in the right format.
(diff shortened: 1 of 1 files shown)


AbdiTolesa (approved):
This looks good, thanks @Crabcyborg!

Crabcyborg:
Thanks Abdi!

PR #1744, opened 2024-05-14
Fix JS translation

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/5056

I ran `npm run makepot` to update the pot file. I tested and it worked.

@Crabcyborg In the old pot file: `POT-Creation-Date: 2024-05-06T15:22:32+00:00\n`. So this file is generated in the latest release. `npm run makepot` is called in `bin/build-plugin.sh` file.

I'm not sure what's wrong with the previous pot file.

@garretlaxton Can you please test this? You need to deactivate Formidable Pro and Formidable Bootstrap Modal plugins. Then use Loco Translate (or any translation tools) to translate block strings mentioned in the issue. And then check if those strings are translated in the block.

### DIFF
--- a/languages/formidable.pot
+++ b/languages/formidable.pot
@@ -9,9 +9,9 @@ msgstr ""
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"POT-Creation-Date: 2024-05-06T15:22:32+00:00\n"
+"POT-Creation-Date: 2024-05-14T04:17:35+00:00\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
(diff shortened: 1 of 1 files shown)


garretlaxton:
I'm not able to find these strings for some reason.
image
image

In Loco Translate, it shows `Display a Formidable Form`, but on the page, it shows `Display a Form` I tried to translate it, but it doesn't show.

truongwp:
@garretlaxton Can you give me the full screen of that image? The `Sync` button should be clickable.

garretlaxton:
Sure thing!
image

truongwp:
@garretlaxton Can I access that site? It works on my local.

truongwp:
@Crabcyborg Abdi also created #1745 to fix this issue using PHP translation. Do you think we should continue this one?

PR #1746, opened 2024-05-14
Use FrmDb::get_results to avoid issue with repeated wp_unslash calls on a field

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5064

The issue seems to be caused by cached `FrmFields::get_all_for_form()` result being passed to `wp_unslash` more than once, causing backslashes to be removed totally. That is because `FrmFields::getAll` calls `FrmDb::getVar` which checks cache first (which is already passed to wp_unslash()), and then strips the slashes again before returning the result. In db backslashes are escaped when saved (`\\`) so that a single `wp_unslash` call should only strip a single slash at a time.

### Test steps

1. Add a text field with the format ^(21|22|23|24)\d{5}$.
2. Enable "Load and save form builder page with AJAX" in form settings.
3. Save the field and reload the form.
4. Select the field again.
5. Confirm that the format is not changed.
6. Preview the form and test the format validation inputting a text in the field.


garretlaxton (approved):
This works great!

truongwp:
@AbdiTolesa Can we fix this issue inside the `get_all_for_form()` method? With this PR, other places which use `get_all_for_form()` could get this issue.

AbdiTolesa:
> @AbdiTolesa Can we fix this issue inside the `get_all_for_form()` method? With this PR, other places which use `get_all_for_form()` could get this issue.

@truongwp That would need avoiding cache for all `get_all_for_form` which impacts the performance. It might be fine to call that function multiple times in other places (to leverage the cached result) unless backslashed strings are to be used from the result which is very specific to issues like this one. Changing that function could probably lead to other issues since it is used in many places.

truongwp (approved):
Thanks @AbdiTolesa! I think we will need Mike's opinion for my previous reply. Other things are good.

Crabcyborg (commented):
This isn't backward compatible since `FrmField::get_all_for_form` includes repeater fields as well and this doesn't.

But this also seems like a workaround that leaves the bug as it is.

Can we explain exactly where the new `wp_unslash` calls are happening here?

I think we can come up with a better fix. Likely something in the `FrmField::get_all_for_form` function or one of the functions it calls.

Crabcyborg (changes requested):
I also see this PHP warning repeating for a ton of my fields.

> PHP Warning:  foreach() argument must be of type array|object, string given in /var/www/src/wp-content/plugins/formidable/classes/views/frm-fields/front-end/dropdown-field.php on line 40

AbdiTolesa:
> Can we explain exactly where the new `wp_unslash` calls are happening here?

@Crabcyborg As I have mentioned in the PR description, `FrmField::get_all_for_form` calls `FrmField::getAll` which unslashes the query result. `FrmField::getAll` calls `FrmDb::get_var`, which uses caching. Because of that, subsequent `FrmField::get_all_for_form` keeps removing slashes repeatedly, one at a time. I believe removing `wp_unslash` call from `FrmField::getAll` could fix the issue but would introduce other issues. Also, avoiding `FrmDb::get_var` call for an alternative function could impact performance since it leads to avoiding caching at all.

I have updated the new function to fix the potential issues you mentioned it could cause though.

AbdiTolesa:
> I also see this PHP warning repeating for a ton of my fields.
> 
> > PHP Warning:  foreach() argument must be of type array|object, string given in /var/www/src/wp-content/plugins/formidable/classes/views/frm-fields/front-end/dropdown-field.php on line 40

That was because I didn't unserialize `$field->options` and fixed now.

Crabcyborg:
@AbdiTolesa Can you try an alternative PR that fixes this at a root then?

I need to figure out how to avoid double unslashing everywhere, not just in one place.

Crabcyborg:
I think I figured out the root of the issue and fixed it in a new PR here https://github.com/Strategy11/formidable-forms/pull/1999

I'm going to close this as it won't be needed, and isn't an ideal solution since it leaves the root of the issue and just works around it.

PR #1747, opened 2024-05-17
Fix deprecated notice when calling trim

AbdiTolesa:
Ticket URL: https://secure.helpscout.net/conversation/2599522042/199655

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -2001,7 +2001,9 @@ class FrmAppHelper {
 	public static function check_selected( $values, $current ) {
 		$values  = self::recursive_function_map( $values, 'trim' );
 		$values  = self::recursive_function_map( $values, 'htmlspecialchars_decode' );
-		$current = htmlspecialchars_decode( trim( $current ) );
+		if ( $current ) {
+			$current = htmlspecialchars_decode( trim( $current ) );
+		}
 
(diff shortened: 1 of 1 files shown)


classes/helpers/FrmAppHelper.php
```diff
@@ -1999,9 +1999,11 @@ public static function checked( $values, $current ) {
...
 
 	public static function check_selected( $values, $current ) {
-		$values  = self::recursive_function_map( $values, 'trim' );
-		$values  = self::recursive_function_map( $values, 'htmlspecialchars_decode' );
-		$current = htmlspecialchars_decode( trim( $current ) );
+		$values = self::recursive_function_map( $values, 'trim' );
+		$values = self::recursive_function_map( $values, 'htmlspecialchars_decode' );
+		if ( $current ) {
+			$current = htmlspecialchars_decode( trim( $current ) );
+		}
```
truongwp, line 2006: `trim( null )` returns `''`. So I think we should add an `else` here and assign `$current = ''`.

AbdiTolesa: I didn't get it, we are not calling `trim()` on `null` at all now right?

truongwp: If it's `null`, we should set it to `''`.

truongwp (approved):

PR #1748, opened 2024-05-17
duplicating a form from the form list page

lauramekaj1:


Crabcyborg (approved):
I think this looks good.

We just need to get it passing the workflows.

Then it will be good to merge.

Thanks Laura!

🥳

PR #1754, opened 2024-05-24
Fix frm_content filter doesn't work with repeater actions

truongwp:
The issue: When running a API form action with a repeater entry, other field shortcodes don't work.

You can use this form to test:
[repeater-actions-form.xml.zip](https://github.com/Strategy11/formidable-forms/files/15427173/repeater-actions-form.xml.zip)

That form contains Text, Repeater 1, and Repeater 2 fields. It has the API form action that sends the request to itself. I use this code to log the request:
```
add_action( 'init', function() {
	if ( isset( $_GET['repeater-actions-test'] ) ) {
		error_log( print_r( file_get_contents('php://input'), true ) );
	}
} );
```

When set the `Run this action for` to `Repeater 1`, the shortcodes for other fields should work.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1693,8 +1693,11 @@ class FrmFormsController {
 			$form = $form->id;
 		}
 
+        /*
+         * Repeater actions adds `parent_entry` to `$entry` store the parent entry data. If `parent_entry` is not empty,
+         * use the parent form ID instead of repeater form ID to fix the parent form field shortcodes doesn't work.
+         */
 		if ( ! empty( $entry->parent_entry ) ) {
-			// If $entry is a repeater entry, the main form field shortcodes won't work. So we use the main form ID.
 			$form = $entry->parent_entry->form_id;
 		}
 
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1693,6 +1693,11 @@ class FrmFormsController {
 			$form = $form->id;
 		}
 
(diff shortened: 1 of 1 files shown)


classes/controllers/FrmFormsController.php
```diff
@@ -1693,6 +1693,11 @@ public static function filter_content( $content, $form, $entry = false ) {
 			$form = $form->id;
 		}
 
+		if ( ! empty( $entry->parent_entry ) ) {
+			// If $entry is a repeater entry, the main form field shortcodes won't work. So we use the main form ID.
+			$form = $entry->parent_entry->form_id;
```
Crabcyborg, line 1698: @truongwp Are you sure this won't introduce any issues with repeaters?

It looks like `FrmFieldsHelper::get_shortcodes` just uses `FrmField::getAll` so it would only get fields for the parent form and not for the repeater itself.

I'm not confident that merging this update is safe.

truongwp: @Crabcyborg This change only happens when `$entry->parent_entry` is not empty. Only repeater actions code adds the `parent_entry` to `$entry`. So I think this is safe.

Crabcyborg: Thank you @truongwp. That's good to know. Can we add a comment here about that so it's more obvious?

truongwp: @Crabcyborg I updated the comment. Is it clear enough?

Crabcyborg (approved):
Thank you @truongwp!

I'll go ahead and merge this.

PR #1756, opened 2024-05-27
Fix issue with "All Templates" click resulting in no visible templates

shervElmi:
It seems that by releasing the new Dashboard, we refactored the animation and introduced the `frmAnimate` JS class. Additionally, we refactored the `onCategoryClick` module inside Form Templates, which caused this problem. In this PR, we move them back to the bottom after the early return condition to prevent any further execution if the new category item is equal to the current category.

### QA URL:
https://qa.formidableforms.com/sherv2/wp-admin/admin.php?page=formidable-form-templates

### Issue URL:
[https://github.com/Strategy11/formidable-pro/issues/5077](https://github.com/Strategy11/formidable-pro/issues/5077)

### Testing Instructions:
Look at the related issue; there is a video about the issue. Confirm that the issue no longer exists.


Crabcyborg (approved):
Thanks Sherv!

🚀

PR #1757, opened 2024-05-27
create multiple forms and bulk delete them from form list page

lauramekaj1:


tests/cypress/support/commonActions.cy.js
```diff
@@ -0,0 +1,26 @@
+export const createNewForm = () => {
```
Crabcyborg, line 1: Thanks for working on this @lauramekaj1!

This looks good.

I have one thought here. There is a `support/commands.js` file already. It's currently empty but it's similar to this.

I think the idea is to make it look more like the other Cypress functions.

For example, if you use `Cypress.Commands.add( 'createNewForm', ...` here, you should be able to call `cy.createNewForm`, and I don't think it would require the import that way either.

lauramekaj1: @Crabcyborg good call. I am just used to add them in a separate file when I used functions which handled data manipulation, calculations, or any logic that did not directly interact with the Cypress API. But I agree that a better practice in this case is to add these functions in the _commands.js_ file since it uses solely Cypress commands!

Will update the PR soon. Thank you for your feedback.

Crabcyborg (approved):
Thank you @lauramekaj1!

I think this one is ready to merge!

🚀

PR #1771, opened 2024-05-31
Fix validation error of name field inside repeater when using API

truongwp:
Fixes https://github.com/Strategy11/formidable-api/issues/176

Steps to replicate are in the issue.

**Before:**

https://github.com/Strategy11/formidable-forms/assets/19748318/dfe89d89-8e0d-46ea-b794-9985e1a2cd51

**After:**

https://github.com/Strategy11/formidable-forms/assets/19748318/47a4942a-8ae0-4463-8293-048d22cd5789

### DIFF
--- a/classes/models/fields/FrmFieldCombo.php
+++ b/classes/models/fields/FrmFieldCombo.php
@@ -398,6 +398,10 @@ class FrmFieldCombo extends FrmFieldType {
 	 * @return array Errors array.
 	 */
 	public function validate( $args ) {
+		if ( isset( $args['value'] ) && ! is_array( $args['value'] ) ) {
+			return parent::validate( $args );
+		}
+
 		$errors = isset( $args['errors'] ) ? $args['errors'] : array();
 
 		if ( ! $this->field->required ) {


garretlaxton (approved):
This works great!

Crabcyborg:
@truongwp Is this ready for review?

It looks like it's been sitting for a while.

Also, what happens to the entry after this API request is sent? The name field expects an array, so shouldn't we be sending something like `{"first": "Updated", "last": "Last name"}` instead?

truongwp:
@Crabcyborg Sory for not requesting your review. This PR is ready for review.

Name field data won't be sent in JSON format. First and last name are sent in different keys: `xxx-first`, `xxx-last` (you can see it in the `Before` screencast).

This PR doesn't change the name field data format. The old validation expects the name field value to be an array, but it isn't in the API request, and this causes the issue. I added a check to not run name field validation if the value isn't an array.

Crabcyborg:
@truongwp I spent some time on this issue. I don't believe this PR really fixes the root of the issue, so I took a deeper look.

I have a new PR here https://github.com/Strategy11/formidable-forms/pull/2009

It just looks like the issue is that we're converting the name to a string when `$atts['return_array']` is set to `1`.

I'm going to close this PR. I requested your review on the new fix.

Thank you for working on this issue!

PR #1773, opened 2024-06-03
fix: prevent php warning when fields are removed

epigeyre:
When fields are disabled with the `frm_pro_available_fields` or `frm_available_fields` filters it can generate php warnings within the `field_types_for_input()` method. This commit prevent it from happening by checking if the key is set before dealing with it.


Crabcyborg:
Thank you @epigeyre for reporting this. I understand your issue and this will be fixed in next release.

I made a new PR with a slightly different fix https://github.com/Strategy11/formidable-forms/pull/1785

`$field_types` is actually an empty array that we are populating in this function. This check here prevents any new values from getting added, which breaks the option to change field type.

In my update I check if the key is set in `$fields` instead.

PR #1784, opened 2024-06-04
Fix dropdown styles

AbdiTolesa:


garretlaxton (approved):
This looks great!

truongwp (approved):

css/custom_theme.css.php
```diff
@@ -227,6 +227,7 @@
 .with_frm_style input[type=phone],
 .with_frm_style input[type=search],
 .with_frm_style select,
+.wp-admin .with_frm_style select,
```
Crabcyborg, line 230: @AbdiTolesa It's nice not to include anything admin related in these front end files.

These also do nothing in Lite so it would be nice to get this into Pro instead.

It doesn't look like there are that many styles here to just copy/paste some of them (whatever we actually need) over to the `formidable-entries.css` file in Pro.

PR #1789, opened 2024-06-05
Check that applications elements are not undefined fix

Crabcyborg:
The form templates page was throwing JavaScript error because my applications API data is empty. It must have hit an API error.

These checks prevent the form templates page from breaking when there is no application data.

Screen Shot 2024-06-05 at 10 37 25 AM


shervElmi (approved):
Thanks for this fix, @Crabcyborg. It looks good to me! 🚀

Crabcyborg:
Thanks Sherv!

PR #1796, opened 2024-06-10
Lite changes required for new abandonment stats in monthly email

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-abandonment/issues/49


classes/models/FrmEmailSummary.php
```diff
@@ -120,6 +120,10 @@ protected function get_content_args() {
 			'support_url'      => FrmEmailSummaryHelper::get_frm_url( 'new-topic', 'contact_support' ),
 		);
 
+		if ( is_callable( 'FrmAbandonmentReportsController::update_email_args' ) ) {
```
truongwp, line 123: Why don't we use the below filter for this?

AbdiTolesa: The thing is this function is called on cron job and we cannot hook into the filter to update this from the Abandonment addon. I really didn't want to include addon code in lite but couldn't find a better approach.

truongwp: Probably the hook controller of Abandonment addon hasn't loaded. Can you try adding hook outside of that class? Maybe in `formidable-abandonment.php` file?

truongwp:
@Crabcyborg Look like the current code of summary email isn't good for customizing. Any custom data via the filter could be overwritten in the `FrmEmailStats::get_content_args()`. Do you think we should add another filter for `FrmEmailStats::get_content_args()`? Or we can add filter for summary email class names so other plugins can override using a child class.

This PR is good for the abandonment PR. But we might get the same problem with other data in the future.

Crabcyborg:
@truongwp I don't think using classes would work well because multiple plugins wouldn't all be able to overwrite the class - we'd be limited to one class and it wouldn't be able to realistically inherit all of the other add-on's features.

I think the current fix looks good. If we don't need to add any new hooks that helps keep this more simple for now.

I don't think we'll run into needing to make these sorts of updates too often.

truongwp (approved):
@Crabcyborg Thanks for the reply.

@AbdiTolesa This looks good. Thank you.

classes/models/FrmEmailStats.php
```diff
@@ -235,4 +236,17 @@ public function get_compare_diff( $current, $prev ) {
...
+	 * Allows reading class object properties.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $prop
+	 *
+	 * @return mixed
+	 */
+	public function get_prop( $prop ) {
+		return $this->$prop;
```
Crabcyborg, line 250: @AbdiTolesa Does this need to be so dynamic?

It would be nice if we didn't have something that returned `mixed` since it makes it harder to understand the type we're working with.

AbdiTolesa: @Crabcyborg It currently returns `array|string` but it is called for about 5 properties right now, that could probably be too much to created individual functions for. Should we try limiting the properties that can be accessed calling this function and return `null` for the otherwise?

Crabcyborg: It looks like there are only 8 or 9 total properties?

I don't think it's too big of a deal if we need to add 5 public getters. This is pretty standard OOP. If we need to access this data it's good to make sure we're not giving too much access by makign this work with everything.

classes/models/FrmEmailStats.php
```diff
@@ -235,4 +236,39 @@ public function get_compare_diff( $current, $prev ) {
 	 * @return string
 	 */
 	abstract protected function get_top_forms_label();
+
+	/**
+	 * @since x.x
+	 */
+	public function get_date_range() {
```
Crabcyborg, line 243: @AbdiTolesa Can we also specify the `@return` comment with the proper type specified?

Crabcyborg (approved):
This looks good to merge! I'll include this in today's release.

🚀

PR #1817, opened 2024-06-19
Replace nbsp from plain email content

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/1970

**Before**
image

**After**
image

### Test steps
1. Create a form and adda an RTE field to it.
2. Create an email action, checking `Send Emails in Plain Text`.
3. Preview the form, switch to "Visual" mode and fill the RTE with a text that has line breaks.
4. View the email sent and confirm that there are no `&nbsp;` strings displayed in the email body.

### DIFF
--- a/classes/models/FrmEmail.php
+++ b/classes/models/FrmEmail.php
@@ -420,7 +420,7 @@ class FrmEmail {
 
 		if ( $this->is_plain_text ) {
 			$this->message = wp_specialchars_decode( strip_tags( $this->message ), ENT_QUOTES );
-			$this->message = str_replace( '&nbsp;', "\r\n", $this->message );
+			$this->message = str_replace( '&nbsp;', '', $this->message );
 		} else {
 			$this->add_autop();
 		}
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This works great!

classes/models/FrmEmail.php
```diff
@@ -420,6 +420,7 @@ private function set_message() {
 
 		if ( $this->is_plain_text ) {
 			$this->message = wp_specialchars_decode( strip_tags( $this->message ), ENT_QUOTES );
+			$this->message = str_replace( '&nbsp;', "\r\n", $this->message );
```
Crabcyborg, line 423: @AbdiTolesa I think this is adding too many line breaks? A non breakable space is probably better replaced with an empty string.

AbdiTolesa: Thanks @Crabcyborg! Empty string is more accurate to replace with here.

truongwp (approved):

Crabcyborg (approved):

🚀

PR #1819, opened 2024-06-20
Add a condition to displaying name your form modal

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5036

We shouldn't show the form name popup when a form name is entered.
CleanShot 2024-07-02 at 21 03 35

### Test steps
1. Create a new form
2. Go to the form settings without saving the new form
3. Enter something in the form title input
4. Click the update button
5. Confirm that the form name modal does not show up in the process.


garretlaxton (approved):
This works great!

js/formidable_admin.js
```diff
@@ -6427,6 +6427,19 @@ function frmAdminBuildJS() {
...
+	 * Returns true if 'Name Your Form' modal should be displayed.
+	 *
+	 * @returns {Boolean}
+	 */
+	function shouldShowNameYourFormNameModal() {
+		const formNameInput = document.getElementById( 'frm_form_name' );
+		if ( formNameInput && formNameInput.value.trim() !== '' ) {
+			return false;
+		}
+		return 'true' === urlParams.get( 'new_template' );
```
truongwp, line 6440: I think we should check urlParams first. It's faster than querying the form name input.

AbdiTolesa: I think that would change the logic, we want to check that only if there is no form name input or that input exists and it is empty.

truongwp: My idea is checking `if ( 'true' !== urlPArams... ) { return false; }` before the `const formNameInput` line.

AbdiTolesa: I got your idea but we want to give that check a priority only in the absence of the form name input. If I changed it to what you are saying, we would never show the form name popup if we are not in a new form template, which is not desired.

truongwp: It's ok. I'm approving this.

truongwp (approved):

Crabcyborg (approved):

🚀

PR #1826, opened 2024-06-22
do not store entries submitted from a form

lauramekaj1:


Crabcyborg (approved):
This looks great!

Thank you @lauramekaj1!

🥳

PR #1828, opened 2024-06-25
Keep Stripe pro tip style consistent

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5133
**Before**
image

**After**:
image


Crabcyborg:
Let's hold off on this change.

I'd like to run this by Steve before we make any changes here.

The Stripe tip is newer. It may be that we want to change all of the other tips instead.

PR #1829, opened 2024-06-25
Do not enqueue popper js on WP Bakery builder

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5150

### Test steps
1. Install and activate WP Bakery builder. Nathanael has put a copy of it [here](https://github.com/Strategy11/formidable-pro/issues/5150#issuecomment-2186509154) if you don't have one.
2. Open a page with WP Bakery builder and try adding elements to your page.
3. Confirm that the following js error is not triggered.
image

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -916,6 +916,9 @@ class FrmAppController {
 	 * @return void
 	 */
 	private static function register_popper1() {
+		if ( defined( 'WPB_VC_VERSION' ) ) {
+			return;
+		}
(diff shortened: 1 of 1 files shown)


Crabcyborg:
@AbdiTolesa This "fix" breaks all of our uses of popper.

For example, I can't open the preview dropdown in the form builder, at least when surveys is active. I get this error,

Screen Shot 2024-06-25 at 9 21 34 AM

We have to always register popper on our pages. We don't want any hard coded references to WP Bakery in this add-on.

They real fix to this issue, is that we need to make sure we're only registering popper on the pages where we're trying to enqueue a script that has popper as a dependency only.

I'm going to close this because it's going in the wrong direction.

PR #1830, opened 2024-06-26
Output screen reader texts on forms and entries list pages

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4901

image

### DIFF
--- a/classes/helpers/FrmEntriesListHelper.php
+++ b/classes/helpers/FrmEntriesListHelper.php
@@ -17,11 +17,12 @@ class FrmEntriesListHelper extends FrmListHelper {
 
 
 	public function __construct( $args ) {
-		$this->screen_reader_content = array(
-			'heading_list' => __( 'Entries list', 'formidable-forms' ),
-		);
-
 		parent::__construct( $args );
+		$this->screen->set_screen_reader_content(
+			array(
+				'heading_list' => __( 'Entries list', 'formidable-forms' ),
+			)
+		);
 	}
 
 	/**
--- a/classes/helpers/FrmEntriesListHelper.php
+++ b/classes/helpers/FrmEntriesListHelper.php
@@ -15,6 +15,15 @@ class FrmEntriesListHelper extends FrmListHelper {
 	 */
 	public $total_items = 0;
 
+
+	public function __construct( $args ) {
+		$this->screen_reader_content = array(
+			'heading_list' => __( 'Entries list', 'formidable-forms' ),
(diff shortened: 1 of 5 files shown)


classes/helpers/FrmEntriesListHelper.php
```diff
@@ -15,6 +15,15 @@ class FrmEntriesListHelper extends FrmListHelper {
 	 */
 	public $total_items = 0;
 
+
+	public function __construct( $args ) {
+		$this->screen_reader_content = array(
```
truongwp, line 20: I think we just need to call `$this->screen->set_screen_reader_content()` instead of adding a new `screen_reader_content` property.

AbdiTolesa: We can't do that here since `$this->screen` is not set (null) yet and I think refactoring all that doesn't seem worth it.

truongwp: @AbdiTolesa Did you try calling it after `parent::__construct()`? I see that the extra property doesn't make the code simpler much.

AbdiTolesa: That worked for me now @truongwp! Thanks!

truongwp (approved):

classes/helpers/FrmListHelper.php
```diff
@@ -997,6 +997,8 @@ public function display( $args = array() ) {
 		if ( $this->should_display( $args, 'display-top-nav' ) ) {
 			$this->display_tablenav( 'top' );
 		}
+		$this->screen->render_screen_reader_content( 'heading_list' )
```
Crabcyborg, line 1000: It looks like there is a missing semi-colon here.

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -15,6 +15,15 @@ class FrmEntriesListHelper extends FrmListHelper {
 	 */
 	public $total_items = 0;
 
+	public function __construct( $args ) {
+		parent::__construct( $args );
+		$this->screen->set_screen_reader_content(
+			array(
+				'heading_list' => __( 'Entries list', 'formidable' ),
```
Crabcyborg, line 22: @AbdiTolesa It doesn't look like these get escaped automatically. I think we need to use `esc_html__` here.

Crabcyborg (approved):

I think we're good to merge this now.

🚀

PR #1834, opened 2024-07-02
Support string form ID - Lite

engahmeds3ed:
Part of https://github.com/Strategy11/formidable-pro/issues/5160

### DIFF
--- a/classes/models/FrmField.php
+++ b/classes/models/FrmField.php
@@ -802,11 +802,11 @@ class FrmField {
 	 * @return array
 	 */
 	public static function get_all_for_form( $form_id, $limit = '', $inc_embed = 'exclude', $inc_repeat = 'include' ) {
-		if ( is_string( $form_id ) ) {
+		if ( ! is_numeric( $form_id ) ) {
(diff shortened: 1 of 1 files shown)


classes/models/FrmField.php
```diff
@@ -802,7 +802,11 @@ public static function get_all_types_in_form( $form_id, $type, $limit = '', $inc
 	 * @return array
 	 */
 	public static function get_all_for_form( $form_id, $limit = '', $inc_embed = 'exclude', $inc_repeat = 'include' ) {
-		if ( ! (int) $form_id ) {
+		if ( is_string( $form_id ) ) {
+			$form_id = self::get_id_by_key( $form_id );
+		}
+
+		if ( ! is_int( $form_id ) ) {
```
Crabcyborg, line 809: I would also make sure here that if `$form_id` is a string number that it doesn't break here. And we should still check for 0.

engahmeds3ed:
I'm going to close this PR, it's not needed as per discussion with @Crabcyborg

PR #1841, opened 2024-07-08
Remove width limit from h1 title

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5173
### Test steps
Switch your site language to Portuguese/Brazillian and go to the form entries page.

**Before**:
CleanShot 2024-07-08 at 11 08 50@2x

**After**:
CleanShot 2024-07-08 at 11 08 12@2x


AbdiTolesa:
@Crabcyborg I couldn't switch to Portuguese (Brazilian) locally but I can see the issue in QA and just created this PR quickly. The parent `div` has `frm_top_left` which constraints the width and I assume it is safe to just get rid off the width limitation on the `h1`.

AbdiTolesa:
> @AbdiTolesa I think we may want to force a minimum width at least. Changing it to `min-width` is probably better than removing the width style.

Thanks, done!

truongwp (approved):

Crabcyborg (approved):
I'm not seeing any issues with this change.

🚀

PR #1850, opened 2024-07-08
Add condition checks before registering popper script

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5150


AbdiTolesa:
@Crabcyborg What do you think about this alternative PR to https://github.com/Strategy11/formidable-forms/pull/1829? I'm trying to avoid registering popper if it is not going to be enqueued in our pages.

Crabcyborg (approved):
Thanks @AbdiTolesa!

I think this should be safe.

PR #1863, opened 2024-07-11
Hide submit field button from field search result

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/5192

The submit field button shouldn't show in the field type search result.


Crabcyborg (approved):
Thank you @truongwp!

🚀

PR #1864, opened 2024-07-11
Lite Version: Add Format Settings

shervElmi:
This update introduces a new feature that adds **Format settings** to the following field types: **Text**, **Paragraph**, **Hidden**, **Slider**, and **Number**.  

The feature provides users with the ability to:  
- Apply number formatting within text.  
- Format pure numeric values.  

This enhancement ensures seamless integration with existing functionality and does not disrupt the experience for existing users.

### For a detailed demonstration, please watch the video:
https://share.cleanshot.com/K8RsgbTz

### Pro version PR:
https://github.com/Strategy11/formidable-pro/pull/5195

### QA URL:
https://qa.formidableforms.com/sherv7/wp-admin/admin.php?page=formidable&frm_action=edit&id=19


classes/views/frm-fields/back-end/format-dropdown.php
```diff
@@ -0,0 +1,71 @@
...
+		 * Filter the number option HTML for the format dropdown.
+		 *
+		 * @since x.x
+		 *
+		 * @param string $number_option       The HTML for the number option.
+		 * @param string $number_option_text The text for the number option.
+		 * @param array  $field              The field array.
+		 * @return string The filtered HTML for the number option.
+		 */
+		echo apply_filters( 'frm_print_format_number_option', $number_option, $number_option_text, $field ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
Crabcyborg, line 51: @shervElmi Can we do this in a safer way? We ideally never want to add more `// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped` comments into Lite, especially without any filtering at all.

Can we run this through `FrmAppHelper::kses` at least?

classes/controllers/FrmFieldsController.php
```diff
@@ -527,6 +529,10 @@ private static function add_input_classes( $field, array &$class ) {
 		if ( isset( $field['size'] ) && $field['size'] > 0 ) {
 			$class[] = 'auto_width';
 		}
+
+		if ( isset( $field['format'] ) && 'currency' === $field['format'] && empty( $field['calc'] ) && 'range' !== $field['type'] ) {
+			$class[] = 'frm-has-number-format';
```
Crabcyborg, line 534: It looks like this line is pretty pro-specific. Is it possible to move this into Pro?

shervElmi: Addressed with the following commits:

- Pro PR: [Move the format fields input class to the Pro version](https://github.com/Strategy11/formidable-pro/pull/5195/commits/3aba89f04755fbf5887641d8e8909a6c29034f4a)
- [Move the format fields input class to the Pro version](https://github.com/Strategy11/formidable-forms/pull/1864/commits/703058666c394386b80a1a0f3dd57e73d6644684)

classes/helpers/FrmAppHelper.php
```diff
@@ -861,13 +861,19 @@ private static function decode_amp( &$string ) {
 	 *
 	 * @param string       $value
 	 * @param array|string $allowed 'all' for everything included as defaults.
+	 * @param bool         $echo    Whether to output the sanitized value directly.
 	 *
 	 * @return string
 	 */
-	public static function kses( $value, $allowed = array() ) {
-		$allowed_html = self::allowed_html( $allowed );
+	public static function kses( $value, $allowed = array(), $echo = false ) {
```
Crabcyborg, line 868: @shervElmi Instead of modifying this function, which makes the function much more complex, can we just add a new function?

I was thinking about this before. I think we can call it `kses_echo` instead.

It gets really ugly with static analysis, and code readability to have a function that sometimes returns `null` and sometimes returns a `string`.

classes/models/FrmEntryValidate.php
```diff
@@ -136,6 +136,8 @@ public static function validate_field( $posted_field, &$errors, $values, $args =
 			$value = $values;
 		}
 
+		self::maybe_normalize_formatted_numbers( $posted_field, $value );
```
Crabcyborg, line 139: @shervElmi Is this something we can put in Pro?

classes/models/FrmEntryValidate.php
```diff
@@ -336,6 +339,24 @@ public static function spam_check( $exclude, $values, &$errors ) {
...
+	 * Normalizes formatted numbers in a value based on the field format settings, if applicable
+	 *
+	 * @since x.x
+	 *
+	 * @param array  $field The field settings.
+	 * @param string $value The value to be normalized (passed by reference).
+	 * @return void
+	 */
+	private static function maybe_normalize_formatted_numbers( $field, &$value ) {
+		if (
```
Crabcyborg, line 352: This should really be in Pro, using a filter.

classes/models/fields/FrmFieldHidden.php
```diff
@@ -41,6 +41,7 @@ protected function field_settings_for_type() {
 		$settings            = parent::field_settings_for_type();
 		$settings['css']     = false;
 		$settings['default'] = true;
+		$settings['format']  = true;
```
Crabcyborg, line 44: Can we put this in `FrmProFieldHidden` instead? I don't think we really want the format setting in Lite.

js/formidable_admin.js
```diff
@@ -8567,7 +8560,7 @@ function frmAdminBuildJS() {
 	 * @returns {Boolean}
 	 */
 	function isContextualShortcode( item ) {
-		const shortcode = item.querySelector( 'a' ).dataset.code;
+		const shortcode = item.querySelector( 'a' )?.dataset.code;
```
Crabcyborg, line 8563: We have a different check for this now in master. We should be able to merge master and revert this change.

classes/views/frm-fields/back-end/format-dropdown.php
```diff
@@ -0,0 +1,79 @@
...
+			return $allowed_html;
+		};
+
+		add_filter( 'frm_striphtml_allowed_tags', $add_allowed_html );
+		FrmAppHelper::kses( $number_option, array( 'option' ), true );
+		remove_filter( 'frm_striphtml_allowed_tags', $add_allowed_html );
+		?>
+
+		<?php if ( 'text' === $field_type ) { ?>
+			<option value="custom" data-dependency="#frm-field-format-custom-<?php echo esc_attr( $field_id ); ?>" <?php selected( ! empty( $format ) && 'currency' !== $format, true ); ?>>
```
Crabcyborg, line 64: The `true` here for the `$current` param is already `true` by default, so we could simply this and just remove the `, true` here.

classes/views/frm-fields/back-end/format-dropdown.php
```diff
@@ -0,0 +1,79 @@
...
+		data-field-id="<?php echo intval( $field_id ); ?>"
+	>
+		<option value="none" <?php selected( $format, '' ); ?>>
+			<?php esc_html_e( 'None', 'formidable' ); ?>
+		</option>
+
+		<?php
+		$number_option_text = in_array( $field_type, array( 'number', 'range' ) ) ? esc_html__( 'Custom', 'formidable' ) : esc_html__( 'Number', 'formidable' );
+		$number_option      = '
+			<option value="" class="frm_show_upgrade frm_noallow" data-upgrade="' . esc_attr__( 'Format number field', 'formidable' ) . '" data-medium="format-number-field">
```
Crabcyborg, line 37: Can we put this in a view or something? Generally I want to avoid any large strings of HTML since they don't work well with syntax highlighting and can easily cause errors.

classes/models/fields/FrmFieldNumber.php
```diff
@@ -28,6 +28,7 @@ protected function field_settings_for_type() {
 			'clear_on_focus' => true,
 			'invalid'        => true,
 			'range'          => true,
+			'format'         => true,
```
Crabcyborg, line 31: @shervElmi Do we want to avoid showing the format option when Pro hasn't been updated yet? A format setting for numbers doesn't really make sense when Pro isn't updated.

I also don't think it makes sense when Pro isn't active. We didn't support the old type of custom formats for numbers (just for currency).

Can we move this to `FrmProFieldNumber.php` instead?

classes/models/fields/FrmFieldTextarea.php
```diff
@@ -26,6 +26,7 @@ protected function field_settings_for_type() {
 		return array(
 			'size'           => true,
 			'clear_on_focus' => true,
+			'format'         => true,
```
Crabcyborg, line 29: We don't need a format setting in Lite for Paragraph fields either.

The only field types in Lite that should have a format setting are Text and Sliders.

js/formidable_admin.js
```diff
@@ -6691,32 +6691,25 @@ function frmAdminBuildJS() {
...
-			if ( 'none' === value ) {
-				formatInput.setAttribute( 'value', '' );
-			}
-			if ( 'international' === value ) {
-				formatInput.setAttribute( 'value', 'international' );
+		formatTypes.forEach( formatType => {
+			const value = formatType.value;
+			if ( value in valueMap ) {
+				const formatInput = document.getElementById( `frm_format_${formatType.dataset.fieldId}` );
+				formatInput.setAttribute( 'value', valueMap[value] );
```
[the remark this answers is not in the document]
Crabcyborg: I also think we typically try to put spaces in braces when there is a variable key.

Like `valueMap[ value ]`.

classes/views/frm-fields/back-end/format-dropdown.php
```diff
@@ -0,0 +1,79 @@
...
+		id="frm_format_dropdown_<?php echo esc_attr( $field_id ); ?>"
+		class="frm_format_dropdown frm_select_with_upgrade frm_select_with_dependency"
+		data-field-id="<?php echo intval( $field_id ); ?>"
+	>
+		<option value="none" <?php selected( $format, '' ); ?>>
+			<?php esc_html_e( 'None', 'formidable' ); ?>
+		</option>
+
+		<?php
+		$number_option_text = in_array( $field_type, array( 'number', 'range' ) ) ? esc_html__( 'Custom', 'formidable' ) : esc_html__( 'Number', 'formidable' );
```
Crabcyborg, line 35: We generally try to use `$strict = true` when calling `in_array`. It's faster and more accurate.

Crabcyborg: It's also better to use `esc_html` when outputting `$number_option_text` rather than escaping it here.

classes/helpers/FrmAppHelper.php
```diff
@@ -1089,6 +1095,10 @@ private static function safe_html() {
 			'legend'     => array(
 				'class' => true,
 			),
+			'option'     => array(
+				'value'    => true,
+				'selected' => true,
```
Crabcyborg, line 1100: @shervElmi It looks like the upgrades are not working on the options at all.

It looks like `'class' => true` is missing here.

classes/models/fields/FrmFieldTextarea.php
Crabcyborg: It might be that we're trying to show an upgrade when these options are selected (that just isn't working).

Can we put a Pro version check here?

Ideally, if Pro is active, but not updated, we won't show this setting.

classes/views/frm-fields/back-end/format-dropdown.php
Crabcyborg: @shervElmi We might release this a week before we make the Pro release. Could we possibly check the `FrmFormApi` API info for the current Pro version (checking the `version` key), and only show this upgrade if the new version of Pro is available?

classes/helpers/FrmAppHelper.php
shervElmi: Addressed with the [Add kses_echo helper method](https://github.com/Strategy11/formidable-forms/pull/1864/commits/7383229e245d8acc6494f9eaf5466ee4f58f7df1) commit.

classes/models/FrmEntryValidate.php
shervElmi: I need to add a new `apply_filters`, then. Does that sound good to you?

Since we pass the value to `set_posted_value`, I need to normalize it beforehand.

classes/models/fields/FrmFieldHidden.php
shervElmi: Addressed with the following commits:
- [Remove format settings](https://github.com/Strategy11/formidable-forms/pull/1864/commits/38047889922e86ffbf3430f24770da7c078cdae2)
- [Add format settings](https://github.com/Strategy11/formidable-pro/pull/5195/commits/5ff69ee4d685ac0a1c94883029cc5394f213b18a)

classes/models/FrmEntryValidate.php
Crabcyborg: @shervElmi What about just using `frm_modify_posted_field_value`?

Do you need to actually modify the `$_POST` value or does just updating the value we're validating work?

classes/models/fields/FrmFieldTextarea.php
shervElmi: I’ve moved the format settings for all fields except Text and Phone to Pro.

Could you clarify what I should do for backward compatibility? Thanks!

classes/views/frm-fields/back-end/format-dropdown.php
shervElmi: I’ve moved all options HTML code to a new view file and completely refactored this file.

Could you provide me with a version of the next Pro for comparison? Thanks!

classes/models/FrmEntryValidate.php
shervElmi: @Crabcyborg, I need to modify the `$_POST` value, but `frm_modify_posted_field_value` doesn’t seem safe.

If you confirm it’s safe, I’ll use that filter.

Note: I need to change the field value before any processing, meaning before it’s saved to the database or any other actions.

classes/views/frm-fields/back-end/format-dropdown.php
shervElmi: @Crabcyborg Just following up on my last message.

classes/models/FrmEntryValidate.php
shervElmi: @Crabcyborg Just following up on my last message.

classes/helpers/FrmCurrencyHelper.php
```diff
@@ -26,6 +26,16 @@ public static function get_currency( $currency ) {
...
 	}
 
+	/**
+	 * Checks if the given format is a valid currency format.
+	 *
+	 * @param string $format_value The format value to check.
+	 * @return bool
+	 */
+	public static function is_currency_format( $format_value ) {
+		return ! empty( $format_value ) && in_array( $format_value, array( 'currency', 'number' ), true );
```
Crabcyborg, line 36: This could be simplified to just `return $format_value && in_array( $format_value, array( 'currency', 'number' ), true );` or even just really just `in_array( $format_value, array( 'currency', 'number' ), true );`, since anything falsey wouldn't match the array values anyway.

classes/models/FrmEntryValidate.php
Crabcyborg: @shervElmi `frm_modify_posted_field_value` should be safe. Using filters that already exist is always ideal, and setting `$_POST` is worst case scenario.

It's easy to modify the value before it is saved to the database. We have a function `set_value_before_save` in `FrmFieldType`, that you can overwrite in any of the children classes to make sure your value is always saved in the DB properly - this is the best place for it, but we'll likely want to normalize numbers in various places, like conditional logic, as well.

classes/controllers/FrmSettingsController.php
```diff
@@ -207,6 +207,11 @@ public static function general_settings() {
 	 * @return void
 	 */
 	public static function maybe_render_currency_selector( $frm_settings, $more_html ) {
+		if ( is_callable( 'FrmProSettingsController::render_currency_selector' ) ) {
+			FrmProSettingsController::render_currency_selector();
+			return;
+		}
+
 		if ( false !== strpos( $more_html, 'id="frm_currency"' ) ) {
```
Crabcyborg, line 215: I think we should be safe to remove this backward compatibility condition at this point now.

stubs.php
```diff
@@ -421,8 +424,8 @@ class FrmProSettingsController {
 		 * @param string $count
 		 * @return string
 		 */
-		public static function inbox_badge( $count ) {
-		}
+		public static function inbox_badge( $count ) {}
+		public static function render_currency_selector() {}
```
Crabcyborg, line 428: Can we add an actual signature to this? Even if it is `@return void`. It's useful for the static analysis to have accurate type data.

js/admin/settings.js
```diff
@@ -23,6 +27,19 @@
 		e.target.nextElementSibling.setAttribute( 'aria-checked', e.target.checked ? 'true' : 'false' );
 	}
 
+	/**
+	 * Updates the currency formatting options based on the selected currency.
+	 *
+	 * @param {HTMLSelectElement} currencySelect The currency select element.
+	 */
+	function syncCurrencyOptoins( currencySelect ) {
```
Crabcyborg, line 35: `syncCurrencyOptoins` has a typo in it. It should be `Options`, not `Optoins`.

classes/views/frm-settings/general.php
```diff
@@ -74,3 +76,11 @@
...
 	</label>
 </p>
+
+<?php
+/**
+ * Trigger an action so Pro can display additional General settings in the Other section.
+ *
+ * @param FrmSettings $frm_settings
+ */
+do_action( 'frm_other_settings_form', $frm_settings );
```
Crabcyborg, line 86: Can we add a `@since` comment here? It's good to have a quick reference for when a new hook was added.

classes/models/FrmEntryValidate.php
shervElmi: I found that `frm_modify_posted_field_value` works well in my case (across multiple pages, AJAX, and complex scenarios), whereas `set_value_before_save` didn’t provide a solution in this instance.

Could you review a scenario where you think this might cause an issue? I mean just using the `frm_modify_posted_field_value` filter.

Related commit:
https://github.com/Strategy11/formidable-pro/pull/5195/commits/66dc8b7d55e41b47a30fe323c4eb4ec4824efa9e

Crabcyborg:
@shervElmi It looks like there are issues with failing unit tests now.

classes/models/FrmEntryValidate.php
```diff
@@ -228,8 +228,9 @@ public static function validate_field_types( &$errors, $posted_field, $value, $a
 	}
 
 	public static function validate_phone_field( &$errors, $field, $value, $args ) {
-		if ( $field->type === 'phone' || ( $field->type === 'text' && FrmField::is_option_true_in_object( $field, 'format' ) ) ) {
+		$format_value = FrmField::get_option( $field, 'format' );
 
+		if ( $field->type === 'phone' || ( $field->type === 'text' && ! FrmCurrencyHelper::is_currency_format( $format_value ) ) ) {
```
Crabcyborg, line 233: @shervElmi Do we still want to also check `FrmField::is_option_true_in_object( $field, 'format' )`? I'm concerned this might validate too often without that check still.

js/admin/settings.js
```diff
@@ -23,6 +27,19 @@
...
 
+	/**
+	 * Updates the currency formatting options based on the selected currency.
+	 *
+	 * @param {HTMLSelectElement} currencySelect The currency select element.
+	 */
+	function syncCurrencyOptions( currencySelect ) {
+		const currency = frmSettings.currencies[ currencySelect.value ];
+
+		document.getElementById( 'frm_thousand_separator' ).setAttribute( 'value', currency.thousand_separator );
```
Crabcyborg, line 38: Is there a reason we're using `setAttribute` here? We typically just use `.value =`, since it's shorter and easier to read. It's also more consistent if we change it to use .value instead.

Crabcyborg (approved):

I'll keep testing, but anything else that comes up can happen in a new PR.

Thank you @shervElmi!

🚀

js/admin/settings.js
Crabcyborg: I'll just merge without this change. It isn't a big deal.

classes/models/FrmEntryValidate.php
shervElmi: Yes, we’re using the `frm_modify_posted_field_value` filter.

Crabcyborg: Thanks. I missed that this must already be gone.

PR #1865, opened 2024-07-12
Fix issue with enabling draft button after lookup is finished

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5194

The issue happens because of an outdated element selector in `enableSaveDraft` function.

**Before**
CleanShot 2024-07-12 at 22 20 54

**After**:
CleanShot 2024-07-12 at 22 20 07

### Test steps

1. Create a simple form and enable draft saving
2. Add a lookup field to get value from another form
3. Add another lookup field to watch the first field
4. Save the form and preview the form
5. Change the value of the first Lookup field and confirm that the draft button is enabled back when the lookup is done

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -153,7 +153,8 @@ function frmFrontFormJS() {
 	 * @param {Object} $form
 	 */
 	function enableSaveDraft( $form ) {
-		$form.find( 'a.frm_save_draft' ).css( 'pointer-events', '' );
+		$form.find( '.frm_save_draft' ).css( 'pointer-events', '' );
+		$form.find( '.frm_save_draft' ).prop( 'disabled', false );
 	}
 
 	function validateForm( object ) {
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -153,12 +153,10 @@ function frmFrontFormJS() {
 	 * @param {Object} $form
 	 */
 	function enableSaveDraft( $form ) {
-		const saveDraftButton = $form[0].querySelector( '.frm_save_draft' );
-		if ( ! saveDraftButton ) {
-			return;
-		}
-		saveDraftButton.disabled            = false;
-		saveDraftButton.style.pointerEvents = '';
+		$form[0].querySelectorAll( '.frm_save_draft' ).forEach( saveDraftButton => {
+			saveDraftButton.disabled            = false;
+			saveDraftButton.style.pointerEvents = '';
+		});
 	}
 
 	function validateForm( object ) {
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -153,8 +153,12 @@ function frmFrontFormJS() {
 	 * @param {Object} $form
 	 */
 	function enableSaveDraft( $form ) {
-		$form.find( '.frm_save_draft' ).css( 'pointer-events', '' );
-		$form.find( '.frm_save_draft' ).prop( 'disabled', false );
+		const saveDraftButton = $form[0].querySelector( '.frm_save_draft' );
+		if ( ! saveDraftButton ) {
+			return;
+		}
+		saveDraftButton.disabled            = false;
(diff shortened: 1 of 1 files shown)


truongwp (approved):

garretlaxton (approved):
This works great!

js/formidable.js
```diff
@@ -153,7 +153,12 @@ function frmFrontFormJS() {
 	 * @param {Object} $form
 	 */
 	function enableSaveDraft( $form ) {
-		$form.find( 'a.frm_save_draft' ).css( 'pointer-events', '' );
+		const saveDraftButton = $form[0].querySelector( '.frm_save_draft' );
+		if ( ! saveDraftButton ) {
+			return;
+		}
+		saveDraftButton.disabled            = false;
+		saveDraftButton.style.pointerEvents = ''
```
Crabcyborg, line 161: There should be a semi-colon on this line. We never omit them.

js/formidable.js
```diff
@@ -153,7 +153,12 @@ function frmFrontFormJS() {
 	 * @param {Object} $form
 	 */
 	function enableSaveDraft( $form ) {
-		$form.find( 'a.frm_save_draft' ).css( 'pointer-events', '' );
+		const saveDraftButton = $form[0].querySelector( '.frm_save_draft' );
+		if ( ! saveDraftButton ) {
+			return;
+		}
+		saveDraftButton.disabled            = false;
+		saveDraftButton.style.pointerEvents = '';
```
engahmeds3ed, line 161: Maybe it's edgy case, but if we have two submit buttons in the form, this will enable only the first one.

image

What about something like:

```suggestion
		const saveDraftButtons = jQuery($form[0]).find( '.frm_save_draft' );
		if ( saveDraftButtons.length === 0 ) {
			return;
		}
		saveDraftButtons.each( (key, item) => {
			jQuery(item).prop( 'disabled', false ).css( 'pointer-events', '' );
		} );
```

You know the code more than me, it's just an idea, what do u think @AbdiTolesa ?

engahmeds3ed (changes requested):
Awesome idea, just a minor request.

js/formidable.js
AbdiTolesa: @engahmeds3ed Thanks for reviewing this! I have thought of that too when I was writing the fix but having two draft buttons for a single for didn't make any sense to me and I don't think that is something to bother for. So I believe we are fine with this.

engahmeds3ed: Yes I think so :) but I consulted Mike here:

https://strategy11.slack.com/archives/C799A2R61/p1721407141219699?thread_ts=1721340597.429319&cid=C799A2R61

and he said:

> Ha. Fair point. Making it support multiple is probably smart even if we don't anticipate it

What do u think?

AbdiTolesa: No problem, I have covered that too now.

engahmeds3ed (approved):
Works like a charm, Great job @AbdiTolesa

js/formidable.js
```diff
@@ -153,7 +153,10 @@ function frmFrontFormJS() {
 	 * @param {Object} $form
 	 */
 	function enableSaveDraft( $form ) {
-		$form.find( 'a.frm_save_draft' ).css( 'pointer-events', '' );
+		$form[0].querySelectorAll( '.frm_save_draft' ).forEach( saveDraftButton => {
```
Crabcyborg, line 156: @AbdiTolesa Maybe we should add a check if `$form.length` is 0 first, just in case? I think this code could potentially break the JS if there were no form for whatever reason.

Crabcyborg (approved):
Thanks @AbdiTolesa @truongwp @engahmeds3ed and @garretlaxton for working on this!

🙌

PR #1866, opened 2024-07-13
allow & to work in format validation

andrewklimek:
using a regex like `^[^&]*$` in a field Format setting does not work because preg_match is faced with `&amp;`


Crabcyborg (approved):
Thank you @andrewklimek!

This violates our PHPCS styling because we require additional spaces.

It looks easier to fix that separately, so I'll just make a follow-up pull request to follow this.

PR #1869, opened 2024-07-16
Print hidden field to track current post id

AbdiTolesa:


engahmeds3ed (commented):
This is working as expected, but again I have another suggestion that I want to discuss with you.

we have a hook here called `frm_after_title` would it be better to add a callback to print the post ID there? something like:

```
add_action( 'frm_after_title', function ( $form ) {
	global $post;
	if ( ! $post ) {
		return;
	}
	?>
	<input type="hidden" name="frm_post_id" value="<?php echo esc_attr( $post->ID ); ?>" />
	<?php
} );
```

@Crabcyborg In case we used this hook, should we add it in pro code or lite?

Thanks.

AbdiTolesa:
> This is working as expected, but again I have another suggestion that I want to discuss with you.
> 
> 
> 
> we have a hook here called `frm_after_title` would it be better to add a callback to print the post ID there? something like:
> 
> 
> 
> ```
> 
> add_action( 'frm_after_title', function ( $form ) {
> 
> 	global $post;
> 
> 	if ( ! $post ) {
> 
> 		return;
> 
> 	}
> 
> 	?>
> 
> 	<input type="hidden" name="frm_post_id" value="<?php echo esc_attr( $post->ID ); ?>" />
> 
> 	<?php
> 
> } );
> 
> ```
> 
> 
> 
> @Crabcyborg In case we used this hook, should we add it in pro code or lite?
> 
> 
> 
> Thanks.

@engahmeds3ed Thanks, I think the hidden element should be in Pro since we are trying to fix a Pro feature.

PR #1873, opened 2024-07-19
Make sure popper is registered for legacy Views

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5206

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -933,10 +933,21 @@ class FrmAppController {
 	private static function should_register_popper() {
 		global $pagenow;
 
-		$page      = FrmAppHelper::simple_get( 'page', 'sanitize_title' );
+		$post_id   = FrmAppHelper::simple_get( 'post', 'absint' );
 		$post_type = FrmAppHelper::simple_get( 'post_type', 'sanitize_title' );
 
-		return strpos( $page, 'formidable' ) === 0 || ( $pagenow === 'edit.php' && $post_type === 'frm_display' ) || ( in_array( $pagenow, array( 'term.php', 'edit-tags.php' ), true ) && FrmAppHelper::simple_get( 'taxonomy' ) === 'frm_application' );
+		if ( 'post.php' === $pagenow && $post_id && 'frm_display' === get_post_type( $post_id ) ) {
+			return true;
+		}
+		$is_views_post_type = 'frm_display' === $post_type;
+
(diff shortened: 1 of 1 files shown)


classes/controllers/FrmAppController.php
```diff
@@ -933,10 +933,21 @@ private static function register_popper1() {
...
+		}
+		$is_views_post_type = 'frm_display' === $post_type;
+
+		if ( 'post-new.php' === $pagenow && $is_views_post_type ) {
+			return true;
+		}
+
+		$page = FrmAppHelper::simple_get( 'page', 'sanitize_title' );
+
+		return strpos( $page, 'formidable' ) === 0 || ( $pagenow === 'edit.php' && $is_views_post_type ) || ( in_array( $pagenow, array( 'term.php', 'edit-tags.php' ), true ) && FrmAppHelper::simple_get( 'taxonomy' ) === 'frm_application' );
```
Crabcyborg, line 950: @AbdiTolesa I think we should probably probably do the `strpos( $page, 'formidable' ) === 0` check first and return early.

We don't want this line to get too big where it becomes difficult to read. It looks like it should be broken up a bit more now.

AbdiTolesa: @Crabcyborg I just saw your comment after I have already refactored it and added a bit of performance improvement.

classes/controllers/FrmAppController.php
```diff
@@ -933,10 +933,23 @@ private static function register_popper1() {
...
-		$page      = FrmAppHelper::simple_get( 'page', 'sanitize_title' );
-		$post_type = FrmAppHelper::simple_get( 'post_type', 'sanitize_title' );
+		$post_id = FrmAppHelper::simple_get( 'post', 'absint' );
+		if ( 'post.php' === $pagenow && $post_id && 'frm_display' === get_post_type( $post_id ) ) {
+			return true;
+		}
+
+		$post_type          = FrmAppHelper::simple_get( 'post_type', 'sanitize_title' );
+		$is_views_post_type = 'frm_display' === $post_type;
+		if ( 'post-new.php' === $pagenow && $is_views_post_type ) {
```
Crabcyborg, line 943: @AbdiTolesa I think the `edit.php` check makes more sense here. We could just check if `$pagenow` matched `post-new.php` or `edit.php` with an `in_array` check. These are basically the same check so I think they make sense together. It's basically `$is_on_views_page`.

engahmeds3ed (approved):
Good job @AbdiTolesa as usual.

PR #1874, opened 2024-07-19
Exclude repeater forms and entries from summary email

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/5205

To test:
- Use `FrmEmailSummaryHelper::send_monthly();` to send the monthly email. Make sure that there is no repeaters in the top forms list.
- Choose a form with a repeater, then submit an entry with some repeater items.
- Use the above function to send the monthly email again.
- Make sure that the `Entries created` increases by 1.

### DIFF
--- a/classes/helpers/FrmEmailSummaryHelper.php
+++ b/classes/helpers/FrmEmailSummaryHelper.php
@@ -295,10 +295,10 @@ class FrmEmailSummaryHelper {
 			'frm_items',
 			array(
 				// The `=` is added after `>` in the query.
-				'created_at >' => $from_date,
-				'created_at <' => $to_date . ' 23:59:59',
-				'is_draft'     => 0,
-				'parent_item_id' => 0,
+				'created_at >'   => $from_date,
+				'created_at <'   => $to_date . ' 23:59:59',
+				'is_draft'       => 0,
+				'parent_item_id' => 0, // Do not count repeater entries.
 			)
 		);
 	}
--- a/classes/helpers/FrmEmailSummaryHelper.php
+++ b/classes/helpers/FrmEmailSummaryHelper.php
@@ -298,6 +298,7 @@ class FrmEmailSummaryHelper {
 				'created_at >' => $from_date,
 				'created_at <' => $to_date . ' 23:59:59',
 				'is_draft'     => 0,
+				'parent_item_id' => 0,
 			)
 		);
 	}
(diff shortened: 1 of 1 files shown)


classes/helpers/FrmEmailSummaryHelper.php
```diff
@@ -298,6 +298,7 @@ public static function get_entries_count( $from_date, $to_date ) {
 				'created_at >' => $from_date,
 				'created_at <' => $to_date . ' 23:59:59',
 				'is_draft'     => 0,
+				'parent_item_id' => 0,
```
Crabcyborg, line 301: @truongwp PHPCS is throwing an error here because the `=>` all need to be aligned.

truongwp: I fixed it.

classes/helpers/FrmEmailSummaryHelper.php
```diff
@@ -317,7 +319,7 @@ public static function get_top_forms( $from_date, $to_date, $limit = 5 ) {
 			$wpdb->prepare(
 				"SELECT fr.id AS form_id, fr.name AS form_name, COUNT(*) as items_count
 						FROM {$wpdb->prefix}frm_items AS it INNER JOIN {$wpdb->prefix}frm_forms AS fr ON it.form_id = fr.id
```
Crabcyborg, line 321: @truongwp I think this SQL query may have compatibility issues.

I tried running this on my local database.

I ran this query,
```
SELECT fr.id AS form_id, fr.name AS form_name, COUNT(*) as items_count
	FROM wp_frm_items AS it INNER JOIN wp_frm_forms AS fr ON it.form_id = fr.id
	WHERE it.created_at BETWEEN '2024-01-01' AND '2024-12-31' AND it.is_draft = 0
```

I get back an error `In aggregated query without GROUP BY, expression #1 of SELECT list contains nonaggregated column 'wordpress_develop.fr.id'; this is incompatible with sql_mode=only_full_group_by`.

It seems that I need an explicit `GROUP BY`. If I add a `GROUP BY fr.id` to the end of the query, the error goes away.

truongwp: We have the GROUP BY below that line. I think you didn't copy the full SQL query.

Crabcyborg: @truongwp Sorry about that! I think I only copied what GitHub was showing.

Crabcyborg (approved):
Thank you @truongwp!

🚀

PR #1878, opened 2024-07-23
add more accessibility test coverage and update existing tests

lauramekaj1:


tests/cypress/e2e/admin-a11y.cy.js
```diff
@@ -14,20 +14,115 @@ describe( 'Run some accessibility tests', function() {
...
+
+    it('Check the payments page is accessible', () => {
+        cy.visit( '/wp-admin/admin.php?page=formidable-settings&t=stripe_settings' );
+        cy.injectAxe();
+        configureAxeWithBaselineIgnoredRuleset();
+        cy.checkA11y();
+    });
+
     it('Check the import/export page is accessible', () => {
-        cy.visit( '/wp-admin/admin.php?page=formidable-import' );
```
Crabcyborg, line 30: @lauramekaj1 It looks like this got a bit messed up.

This test is checking the dashboard but the title says import/export. The import/export page check was removed as well. We need to put this line back, and add a new test that covers the dashboard page as well.

tests/cypress/e2e/admin-a11y.cy.js
```diff
@@ -14,20 +14,115 @@ describe( 'Run some accessibility tests', function() {
                 { id: 'empty-table-header', enabled: false },
                 { id: 'heading-order', enabled: false },
                 { id: 'no-dup-id', enabled: false },
-                { id: 'aria-allowed-role', enabled: false }
+                { id: 'aria-allowed-role', enabled: false },
+                {id: 'label-title-only', enabled:false },
```
Crabcyborg, line 18: Is it possible to only ignore these on the pages that are violating them? I would like to avoid more exceptions on the pages that were already passing the baseline.

lauramekaj1: @Crabcyborg Thank you for the comments. I added fixes accordingly!

Crabcyborg (approved):
Thank you @lauramekaj1! This looks great!

I pushed a small commit to rename the function. Since the baseline is no longer applied by default, the `baseline` part of the function name didn't seem accurate anymore.

🚀

PR #1883, opened 2024-07-25
Fix NEW style for Payments item when hovering over the formidable admin menu.

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5212
**Before**
CleanShot 2024-07-25 at 22 26 43@2x

**After**
CleanShot 2024-07-25 at 22 23 57@2x


AbdiTolesa:
@Crabcyborg I have left some styles that use css vars in `frm_admin.css` since they are not defined if we are only loading `frm_admin_global.css`. I used hard coded values instead in that case. Let me know if you have a different thought on this.

Crabcyborg:
@shervElmi Since this is CSS related, I'd like your review on this.

Thank you!

shervElmi:
Thanks for considering me @Crabcyborg!

@AbdiTolesa, there is a hover style issue with the background color, could you also consider fixing that? Thanks!

CleanShot 2024-07-26 at 16 30 21@2x

shervElmi:
The menu item text and tag aren't vertically aligned:

CleanShot 2024-07-26 at 16 30 21@2x

css/frm_admin.css
```diff
@@ -2942,10 +2942,6 @@ body.frm-body-with-open-modal {
 .frm-meta-tag:hover {
 	padding: var(--gap-2xs) var(--gap-xs);
```
shervElmi, line 2943: Remove padding styles from `frm_admin.css` since they are already in `frm_admin_global.css` now.

css/frm_admin.css
```diff
@@ -2981,7 +2977,6 @@ body.frm-body-with-open-modal {
 
 .frm-meta-tag.frm-new-pill {
 	background-color: var(--success-500);
```
shervElmi, line 2979: You should also remove `background-color` here and prevent overwriting them in the `frm_admin.css` file.

AbdiTolesa:
> @AbdiTolesa, there is a hover style issue with the background color. Could you also consider fixing that? Thanks!

@shervElmi I can't see a background applied when hovering over the items as in this screenshot:
CleanShot 2024-07-30 at 16 36 53@2x

shervElmi:
@AbdiTolesa, you should hover over the badge element, I mean the "NEW" badge.

css/frm_admin.css
AbdiTolesa: I left those here intentionally since we don't have the css vars defined in `frm_admin_global.css` and we are forced to hard code values there and maintain them manually if there is any color definition change in the future. It might be ok for now since there are a few deviations there.

AbdiTolesa:
> @AbdiTolesa, you should hover over the tag element, I mean the "NEW" tag.

@shervElmi I tried that but see no issues there as well:
CleanShot 2024-07-30 at 16 41 56@2x

css/frm_admin.css
shervElmi: In this case, we should use a hard value to ensure consistent design and prevent multiple designs on different pages.

shervElmi:
This is the CSS on hover for tag elements, just rewrite it for `.frm-new-pill`.

CleanShot 2024-07-30 at 16 47 15@2x

AbdiTolesa:
@Crabcyborg One thing I think is better to confirm you,
Should we apply background style to the "NEW" badge on hover, as Sherv suggested like in this screenshot below?
CleanShot 2024-07-30 at 17 04 21@2x

Crabcyborg:
Thanks @AbdiTolesa. My gut is that we don't need to change the background colour of the "NEW" pill on hover. It isn't really a unique clickable object, so it wouldn't make a lot of sense to me.

But I think we could also pass this question by @tuguirazvan.

Razvan, what do you think? Should we try to change the background color on this "NEW" element when you hover over it?

Screen Shot 2024-07-30 at 12 19 20 PM

shervElmi:
@AbdiTolesa, actually my review aligns with Mike's input on this matter. I found an incorrect hover background. It seems the parent class you're using, `.wp-submenu.wp-submenu-wrap...`, changes in some situations like response, etc.

So, you should use a higher order ID for this purpose to ensure you can even overwrite the hover state (ID selectors are always stronger than class selectors, even if the hover selector has a different style) or overwrite the hover state with the current parent selector you chose.

tuguirazvan:
@Crabcyborg I have discussed this with Steve [here](https://strategy11.slack.com/archives/C79564K98/p1720544896864619) and decided not to use the green badge as it seems a bit too much. Instead, we could opt for something slimmer and simpler, such as an orange text label with 'NEW'. I will attach an example below. Not sure if Steve communicated this decision with you all. 

image

Crabcyborg:
Thank you @tuguirazvan. I'm good with making that change.

@AbdiTolesa This will require a lot of changes to this current update. When you have the chance, can you try to get this looking more like the design from Razvan shared above?

Thank you!

Crabcyborg:
@AbdiTolesa I'm not sure if you missed the design updates it this PR or not but Razvan came back with new design changes (https://github.com/Strategy11/formidable-forms/pull/1883#issuecomment-2276041218).

Can you prioritize this update this week? I'd like to make sure we have this update for the next core release (in 2 weeks). It should be a quick update.

Thank you!

AbdiTolesa:
> @AbdiTolesa I'm not sure if you missed the design updates it this PR or not but Razvan came back with new design changes ([#1883 (comment)](https://github.com/Strategy11/formidable-forms/pull/1883#issuecomment-2276041218)).
> 
> Can you prioritize this update this week? I'd like to make sure we have this update for the next core release (in 2 weeks). It should be a quick update.
> 
> Thank you!

@Crabcyborg Sure, I just got back to this and here is how the NEW tag looks currently:
image

Crabcyborg:
Thanks @AbdiTolesa!

@shervElmi Can you review the new changes? Thank you!

css/admin/frm_admin_global.css
```diff
@@ -7,3 +7,12 @@
 	font-size: 9px;
 	vertical-align: middle;
 }
+
+.wp-submenu.wp-submenu-wrap .frm-meta-tag.frm-new-pill {
```
shervElmi, line 11: To make this CSS selector more optimized and concise, you can use the following:

```css
#adminmenu .frm-new-pill
```

css/admin/frm_admin_global.css
```diff
@@ -7,3 +7,12 @@
...
+	color: #F47449;
+	background: inherit;
+	margin-left: 0;
+	font-size: unset;
```
shervElmi, line 15: In Razvan's design, the font size is smaller than in the implementation. I tested the following, and it works perfectly:

```css
font-size: 85%;
```

### Before
CleanShot 2024-08-26 at 21 03 37@2x

### After
CleanShot 2024-08-26 at 21 04 10@2x

css/admin/frm_admin_global.css
```diff
@@ -7,3 +7,12 @@
 	font-size: 9px;
 	vertical-align: middle;
 }
+
+.wp-submenu.wp-submenu-wrap .frm-meta-tag.frm-new-pill {
+	color: #F47449;
+	background: inherit;
+	margin-left: 0;
```
shervElmi, line 14: You should remove padding here since we don't have any background, and change the margin-left to the following for perfect pixel alignment:

```css
margin-left: 5px;
```

css/admin/frm_admin_global.css
```diff
@@ -7,3 +7,12 @@
 	font-size: 9px;
 	vertical-align: middle;
 }
+
+.wp-submenu.wp-submenu-wrap .frm-meta-tag.frm-new-pill {
+	color: #F47449;
+	background: inherit;
```
shervElmi, line 13: Since we don't have any background, could you change `inherit` to `none`?

css/admin/frm_admin_global.css
```diff
@@ -7,3 +7,12 @@
...
+	margin-left: 0;
+	font-size: unset;
+	font-weight: unset;
```
shervElmi, line 16: Changing `unset` to `inherit` is a more future-proof choice here.

css/admin/frm_admin_global.css
AbdiTolesa: I think this looks more like the design with 0 left padding.

shervElmi: Thanks for the update, @AbdiTolesa!

However, I noticed that there's more space between the badge and the original text in the implemented design.

Implemented design:
CleanShot 2024-08-30 at 14 31 13@2x

Razvan's design:
CleanShot 2024-08-30 at 14 32 17@2x

shervElmi:
Thanks for the update again, @AbdiTolesa.

You should also remove the current padding by adding:

```css
padding: 0;
```

shervElmi (approved):
Thank you, @AbdiTolesa! It looks great now! 🎉

Crabcyborg (approved):
Thank you @AbdiTolesa @shervElmi and @tuguirazvan for your work on this issue!

🚀

PR #1885, opened 2024-07-25
Try to optimize FrmEntryMeta::get_ids_query

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5167

**Testing instructions**
This update changes how the database is queried when running many queries that check for entry IDs. This includes the `[frm-search]` functionality in Formidable Views.

You can view the database queries by activating the Query Monitor plugin.

**Before**
The query when searching for entries before included two `JOIN` statements. Getting the data from three tables is quite slow.

**After**
The slow `LEFT OUTER JOIN` that joins data from the fields table is now removed from the query when there is no required field data in the query. This changes the query when searching using `[frm-search]` so it only includes a single `JOIN`. The new query (when searching data for ~50k entries), dropped from ~1.5s to ~1.1s, saving approximately 0.4 seconds from the page load (which is quite significant).

**What to test for**
- Confirming the queries when searching are still accurate.
- Confirming there are no entry ID queries that are now broken.
- Confirming the queries when searching are more efficient.


classes/models/FrmEntryMeta.php
```diff
@@ -399,6 +399,33 @@ public static function get_top_level_entry_ids( $query, $args ) {
...
+	 * @return bool
+	 */
+	private static function should_join_fields_table( &$where ) {
+		if ( is_string( $where ) ) {
+			if ( preg_match( '/\bfi\.(?!form_id)\w+/i', $where ) ) {
+				return true;
+			}
+			$where = str_replace( 'fi.form_id', 'e.form_id', $where );
+			return false;
+		}
```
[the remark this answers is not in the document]
truongwp: @AbdiTolesa Since this where could be modified via filter hook, we should consider this suggestion.

tests/phpunit/entries/test_FrmEntryMeta.php
```diff
@@ -52,4 +52,23 @@ public function test_update_entry_metas() {
 
 		$this->assertNull( $meta );
 	}
+
+	public function test_should_join_fields_table() {
+		$where = 'fi.form_id=123';
+		$this->assertFalse( $this->run_private_method( array( 'FrmEntryMeta', 'should_join_fields_table' ), array( &$where ) ) );
```
truongwp, line 58: We should assert the `$where` after this as well.

truongwp: I mean asserting the `$where` value like you do on line 72.

truongwp (approved):

classes/models/FrmEntryMeta.php
```diff
@@ -425,8 +452,13 @@ private static function get_ids_query( $where, $order_by, $limit, $unique, $args
 			$query[] = 'it.item_id';
 		}
 
-		$query[] = 'FROM ' . $wpdb->prefix . 'frm_item_metas it LEFT OUTER JOIN ' . $wpdb->prefix . 'frm_fields fi ON it.field_id=fi.id';
+		$from = 'FROM ' . $wpdb->prefix . 'frm_item_metas it';
+
+		if ( self::should_join_fields_table( $where ) ) {
```
Crabcyborg, line 457: @AbdiTolesa I think this is probably good to go as-is, but what are your thoughts on also checking/possibly updating `$order_by` for references to the fields table?

AbdiTolesa: I think that may be worth it and just added an additional check for `order_by` now.

classes/models/FrmEntryMeta.php
```diff
@@ -425,8 +452,13 @@ private static function get_ids_query( $where, $order_by, $limit, $unique, $args
 			$query[] = 'it.item_id';
 		}
 
-		$query[] = 'FROM ' . $wpdb->prefix . 'frm_item_metas it LEFT OUTER JOIN ' . $wpdb->prefix . 'frm_fields fi ON it.field_id=fi.id';
+		$from = 'FROM ' . $wpdb->prefix . 'frm_item_metas it';
+
+		if ( self::should_join_fields_table( $where ) || self::should_join_fields_table( $order_by ) ) {
```
Crabcyborg, line 457: @AbdiTolesa I think this could still have issues.

If I both have a `$where` that includes `fi.form_id`, and a `$order_by` that includes `fi.form_id`, then the first check will return true, and it will never update the `$order_by`, leaving it referencing the table that isn't in the query.

I think we need to always call both functions, then check the results.

Crabcyborg (approved):

lauramekaj1 (approved):
@AbdiTolesa I tested this and it looks good to me. Thank you!

Crabcyborg:
Thank you Laura!

🙌

classes/models/FrmEntryMeta.php
```diff
@@ -399,6 +399,33 @@ public static function get_top_level_entry_ids( $query, $args ) {
...
+			$where = str_ireplace( 'fi.form_id', 'e.form_id', $where );
+			return false;
+		}
+		$where_fields = array_keys( $where );
+		foreach ( $where_fields as $where_field ) {
+			if ( strpos( $where_field, 'fi.' ) === 0 && 'fi.form_id' !== $where_field ) {
+				return true;
+			}
+		}
+		$where['e.form_id'] = $where['fi.form_id'];
```
Crabcyborg, line 424: @AbdiTolesa It looks like this is breaking a unit test in Pro now.

It's throwing a `Undefined index: fi.form_id` error.

I think we only want to add `$where['e.form_id']` when `$where['fi.form_id']` is actually set. It looks like that check was missed.

Crabcyborg: I made a small PR that should resolve this https://github.com/Strategy11/formidable-forms/pull/2087

PR #1893, opened 2024-07-30
search valid and invalid form title

lauramekaj1:


tests/cypress/e2e/Forms/searchFunctionality.cy.js
```diff
@@ -0,0 +1,34 @@
...
+
+    });
+
+    it("should search valid and invalid form title", () => {
+        cy.createNewForm();
+
+        cy.log("Search the newly created form");
+        cy.get('#entry-search-input').type("Test Form {enter}");
+        cy.get('#entry-search-input').type("Test Form").clear();
+        cy.get('#entry-search-input').type("Test Form");
```
Crabcyborg, line 15: @lauramekaj1 What exactly is going on on these three lines. It types Test Form there times. Once with enter, once with clear, and one with nothing.

Do we need all of these? Is it adding anything to have all three of these?

tests/cypress/e2e/Forms/searchFunctionality.cy.js
```diff
@@ -0,0 +1,34 @@
...
+        cy.get('#search-submit').click();
+
+        cy.get('.published > .current').should("exist");
+        cy.get('.current > .count').should("contain", "1");
+        cy.get('.displaying-num').should("contain", "1");
+        cy.contains('#the-list tr', 'Test Form').should("exist");
+
+        cy.log("Search for an invalid form title");
+        cy.get('#entry-search-input').type("Test Form").clear();
```
Crabcyborg, line 24: Do we need this line first or can we just type the other? Does having this line add anything?

tests/cypress/e2e/Forms/searchFunctionality.cy.js
lauramekaj1: @Crabcyborg thanks for the review.
So,on the first line it tests that a form can be searched by using enter. After that it clears the field and tests that it can search a form by using the Search button (line 16)!

tests/cypress/e2e/Forms/searchFunctionality.cy.js
lauramekaj1: @Crabcyborg yes I think I can remove this line!

tests/cypress/e2e/Forms/searchFunctionality.cy.js
Crabcyborg: Should we have assertions between those? If it just tests both back to back without more tests, is it covering much?

lauramekaj1: Yes you are right, I will add assertions. Thank you!

Crabcyborg (approved):
Thanks @lauramekaj1!

🚀

PR #1894, opened 2024-08-01
Do not add a new option to conditional logic if renaming an option

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/2909

### Test steps
1. Create a Product field with some product options.
2. Add another field that conditionally shows based off the product field's value and save.
3. Go the the Product field and rename an option.
4. Go back to the conditional field and confirm that there is no new option added in the conditional logic row. The existing options should have been renamed instead.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -5636,8 +5636,6 @@ function frmAdminBuildJS() {
 					if ( hasMatch ) {
 						continue;
 					}
-					prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption );
-					continue;
 				}
 				prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption );
 			}
@@ -5651,10 +5649,7 @@ function frmAdminBuildJS() {
 
 	function prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption ) {
 		if ( optionMatch === null ) {
-			optionMatch = frmDom.tag( 'option', {
-				text: expectedOption,
-				value: expectedOption
-			});
+			optionMatch = frmDom.tag( 'option', { text: expectedOption });
 			optionMatch.value = expectedOption;
 		}
 
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -5632,12 +5632,13 @@ function frmAdminBuildJS() {
 				const expectedChoiceEl = Array.from( fieldChoices ).find( element => element.value === expectedOption );
 				if ( expectedChoiceEl ) {
 					const hasMatch = valueSelect.querySelector( 'option[value="' + expectedChoiceEl.getAttribute( 'data-value-on-focus' ) + '"]' );
-					if ( ! hasMatch ) {
-						prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption );
+					if ( hasMatch ) {
+						continue;
 					}
-				} else {
 					prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption );
+					continue;
 				}
+				prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption );
 			}
 
 			optionMatch = valueSelect.querySelector( 'option[value=""]' );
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This fix works great!

js/formidable_admin.js
```diff
@@ -5644,6 +5649,16 @@ function frmAdminBuildJS() {
 		}
 	}
 
+	function prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption ) {
+		if ( optionMatch === null ) {
+			optionMatch = document.createElement( 'option' );
```
truongwp, line 5654: Can we use the `tag()` function of `frmDom` here?

js/formidable_admin.js
```diff
@@ -5628,13 +5628,18 @@ function frmAdminBuildJS() {
...
 				optionMatch = valueSelect.querySelector( 'option[value="' + expectedOption + '"]' );
 
-				if ( optionMatch === null ) {
-					optionMatch = document.createElement( 'option' );
-					optionMatch.setAttribute( 'value', expectedOption );
-					optionMatch.textContent = expectedOption;
+				const fieldChoices     = document.querySelectorAll( '#frm_field_' + fieldId + '_opts input[data-value-on-focus]' );
+				const expectedChoiceEl = Array.from( fieldChoices ).find( element => element.value === expectedOption );
+				if ( expectedChoiceEl ) {
+					const oldValue = expectedChoiceEl.getAttribute( 'data-value-on-focus' );
```
truongwp, line 5634: I think we can use `dataset.valueOnFocus` instead of `getAttribute()`.

truongwp (approved):

js/formidable_admin.js
```diff
@@ -5628,13 +5628,18 @@ function frmAdminBuildJS() {
...
+				const fieldChoices     = document.querySelectorAll( '#frm_field_' + fieldId + '_opts input[data-value-on-focus]' );
+				const expectedChoiceEl = Array.from( fieldChoices ).find( element => element.value === expectedOption );
+				if ( expectedChoiceEl ) {
+					const oldValue = expectedChoiceEl.dataset.valueOnFocus;
+					const hasMatch = oldValue && valueSelect.querySelector( 'option[value="' + oldValue + '"]' );
+					if ( hasMatch ) {
+						continue;
+					}
+					prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption );
+					continue;
```
Crabcyborg, line 5640: @AbdiTolesa Having two `continue` statements in this `if` condition feels a bit weird.

In this case, I think it would look a lot cleaner if you removed a `continue`. Something like,

```
const hasMatch = oldValue && valueSelect.querySelector( 'option[value="' + oldValue + '"]' );
if ( ! hasMatch ) {
    prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption );
}
continue;
```

js/formidable_admin.js
```diff
@@ -5644,6 +5649,18 @@ function frmAdminBuildJS() {
 		}
 	}
 
+	function prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption ) {
+		if ( optionMatch === null ) {
+			optionMatch = frmDom.tag( 'option', {
+				text: expectedOption,
+				value: expectedOption
```
Crabcyborg, line 5656: @AbdiTolesa `value: expectedOption` isn't a support option for the `tag` function. You fix it in the following `optionMatch.value = expectedOption;` line below. It shouldn't also be here. This line should be removed.

js/formidable_admin.js
Crabcyborg: I updated this in a different way. The `prependValueSelectWithOptionMatch( valueSelect, optionMatch, expectedOption );` was getting called twice.

There's really no reason to `continue` after calling the method either, so I just removed some code here instead.

js/formidable_admin.js
Crabcyborg: I updated this.

Crabcyborg (approved):
I made some small tweaks to fix the issues I commented on.

I just tested this and it looks like it solves the issue.

🚀

PR #1895, opened 2024-08-01
Prevent user id field from being dragged into field group

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5225

CleanShot 2024-08-01 at 22 36 38

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1873,8 +1873,10 @@ function frmAdminBuildJS() {
 		}
 
 		const isHiddenField = draggable.classList.contains( 'edit_field_type_hidden' );
-		if ( isHiddenField ) {
-			// Hidden fields should not be added to field groups since they're not shown and don't make sense with the grid distribution.
+		const isUserIdField = draggable.classList.contains( 'edit_field_type_user_id' );
+		if ( isHiddenField || isUserIdField ) {
+			// Hidden fields and user id fields should not be added to field groups since they're not shown
+			// and don't make sense with the grid distribution.
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -788,7 +788,12 @@ function frmAdminBuildJS() {
 		document.querySelectorAll( sortableSelector ).forEach(
 			list => {
 				makeDroppable( list );
-				Array.from( list.children ).forEach( child => makeDraggable( child, '.frm-move' ) );
+				Array.from( list.children ).forEach( child => {
```
Crabcyborg, line 791: @AbdiTolesa I don't think you'll want to update the initialization code to fix this.

Your update should be contained to the `allowDrop` function.

Crabcyborg: Going further, you should really only need to update `allowNewFieldDrop`.

AbdiTolesa: @Crabcyborg Yeah I was just updating that already but just saw your comment now.

Crabcyborg (approved):
Thanks for working on this @AbdiTolesa.

I moved some code around and fixed an issue where you could still drag fields beside a user ID field in a field group that way.

The user ID check makes more sense where the hidden field check is, and the function `groupIncludesBreakOrHidden` needed to also start checking for user id fields since that helper function covers the similar issue with hidden fields.

🚀

AbdiTolesa:
> Thanks for working on this @AbdiTolesa.
> 
> I moved some code around and fixed an issue where you could still drag fields beside a user ID field in a field group that way.
> 
> The user ID check makes more sense where the hidden field check is, and the function `groupIncludesBreakOrHidden` needed to also start checking for user id fields since that helper function covers the similar issue with hidden fields.
> 
> I think this is good to go now.
> 
> 🚀

Thanks @Crabcyborg!

PR #1899, opened 2024-08-05
Add label to entries list checkboxes

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5239

CleanShot 2024-08-05 at 16 52 47@2x

### DIFF
--- a/classes/helpers/FrmEntriesListHelper.php
+++ b/classes/helpers/FrmEntriesListHelper.php
@@ -307,7 +307,7 @@ class FrmEntriesListHelper extends FrmListHelper {
 		// Set up the checkbox ( because the user is editable, otherwise its empty )
 		$checkbox = "<input type='checkbox' name='item-action[]' id='cb-item-action-{$item->id}' value='{$item->id}' />";
 		/* translators: %s: Form name */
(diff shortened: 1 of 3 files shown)


classes/helpers/FrmEntriesListHelper.php
```diff
@@ -292,6 +292,8 @@ public function single_row( $item, $style = '' ) {
 
 		// Set up the checkbox ( because the user is editable, otherwise its empty )
 		$checkbox = "<input type='checkbox' name='item-action[]' id='cb-item-action-{$item->id}' value='{$item->id}' />";
+		/* translators: %s: Form name */
+		$checkbox .= "<label for='cb-item-action-{$item->id}'/><span class='screen-reader-text'>" . sprintf( __( 'Select %s', 'formidable' ), FrmFormsHelper::edit_form_link_label( $item->form_id ) ) . '</span></label>';
```
truongwp, line 296: There is a `/>` in the label tag, which is wrong.

AbdiTolesa: I think `screen-reader-text` should be kept as is to be consistent with how core WP does it.

truongwp (approved):

Crabcyborg (approved):
Thanks @AbdiTolesa!

I added some escaping to the translations.

I also removed the Cypress rule that was breaking.

I think we're good to merge this now.

🚀

PR #1900, opened 2024-08-05
validate data in list and excerpt view in forms page

lauramekaj1:


.eslintrc.json
```diff
@@ -46,6 +46,7 @@
 		"camelcase": ["error", { "properties": "never" }],
 		"lines-around-comment": "off",
 		"react/display-name": "off",
+		"no-unused-expressions": "off",
```
Crabcyborg, line 49: @lauramekaj1 I probably wouldn't add a universal rule that isn't exclusive to cypress to our list here.

What line are you having difficulty with passing this rule?

I would first try to fix the issue. If you can't, you can disable the rule in a specific file by adding a comment to the top of the file.

We have this example in our `formidable_admin.js` file at the moment. It needs quite a few exceptions at the moment,

/* eslint-disable jsdoc/require-param, prefer-const, no-redeclare, @wordpress/no-unused-vars-before-return, jsdoc/check-types, jsdoc/check-tag-names, @wordpress/i18n-translator-comments, @wordpress/valid-sprintf, jsdoc/require-returns-description, jsdoc/require-param-type, no-unused-expressions, compat/compat */

tests/cypress/e2e/Forms/formPageDataValidation.cy.js
```diff
@@ -0,0 +1,145 @@
...
+        cy.get(`[id^="item-action-"] > .name > strong > .row-title:contains("${formTitle}")`)
+            .parents('[id^="item-action-"]')
+            .within(() => {
+                cy.get('.entries > a').should("contain", "0");
+                cy.get('.name > strong > .row-title').should("contain", formTitle);
+                cy.get('.form_key').should("contain", "test-form");
+                cy.get('.shortcode > div').should("exist");
+
+                const currentDate = new Date();
+                const formattedDate = currentDate.toISOString().split('T')[0].replace(/-/g, '/');
```
Crabcyborg, line 79: @lauramekaj1 I notice these two lines are identical in two places. Maybe we could move this out into a new function? Like `getFormattedCurrentDate()`?

Crabcyborg (commented):
I think this looks great!

Thanks Laura!

Once we address the one comment I left, I think this should be good to merge.

.eslintrc.json
lauramekaj1: Thank you @Crabcyborg! Great call, I shouldn't have added that as a universal rule. I tried to fix the issue but I couldn't so I added eslint-enable no-unused-expressions for that line!

Crabcyborg (approved):
Thank you @lauramekaj1!

This looks awesome!

These tests are great!

🚀

Crabcyborg:
@lauramekaj1 There's another difference between the list and excerpt views that might be worth adding tests for.

When in "Excerpt" mode, the list should show the first 50 characters of the form description setting. A test that adds a form description, and then tests that it appears in Excerpt mode would be great.

lauramekaj1:
@Crabcyborg yes true I missed that! Will add coverage for that part as well! Thank you.

PR #1902, opened 2024-08-07
Prevent dragging new sections into sections

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5230

**Before**:
CleanShot 2024-08-07 at 09 37 40

**After**:
CleanShot 2024-08-07 at 09 39 00

### Test procedure
1. Create a form and add a section field to it.
2. Drag any field to the section field except a Section field.
3. Now, try dragging and dropping a Section field to the section, beside the field just added above.
4. Confirm that the action is rejected and you cannot add a Section inside a Section field.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -1834,7 +1834,7 @@ function frmAdminBuildJS() {
 			return true;
 		}
 
-		const newFieldWillBeAddedToASection = droppable.classList.contains( 'start_divider' ) || null !== droppable.closest( '.start_divider' );
+		const newFieldWillBeAddedToASection = droppable.classList.contains( 'start_divider' ) || droppable.classList.contains( '.start_divider' );
 		if ( newFieldWillBeAddedToASection ) {
 			// Don't allow a section or an embedded form in a section.
 			return ! newEmbedField && ! newSectionField;
@@ -1849,7 +1849,7 @@ function frmAdminBuildJS() {
 
 			return ! newHiddenField && ! newPageBreakField && ! newUserIdField;
 		}
-	
+
 		return true;
 	}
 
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -1854,6 +1844,16 @@ function frmAdminBuildJS() {
 			return ! newEmbedField && ! newSectionField;
 		}
 
+		const newFieldWillBeAddedToAGroup = ! ( 'frm-show-fields' === droppable.id || droppable.classList.contains( 'start_divider' ) );
+		if ( newFieldWillBeAddedToAGroup ) {
```
Crabcyborg, line 1848: @AbdiTolesa What is the reason for moving the code?

Crabcyborg: Can you make this fix in a separate PR? We should merge this fix right away.

AbdiTolesa: > Can you make this fix in a separate PR? We should merge this fix right away so it isn't like this on master. It shouldn't rely on this other update.

@Crabcyborg I have created a new PR for the undefined function error fix instead: https://github.com/Strategy11/formidable-forms/pull/1905

AbdiTolesa: > @AbdiTolesa What is the reason for moving the code?

@Crabcyborg The reason for moving the code is `newFieldWillBeAddedToAGroup` would be evaluated first and validates when you drag a section into a field group and the function returns true even if the field group you are dragging into is found in a section. Running `if ( newFieldWillBeAddedToASection ) {...` first makes sure to return `false` when dragging a section into a section. I don't know my explanation makes sense though...

garretlaxton (approved):
This works great!

truongwp (approved):

Crabcyborg:
@AbdiTolesa I messed up the commits on this. I was trying to a fix an issue but it seemed to introduce new issues.

I'm just going to close this and we can try another approach.

The issue with this update is that you can now drag several disallowed field types (user id, hidden field, page break) into field groups in sections.

This was previously fine because all field group checks would happen before that check, so there would be conflicts with field groups matching the section check.

We need to figure that out still.

PR #1903, opened 2024-08-07
Exclude submit field from entry table

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/5253


Crabcyborg (approved):
Thank you for fixing that @truongwp!

🚀

PR #1904, opened 2024-08-07
Check if openssl_encrypt function is defined before attempt to call it

AbdiTolesa:
Related issue: https://github.com/Strategy11/formidable-pro/issues/5234

### DIFF
--- a/classes/models/FrmFormState.php
+++ b/classes/models/FrmFormState.php
@@ -149,6 +149,9 @@ class FrmFormState {
 			return;
 		}
 		$state_string = $this->get_state_string();
+		if ( ! $state_string ) {
+			return;
+		}
 		echo '<input name="frm_state" type="hidden" value="' . esc_attr( $state_string ) . '" />';
 	}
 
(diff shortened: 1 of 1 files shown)


classes/models/FrmFormState.php
```diff
@@ -159,7 +162,10 @@ private function get_state_string() {
 		$secret           = self::get_encryption_secret();
 		$compressed_state = $this->compressed_state();
 		$json_encoded     = json_encode( $compressed_state );
-		$encrypted        = openssl_encrypt( $json_encoded, 'AES-128-ECB', $secret );
+		if ( ! function_exists( 'openssl_encrypt' ) ) {
```
Crabcyborg, line 165: @AbdiTolesa I think it would make sense to move this check earlier than right before.

If someone doesn't have it available, they don't need to call any of the other code before this.

We could put the check at the top of this function instead.

classes/models/FrmFormState.php
```diff
@@ -149,6 +149,9 @@ public function render_state_field() {
 			return;
 		}
 		$state_string = $this->get_state_string();
+		if ( ! $state_string ) {
+			return;
```
Crabcyborg, line 153: A comment here explaining that this would occur when open SSL isn't installed might be nice.

classes/models/FrmFormState.php
AbdiTolesa: Thanks @Crabcyborg! I have moved the check to the top of both functions and also created a small function for better readability and reducing comment repetition.

PR #1905, opened 2024-08-07
Fix undefined function: groupIncludesBreakOrHidden

AbdiTolesa:
Fixes an error introduced in https://github.com/Strategy11/formidable-forms/commit/9e1b2cda5e40a26472158811f1a522d970883c36
image


Crabcyborg (approved):
Thanks for fixing that and moving it into a separate update @AbdiTolesa!

PR #1906, opened 2024-08-07
Avoid corrupted custom html when duplicating field

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5248

### Test steps
0. Download the XML from https://drive.google.com/file/d/1zJL2dhe0XdbIZ8K_3xBp2_nsYgfOni8C/view?usp=sharing and import it.
1. Try duplicating the form with title: `Towards Accessible Waterways 2023 Exhibitor Form`
2. Preview the duplicate
3. Confirm that there is no `[if field_ix]...` shortcodes in place of the field descriptions like below:
image
4. Try updating the field descriptions for the imported field and check everything works as expected on the form preview.

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -1512,7 +1512,7 @@ class FrmFieldsHelper {
 			foreach ( $val as $k => $v ) {
 				if ( is_string( $v ) ) {
 					if ( 'custom_html' === $k ) {
-						$val[ $k ] = self::switch_ids_except_strings( $replace, $replace_with, array( '[if description]', '[/if description]' ), $v );
+						$val[ $k ] = self::switch_ids_except_strings( $replace, $replace_with, array( '[if description]', '[description]', '[/if description]' ), $v );
 						unset( $k, $v );
 						continue;
 					}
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This fix works great!

classes/helpers/FrmFieldsHelper.php
```diff
@@ -1522,6 +1527,31 @@ public static function switch_field_ids( $val ) {
...
+	 */
+	private static function switch_ids_except_strings( $replace, $replace_with, $exceptions, $value ) {
+		foreach ( $exceptions as $exception ) {
+			$index = array_search( $exception, $replace, true );
+			if ( false === $index ) {
+				continue;
+			}
+			unset( $replace[ $index ] );
+			unset( $replace_with[ $index ] );
+		}
```
truongwp, line 1550: When there are multiple HTML fields, this loop will run multiple times. It's better if we can make this loop run once.

AbdiTolesa: @truongwp I didn't get you, we can only have a single `custom_html` for a field, right? This issue is only related to that field... or did I miss something here?

truongwp: Sorry. I was wrong. I thought we're looping through a list of fields.

truongwp (approved):

Crabcyborg (approved):
This looks good to me.

🙌

PR #1907, opened 2024-08-08
Stop using font icons for form actions collapsible

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5249
CleanShot 2024-08-08 at 11 04 37

### DIFF
--- a/classes/views/frm-form-actions/form_action.php
+++ b/classes/views/frm-form-actions/form_action.php
@@ -24,7 +24,7 @@ if ( FrmOnSubmitAction::$slug === $form_action->post_excerpt ) {
 	<div class="widget-top">
 		<div class="widget-title-action">
 			<button type="button" class="widget-action hide-if-no-js" aria-expanded="false">
-				<i class="frm_icon_font frm_arrow_right_icon" aria-hidden="true"></i>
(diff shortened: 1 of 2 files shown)


truongwp:
@AbdiTolesa I think we can remove this CSS https://github.com/Strategy11/formidable-forms/blob/62e5898163b38ec797cd87ea05f2b2df79c1f493/css/frm_admin.css#L6799

truongwp:
We should add `transition` for the icon, the previous `<i>` tag has a transition CSS rule. We can copy it for the svg and probably remove the CSS rule for that `<i>` tag.

AbdiTolesa:
> We should add `transition` for the icon, the previous `<i>` tag has a transition CSS rule. We can copy it for the svg and probably remove the CSS rule for that `<i>` tag.

@truongwp I have removed the old css, thanks! I didn't see any significant transition effect for the old `<i>` tag, are you referring to the the transition below?
image

truongwp:
@AbdiTolesa Sorry, the old <i> tag doesn't have the transition for rotation.

truongwp (approved):

Crabcyborg (approved):

🚀

PR #1917, opened 2024-08-12
validate elements and data of entries of submitted forms

lauramekaj1:


Crabcyborg (approved):
Thank you @lauramekaj1!

I think this looks good.

🙌

PR #1918, opened 2024-08-14
Replace font icons with SVGs for edd license status

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5251

Enhancements:
- Switched from using font icons to SVGs.
- User can add/remove license without refreshing the page.
- The activate/deactivate button is disabled until response is received for the prior request

### Test steps
1. Go to the Plugin Licenses tab in the Global Settings and try adding and removing add-on licenses.
2. Confirm that the icons reflect the action of success/error action as expected.
3. Also confirm that, a user can add/remove a license for an add-on without leaving the page, since the 'Activate' and 'Deactivate' elements toggle after few seconds of action.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -5204,7 +5204,7 @@ label.frm-example-icon {
 }
 
 .edd_frm_license_row .frm_action_success .frmsvg {
-	fill: green;
+	color: green;
 }
 
 .edd_frm_license_row .frm_action_error .frmsvg {
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -5204,8 +5204,16 @@ label.frm-example-icon {
 	margin: 0 5px;
 }
 
-.edd_frm_status_icon.frm_icon_font.frm_inactive_icon {
-	color: green;
+.edd_frm_status_icon.frmsvg {
+	margin: 0;
+}
+
+.edd_frm_status_icon.frmsvg {
+	fill: #D54E21;
+}
+
+.edd_frm_status_icon.frmsvg.frm_inactive_icon {
+	fill: green;
 }
 
 /* Hide the install steps for solutions */
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -5208,7 +5208,7 @@ label.frm-example-icon {
 }
 
 .edd_frm_license_row .frm_action_error .frmsvg {
-	fill: #D54E21;
+	color: #D54E21;
 }
 
 /* Hide the install steps for solutions */
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -5204,7 +5204,7 @@ label.frm-example-icon {
 	margin: 0 5px;
 }
 
-.edd_frm_status_icon.frmsvg.frm_inactive_icon {
+.edd_frm_status_icon.frmsvg.frm_check1_icon {
 	fill: green;
 }
 
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -5199,17 +5199,9 @@ label.frm-example-icon {
 	display: none;
 }
 
-.edd_frm_status_icon.frm_icon_font {
+.edd_frm_status_icon.frmsvg {
 	color: #D54E21;
 	margin: 0 5px;
-}
-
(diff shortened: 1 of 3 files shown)


AbdiTolesa:
@Crabcyborg I have replaced the icons on load (PHP) now. Are those icons changed with js when the license for a given plugin is activated/deactivated. If so, is there a plugin license I can use for test that you can share me?

Crabcyborg:
@AbdiTolesa I left a reply on the issue. Since this is a public repo I didn't want to share any license keys here.

https://github.com/Strategy11/formidable-pro/issues/5251#issuecomment-2288733953

garretlaxton (approved):
This looks great!

css/frm_admin.css
```diff
@@ -5199,13 +5199,16 @@ label.frm-example-icon {
...
-.edd_frm_status_icon.frm_icon_font {
-	color: #D54E21;
+.edd_frm_status_icon.frmsvg {
 	margin: 0 5px;
 }
 
-.edd_frm_status_icon.frm_icon_font.frm_inactive_icon {
-	color: green;
+.edd_frm_license_row .frm_action_success .frmsvg {
+	fill: green;
```
truongwp, line 5207: We had the CSS `fill: currentColor` for `.frmsvg`. So I think keeping the `color` of `.frm_icon_font` as the old version will work.

css/frm_admin.css
```diff
@@ -5199,15 +5199,18 @@ label.frm-example-icon {
...
 	margin: 0 5px;
 }
 
-.edd_frm_status_icon.frm_icon_font.frm_inactive_icon {
+.edd_frm_license_row .frm_action_success .frmsvg {
 	color: green;
 }
 
+.edd_frm_license_row .frm_action_error .frmsvg {
+	fill: #D54E21;
```
truongwp, line 5211: We should use `color` here too.

truongwp (approved):

js/formidable_admin.js
```diff
@@ -8676,27 +8676,35 @@ function frmAdminBuildJS() {
...
 					setTimeout( function() {
 						messageBox.html( '' );
-					}, 15000 );
+						thisRow.find( '.frm_icon_font' ).addClass( 'frm_hidden' );
+						if ( actionIsSuccess ) {
+							const actionIsActivate = action === 'activate';
+							thisRow.get(0).querySelector( '.edd_frm_unauthorized' ).classList.toggle( 'frm_hidden', actionIsActivate )
+							thisRow.get(0).querySelector( '.edd_frm_authorized' ).classList.toggle( 'frm_hidden', ! actionIsActivate )
+						}
+					}, 5000 );
```
Crabcyborg, line 8707: @AbdiTolesa Can we update the icons quicker than this 5 second timeout? It feels jarring that it shows "Activate" still and an open input for 5 seconds. It feels like those things should change more immediately.

AbdiTolesa: @Crabcyborg I just cut that down to 2s and feel like that is optimal now.

js/formidable_admin.js
```diff
@@ -8676,27 +8676,35 @@ function frmAdminBuildJS() {
...
 				const messageBox = thisRow.find( '.frm_license_msg' );
 				messageBox.html( msg.message );
 				if ( msg.message !== '' ) {
 					setTimeout( function() {
 						messageBox.html( '' );
-					}, 15000 );
+						thisRow.find( '.frm_icon_font' ).addClass( 'frm_hidden' );
+						if ( actionIsSuccess ) {
+							const actionIsActivate = action === 'activate';
+							thisRow.get(0).querySelector( '.edd_frm_unauthorized' ).classList.toggle( 'frm_hidden', actionIsActivate )
```
Crabcyborg, line 8704: @AbdiTolesa Neither of these lines have a semi-colon at the end.

Crabcyborg: I added these.

Crabcyborg (approved):
I think this is probably good enough. I was going to try making a PR with some improvements but I see that it's probably more work than it's worth.

🚀

PR #1922, opened 2024-08-17
Hide notice from some plugins

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/5178

To replicate, I modified the plugin code to always show their notices.

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -2601,13 +2601,15 @@ h2.frm-h2 + .howto {
 #wpbody-content > .update-nag,
 #wpbody-content > .notice,
 #wpbody-content > .error:not(.frm_previous_install),
-#wpbody-content > .cky-admin-notice,
+#wpbody-content > [class*="admin-notice"],
 .frm-white-body .updated,
 .frm-white-body .notice,
-.frm-white-body .cky-admin-notice,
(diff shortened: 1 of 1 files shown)


css/frm_admin.css
```diff
@@ -2601,9 +2601,13 @@ h2.frm-h2 + .howto {
 #wpbody-content > .update-nag,
 #wpbody-content > .notice,
 #wpbody-content > .error:not(.frm_previous_install),
+#wpbody-content > .cky-admin-notice,
 .frm-white-body .updated,
 .frm-white-body .notice,
+.frm-white-body .cky-admin-notice,
+.frm-white-body #codepeople-review-banner,
```
Crabcyborg, line 2608: @truongwp Can we do this with less specific classes?

Something like `.frm-white-body [class$="-admin-notice"]`, `.frm-white-body [id$="-review-banner"]`?

That should help to catch more.

truongwp: I updated it. I keep `#codepeople-review-banner` to override its `display`. I use `[class*=""]` instead of `$=` because those classes might be in the middle of `class` attribute.

Crabcyborg (approved):
Thank you @truongwp!

🚀

PR #1924, opened 2024-08-19
Use SVG for tooltip icon

truongwp:
Related issue: https://github.com/Strategy11/formidable-pro/issues/5250


truongwp:
@Crabcyborg Do you think it's worth adding a helper function? In other plugins, we will need to check if it exists.

Crabcyborg:
@truongwp I think a helper function would be great.

As long as we get Lite out first before the other plugins it shouldn't be too much of an issue to not include tooltips when Lite is not up-to-date.

Crabcyborg (approved):
This looks great!

Thank you so much Truong!

🥳

PR #1925, opened 2024-08-19
Sync conditional logic options after field option deletion

AbdiTolesa:
Related issue: https://github.com/Strategy11/formidable-pro/issues/4254
Requires: https://github.com/Strategy11/formidable-pro/pull/5313

### Test steps
1. Add one of those fields mentioned in the title and add some options/choices
2. Add another field to the form and create a conditional logic that uses one of those option fields mentioned in the title
3. Select the option field in Step 1 and delete some of options
4. Go back to the field dependent on this ( the one created in Step 2 ), and confirm the conditional logic, the deleted options should not be available anymore.

CleanShot 2024-08-19 at 16 18 07

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -3589,6 +3589,10 @@ function frmAdminBuildJS() {
 			wp.hooks.doAction( 'frm_before_delete_field_option', this );
 			jQuery( parentLi ).remove();
 
+			setTimeout( () => {
+				resetDisplayedOpts( fieldId );
+			}, 1000 );
+
 			const hasOther = jQuery( parentUl ).find( '.frm_other_option' );
 			if ( hasOther.length < 1 ) {
 				otherInput = document.getElementById( 'other_input_' + fieldId );
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -3587,9 +3587,7 @@ function frmAdminBuildJS() {
 
 		jQuery( parentLi ).fadeOut( 'slow', function() {
 			wp.hooks.doAction( 'frm_before_delete_field_option', this );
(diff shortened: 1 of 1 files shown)


garretlaxton:
This doesn't work for me if I have a field selected. Below, you can see I deleted the `Under 18` option, but it still shows in the dropdown. If I don't select the checkbox field, then it will work fine.
image

AbdiTolesa:
> This doesn't work for me if I have a field selected. Below, you can see I deleted the `Under 18` option, but it still shows in the dropdown. If I don't select the checkbox field, then it will work fine.

@garretlaxton I didn't get it when you say "when I have a field selected". Could you share a screen record of the case not working for you well?

garretlaxton:
Here is a video: https://www.loom.com/share/e053d87d43ac426e8583442aa3acef97

Notice the dropdown field works since it wasn't selected in the conditional logic section.

AbdiTolesa:
> Here is a video: https://www.loom.com/share/e053d87d43ac426e8583442aa3acef97
> 
> Notice the dropdown field works since it wasn't selected in the conditional logic section.

Interesting that we see different results. Here is my screen record that works even if the option field is selected in the conditional logic for the other field.
CleanShot 2024-08-22 at 08 57 35

Could you share your form export too?

garretlaxton:
Thanks Abdi, this works for me on the QA site. One thing I did find is when I bulk edit the field options, the previous options show in the conditional logic section.

I made a quick video of this: https://www.loom.com/share/e226170396ba4c3094eb0e6aed0ff794

garretlaxton (approved):
That worked great!

Crabcyborg (approved):
This looks good.

Thanks @AbdiTolesa!

🚀

PR #1928, opened 2024-08-20
Implement AJAX Submission Notification for Formidable Modal - Lite Version

shervElmi:
### Main PR: https://github.com/Strategy11/formidable-modal/pull/40


classes/controllers/FrmSimpleBlocksController.php
```diff
@@ -192,7 +192,15 @@ public static function simple_form_render( $attributes ) {
...
 		unset( $params['formId'] );
 
-		$form = FrmFormsController::get_form_shortcode( $params );
+		/**
+		 * @since x.x
+		 *
+		 * @param string $form
+		 * @param array $params
+		 */
+		$form = apply_filters( 'frm_before_simple_form_render_filter', '', $params );
```
Crabcyborg, line 201: @shervElmi I don't think our filter needs the `_filter` in its name.

shervElmi: Addressed in the [Remove "_filter" in filter name](https://github.com/Strategy11/formidable-forms/pull/1928/commits/b6fe7df2a156f6311a5fefe0e005977a387111ad) commit.

shervElmi: Since there is an action named "frm_before_simple_form_render," removing "_filter" caused a PHP Fatal error. So, I reverted to using that word again.

@Crabcyborg, do you have any other suggestions, or is it fine to go with the name "frm_before_simple_form_render_filter"?

Commit: [Fix conflict between action and filter hooks in simple_form_render](https://github.com/Strategy11/formidable-forms/pull/1928/commits/5020164c1de29198368eb329caee0daade613533)

Crabcyborg: @shervElmi Can we just use `frm_before_simple_form_render` then? I don't think we necessarily need a new hook if we already have that one.

Crabcyborg: I think we could use `ob_*` functions to get whatever HTML is output when the `frm_before_simple_form_render` action is triggered, and we can prepend that HTML to the returned HTML, since I imagine that's out problem (returning the extra HTML).

shervElmi: Addressed in the [Capture form render output using ob_start()](https://github.com/Strategy11/formidable-forms/pull/1928/commits/db373ebbaea150b8e0cdab4782a11ca8c6285518) commit.

classes/controllers/FrmSimpleBlocksController.php
```diff
@@ -182,17 +182,21 @@ public static function simple_form_render( $attributes ) {
...
 
+		ob_start();
+
 		/**
 		 * @since 5.5.2
 		 * @param array $attributes
 		 */
 		do_action( 'frm_before_simple_form_render', $attributes );
 
+		$form = ob_get_clean();
```
Crabcyborg, line 193: Thanks @shervElmi!

This looks great.

There's just one more thing here. A type issue that would likely come up in our static analysis at some point in the future.

`ob_get_clean` technically has a return type of `string|false`.

So if `$form === false` here, it would be good to set it to `''` to enforce the string type.

shervElmi: Addressed in the [Add validation for `$form`](https://github.com/Strategy11/formidable-forms/pull/1928/commits/1e397260ca5f46d9aba4e544741efa9e1cf36a5d) commit.

Crabcyborg (approved):
Thanks @shervElmi!

🚀

PR #1929, opened 2024-08-20
Use FrmAppHelper::mb_function to use mb_strlen when it is available

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5295


Crabcyborg (approved):
Thank you @AbdiTolesa!

🚀

PR #1934, opened 2024-08-21
delete forms permanently

lauramekaj1:
Empty trash for the deleted forms - Restore and delete permanently options)


Crabcyborg (approved):
Thank you @lauramekaj1!

This looks good to me!

🚀

PR #1938, opened 2024-08-22
Fix: Improve compatibility with old versions of Lite when checking for Chosen

shervElmi:
Ensure the `FrmProAppHelper::use_chosen_js` is called only if it callable.

### Related Issue:
https://github.com/Strategy11/formidable-pro/issues/5309

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -4325,4 +4325,19 @@ class FrmAppHelper {
 		</span>
 		<?php
 	}
+
+	/**
+	 * Determines if the chosen JavaScript library should be used.
+	 *
+	 * @return bool
+	 */
+	public static function use_chosen_js() {
+		if ( ! self::pro_is_installed() ) {
+			return false;
+		}
+
+		return is_callable( 'FrmProAppHelper::use_chosen_js' )
+			? FrmProAppHelper::use_chosen_js()
+			: true;
+	}
 }
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -4325,19 +4325,4 @@ class FrmAppHelper {
 		</span>
 		<?php
 	}
-
-	/**
-	 * Determines if the chosen JavaScript library should be used.
-	 *
-	 * @return bool
-	 */
-	public static function use_chosen_js() {
-		if ( ! self::pro_is_installed() ) {
-			return false;
-		}
-
-		return is_callable( 'FrmProAppHelper::use_chosen_js' )
-			? FrmProAppHelper::use_chosen_js()
-			: true;
-	}
 }
--- a/css/custom_theme.css.php
+++ b/css/custom_theme.css.php
@@ -22,7 +22,7 @@ $default_style    = $frm_style->get_default_style( $styles );
 $defaults         = FrmStylesHelper::get_settings_for_output( $default_style );
 $important        = empty( $defaults['important_style'] ) ? '' : ' !important';
 $pro_is_installed = FrmAppHelper::pro_is_installed();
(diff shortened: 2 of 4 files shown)


css/custom_theme.css.php
```diff
@@ -22,7 +22,10 @@
 $defaults         = FrmStylesHelper::get_settings_for_output( $default_style );
 $important        = empty( $defaults['important_style'] ) ? '' : ' !important';
 $pro_is_installed = FrmAppHelper::pro_is_installed();
-$use_chosen_js    = $pro_is_installed && FrmProAppHelper::use_chosen_js();
+
+$use_chosen_js =
```
Crabcyborg, line 26: @shervElmi This isn't quite right.

When `$pro_is_installed` is false, we want `$use_chosen_js` to also be false.

Crabcyborg: Also @shervElmi. I'd like to fix https://github.com/Strategy11/formidable-pro/issues/5308 as well.

It will need a similar check. I think a new `FrmStylesHelper::use_chosen_js` helper function might be nice.

shervElmi: It's doing that actually, @Crabcyborg!

The first condition checks `$pro_is_installed` and then `&& is_callable(...`.

shervElmi: @Crabcyborg this PR will fix that issue automatically since the "_single_theme.css.php" is required in the "custom_theme.css.php" file and inherits this fixed version.

Crabcyborg: @shervElmi That's the problem. Look at https://github.com/Strategy11/formidable-pro/issues/5308

`custom_theme.css.php` is never included in that code. `_single_theme.css.php`is included directly.

Crabcyborg: Also, it's never false unless `FrmProAppHelper::use_chosen_js()` returns false.

Your code is the same as,

```
if ( $pro_is_installed && is_callable( 'FrmProAppHelper::use_chosen_js' ) ) {
    $use_chosen_js = FrmProAppHelper::use_chosen_js();
} else {
    $use_chosen_js = true;
}
```

In order for `$use_chosen_js` to be false, this would need to be structured totally differently. More like,

```
$use_chosen_js = $pro_is_installed ? ( is_callable( 'FrmProAppHelper::use_chosen_js' ) ? FrmProAppHelper::use_chosen_js() : true ) : false;
```

But that's kind of ugly, especially since we'll need it in at least 2 please. Which is why I think we need a helper function too.

That way we can break it up more and make it look nicer. I would start with the `pro_is_installed` check. If that's false, I'd `return false` early.

classes/helpers/FrmAppHelper.php
```diff
@@ -4325,4 +4325,19 @@ public static function tooltip_icon( $tooltip_text, $atts = array() ) {
 		</span>
 		<?php
 	}
+
+	/**
+	 * Determines if the chosen JavaScript library should be used.
+	 *
+	 * @return bool
+	 */
+	public static function use_chosen_js() {
```
Crabcyborg, line 4334: @shervElmi I asked to put this in `FrmStylesHelper` intentionally.

The only use of this function in Lite will be for styling purposes.

I'd like to avoid adding more public functions to `FrmAppHelper.` It's far too large as it is.

shervElmi: I'm feeling a bit sick and off today @Crabcyborg! Sorry!

Addressed in the [Move "use_chosen_js" method to "FrmStylesHelper"](https://github.com/Strategy11/formidable-forms/pull/1938/commits/5addb0c908abe8bcd3bcb9a99994e7d9e1ed061b) commit.

Crabcyborg (approved):
Thank you @shervElmi!

🚀

Issue #1781, opened 2024-06-04
Akismet Integration: the comment_author parameter should be restricted to the name field


kbrown9:
The `FrmEntryValidate::is_akismet_guest_info_value` [function](https://github.com/Strategy11/formidable-forms/blob/master/classes/models/FrmEntryValidate.php#L615-L620) attempts to find a value to use as the `comment_author` parameter for Akismet. First, it checks for name field ids. If none exist, it uses the value of the first field with a length less than 200.

Preferably, the end user would be able to select which field to use for the `comment_author` parameter. If that's not possible, the form should send a `comment_author` parameter only when the form contains a `Name` field. If the form does not contain a `Name` field, the `comment_author` value should be empty.

PR #1945, opened 2024-08-23
Restrict comment_author to name fields

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-forms/issues/1781


AbdiTolesa:
@truongwp I accidentally took over your task as it was in my list and I thought I didn't notice it was unassigned from me. :)
Anyway, I think the solutions is simple enough, I haven't tested it though. I think it could use your review.

truongwp (approved):
Thanks @AbdiTolesa so much! I haven't started working on this issue. This looks good.

classes/models/FrmEntryValidate.php
```diff
@@ -620,7 +620,6 @@ private static function is_akismet_guest_info_value( $key, $value, $field_id, $n
 					// If there is name field in the form, we should always use it as author name.
 					return in_array( $field_id, $name_field_ids, true );
 				}
-				return strlen( $value ) < 200;
```
Crabcyborg, line 623: @AbdiTolesa We do have a lot of non-Name field Name fields. What do you think about making a label check as well? If the field name includes "Name", we call it a Name field?

Crabcyborg: Trying to support our Contact form would be nice.

It doesn't currently use a name field. Instead it's two consecutive fields where one is "Name" and the other is "Last". Maybe we could check for that pattern?

AbdiTolesa: > @AbdiTolesa We do have a lot of non-Name field Name fields. What do you think about making a label check as well? If the field name includes "Name", we call it a Name field?

@Crabcyborg I think that could easily lead to unexpected behavior as "Name" is a very common verb to use in other objects like animal, place etc

AbdiTolesa: > Trying to support our Contact form would be nice.
> 
> It doesn't currently use a name field. Instead it's two consecutive fields where one is "Name" and the other is "Last". Maybe we could check for that pattern and build the name?

This makes more sense to me and I have tried to build from that pattern.

classes/models/FrmEntryValidate.php
```diff
@@ -616,12 +618,22 @@ private static function is_akismet_guest_info_value( $key, $value, $field_id, $n
...
-		}
+				$form_id = FrmAppHelper::get_post_param( 'form_id', 0, 'absint' );
+				$fields  = FrmDb::get_results( 'frm_fields', array( 'form_id' => $form_id ), 'id,name', array( 'order_by' => 'field_order ASC' ) );
+				foreach ( $fields as $index => $field ) {
+					if ( 'Name' !== $field->name ) {
+						continue;
+					}
+					if ( isset( $fields[ $index + 1 ] ) && 'Last' === $fields[ $index + 1 ]->name ) {
+						$value .= ' ' . $values[ $fields[ $index + 1 ]->id ];
```
Crabcyborg, line 632: It's probably good to check that `$values[ $fields[ $index + 1 ]->id ]` is really set, just in case.

AbdiTolesa:
I see an error with PHPStan, which I can get more text on when googling it.
image

Crabcyborg:
> I see an error with PHPStan, which I can get more text on when googling it.

@AbdiTolesa Sorry about that. I believe the rector rule just no longer exists. Removing the exception didn't introduce new issues with Rector.

This was already fixed in master. I merged master into this branch and it's passing now. The issue was unrelated to your branch.

classes/models/FrmEntryValidate.php
```diff
@@ -617,12 +619,25 @@ private static function is_akismet_guest_info_value( $key, $value, $field_id, $n
...
-				if ( $name_field_ids ) {
+				if ( $name_field_ids && in_array( $field_id, $name_field_ids, true ) ) {
 					// If there is name field in the form, we should always use it as author name.
-					return in_array( $field_id, $name_field_ids, true );
+					return true;
 				}
-				return strlen( $value ) < 200;
-		}
+				$form_id = FrmAppHelper::get_post_param( 'form_id', 0, 'absint' );
+				$fields  = FrmDb::get_results( 'frm_fields', array( 'form_id' => $form_id ), 'id,name', array( 'order_by' => 'field_order ASC' ) );
```
Crabcyborg, line 627: Can we query on this line just for text fields to avoid getting more data than we need?

classes/models/FrmEntryValidate.php
```diff
@@ -617,12 +619,34 @@ private static function is_akismet_guest_info_value( $key, $value, $field_id, $n
...
+					return true;
 				}
-				return strlen( $value ) < 200;
-		}
+				$form_id = FrmAppHelper::get_post_param( 'form_id', 0, 'absint' );
+				$fields  = FrmDb::get_results(
+					'frm_fields',
+					array(
+						'form_id' => $form_id,
+						'type'    => 'text',
```
Crabcyborg, line 631: Would it make sense to also filter the `name` here if we're only checking for `First` and `Last`?. It still seems like we're getting too much here.

Crabcyborg (approved):

I think this should be good to go for next release.

🚀

classes/models/FrmEntryValidate.php
```diff
@@ -617,12 +619,35 @@ private static function is_akismet_guest_info_value( $key, $value, $field_id, $n
...
-				if ( $name_field_ids ) {
+				if ( $name_field_ids && in_array( $field_id, $name_field_ids, true ) ) {
 					// If there is name field in the form, we should always use it as author name.
-					return in_array( $field_id, $name_field_ids, true );
+					return true;
 				}
-				return strlen( $value ) < 200;
-		}
+				$form_id = FrmAppHelper::get_post_param( 'form_id', 0, 'absint' );
+				$fields  = FrmDb::get_results(
```
Crabcyborg, line 627: @AbdiTolesa I think I might have spoke too soon.

I'm concerned this query might get called multiple times. Could we possibly store this value in memory/in a variable and re-use it if the value has already been set in the request?

AbdiTolesa: @Crabcyborg It looks this is called once per form submission as my debug goes but we are caching the db result when we use `FrmDb::get_results` anyway, right?

Crabcyborg: @AbdiTolesa We cache it, but there's no guarantee people have object caching on. We don't want to rely much on caching.

I don't totally know when this gets called multiple times but it's part of this `recursive_add_akismet_guest_info` function. Since it's recursive, it may get called many times. I haven't looked too much into it though.

AbdiTolesa: > I don't totally know when this gets called multiple times but it's part of this `recursive_add_akismet_guest_info` function. Since it's recursive, it may get called many times. I haven't looked too much into it though.

Thanks for explaining that, I couldn't see any reason that this might be called multiple times since we the `switch` case validates only once for `comment_author`. I'm concerned that could just be adding a complication but we can leverage transient if you could see a practical reason to use it.

Crabcyborg: No I wouldn't use a transient. We don't want anything extra in the database here.

I just mean storing in a variable. Like a `self::$name_text_fields` private static variable.

classes/models/FrmEntryValidate.php
```diff
@@ -617,16 +626,57 @@ private static function is_akismet_guest_info_value( $key, $value, $field_id, $n
...
+	 * @since x.x
+	 *
+	 * @param int $form_id
+	 * @return array
+	 */
+	private static function get_name_text_fields( $form_id ) {
+		if ( null !== self::$name_text_fields ) {
+			return self::$name_text_fields;
+		}
+		$fields = FrmDb::get_results(
```
Crabcyborg, line 665: This could be reduced further a bit. I wouldn't bother with having a separate `$fields` variable here. We can just set `self::$name_text_fields` here and then return `self::$name_text_fields`.

classes/models/FrmEntryValidate.php
```diff
@@ -617,16 +626,57 @@ private static function is_akismet_guest_info_value( $key, $value, $field_id, $n
...
+	 * Returns fields that have 'Name' and 'Last' as their name.
+	 *
+	 * @since x.x
+	 *
+	 * @param int $form_id
+	 * @return array
+	 */
+	private static function get_name_text_fields( $form_id ) {
+		if ( null !== self::$name_text_fields ) {
+			return self::$name_text_fields;
```
Crabcyborg, line 663: This also wouldn't work if there were multiple forms getting submitted.

I'd probably index `$name_text_fields` by `$form_id` and add to the array as keys are missing.

Crabcyborg (approved):

🙌

PR #1951, opened 2024-08-27
Add frm_get_ajax_form_errors event

truongwp:
This is required to add JS validation for minimum selections.


Crabcyborg (approved):
Thanks @truongwp! This looks good to me.

🚀

PR #1954, opened 2024-08-28
Add classic style template to styles list

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/5302

I still need to handle the logic where it actually properly installs the old style template.


Crabcyborg:
Sounds like we might not go in this direction.

I'm going to just close this for now.

PR #1955, opened 2024-08-28
Only add Cloudflare Turnstile captcha for visible elements

engahmeds3ed:
## The issue

In a customer's site we found a case where on the same page we may have one form twice, one for desktop and the other for mobile and this can be customized in the page builder (SiteOrigin page builder in our case)
Here we will have only one visible form but if the form contains Cloudflare Turnstile captcha, we will have two captcha fields visible.

## The solution

Here I'm checking if the captcha field is visible before initiating the captcha field.

## Screencast

https://github.com/user-attachments/assets/60f80ccd-9354-4135-8900-532e722e7f59

## Ticket(s)

https://secure.helpscout.net/conversation/2685885705/207895/

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1779,16 +1779,13 @@ function frmTurnstile() {
 	frmCaptcha( '.cf-turnstile' );
 }
 
-function frmElementIsVisible(element) {
-	return element.offsetParent !== null;
-}
-
 function frmCaptcha( captchaSelector ) {
 	let c;
 	const captchas = document.querySelectorAll( captchaSelector );
 	const cl       = captchas.length;
 	for ( c = 0; c < cl; c++ ) {
-		if ( ! frmElementIsVisible( captchas[c] ) ) {
+		const isVisible = captchas[c].offsetParent !== null;
+		if ( ! isVisible ) {
 			continue;
 		}
 		frmFrontForm.renderCaptcha( captchas[c], captchaSelector );
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1779,11 +1779,18 @@ function frmTurnstile() {
 	frmCaptcha( '.cf-turnstile' );
 }
 
+function frmElementIsVisible(element) {
+	return element.offsetParent !== null;
+}
+
 function frmCaptcha( captchaSelector ) {
 	let c;
 	const captchas = document.querySelectorAll( captchaSelector );
 	const cl       = captchas.length;
 	for ( c = 0; c < cl; c++ ) {
+		if ( ! frmElementIsVisible( captchas[c] ) ) {
+			continue;
+		}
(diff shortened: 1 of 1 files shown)


js/formidable.js
```diff
@@ -1779,11 +1779,18 @@ function frmTurnstile() {
 	frmCaptcha( '.cf-turnstile' );
 }
 
+function frmElementIsVisible(element) {
```
Crabcyborg, line 1782: @engahmeds3ed Can we avoid another global function?

I think since this is just being used once we're fine just making it a one liner like `const isVisible = element.offsetParent !== null;`.

engahmeds3ed: Sure thing

engahmeds3ed: Done

engahmeds3ed:
@Crabcyborg 

Here:
https://github.com/Strategy11/formidable-forms/blob/c946499f5d87efc833d83c5db44e71c8b535ca20/js/formidable.js#L1487

In our case here, this container ID will return only the first element even if it's not visible, what do u think?

engahmeds3ed:
Yes I can see an issue with the customer's site in the case when using a page builder to add the form twice, one for the desktop version and the other one for mobile so both forms are on the dom but only one will be visible at a time so we will have every field twice with the same ID so for the line that I mentioned above, it'll be something like `#field_xxx` replacing `xxx` with the field key.
This leads to getting the first field on the page with that ID but this may not be the correct one here.

For this specific customer, I used the following snippet to remove the not visible captcha field from the page before the captcha script runs:

https://secure.helpscout.net/conversation/2696062256/208907/

```
add_action( 'wp_footer', function() {
	?>
	<script>
		function frm_helper_multiple_captchas() {
			let allElements = document.querySelectorAll('#field_65mzm');
			if ( allElements.length <= 0 ) {
				return;
			}
			allElements.forEach(
				(element) => {
					const isVisible = element.offsetParent !== null;
					if ( ! isVisible ) {
						element.remove();
					}
				}
			);
		}
		frm_helper_multiple_captchas();
	</script>
	<?php
}, 1 );
```

Crabcyborg:
Thanks @engahmeds3ed.

I think you'll likely come across other issues as well.

We don't really support the same form twice on a page. There are a lot of issues that come up (like you said - there are unique ID conflicts).

I don't think we need to worry too much if this isn't working, as there are likely other issues as well.

They might be better off doing something like calling `wp_is_mobile()` and only including the HTML for mobile/desktop. Including the same form twice when only one is ever used doesn't sound very efficient either, though that might not work very well with caching unless they can have a cache for desktop and a cache for mobile.

engahmeds3ed:
Totally agree with u, so we deal with those cases as edge cases, that's good to know and the snippet is working as a workaround in cases this is needed in the future.
Thanks @Crabcyborg

PR #1960, opened 2024-09-03
Move Styles & Scripts settings to Manage Styles tab

truongwp:
Fixes: https://github.com/Strategy11/formidable-pro/issues/5187

This PR will move 2 settings mentioned in the issue to the Manage Styles tab. We will move the `Fade in forms with conditional logic on page load` setting to Other section in General Settings tab in another PR in Pro.


Crabcyborg (approved):
This looks good!

Since the other update is in v6.14 we're less likely to get the other setting showing up on the styles page now.

PR #1802, opened 2024-06-10
New Add-Ons Page

shervElmi:

PR #1963, opened 2024-09-03
add-ons page validations

lauramekaj1:


Crabcyborg:
@lauramekaj1 Does this update validate the old add-ons page we're replacing soon?

Or does it work with the new add-ons update as well? The page is changing quite a bit so I'm concerned these tests could immediately break.

The add-ons update is in this PR https://github.com/Strategy11/formidable-forms/pull/1802

Have you run these tests against that branch?

lauramekaj1:
@Crabcyborg Yes, it validates the old page only. I tried to run the tests against the new add-ons PR https://github.com/Strategy11/formidable-forms/pull/1802 but the tests fail.  

Do you think I should update my tests based on the PR https://github.com/Strategy11/formidable-forms/pull/1802 and as soon as the new add-ons will be released then we can add our tests as well?

Crabcyborg:
@lauramekaj1 Yeah, I think we could branch off of Sherv's branch and merge it even into his branch ahead of time possibly.

We're likely going to merge his update some time this week.

lauramekaj1:
Hi @Crabcyborg, I updated it based on the new add-ons. If everything looks ok then we can merge against the new-add-ons branch.

lauramekaj1:
@Crabcyborg Also, I think the search function is based on the pattern e.g. data-slug="signature". For example, when I try to search for "Digital signatures," no results are shown. However, when I search for just "signature," the search results are displayed.

Crabcyborg:
Thanks @lauramekaj1. I tried to come up with a solution to the plural search issue. I have a new pull request here https://github.com/Strategy11/formidable-forms/pull/2007

Could you help test it? I assigned you for review.

Crabcyborg:
> Also, do you think it's better if I create a branch out of the new-add-ons branch and then will push the changes there?

@lauramekaj1 That's how I would probably do it. If we just merge the add-ons branch into this one it should be fine. We wouldn't merge this before the other update.

lauramekaj1:
> > Also, do you think it's better if I create a branch out of the new-add-ons branch and then will push the changes there?
> 
> @lauramekaj1 That's how I would probably do it. If we just merge the add-ons branch into this one and change the base merge branch here in GitHub to the add-ons branch it should be fine. We wouldn't merge this before the other update.


@Crabcyborg I merged new-add-ons branch into this one and also changed the base merge. There are some tests which are failing here. I think it's because that the Cypress check is being run on master!

Crabcyborg (approved):
@lauramekaj1 I think the only issue was a little mistake with the search checks. It checks for the signatures add-on, but still had the validation for the PayPal check before that.

I made a small commit and it's passing now https://github.com/Strategy11/formidable-forms/pull/1963/commits/ce26573b2c1f88f112d9232a6ecdc26fdc9f9319

PR #1964, opened 2024-09-04
Change conditional logic row's value select element type conditionally

AbdiTolesa:
Related issue: https://github.com/Strategy11/formidable-pro/issues/5307
Related PR: https://github.com/Strategy11/formidable-pro/pull/5338

### Test procedure
1. Create a form and add an option field like dropdown field and another field, like a text field.
2. Start adding a conditional logic to the text field and select the dropdown field as value source and select comparison among `contains`, `does not contain`, `starts with` or `ends with`
3. Confirm that the control in which you define the match value is changed to text input.
4. Enter a value in the text field and save your form.
5. Confirm that the value select control is still a text input even after refreshing your page.
6. Preview your form and then test the conditional logic works as expected.
7. Get back to the form builder and change the comparison type to something else except those mentioned in step 2 above.
8. Save and preview your form and confirm the conditional logic works as expected.


js/formidable_admin.js
```diff
@@ -6143,7 +6144,7 @@ function frmAdminBuildJS() {
 
 			if ( showSelect || showText ) {
 				fill.innerHTML = '';
-				if ( showSelect ) {
+				if ( showSelect && ! [ 'LIKE', 'not LIKE', 'LIKE%', '%LIKE' ].includes( comparison ) ) {
```
truongwp, line 6147: Should we set `showSelect` value based on the comparison before the `if` on line https://github.com/Strategy11/formidable-forms/pull/1964/files#diff-23fb66f34c625a76ac7564df177d3ab58c86b1f5d071f745aa50e028fb0209c4R6137?

AbdiTolesa: @truongwp Thanks! I think we can avoid attempt to fill dropdown with values when value select is of text input type and I have added another condition and used it in both places without altering `showSelect`.

js/formidable_admin.js
```diff
@@ -6132,6 +6132,7 @@ function frmAdminBuildJS() {
 			let input = false;
 			let showSelect = ( valueFieldType === 'select' || valueFieldType === 'checkbox' || valueFieldType === 'radio' );
 			const showText = ( valueFieldType === 'text' || valueFieldType === 'email' || valueFieldType === 'phone' || valueFieldType === 'url' || valueFieldType === 'number' );
+			const comparison = document.querySelector( `#frm_logic_${fieldID}_${metaKey} [name="field_options[hide_field_cond_${fieldID}][]"]` ).value;
```
truongwp, line 6135: Please move this line into the `if` it's used.

truongwp (approved):

Crabcyborg (approved):
Thank you @AbdiTolesa and @truongwp!

🚀

PR #1968, opened 2024-09-06
Do not pass non-customizable fields to Akismet guest info

### LINKED ISSUE #1782: Akismet Integration: non-customizable fields should not be used for guest info
The `FrmEntryValidate::is_akismet_guest_info_value` function sets the `comment_author`, `comment_author_email`, and `comment_author_url` parameters by checking for certain features in the field values. This makes it possible for values from non-customizable fields, such as dropdowns and radio boxes, to be sent to Akismet. Non-customizable field values do not affect the spaminess of a contact form submission and should not be sent to Akismet, as described in #560.

Ideally, the function would skip values from non-customizable fields when attempting to find values for the `author` parameters.

truongwp:
Fixes #1782 

We did the same thing for comment content. I don't know how to tell Garret to test. I use `error_log()` to test this.

### DIFF
--- a/classes/models/FrmEntryValidate.php
+++ b/classes/models/FrmEntryValidate.php
@@ -477,6 +477,7 @@ class FrmEntryValidate {
 		self::add_server_values_to_akismet( $datas );
 
 		self::prepare_values_for_spam_check( $values );
+		self::skip_adding_values_to_akismet( $values );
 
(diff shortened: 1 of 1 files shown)


Crabcyborg (approved):
I haven't tested this but this update makes sense and looks quite straight forward.

I think this should be good to merge.

Thank you @truongwp!

PR #1979, opened 2024-09-12
Fix HTML display issue for checkbox with text and links

shervElmi:
This PR resolves an issue where checkboxes with longer text, including `<br>` and `<a>` tags, were not rendering correctly, causing misalignment.

### Related Issue
https://github.com/Strategy11/formidable-pro/issues/5363

### Steps to Reproduce
1. Create a checkbox field with long text and include `<br>`, `<a>` tags in the text.
2. Verify that the issue is fixed and ensure your output matches the "After" section screenshot.

### Before
CleanShot 2024-09-12 at 17 17 31@2x

### After
CleanShot 2024-09-12 at 17 17 00@2x

### DIFF
--- a/css/custom_theme.css.php
+++ b/css/custom_theme.css.php
@@ -816,11 +816,15 @@ a.frm_save_draft{
 }
 
 .with_frm_style .frm_checkbox label,
-.with_frm_style .frm_radio label{
-	display: flex;
-	align-items: center;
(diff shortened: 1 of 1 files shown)


Crabcyborg (approved):
Thanks @shervElmi!

I think this looks good.

🚀

PR #1982, opened 2024-09-13
fix error fields style when underline is used as field shape type

Liviu-p:


classes/views/styles/_quick-settings.php
```diff
@@ -132,7 +132,7 @@
 <div class="frm5 frm_form_field"><label class="frm-style-item-heading"><?php esc_html_e( 'Base Font Size', 'formidable' ); ?></label></div>
 <div class="frm7 frm_form_field">
 	<?php if ( ! FrmStylesHelper::is_advanced_settings() ) { ?>
-		<input type="hidden" name="<?php echo esc_attr( $frm_style->get_field_name( 'use_base_font_size' ) ); ?>" value="true" />
+		<input type="hidden" name="<?php echo esc_attr( $frm_style->get_field_name( 'use_base_font_size' ) ); ?>" value="false" />
```
Crabcyborg, line 135: @Liviu-p Don't we want to support the use base font size setting in quick settings?

The base font size setting is part of quick settings, so this should be true, shouldn't it?

Crabcyborg: What about on load? If the base font size setting was changed before, and it was on, then you save again, it would reset to false, right?

Liviu-p: that's right, it will reset to false. That's the desired behaviour. Every time when the "Base Font Size" is changed and saved, it will change all sizes settings from "Advanced Settings". When it's false we keep the values that are already saved in "Advanced Settings".

Crabcyborg (approved):
I haven't tested this yet but it looks good to me.

I assigned @shervElmi to review as well to get some additional eyes on this.

Thanks @Liviu-p!

🚀

shervElmi:
@Liviu-p, I'm working on QA and reviewing the changes as well. Could you provide some details about this PR, please? Thanks.

Liviu-p:
> @Liviu-p, I'm working on QA and reviewing the changes as well. Could you provide some details about this PR, please? Thanks.

Hey @shervElmi, it provides fixes for few reported issues:
https://github.com/Strategy11/formidable-pro/issues/5369
https://github.com/Strategy11/formidable-pro/issues/5368
https://github.com/Strategy11/formidable-pro/issues/5360
https://github.com/Strategy11/formidable-pro/issues/5375

PR #1985, opened 2024-09-16
form templates functionalities

lauramekaj1:


Crabcyborg (approved):
Thanks @lauramekaj1!

@shervElmi Since you worked on the form templates page, do you think there's anything else here that should have e2e test coverage?

tests/cypress/e2e/Form Templates/FormTemplates.cy.js
```diff
@@ -0,0 +1,337 @@
...
+        cy.get('#template-search-input').should("exist");
+
+        cy.log("Validate data categories");
+
+        const categories = [
+            { category: 'all-templates', text: 'All Templates' },
+            { category: 'favorites', text: 'Favorites' },
+            { category: 'custom', text: 'Custom' },
+            { category: 'available-templates', text: 'Available Templates' },
+            { category: 'ai', text: 'AI' },
```
shervElmi, line 20: Hi @lauramekaj1,

The following items are static, and others come from an external API (Formidable Website), so it seems we shouldn't list them below:

`'all-templates', 'favorites', 'custom', 'available-templates'`

shervElmi: @Crabcyborg I need your input on my point as well :)

tests/cypress/e2e/Form Templates/FormTemplates.cy.js
```diff
@@ -0,0 +1,337 @@
...
+        cy.log("Release Template");
+        cy.get('[frm-search-text="release"] .frm-form-templates-item-title-text')
+            .should('contain', 'Release');
+        cy.get('[frm-search-text="release"] .frm-form-templates-item-description')
+            .should('contain.text', 'A simple standard release form template.');
+
+    });
+
+    it("search for valid and invalid templates", () => {
+
```
shervElmi, line 184: Could you also cover applications? Pro installation is required to show them as search results. For example:

CleanShot 2024-09-17 at 21 07 12@2x

tests/cypress/e2e/Form Templates/FormTemplates.cy.js
```diff
@@ -0,0 +1,337 @@
...
+        ];
+        
+        categories.forEach(({ category, text }) => {
+            cy.get(`li[data-category="${category}"]`).within(() => {
+                cy.get('.frm-form-templates-cat-text').should("have.text", text);
+            });
+        });
+        
+
+        cy.log("Check the items on the All Templates page");
```
shervElmi, line 60: Could you cover whether the featured templates list has an icon?

CleanShot 2024-09-17 at 21 17 42@2x

tests/cypress/e2e/Form Templates/FormTemplates.cy.js
shervElmi: And also, there is another category when Pro isn't activated called `free-templates`. It is also static and not dynamic, so it should be covered as well.

shervElmi:
@lauramekaj1, it seems the PR doesn't cover the "Get 20+ Free Form Templates" modal. Could you add that as well?

**Note:** If the Pro version is activated, you should deactivate it.

CleanShot 2024-09-17 at 21 28 42@2x

shervElmi:
@lauramekaj1, I'm not sure if covering the "Name your form" modal is within the scope of this PR, but after creating a new form and clicking save, the modal will appear.

CleanShot 2024-09-17 at 21 37 09@2x

shervElmi:
@lauramekaj1, same for the renew modal, could you also cover that?

CleanShot 2024-09-17 at 21 53 50@2x

**Note:** To have an expired license, you can install the [ff-qa-tools](https://github.com/Strategy11/ff-qa-tools) plugin and then expire the license like below.

CleanShot 2024-09-17 at 21 56 02@2x

tests/cypress/e2e/Form Templates/FormTemplates.cy.js
lauramekaj1: Hi @shervElmi! Thank you for the feedback. I will start by making some updates according to the comments.

However, about the "Free Templates" category it's not showing on my site anymore. It appeared at first, and after clicking 'Use Template' and got the code when the "Get 20+ Free Form Templates" modal was displayed, then the category stopped showing. I was just wondering if it still should show as a category even after getting the 20+ free templates?
image

shervElmi: Hi @lauramekaj1, you should remove the saved key from the data. I don't recall exactly, but it might be `free_templates`. You could either install a fresh WordPress that doesn’t save this option or reset your current WordPress installation database.

lauramekaj1:
Hi @Crabcyborg, I went through all comments and I updated form templates tests. I just wanted to ask you about the one with the [ff-qa-tools](https://github.com/Strategy11/ff-qa-tools) plugin. Will it work if it runs through the Cypress check, given that the tool should also be installed in that environment? Do we need to perform any setup to support tests running against that tool on Lite?


> @lauramekaj1, same for the renew modal, could you also cover that?
> 
> CleanShot 2024-09-17 at 21 53 50@2x
> 
> **Note:** To have an expired license, you can install the [ff-qa-tools](https://github.com/Strategy11/ff-qa-tools) plugin and then expire the license like below.
> 
> CleanShot 2024-09-17 at 21 56 02@2x

Crabcyborg:
Thanks @lauramekaj1. That wouldn't happen in Lite. We can consider checking for expired licenses once we have tests running for Pro.

We can skip that request for now.

lauramekaj1:
Hi @shervElmi, I reviewed all the comments and updated everything accordingly. Please let me know if I missed anything that needs to be covered. Thank you!

tests/cypress/e2e/Form Templates/FormTemplates.cy.js
shervElmi: Oops!

I mean we should keep only the 'all-templates', 'favorites', 'custom', 'available-templates', and 'free-templates' categories. The others come from an external API and aren't static.

Sorry for the confusion!

tests/cypress/e2e/Form Templates/FormTemplates.cy.js
```diff
@@ -10,13 +10,33 @@ describe("Form Templates page", () => {
...
+        });
+
+        cy.get('li[data-category="available-templates"]').within(() => {
+            cy.get('.frm-form-templates-cat-text').should("have.text", "Available Templates");
+            cy.get('.frm-form-templates-cat-count').should("have.text", "0");
+        });
+
+        cy.get('li[data-category="all-templates"]').within(() => {
+            cy.get('.frm-form-templates-cat-text').should("have.text", "All Templates");
+            cy.get('.frm-form-templates-cat-count').should("have.text", "319");
```
shervElmi, line 31: The count number for "All Templates" and "Free Templates" could change from the external API. We should remove this line for both "All Templates" and "Free Templates" or come up with another approach.

shervElmi (approved):
Hi @lauramekaj1,  
Thanks for the update, it looks great to me now! 🚀 

I just have two new comments, but they're not major.

lauramekaj1:
Hi @shervElmi, since all those categories are dynamic on the count then I think I should better remove the count. Thank you!

Crabcyborg:
Thanks for your help reviewing this @shervElmi!

And thank you for all your work improving these tests @lauramekaj1!

🙌

PR #1986, opened 2024-09-16
Address Styling Issues on Entry (Edit/New) and Style Pages

shervElmi:
This PR resolves styling issues on both the Entry (Edit/New) and Style pages, ensuring a consistent and polished appearance across these sections.

### Related Pro PR:
https://github.com/Strategy11/formidable-pro/pull/5501

### Related Issue:
https://github.com/Strategy11/formidable-pro/issues/5359

### Reproduce Steps:
1. Change current styles to something similar to the image below:

CleanShot 2024-09-16 at 16 08 38@2x

2. Ensure that the Entry (Edit/New) and Style pages match the frontend style and no longer display the **Fixes** section issues.

### Fixes:
1. Style of Disabled/Readonly input, select, and textarea fields (specifically border color on focus).
2. Checkbox and Radio Button styles.
3. Entry Edit/New navigation button styles.
4. Focus state of fields.

### Before:
CleanShot 2024-09-16 at 15 56 02

### After:
CleanShot 2024-09-16 at 16 01 37


css/frm_admin.css
```diff
@@ -1406,7 +1406,7 @@ body.frm-hidden-overflow {
 .post-type-frm_display.wp-core-ui .button-primary,
 .frm-button-primary,
 .wp-core-ui .button-primary.frm-button-primary,
-.frm_single_entry_page .button-primary {
+.frm_single_entry_page .button-primary:not(.frm-fields-item) {
```
Crabcyborg, line 1409: @shervElmi I don't love this class name. It's pretty generic.

Also, it looks like this is only referred to in Pro. Is it possible to just fix it from the `formidable-entries.css` instead so we can isolate the logic to Pro?

shervElmi: Since the button styles come from the Style page and are dynamic.

shervElmi: How about using a class name like `frm-entry-page-footer-button`?

Crabcyborg: It looks like it's just a `background-color`, `color` style here. I don't see why overwriting wouldn't work?

css/frm_admin.css
```diff
@@ -8258,19 +8242,30 @@ button.frm_choose_image_box,
...
-.frm_wrap textarea:focus {
+body:not(.frm-admin-page-styles):not(.frm-admin-page-entries) .frm_wrap textarea:focus {
 	border-color: var(--primary-500) !important;
 	box-shadow: none !important;
 }
 
+.frm-admin-page-styles textarea:focus,
+.frm-admin-page-styles input:focus,
+.frm-admin-page-styles select:focus,
+.frm-admin-page-entries textarea:focus,
```
Crabcyborg, line 8262: Can we target front end form styles here instead of the pages themselves?

Any inputs on the style/entries pages should be the same as other admin pages, except for the inputs inside of form previews themselves.

css/frm_admin.css
shervElmi: I don't have access to the dynamic color in the "formidable-entries.css" file. For example, if you look at the after section in the description, the button color is red, and I just need to exclude the always-applied blue CSS from those buttons.

css/frm_admin.css
shervElmi: I'm concerned that forcing styles on the frontend might create new issues. Does that sound like a problem to you? We could likely resolve it by importing the styles.

Crabcyborg: I don't mean changing anything on the front end.

I just mean targeting these so we don't need 2 rules for each input (one for styles and one for entries). This could grow to 3, 4, 5, over time, as we preview forms in other places. Already, I think this is missing the visual views editor preview.

I mean something like,

```
body.wp-admin .frm_form_fields .frm_fields_container input:focus,
body.wp-admin .frm_form_fields .frm_fields_container select:focus,
body.wp-admin .frm_form_fields .frm_fields_container textarea:focus,
```

I'm basically trying to remove page specific logic and duplication.

css/frm_admin.css
shervElmi: Addressed with the following commits:

- [Update "frm-fields-item" class to "frm-button-entry-nav"](https://github.com/Strategy11/formidable-forms/pull/1986/commits/af45f803489272b083d6818dd25a1768840a852c)
- Pro PR: [Update class name](https://github.com/Strategy11/formidable-pro/pull/5376/commits/990d5949919d7218e6bc744580e6579ac5bf204b)

css/frm_admin.css
shervElmi: Addressed in the [Revert the addition of the ".frm_forms" class to ".with_frm_style".](https://github.com/Strategy11/formidable-forms/pull/1986/commits/7e5bd44cf42ce34d3cdc4c8881cef8629cc1e78a) commit.

css/custom_theme.css.php
```diff
@@ -816,15 +816,11 @@
...
-	display: inline-block;
-	vertical-align: middle;
-	white-space:normal;
-}
-
-.with_frm_style .frm_checkbox label input[type=checkbox],
-.with_frm_style .frm_radio label input[type=radio] {
-	margin-right: 4px;
+.with_frm_style .frm_radio label{
+	display: flex;
```
Crabcyborg, line 820: Would this re-introduce the issues with using HTML inside of a radio button label? Have you tested this with anchor tags?

shervElmi: I haven't worked on them, and I don't know where they came from!

However, I just reverted them.

css/frm_admin.css
```diff
@@ -8252,19 +8236,27 @@ button.frm_choose_image_box,
...
+.frm-white-body:not(.frm-admin-page-styles):not(.frm-admin-page-entries) textarea:focus,
+.frm-white-body:not(.frm-admin-page-styles):not(.frm-admin-page-entries) input:focus,
+.frm-white-body:not(.frm-admin-page-styles):not(.frm-admin-page-entries) select:focus,
+body:not(.frm-admin-page-styles):not(.frm-admin-page-entries) .frm_wrap input:focus,
+body:not(.frm-admin-page-styles):not(.frm-admin-page-entries) .frm_wrap select:focus,
+.wp-core-ui:not(.frm-admin-page-styles):not(.frm-admin-page-entries) .frm_wrap select:focus,
 #frm-form-templates-modal select:focus,
 .frm-btn-group .multiselect.dropdown-toggle:focus,
-.frm_wrap textarea:focus {
+body:not(.frm-admin-page-styles):not(.frm-admin-page-entries) .frm_wrap textarea:focus {
```
Crabcyborg, line 8248: @shervElmi Is it possible here to just not use `.frm_wrap` on these pages instead of this rule? This rule seems wrong, as this is what a `frm_wrap` element should be like.

css/custom_theme.css.php
```diff
@@ -900,11 +900,13 @@
 <?php if ( $pro_is_installed ) : ?>
 .frm_forms.with_frm_style .frm_fields_container .frm_scale input[type=radio]:not([disabled]):checked,
 <?php endif; ?>
-.frm_forms.with_frm_style .frm_fields_container .frm_checkbox input[type=checkbox]:not([disabled]):checked {
+.frm_forms.with_frm_style .frm_fields_container .frm_checkbox input[type=checkbox]:not([disabled]):checked,
+.frm_forms .with_frm_style .frm_fields_container .frm_checkbox input[type=checkbox]:not([disabled]):checked {
```
Crabcyborg, line 904: @shervElmi If `.frm_forms .with_frm_style` is only relevant on the entry edit page, can we possibly put this CSS somewhere else? I don't want it adding to our front end CSS file size.

css/frm_admin.css
shervElmi: Could you give me an example or the actual selector you mentioned? It would help me a lot. Thanks!

css/custom_theme.css.php
shervElmi: Addressed with the [Optimze frontend CSS](https://github.com/Strategy11/formidable-forms/pull/1986/commits/fa5192ddd5c09adf6e4563f82d40e56a8e736121) commit.

css/frm_admin.css
```diff
@@ -1473,7 +1473,7 @@ body.frm-hidden-overflow {
 .frm-white-body .button:not(.frm-button-primary):active,
 .frm-white-body .frm-button-secondary:hover,
 .wp-core-ui.frm-white-body .frm-button-secondary:hover,
-.wp-core-ui.frm-white-body .button-secondary:hover,
+.wp-core-ui.frm-white-body .button-secondary:not(.frm-button-entry-nav):hover,
```
Crabcyborg, line 1476: It looks like this old reference to `frm-button-entry-nav` is still here.

css/frm_admin.css
Crabcyborg: If I remove `frm_wrap`, it looks like the form dropdown changes slightly. I think we can try to put `frm_wrap` on the `#frm_top_bar` element only, and not the rest of the page.

classes/views/frm-entries/show.php
```diff
@@ -3,18 +3,20 @@
 	die( 'You are not allowed to call this page directly.' );
 }
 ?>
-<div id="form_show_entry_page" class="frm_wrap frm_single_entry_page">
```
Crabcyborg, line 6: @shervElmi We might still want this for the "View entry" page. Does anything in the table change with this removed?

We're really just trying to avoid issues that would come up on the "Edit" page.

Crabcyborg: I just tried testing it, and it does look pretty messed up now.

Screen Shot 2024-11-26 at 11 57 15 AM

css/frm_admin.css
```diff
@@ -8254,6 +8239,13 @@ button.frm_choose_image_box,
 	box-shadow: none !important;
 }
 
+.wp-admin .frm_fields_container textarea:focus,
```
Crabcyborg, line 8242: What exactly is the purpose of this additional CSS?

Is this something we can remove, or add a comment to to provide more context?

classes/views/frm-entries/show.php
shervElmi: Thanks for catching that! I’ve reverted the show template file changes with this commit: [Revert show template](https://github.com/Strategy11/formidable-forms/pull/1986/commits/3a26bcace5801f97a6dea672441d9746605efbe3)

css/frm_admin.css
shervElmi: Addressed with the [Add a comment for clarification](https://github.com/Strategy11/formidable-forms/pull/1986/commits/9ea50f53ad578b3f777842a8b02974e42a94845e) commit.

shervElmi:
@Crabcyborg It seems the E2E test error isn’t related to this PR, correct? I’ve reviewed the code and couldn’t find anything related.

Crabcyborg:
@shervElmi Yes, I believe those are unrelated. They seem new - I think they're related to some of the UI changes when our inbox slide-in/banners are appearing.

lauramekaj1 (approved):
@shervElmi @Crabcyborg I tested this and verified that the issue is fixed. Thank you!

Crabcyborg:
Thank you Laura!

PR #1993, opened 2024-09-16
Hide all admin notices that wasn't added by our code (core and addons) from our pages

engahmeds3ed:
Closes https://github.com/Strategy11/formidable-pro/issues/5178

Here we are removing all admin notices that was added by third parties in our admin pages only.

## How to reproduce?

Use the following snippet:
```
add_action( 'admin_notices', 'test_ahmed' );
function test_ahmed() {
	?>
	<div class="tesssssssssssssst" style="background-color: red;color: white;">
		<h1>This is a test notice.</h1>
	</div>
	<?php
}
```

And visit any form settings page or any formidable page, you will see this notice and this causes some UI problems.

## How to validate the fix?

With the above code activated, you shouldn't see the admin notice in our pages at all.

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -1323,10 +1323,22 @@ class FrmAppController {
 	}
 
 	/**
+	 * Check if we are in our admin pages.
+	 *
+	 * @return bool
+	 */
+	private static function in_our_pages() {
+		global $current_screen;
+		return FrmAppHelper::is_formidable_admin() || ( ! empty( $current_screen->post_type ) && 'frm_logs' === $current_screen->post_type );
+	}
+
+	/**
+	 * Hide all third-parties admin notices only in our admin pages.
+	 *
 	 * @return void
 	 */
 	public static function filter_admin_notices() {
-		if ( ! FrmAppHelper::is_formidable_admin() ) {
+		if ( ! self::in_our_pages() ) {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


classes/controllers/FrmAppController.php
```diff
@@ -1321,4 +1321,61 @@ private static function enqueue_floating_links( $plugin_url, $version ) {
 	public static function page_route( $content ) {
 		return FrmDeprecated::page_route( $content );
 	}
+
+	/**
+	 * @return void
```
Crabcyborg, line 1326: Some `@since x.x` comments would be nice to add to the new functions so it's easier to get a history of when something was added.

Crabcyborg (approved):
I haven't tested this yet myself but the code looks good to me.

I assigned Garret to review it so we can get some help testing.

Thanks Ahmed!

🚀

classes/controllers/FrmAppController.php
engahmeds3ed: Good catch, Done

garretlaxton:
I am still seeing notices on the Logs page.
image

engahmeds3ed:
Checking that now.

engahmeds3ed:
@Crabcyborg 
Do we need to handle the case @garretlaxton mentioned, this happens only when logs addon is activated but it doesn't break the UI, I thought that we need to handle this in our builder pages only because UI issues.

What do u think? It's one line change but I just wanted to make sure that it's needed.

Crabcyborg:
@engahmeds3ed I think it wouldn't hurt to also exclude notices on the logs page.

engahmeds3ed:
@garretlaxton can u plz check now, it should work.

garretlaxton (approved):
This looks great now!

PR #1996, opened 2024-09-18
Set conditional log row's value to the current value after syncing options with bulk edit

AbdiTolesa:
This PR sets the conditional logic row's selected value after options are synced using the dataset value set in Pro (https://github.com/Strategy11/formidable-pro/pull/5313).


Crabcyborg:
This relies on a pretty messy solution.

I'm just going to close this for now. I think we could come up with something better.

PR #1999, opened 2024-09-19
Fix pro issue 5064 / Fix format unslashing issue in field cache

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/5064

Following up from a previous PR https://github.com/Strategy11/formidable-forms/pull/1746. That solution wasn't ideal as it didn't fix the problem and just changed how fields were queried in the form builder to avoid the problem.

The issue here is that the cached object has only a single backslash and then gets passed through `wp_unslash` in `FrmField::getAll`.

In the database, the backslash is already saved as two backslashes (`\\`). When `field_options` are unserialized, the two backslashes become one `\`.

This update makes sure that if the field options have already been unserialized, that we add slashes to the object before it is cached, so the cached result includes the expected double backslashes (`\\`).

The data should never have issues here with extra backslashes as `wp_unslash` is called after `format_field_results` is, so these added slashes get removed again.

**Pre-release**
[formidable-6.14.2b.zip](https://github.com/user-attachments/files/17061358/formidable-6.14.2b.zip)

### DIFF
--- a/classes/models/FrmField.php
+++ b/classes/models/FrmField.php
@@ -1066,7 +1066,7 @@ class FrmField {
 	 * @return void
 	 */
 	private static function add_slashes_to_format_before_setting_field_cache( $result ) {
-		if ( ! is_array( $result->field_options ) || ! isset( $result->field_options ) || empty( $result->field_options['format'] ) ) {
+		if ( ! isset( $result->field_options ) || ! is_array( $result->field_options ) || empty( $result->field_options['format'] ) ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This fix works great!

AbdiTolesa:
I'm not able to replicate this issue on master, @garretlaxton have you tried replicating the issue on master?

Crabcyborg:
@AbdiTolesa I'm not sure why you are having issues replicating this now. It's the same issue you've tried to fix with a previous PR (https://github.com/Strategy11/formidable-forms/pull/1746). And you got pretty close with identifying the issue being related to double unslashing and cache.

I'll just merge this. I think it's good to go.

PR #1961, opened 2024-09-03
Support fields validation with captcha

AbdiTolesa:

PR #2003, opened 2024-09-20
Fix pro issue 1961

Crabcyborg:
This includes a previous update https://github.com/Strategy11/formidable-forms/pull/1961

But fixes a bug I was seeing where invisible captchas would fail validation before the validation would happen before the invisible captchas has executed.

I also cleaned up the `validateRecaptcha` function a little. I didn't change any logic. I just have it return early when there are no captchas, and updated the code to use `const` more.

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -204,7 +204,11 @@ function frmFrontFormJS() {
 			}
 		}
 
-		errors = validateRecaptcha( object, errors );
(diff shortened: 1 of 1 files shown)


AbdiTolesa (approved):
This looks good to me, I found no issues with it.

Crabcyborg:
Thanks @AbdiTolesa!

🚀

PR #2007, opened 2024-09-25
Allow field search to also match plural

Crabcyborg:
Laura noted this in QA https://github.com/Strategy11/formidable-forms/pull/1963#issuecomment-2371697547

If you search for "signatures", it won't match "signature".

This update adds a little check for basic plural matches. I basically do a second check if the `item's value + 's'` matches.

I have a couple small checks in place so searching `'s'` on its own wont return every field, and if you search `ss` it won't match everything with a single `s`.


lauramekaj1 (approved):
@Crabcyborg I tested this and it looks good to me. Thank you! 🚀

Crabcyborg:
Thank you Laura!

🚀

PR #2013, opened 2024-09-30
Fix default email setting is never used

Crabcyborg:
**Related Slack conversation** https://strategy11.slack.com/archives/C799A2R61/p1727465741085109

New email actions now use `[default-email]` and `[default-from-email]` shortcodes.

Screen Shot 2024-10-01 at 2 59 11 PM

These settings are currently only set in the onboarding wizard. You can find them at the `/wp-admin/admin.php?page=formidable-onboarding-wizard` URL.

When these settings are empty or not valid emails, the admin email is used instead.

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -1077,7 +1077,7 @@ class FrmFieldsHelper {
 				break;
 			case 'get':
 				$new_value = self::process_get_shortcode( $atts, $return_array );
-		}
+		}//end switch
 
 		return $new_value;
 	}
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -846,6 +846,8 @@ class FrmFieldsHelper {
 			'siteurl',
 			'sitename',
 			'admin_email',
+			'default-email',
+			'default-from-email',
 			'post[-|_]id',
 			'created[-|_]at',
 			'updated[-|_]at',
@@ -968,7 +970,7 @@ class FrmFieldsHelper {
 			'ip'  => $atts['entry']->ip,
 		);
 
-		$dynamic_default = array( 'admin_email', 'siteurl', 'frmurl', 'sitename', 'get' );
+		$dynamic_default = array( 'admin_email', 'siteurl', 'frmurl', 'sitename', 'get', 'default-email', 'default-from-email' );
 
 		if ( isset( $shortcode_values[ $atts['tag'] ] ) ) {
 			$replace_with = $shortcode_values[ $atts['tag'] ];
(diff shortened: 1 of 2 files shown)


truongwp:
@Crabcyborg If user changes the default email and wants to apply that change to all form actions that have the default email address, it will require updating all actions to change it. I think a shortcode will be better.

Crabcyborg:
@truongwp Since it's only set in the onboarding wizard, I think it won't likely get updated.

I agree shortcodes make sense, but require additional documentation and discussion.

@srwells If we were to add new shortcodes to email actions for the "Default From Email" and the "Default To Email", what do you think the shortcodes should be?

Right now we use `[admin_email]` for the WordPress admin email. We are generally trying to move away from underscores though, so we would generally lean toward hyphens with a new shortcode.

We already have `[default-message]` so for consistency, I think `[default-from-email]`/`[default-to-email]` make sense.

But I think we could possibly go with something shorter here. Possibly `[from-email]`/`[to-email]`, dropping the `default-` prefix, or even dropping the `-email` descriptor and using `[default-from]`/`[default-to]`? If we try something too generic, we risk running into field key conflicts if someone had a field with a key like `to` for example.

Thank you!

srwells:
@Crabcyborg @truongwp. So to make sure I understand. [admin-email] will always point to the WordPress admin email setting, so we are coming up with a different shortcode that will point to the FF global settings email address, and fall back to the admin email if that does not exist or is invalid, Correct?

In that case, I would recommend [default-email]. I don't think we need shortcodes for to/from email addresses. 

The purpose of this feature is simply to allow all forms to send to a different email than is set in their WP settings by default. 

So [admin_email] could point to WP admin email always, and [default-email] would point to the FF global settings first, then fallback to WP admin email.

Crabcyborg:
@srwells

Thanks for your feedback.

> In that case, I would recommend [default-email]. I don't think we need shortcodes for to/from email addresses.

We are adding a "FROM" email setting to avoid deliver-ability issues. So we have two values, which would require two shortcodes. I think `[default-email]` is fine as the default `TO` option, but we would also need an additional `[default-from-email]`.

Screen Shot 2024-09-30 at 6 02 01 PM

srwells:
Yeah, good point. Razvan is currently working on a simplified onboarding to maximize email signups. It does away with a displayed email field at all as part of onboarding. https://www.figma.com/design/GRF0taNpam5N3RYjgD22Pi/Onboarding?node-id=901-586&t=4G6dXcXjPezzIzKQ-1

Crabcyborg:
Thanks @srwells. Long term, I think we should probably have a dedicated non-onboarding place for these settings anyway. What do you think about adding an Email section somewhere to our Global Settings so these can be modified more easily for existing users?

srwells:
Yep, I agree. I’m working on this with @tuguirazvan. I’m thinking maybe a modal that shows the first time they go to the form actions page and a section on the global settings page.

shervElmi (approved):
Thanks for your work on this, @Crabcyborg. It looks great to me! 🚀

Crabcyborg:
@truongwp I just updated this so it now uses `[default-email]` and `[default-from-email]` shortcodes. These are now used instead of the `[admin_email]` placeholders for new email actions. If the onboarding setting is not set or not a valid email, they fallback on the admin email.

Can you take a look? I plan to include this in tomorrow's release.

@garretlaxton This could use testing as well. Do you think you could take a look at this today?

Thank you!

truongwp (approved):

Crabcyborg:
Thanks @truongwp!

🚀

PR #2017, opened 2024-10-02
Show save and reload modal on field type change

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5339
Related update: https://github.com/Strategy11/formidable-pro/pull/5443
### Test Procedure
1. Create a form and add a Product field and a different type field.
2. Go to the field options for those fields one by one and try changing the field type in Advanced > Field Type option.
3. Confirm that the 'Save and Reload' modal shows up when changing the field type and clicking on the each button in the modal does what is expected.

### Screen record
CleanShot 2024-10-02 at 14 31 36


garretlaxton (approved):
This fix works great!

js/formidable_admin.js
```diff
@@ -10521,6 +10601,7 @@ function frmAdminBuildJS() {
 			toggleSectionHolder();
 			handleShowPasswordLiveUpdate();
 			document.addEventListener( 'scroll', updateShortcodesPopupPosition, true );
+			document.addEventListener( 'change', handleChangeEventonDocument );
```
truongwp, line 10604: We can use `frmDom.util.documentOn()` instead.

js/formidable_admin.js
```diff
@@ -8375,6 +8375,86 @@ function frmAdminBuildJS() {
...
+	 * @since x.x
+	 *
+	 * @param {HTMLElement} target
+	 * @returns {Void}
+	 */
+	function maybeShowSaveAndReloadModal( target ) {
+		if ( ! target.id.startsWith( 'field_options_type_' ) && ! target.classList.contains( 'frmjs_prod_data_type_opt' ) ) {
+			return;
+		}
+		const idParts = target.id.split( '_' );
```
truongwp, line 8402: This doesn't work with the Product type setting because it doesn't have an id attribute.

js/formidable_admin.js
AbdiTolesa: Yeah but I think this generic function (`handleChangeEventonDocument `) could be used to handle other future `change` events on `document`, so lets leave it as is.

js/formidable_admin.js
AbdiTolesa: @truongwp Yeah, product field would always show that modal on the field type change because of that, and I think that isn't a big deal for now. But I have added a condition check for the code that helps existing early.

truongwp: Why do we always show that modal for the Product field? Sorry, I might have missed something.

AbdiTolesa: That's just a down side, not something intended 😄 We don't already have an information in the HTML we can use to track the original Product field type when the page loads. For other field types, we have a class `frm-type-{fieldType}` and we use it to avoid the popup, if the user just reverted back to the original field type when page is loaded. I hope that is clear now. So if original field type is `Checkbox` and user change it to `Radio`, we show the popup. If the user cancels the save and reload popup, and then again changed the field type to `Checkbox` type (which is the field type when page initially loaded) we avoid the popup since nothing is really changed.

truongwp: We can add a class or data- attribute for product type. I think it's worth adding it, it might be helpful in the future.
The Product field is in Pro, so we should move the related code to the Pro plugin.

AbdiTolesa: Ok, I have create a separate PR in Pro to handle that case: https://github.com/Strategy11/formidable-pro/pull/5443

truongwp (approved):

js/formidable_admin.js
```diff
@@ -10521,6 +10604,7 @@ function frmAdminBuildJS() {
 			toggleSectionHolder();
 			handleShowPasswordLiveUpdate();
 			document.addEventListener( 'scroll', updateShortcodesPopupPosition, true );
+			document.addEventListener( 'change', handleChangeEventonDocument );
```
Crabcyborg, line 10607: It looks like this function is only called on the builder page.

`handleChangeEventonDocument` isn't really descriptive enough.

I think we should mention `builder` somewhere in this name.

Crabcyborg (approved):

🚀

PR #2023, opened 2024-10-03
validations for applications page

lauramekaj1:


tests/cypress/e2e/Applications/validateApplicationsPage.cy.js
```diff
@@ -0,0 +1,300 @@
...
+        cy.get('.frm-search-result > :nth-child(1) > h3').should('contain', "Business Hours");
+        cy.get('#frm-application-search').clear().type("menu");
+        cy.get('.frm-search-result > :nth-child(1) > h3').should('contain', "Restaurant Menu");
+        cy.get('#frm-application-search').clear().type("business");
+        cy.get(':nth-child(1) > :nth-child(1) > h3').should('contain', "Business Directory");
+        cy.get(':nth-child(2) > :nth-child(1) > h3').should('contain', "Business Hours");
+
+        cy.log("Search for non-valid application templates");
+        cy.get('#frm-application-search').clear().type("Application does not exist");
+        cy.get('#frm_application_templates_grid > :nth-child(17)').should('contain', 'No application templates match your search query.');
```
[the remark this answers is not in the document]
Crabcyborg: @lauramekaj1 What do you think about this comment Code Rabbit has left. I think `:nth-child(17)` might be a bit of a weird selector here. A slight markup chance would likely break this check.

Is there a better way to check this?

tests/cypress/e2e/Applications/validateApplicationsPage.cy.js
```diff
@@ -0,0 +1,300 @@
...
+        });
+
+        cy.get('#frm_application_templates_grid > :nth-child(3)').within(() => {
+            cy.get('h3 .frm-inner-text').should('contain.text', 'Certificate');
+            cy.get('.button.frm-button-secondary.frm-button-sm')
+                .should('contain.text', 'Learn More')
+                .and('have.attr', 'aria-description', 'Certificate Template');
+            cy.get('.frm-application-card-image-wrapper img')
+                .should('have.attr', 'src')
+                .and('include', '/plugins/formidable-forms/images/applications/thumbnails/placeholder.svg');
```
Crabcyborg, line 87: @lauramekaj1 I made a new PR to address this issue with so many applications using the `placeholder.svg`.

New PR https://github.com/Strategy11/formidable-forms/pull/2030

I merged those changes here and broke your tests. Would you be able to help update these checks so they match the new thumbnails? A lot of the new ones are `.webp` files. They should all match the application "keys". All of the images are included in the `applications/thumbnails/` folder as well where they can be referenced.

lauramekaj1: @Crabcyborg Thank you for looking into this.
I didn't know those thumbnails were missing cause I thought usage of placeholder.svg was on purpose. now I update tests to match new thumbnails.
Also, updated the tests not to include selectors like _:nth-child_ because I agree that it can be unreliable.

Crabcyborg: Thanks @lauramekaj1.

The `placeholder.svg` is because WordPress has restrictions about how images are loaded, restricting us from loading the images directly from our site. Instead we need to add the images into the plugin directly, so they can load locally.

This means any new applications will use placeholders, which isn't ideal, but we need to display something if there is no local image available.

Crabcyborg (approved):
Thank you @lauramekaj1!

I think this looks good to merge now.

🚀🙌

PR #2036, opened 2024-10-08
Set aria-invalid to true only on name part input with error

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5409

### Test steps
1. Create a form and add a name field and set it to be required.
2. Preview your form and try submitting the form filling part of the name field at a time (either first, middle or last).
3. Confirm that the submission is failed and the first field with error is focused on.
4. Confirm that `aria-invalid` attribute is only set on fields with error.

### Screen record
CleanShot 2024-10-09 at 15 48 22


lauramekaj1:
Hi @AbdiTolesa, I tested this by doing multiple scenarios using different input fields.

1. It seems that the Name field is not working as expected; after clicking 'Submit,' the fields with errors are not being focused at all. I saw in your screen recording that the Name field worked fine, but could you take another look?
2. For the Paragraph field, after clicking the 'Submit' button, the field is getting focused, but with a red border.
https://www.loom.com/share/8b37bdc839024992b3ab8dea00499a6d?sid=ec2f5d7e-7c2c-4744-8afa-4057848ca33d

AbdiTolesa:
> Hi @AbdiTolesa, I tested this by doing multiple scenarios using different input fields.
> 
> 1. It seems that the Name field is not working as expected; after clicking 'Submit,' the fields with errors are not being focused at all. I saw in your screen recording that the Name field worked fine, but could you take another look?
> 2. For the Paragraph field, after clicking the 'Submit' button, the field is getting focused, but with a red border.
>    https://www.loom.com/share/8b37bdc839024992b3ab8dea00499a6d?sid=ec2f5d7e-7c2c-4744-8afa-4057848ca33d

Hi @lauramekaj1, thank you for catching the issue with Paragraph fields! I have deployed this to the QA so that we can be on the same page while testing. Here is the link: https://qa.formidableforms.com/abdi/wp-admin/admin-ajax.php?action=frm_forms_preview&form=c35sh

Let me know if you can't access it for some reason.

lauramekaj1:
> Hi @lauramekaj1, thank you for catching the issue with Paragraph fields! I have deployed this to the QA so that we can be on the same page while testing. Here is the link: https://qa.formidableforms.com/abdi/wp-admin/admin-ajax.php?action=frm_forms_preview&form=c35sh
> 
> Let me know if you can't access it for some reason.

Hi @AbdiTolesa, I tested this from your QA site and it seems that it's working as expected. Thank you!

lauramekaj1 (approved):

classes/controllers/FrmFieldsController.php
```diff
@@ -915,6 +915,16 @@ private static function add_shortcodes_to_html( $field, array &$add_html ) {
 		}
 
 		foreach ( $field['shortcodes'] as $k => $v ) {
+			if ( $field['type'] === 'name' && 0 === strpos( $k, 'aria-invalid' ) && isset( $field['subfield_name'] ) ) {
+				$subfield_name = $field['subfield_name'];
+				if ( isset( $field['shortcodes'][ 'aria-invalid-' . $subfield_name ] ) ) {
```
truongwp, line 920: We can swap this `if` to run `continue` in the first block to save some indentation.

truongwp (approved):

classes/controllers/FrmFieldsController.php
```diff
@@ -915,6 +915,15 @@ private static function add_shortcodes_to_html( $field, array &$add_html ) {
 		}
 
 		foreach ( $field['shortcodes'] as $k => $v ) {
+			if ( isset( $field['subfield_name'] ) && 'name' === $field['type'] && 0 === strpos( $k, 'aria-invalid' ) ) {
+				$subfield_name = $field['subfield_name'];
+				if ( ! isset( $field['shortcodes'][ 'aria-invalid-' . $subfield_name ] ) ) {
```
Crabcyborg, line 920: @AbdiTolesa Is `$k` here different from `'aria-invalid-' . $subfield_name`?

If these values are the same, this might be easier to follow if we use the same variable everywhere.

Also, I think this could could possibly use a comment, explaining what it's doing here for name fields and why it's required. It isn't very clear right now at a glance.

classes/models/FrmFieldFormHtml.php
```diff
@@ -394,13 +394,31 @@ private function prepare_input_shortcode_atts( $shortcode_atts ) {
 			unset( $shortcode_atts['class'] );
 		}
 
-		$shortcode_atts['aria-invalid'] = isset( $this->pass_args['errors'][ 'field' . $this->field_id ] ) ? 'true' : 'false';
+		if ( $this->field_obj->get_field_column( 'type' ) === 'name' ) {
```
Crabcyborg, line 397: It looks like we already have the `field_obj`/`FrmFieldType` here.

This would work better moving forward if we did this in an object oriented way and made it part of `FrmFieldType`. Ideally we don't want references to a specific field type in files like `FrmFieldFormHtml.php`. It's a bit of a code smell.

js/formidable.js
```diff
@@ -1232,6 +1232,27 @@ function frmFrontFormJS() {
...
+	 *
+	 * @since x.x
+	 *
+	 * @param {HTMLElement} element
+	 */
+	function maybeSetFocusOnNameFieldElement( element ) {
+		if ( 'FIELDSET' !== element.nodeName ) {
+			return;
+		}
+		if ( ! element.querySelector( '.frm_combo_inputs_container[data-name-layout]' ) ) {
```
Crabcyborg, line 1246: Would it make sense to not do anything "name" specific here? It would be nice if this just worked for combo fields so if we added another combo fields it would work as well.

classes/controllers/FrmFieldsController.php
```diff
@@ -915,6 +915,15 @@ private static function add_shortcodes_to_html( $field, array &$add_html ) {
 		}
 
 		foreach ( $field['shortcodes'] as $k => $v ) {
+			if ( isset( $field['subfield_name'] ) && 'name' === $field['type'] && 0 === strpos( $k, 'aria-invalid' ) ) {
```
Crabcyborg, line 918: This would probably be more future-proof if we removed the `'name' === $field['type']` check.

In the future if we added other sub fields, I think we could apply this logic there as well.

classes/controllers/FrmFieldsController.php
AbdiTolesa: Yes they may be different since `$k` is just key for the current iteration of possible subfields while `$field['subfield_name']` is a check for the current subfield we are preparing html for.

js/formidable.js
AbdiTolesa: If we remove that check, it would affect how other combo fields work currently as it focuses on the first input in the combo field whether it is invalid or not since this update does not cover every field yet. So as long as the scope of this issue, the check should be better there.

css/_single_theme.css.php
```diff
@@ -179,6 +179,7 @@
 }
 
 .<?php echo esc_html( $style_class ); ?> .form-field input:not([type=file]):not([type=range]):not([readonly]):focus,
+.<?php echo esc_html( $style_class ); ?> .form-field textarea:focus,
```
Crabcyborg, line 182: There's a similar `.<?php echo esc_html( $style_class ); ?> textarea:focus,` style below this. Is that not working?

I don't think we should probably have this and the other style. We probably want to update the other selector instead.

js/formidable.js
Crabcyborg: Maybe we should be handling address fields at the same time here. It doesn't make a lot of sense to only fix a single field type.

classes/models/fields/FrmFieldName.php
```diff
@@ -172,6 +172,22 @@ public function sanitize_value( &$value ) {
...
+	 *
+	 * @since x.x
+	 *
+	 * @param array $shortcode_atts
+	 * @param array $args
+	 *
+	 * @return void
+	 */
+	public function set_aria_invalid_error( &$shortcode_atts, $args ) {
+		foreach ( array( 'first', 'middle', 'last' ) as $name_part ) {
```
Crabcyborg, line 186: @AbdiTolesa What do you think about moving this into `FrmFieldCombo` and using data from `$this->get_sub_fields` instead? It would be great if we didn't need to basically copy/paste this for another combo field later.

js/formidable.js
Crabcyborg: It looks like this is still hard coded to only support name fields.

AbdiTolesa: Sorry for missing this comment! I have now updated this function so that it would work for other combo fields too.

js/formidable.js
```diff
@@ -1339,6 +1356,8 @@ function frmFrontFormJS() {
 				break;
 			}
 
+			maybeFocusOnComboSubField( element );
```
Crabcyborg, line 1359: I think it might make sense to `break` here if we do focus a field.

`maybeFocusOnComboSubField` could return a boolean, returning true only if we find the element and call `.focus()`.

js/formidable.js
```diff
@@ -1319,6 +1319,25 @@ function frmFrontFormJS() {
...
+	 * @return {boolean} True if the focus was set on a combo field.
+	 */
+	function maybeFocusOnComboSubField( element ) {
+		if ( 'FIELDSET' !== element.nodeName ) {
+			return false;
+		}
+		if ( ! element.querySelector( '.frm_combo_inputs_container' ) ) {
+			return false;
+		}
+		element.querySelector( '[aria-invalid="true"]' )?.focus();
```
Crabcyborg, line 1337: I think we should also check that this `[aria-invalid="true"]` element exists first before returning true.

We should set this to a `const` variable and drop the nullable.

Crabcyborg (approved):
I think this is all good to merge now.

🚀

PR #2046, opened 2024-10-11
Redirect immediately to onboarding when activating a single plugin

Crabcyborg:
…sient

Fixes https://github.com/Strategy11/formidable-pro/issues/5411

Now we only avoid redirecting immediately when we are not bulk-activating. If the plugin is activated when bulk-activating, we fallback to the old behaviour, where we check for a Formidable admin page.


shervElmi (approved):
Thanks @Crabcyborg, It looks great to me! 🚀

Crabcyborg:
Thanks Sherv!

PR #2049, opened 2024-10-14
Usage tracking v2

truongwp:
This PR:
- Changes usage endpoint to `https://usage2.formidableforms.com/snapshot`.
- Adds more usage data and improves some old ones.
- Adds code to track form templates and edit form clicked links.
- Logs the usage request in case it's failed. I did this when debugging. I'm not sure we should keep this.


classes/models/FrmUsage.php
```diff
@@ -89,11 +99,30 @@ public function snapshot() {
 			'actions'           => $this->actions(),
 
 			'onboarding-wizard' => FrmOnboardingWizardController::get_usage_data(),
```
Crabcyborg, line 101: @truongwp In older versions of Lite (before v6.16), it looks like we were sending some data in this like `default_email`. Those settings set set to an option, so they'll continue to stay there if user's added them before.

Since we want the tracking data to be anonymous, can we filter this data and only return the current keys we're trying to track here?

This code here has all of the expected keys now.
```
    $fields_to_update = array(
			'allows_tracking'  => 'rest_sanitize_boolean',
			'installed_addons' => 'sanitize_text_field',
			'processed_steps'  => 'sanitize_text_field',
			'completed_steps'  => 'rest_sanitize_boolean',
		);
```

Crabcyborg: We might want to visit this further. I don't think `'allows_tracking'  => 'rest_sanitize_boolean',` and `'installed_addons' => 'sanitize_text_field',` are likely useful here either.

Crabcyborg (commented):
@truongwp I heard Dalton mentioning the other day in a meeting how we want to track to know if people are using AI features.

Do we have anything like that in place yet?

We'd like to know probably how many AI fields are being used? And how often they're building an AI-generated form.

Can we track those things as well?

Thank you!

truongwp:
@Crabcyborg This PR doesn't track the AI-generated form. It will be in the AI addon. 

There are 2 ways to track it:
- Save a value in the form options for AI-generated forms.
- Update the count of AI-generated form right after form is created (on line https://github.com/Strategy11/formidable-ai/blob/e25af8fc6329fdb0e381237eb81cd5299afed538/classes/controllers/FrmAIAppController.php#L378) in the `frm_usage_tracking_flows` option.

Which way do you prefer?

I will merge AI-generated form data into the form templates. AI-generated forms will have the `ai` template. What do you think about this?

classes/models/FrmUsage.php
truongwp: @Crabcyborg I skipped all unnecessary keys. Any data added in the future will be tracked by default.

classes/controllers/FrmUsageController.php
```diff
@@ -56,4 +65,74 @@ public static function send_snapshot() {
...
+
+		if ( ! isset( $flows_data[ $key ][ $value ] ) ) {
+			$flows_data[ $key ][ $value ] = 0;
+		}
+
+		$flows_data[ $key ][ $value ]++;
+		update_option( self::FLOWS_ACTION_NAME, $flows_data );
+
+		wp_send_json_success();
+	}
```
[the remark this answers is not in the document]
truongwp: @Crabcyborg Do you think we should check `$flows_data[ $key ][ $value ] < PHP_INT_MAX`?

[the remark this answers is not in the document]
Crabcyborg: @truongwp I think we're fine. The lowest `PHP_INT_MAX` should ever be is `2147483647` which is a number we shouldn't expect to ever reach here.

Crabcyborg:
Sorry I missed the notification on this @truongwp!

It looks like the update you made in the AI add-on is a great solution.

classes/controllers/FrmHooksController.php
```diff
@@ -95,6 +95,8 @@ public static function load_hooks() {
 
 		add_filter( 'cron_schedules', 'FrmUsageController::add_schedules' );
 		add_action( 'formidable_send_usage', 'FrmUsageController::send_snapshot' );
+		add_action( 'admin_enqueue_scripts', 'FrmUsageController::load_scripts' );
+		add_action( 'wp_ajax_frm_track_flows', 'FrmUsageController::ajax_track_flows' );
```
Crabcyborg, line 99: @truongwp Can we move this line to our `load_ajax_hooks` function?

classes/controllers/FrmUsageController.php
```diff
@@ -56,4 +65,78 @@ public static function send_snapshot() {
...
+
+	/**
+	 * AJAX handler to track flows.
+	 *
+	 * @since x.x
+	 */
+	public static function ajax_track_flows() {
+		check_ajax_referer( 'frm_ajax', 'nonce' );
+
```
Crabcyborg, line 109: Is there a permission check we should put here?

classes/models/FrmUsage.php
```diff
@@ -31,9 +31,19 @@ public function send_snapshot() {
 				'Content-Length' => strlen( $body ),
 			),
 			'body'    => $body,
+			'timeout' => 45, // Without this, Debug Log catches the `http_request_failed` error.
```
Crabcyborg, line 34: @truongwp Would we also want to increase the server timeout?

We sometimes put some code in cases where a timeout may happen,
```
if ( function_exists( 'set_time_limit' ) ) {
	set_time_limit( 0 );
}
```

js/admin/usage-tracking.js
```diff
@@ -0,0 +1,27 @@
...
+	frmDom.util.documentOn( 'click', '.frm-form-templates-use-template-button', function( event ) {
+		sendData({
+			key: 'form_templates',
+			value: event.target.closest( 'li' ).dataset.slug
+		});
+	});
+
+	const sendData = ( data ) => {
+		const formData = new FormData();
+		Object.keys( data ).forEach( key => {
```
Crabcyborg, line 22: It looks like `data` is always just a `key` and a `value`. This code here seems like it could be more simple.

I would possibly just change this to something like:
```
const sendData = ( key, value ) => {
	const formData = new FormData();
	formData.append( key, value );
	frmDom.ajax.doJsonPost( 'track_flows', formData );
}
```

classes/models/FrmUsage.php
Crabcyborg: Thanks Truong!

classes/controllers/FrmUsageController.php
truongwp: I added the `frm_view_forms` permission check.

js/admin/usage-tracking.js
truongwp: If using `formData.append( key, value );`, in the AJAX handler, the `$_POST` data will be look like:
```
array(
    'form_templates' => 'ui-light',
    '_ajax_nonce'    => '...',
    ... some more AJAX request related data
)
```

I see that it's more complex to get the flows data with this.

truongwp: I simplified the `sendData()` function.

classes/models/FrmUsage.php
```diff
@@ -269,7 +341,19 @@ private function forms() {
...
 			foreach ( $settings as $setting ) {
 				if ( isset( $form->options[ $setting ] ) ) {
-					$new_form[ $setting ] = $this->maybe_json( $form->options[ $setting ] );
+					if ( 'custom_style' === $setting ) {
+						$style->id = $form->options[ $setting ];
+						if ( 1 === intval( $style->id ) ) {
+							$style_name = 'formidable-style';
+						} else {
+							$style_post = $style->get_one();
+							$style_name = $style_post ? $style_post->post_name : 'formidable-style';
```
Crabcyborg, line 350: @truongwp It's possible that someone has disabled styling as well.

Should we account for that? It would be nice to know if people are disabling styling.

js/admin/usage-tracking.js
Crabcyborg: Thanks @truongwp. That's what I meant. I got a bit mixed up on the example.

classes/models/FrmUsage.php
truongwp: I updated the code to track the disabled styling case.

Crabcyborg (approved):
Thanks @truongwp!

I think this is good to go for next release. I'll go ahead and merge this.

🚀

truongwp:
@Crabcyborg I fixed those errors.

This PR logs the failed usage requests. Do you think we should keep it?

Crabcyborg:
> This PR logs the failed usage requests. Do you think we should keep it?

@truongwp I'm not sure. If there are issues, it would help make them more visible. But it also isn't something that our customers shouldn't need to worry about.

I think we're fine to just remove it.

classes/models/FrmUsage.php
```diff
@@ -88,12 +104,74 @@ public function snapshot() {
...
+	private function payments( $table = 'frm_payments' ) {
+		$allowed_tables = array( 'frm_payments', 'frm_subscriptions' );
+		if ( ! in_array( $table, $allowed_tables, true ) ) {
+			return array();
+		}
+
+		global $wpdb;
+		$rows = $wpdb->get_results(
+			$wpdb->prepare(
+				"SELECT amount, status, paysys, created_at FROM %i",
```
Crabcyborg, line 157: @truongwp I believe `%i` support only exists as of WordPress 6.2. We might need to do it with unsafe way for now, or check for WordPress version and have the fallback code for now that usually isn't used.

Right now we're targeting 5.2+.

truongwp: I updated to use `%1$s` now. From the doc:
```
Note: There is one exception to the above: for compatibility with old behavior, numbered or formatted string placeholders (eg, %1$s, %5s) will not have quotes added by this function
```

Crabcyborg (approved):
Thanks @truongwp!

🚀

PR #2052, opened 2024-10-16
Show/Hide some shortcodes contextually

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/829

Related update: https://github.com/Strategy11/formidable-views/pull/592

### Test procedure
1. Create a form and go to its settings. Click on the 'Send email' action and try opening the shortcodes input `...` next to `To`, `From`, `CC` and `BCC` fields.
2. Click on `Advanced` tab in the shortcodes modal.
3. Confirm that `Admin email` and `Default from email` shortcodes are available in the list.
4. Now, go to the form builder and add fields that use the shortcodes modal, like an HTML field.
5. Go to the HTML field's options and try entering content into the `Content` field to open the shortcodes popup.
6. Confirm that `Default Msg`, `Default HTML`, `Default Plain` and `Form name` are not under the `Advanced` tab.

### Screen record
CleanShot 2024-10-16 at 14 39 07


lauramekaj1 (approved):
@AbdiTolesa I tested and verified that shortcodes are added/removed as described in this PR.

js/formidable_admin.js
```diff
@@ -8546,7 +8566,35 @@ function frmAdminBuildJS() {
...
 
+	/**
+	 * Returns shortcodes that are contextual to the current input field.
+	 *
+	 * @since x.x
+	 *
+	 * @returns {Array}
+	 */
+	function getContextualShortcodes() {
+		return [ 'admin_email', 'default-from-email' ];
```
truongwp, line 8577: Can we get this from PHP? If we change this, we will only need to change in one place.

js/formidable_admin.js
```diff
@@ -8546,7 +8566,35 @@ function frmAdminBuildJS() {
...
+	}
+
+	/**
+	 * Hide shortcodes that are contextual to the current input field.
+	 *
+	 * @since x.x
+	 * @returns {Void}
+	 */
+	function hideContextualShortcodes() {
```
truongwp, line 8586: Can we hide the contextual shortcodes in this https://github.com/Strategy11/formidable-forms/pull/2052/files#diff-23fb66f34c625a76ac7564df177d3ab58c86b1f5d071f745aa50e028fb0209c4R8462?

classes/views/shared/mb_adv_info.php
```diff
@@ -210,6 +210,10 @@
 			$classes .= in_array( $skey, array( 'siteurl', 'sitename', 'entry_count' ), true ) ? ' show_before_content show_after_content' : '';
 			$classes .= strpos( $skey, 'default-' ) === 0 ? ' hide_frm_not_email_subject' : '';
 
+			if ( in_array( $skey, array( 'admin_email', 'default-from-email' ), true ) ) {
```
truongwp, line 213: This can also be get from `get_contextual_shortcodes()`.

js/formidable_admin.js
```diff
@@ -8447,6 +8447,64 @@ function frmAdminBuildJS() {
...
+		return JSON.parse( document.getElementById( 'frm_adv_info' ).dataset.contextualShortcodes );
+	}
+
+	/**
+	 * Show shortcodes that are contextual to the current input field.
+	 *
+	 * @since x.x
+	 * @returns {Void}
+	 */
+	function showContextualShortcodes() {
```
truongwp, line 8485: We can change this into `toggleContextualShortcodes()` and merge the `hideContextualShortcodes()` into this.

AbdiTolesa: We already have `showOrHideContextualShortcodes` function that calls this and `hideContextualShortcodes` functions conditionally.

truongwp: Their code is similar so I think we can merge them into one.

AbdiTolesa: Oh, thanks! Sorry for the oversight!

truongwp (approved):

Crabcyborg:
@AbdiTolesa I think this is incomplete.

Referring back to the original issue, it mentions what I believe was the setting for what is now a confirmation action message.

With this update, I can still see `[default-msg]`, `[default-plain]`, etc, inside a confirmation action. We should only show these options for email action types - I don't believe any other action type requires this.

Screen Shot 2024-11-08 at 1 04 31 PM

classes/controllers/FrmFormsController.php
```diff
@@ -1630,6 +1630,19 @@ private static function user_shortcodes() {
...
+	/**
+	 * Returns shortcodes that are shown/hidden based on the context.
+	 *
+	 * @since x.x
+	 * @return array
+	 */
+	public static function get_contextual_shortcodes() {
+		return array(
+			'admin_email'        => __( 'Admin email', 'formidable' ),
+			'default-from-email' => __( 'Default from email', 'formidable' ),
```
Crabcyborg, line 1642: We also have a `[default-email]` shortcode that would be nice to include.

AbdiTolesa:
> @AbdiTolesa I think this is incomplete.
> 
> Referring back to the original issue, it mentions what I believe was the setting for what is now a confirmation action message.
> 
> With this update, I can still see `[default-msg]`, `[default-plain]`, etc, inside a confirmation action. We should only show these options for email action types - I don't believe any other action type requires this.
> 
> Screen Shot 2024-11-08 at 1 04 31 PM

@Crabcyborg I have now added more checks to limit the availability of these shortcodes to email action only. I also tested this again and found searching shortcodes is not working as expected after this update and fixed the conflict when doing so.

classes/controllers/FrmFormsController.php
```diff
@@ -1630,6 +1630,27 @@ private static function user_shortcodes() {
 		return apply_filters( 'frm_user_shortcodes', $options );
 	}
 
+	/**
+	 * Returns shortcodes that are shown/hidden based on the context.
+	 *
+	 * @since x.x
+	 * @return array
+	 */
+	public static function get_contextual_shortcodes() {
```
Crabcyborg, line 1639: I don't think a controller is probably the right place for this. I'd probably go with the form helper file instead.

classes/controllers/FrmFormsController.php
```diff
@@ -1648,14 +1669,13 @@ private static function get_shortcode_helpers( $settings_tab ) {
 			'sitename'   => __( 'Site Name', 'formidable' ),
 		);
 
+		$contextual_shortcodes = self::get_contextual_shortcodes();
+		$entry_shortcodes      = array_merge( $contextual_shortcodes['address'], $contextual_shortcodes['body'], $entry_shortcodes );
```
Crabcyborg, line 1673: I see another `array_merge( $contextual_shortcodes['address'], $contextual_shortcodes['body'] )` later in the code.

It might be good to move this somewhere where we don't need to update both of these lines any time another context is added.

Something that could probably just loop through `$contextual_shortcodes` so it gets every key.

Like the other comment, I think we could try to use the `FrmShortcodeHelper` class.

classes/views/frm-forms/mb_insert_fields.php
```diff
@@ -2,8 +2,18 @@
 if ( ! defined( 'ABSPATH' ) ) {
 	die( 'You are not allowed to call this page directly.' );
 }
+$contextual_shortcodes = FrmFormsController::get_contextual_shortcodes();
+
+$data_attrs = array(
+	'data-contextual-shortcodes' => FrmAppHelper::maybe_json_encode(
+		array(
```
Crabcyborg, line 9: I think we could move this logic into the `FrmShortcodeHelper` class as well, to help remove logic from view files where there shouldn't really be a lot of non-HTML code.

We could also make this easier to maintain in the future. I think we could loop through `$contextual_shortcodes` instead of hard coding all of this.

js/formidable_admin.js
```diff
@@ -9456,12 +9541,16 @@ function frmAdminBuildJS() {
...
 			if ( searchText === '' ) {
 				if ( itemCanBeShown ) {
-					items[i].classList.remove( 'frm_hidden' );
+					if ( checkContextualShortcode( items[i] ) ) {
+						items[i].classList.remove( 'frm_hidden' );
+					}
 				}
 				items[i].classList.remove( 'frm-search-result' );
 			} else if ( ( regEx && new RegExp( searchText ).test( innerText ) ) || innerText.indexOf( searchText ) >= 0 || textMatchesPlural( innerText, searchText ) ) {
 				if ( itemCanBeShown ) {
```
Crabcyborg, line 9550: It looks like this could be merged into a single `if ( itemCanBeShown && checkContextualShortcode( items[i] )  ) {`.

js/formidable_admin.js
```diff
@@ -9456,12 +9541,16 @@ function frmAdminBuildJS() {
 			const itemCanBeShown = ! ( getExportOption() === 'xml' && items[i].classList.contains( 'frm-is-repeater' ) );
 			if ( searchText === '' ) {
 				if ( itemCanBeShown ) {
```
Crabcyborg, line 9543: Here as well, I think we could merge the `if` conditions and check `if ( itemCanBeShown && checkContextualShortcode( items[i] ) ) {`

classes/controllers/FrmFormsController.php
```diff
@@ -1648,14 +1669,13 @@ private static function get_shortcode_helpers( $settings_tab ) {
...
 		if ( ! FrmAppHelper::pro_is_installed() ) {
 			unset( $entry_shortcodes['post_id'] );
 		}
 
 		if ( $settings_tab ) {
```
Crabcyborg, line 1678: If we drop this name for Form Name, it looks like we could drop the `$settings_tab` param entirely, so we can manage this all in JS instead of having two ways like we kind of have now.

js/formidable_admin.js
```diff
@@ -8447,9 +8447,94 @@ function frmAdminBuildJS() {
...
+
+	/**
+	 * Returns shortcodes that are contextual to the current input field.
+	 *
+	 * @since x.x
+	 *
+	 * @returns {Array}
+	 */
+	function getContextualShortcodes() {
+		return JSON.parse( document.getElementById( 'frm_adv_info' ).dataset.contextualShortcodes );
```
Crabcyborg, line 8535: Is this something we could set to a variable on load and re-use? It seems a bit inefficient that is uses `JSON.parse` and queries for an element, when it's used in one-off ways like `getContextualShortcodes().address.includes(`.

js/formidable_admin.js
```diff
@@ -8447,9 +8447,94 @@ function frmAdminBuildJS() {
...
+	 *
+	 * @param {HTMLElement} item
+	 * @returns {Boolean}
+	 */
+	function canShowContextualShortcode( item ) {
+		const shortcode = item.querySelector( 'a' ).dataset.code;
+		const inputId = document.getElementById( 'frm_adv_info' ).dataset.fills;
+		const input   = document.getElementById( inputId );
+		if ( getContextualShortcodes().address.includes( shortcode ) ) {
+			return input.matches( '[id^=email_to], [id^=from_], [id^=cc], [id^=bcc]' );
```
Crabcyborg, line 8491: It looks like these selectors are in multiple places, which makes them difficult to maintain.

Like I mentioned in my previous comment, I think we should put these selectors in the dataset and we can avoid having these hard coded strings in more than one place.

classes/views/shared/mb_adv_info.php
```diff
@@ -210,6 +210,12 @@
 			$classes .= in_array( $skey, array( 'siteurl', 'sitename', 'entry_count' ), true ) ? ' show_before_content show_after_content' : '';
 			$classes .= strpos( $skey, 'default-' ) === 0 ? ' hide_frm_not_email_subject' : '';
 
+			$contextual_shortcodes = FrmFormsController::get_contextual_shortcodes();
```
Crabcyborg, line 213: Can we call `$contextual_shortcodes = FrmFormsController::get_contextual_shortcodes();` outside of the loop? This doesn't seem very efficient where it is.

classes/views/shared/mb_adv_info.php
```diff
@@ -210,6 +210,12 @@
...
+
+			if ( in_array( $skey, array_keys( array_merge( $contextual_shortcodes['address'], $contextual_shortcodes['body'] ) ), true ) ) {
```
Crabcyborg, line 215: We should also handle this `array_merge` before the loop, not in every iteration since it doesn't change.

classes/controllers/FrmFormsController.php
```diff
@@ -1648,14 +1669,13 @@ private static function get_shortcode_helpers( $settings_tab ) {
...
-			$entry_shortcodes['default-message'] = __( 'Default Msg', 'formidable' );
-			$entry_shortcodes['default-html']    = __( 'Default HTML', 'formidable' );
-			$entry_shortcodes['default-plain']   = __( 'Default Plain', 'formidable' );
 			$entry_shortcodes['form_name']       = __( 'Form Name', 'formidable' );
```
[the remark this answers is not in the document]
Crabcyborg: So `$settings_tab` is basically `true` if we're on the Formidable builder or settings pages.

`$settings_tab` is `false` when it's used in the Visual Views editor.

We support `form_name` now in Views. There's actually code there that uses the `frm_helper_shortcodes` filter to add the form name shortcode.

We can remove that special code in Views if we just move this form name shortcode out of the if condition.

PR #2058, opened 2024-10-16
Try to render captcha a second time on a timeout

Crabcyborg:
Related Slack conversation https://strategy11.slack.com/archives/C799A2R61/p1729107464507059

Now if the form immediately becomes visible (within the next second), the captcha will still render.

### DIFF
--- a/js/formidable.js
+++ b/js/formidable.js
@@ -1863,8 +1863,9 @@ function frmCaptcha( captchaSelector ) {
 		const formIsVisible = closestForm && closestForm.offsetParent !== null;
 		const captcha       = captchas[c];
 		if ( ! formIsVisible ) {
-			// If the form is not visible, try again in 1 second.
+			// If the form is not visible, try again later in 400ms.
 			// This fixes issues where the form fades visible on page load.
+			// Or whne the form is inside of a modal.
 			const interval = setInterval(
 				function() {
 					if ( closestForm && closestForm.offsetParent !== null ) {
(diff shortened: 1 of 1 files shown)


engahmeds3ed:
[formidable-6.15.1-beta.zip](https://github.com/user-attachments/files/17415570/formidable-6.15.1-beta.zip)

Zip file from that change.

@Crabcyborg just to confirm, we are sure that one second is enough time, correct?

Crabcyborg:
@engahmeds3ed I'm not totally confident it's perfect, no.

It might be nice to actually use an interval so we can solve the modal issue as well.

Crabcyborg:
@engahmeds3ed I just updated this to try an interval instead. I shortened it to 400ms now as well so it should be more instant when the visibility change happens.

It should solve the modal issue as well.

Could you help test it out?

Thank you!

engahmeds3ed:
I tested this change for that case and it works perfectly.

Let me validate with the modal issue.

engahmeds3ed:
It works perfectly also with the modal issue, u r awesome @Crabcyborg 

The only risk here is that the interval will keep working forever till the modal trigger button is clicked, once it's clicked and the form appears, we will clear that interval.
I think this is acceptable risk as we only do a small check each time.

Crabcyborg:
Thanks @engahmeds3ed!

Yeah, I think it should be fine. It only happens if it's hidden as well and helps to fix the whole class of new tickets that have been coming up because of this visibility check.

PR #2061, opened 2024-10-17
Try to optimize how styles are saved

Crabcyborg:
Related Slack conversation https://strategy11.slack.com/archives/C799A2R61/p1729188618152819

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -10926,6 +10926,21 @@ function frmAdminBuildJS() {
 
 			// Set fieldsUpdated to 0 to avoid the unsaved changes pop up.
 			frmDom.util.documentOn( 'submit', '.frm_settings_form', () => fieldsUpdated = 0 );
+
+			const manageStyleSettings = document.getElementById( 'manage_styles_settings' );
+			if ( manageStyleSettings ) {
+				document.addEventListener(
+					'change',
+					event => {
+						const target = event.originalTarget;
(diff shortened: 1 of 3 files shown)


js/formidable_admin.js
```diff
@@ -10926,6 +10926,21 @@ function frmAdminBuildJS() {
 
 			// Set fieldsUpdated to 0 to avoid the unsaved changes pop up.
 			frmDom.util.documentOn( 'submit', '.frm_settings_form', () => fieldsUpdated = 0 );
+
+			const manageStyleSettings = document.getElementById( 'manage_styles_settings' );
+			if ( manageStyleSettings ) {
+				manageStyleSettings.addEventListener(
+					'change',
+					event => {
+						const target = event.originalTarget;
```
engahmeds3ed, line 10935:
```suggestion
						const target = event.target;
```
Using target is better in our case and I tested it and it's working properly.
https://developer.mozilla.org/en-US/docs/Web/API/Event/originalTarget

What do u think?

engahmeds3ed (changes requested):
Indeed, that's a brilliant idea.
Tested it on a customer's site with the suggested change and it's working properly.

js/formidable_admin.js
Crabcyborg: @engahmeds3ed Thanks. I don't have any reason to use originalTarget in particular. I just updated this.

engahmeds3ed (approved):

Crabcyborg:
Thanks Ahmed!

PR #2069, opened 2024-10-22
Make field options grid clickable on mobile(iphone)

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5431
**Note**: Go to Formidable > Styles and click on 'Update' on the styler page to generate a new stylesheet used on your pages before testing.

### Test steps
1. Add a couple of checkbox fields to a form.
2. Create a Grid Layout using the steps outlined [here](https://formidableforms.com/knowledgebase/form-layout-and-css-classes/#kb-create-a-grid)
3. Preview the form using BrowserStack or available iPhone 14, 15, 16 within Chrome.
4. Confirm that the checkboxes are selectable when you tap on them.

CleanShot 2024-10-22 at 17 43 59


AbdiTolesa:
> @AbdiTolesa checkboxes are selectable but for example when I change the viewport dimensions to iphone15 this is how checkboxes are aligned. Maybe it's only on my end but could you please check this also? image

@lauramekaj1 I couldn't find that issue testing it again. Is that the case for you only a specific screen size or do you see that for any device size?

lauramekaj1 (approved):
@AbdiTolesa sorry, it was just on my end. It seems to be working as expected. Thank you!

truongwp (approved):
I can't replicate this problem. But the code looks good to me. Thanks @AbdiTolesa!

PR #2075, opened 2024-10-25
Add "NEW" to payment field

Crabcyborg:
We have this "NEW" text in our payments submenu.

Screen Shot 2024-10-25 at 1 58 25 PM

But I was think we might draw more attention to payments as well if we put a "NEW" flag directly beside the Payment field as well.
Screen Shot 2024-10-25 at 1 59 19 PM

I put this behind a check so it doesn't appear if you have a payments table already installed. This would apply to anyone using the PayPal add-on, Authorize.Net, or anyone who has used Stripe already.


Crabcyborg:
@srwells What do you think about this update? We were talking the other day about ways to help convert more people to Stripe. We have the "NEW" item in the sub-menu, but what do you think about showing it here in the field as well?

Thank you!

srwells:
I like it! Yeah, the more we can draw attention to the payment tools, the better.

Crabcyborg:
Thanks Steve!

I'll go ahead and merge this then.

🚀

PR #2082, opened 2024-11-05
center form fix

Liviu-p:


Crabcyborg (approved):
Thank you @Liviu-p!

I was checking for `'center_form'`. I forgot to check for the `$center_form` variable since the `extract` code makes it a bit harder to search for.

This looks good then!

🚀

PR #2083, opened 2024-11-05
Fix test to handle dynamic h1 text on formidableforms.com

lauramekaj1:


lauramekaj1:
@Crabcyborg The tests are now working fine. As you mentioned, the issue was due to a change on [formidableforms.com](http://formidableforms.com/) that was causing the failure. Thank you!

Crabcyborg (approved):
Thank you @lauramekaj1! This looks great!

🚀

PR #2084, opened 2024-11-06
Show hidden field validation message along main error message

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5161


AbdiTolesa:
This should really be in Pro.

PR #2085, opened 2024-11-07
Fix submit button auto order change

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5459

### Test steps
1. Create a form and add a field to it.
2. Then click on the Submit button in the builder page and then click on "Update" button at the top right.
3. When the page reloads, confirm that the Submit button's position/order is not changed.

### Screencasts
**Before**:
CleanShot 2024-11-07 at 15 37 27

**After**
CleanShot 2024-11-07 at 15 38 29

### DIFF
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -516,7 +516,7 @@ class FrmForm {
 		foreach ( $field_cols as $col => $default ) {
 			$default = $default === '' ? $field->{$col} : $default;
 
-			if ( 'field_order' === $col && $field->type === 'submit' && '9999' === $values['field_options'][ $col . '_' . $field->id ] ) {
+			if ( '9999' === $values['field_options'][ $col . '_' . $field->id ] && $field->type === 'submit' && 'field_order' === $col ) {
 				$new_field[ $col ] = $field->field_order;
 			} else {
 				$new_field[ $col ] = isset( $values['field_options'][ $col . '_' . $field->id ] ) ? $values['field_options'][ $col . '_' . $field->id ] : $default;
--- a/classes/models/FrmForm.php
+++ b/classes/models/FrmForm.php
@@ -516,7 +516,7 @@ class FrmForm {
 		foreach ( $field_cols as $col => $default ) {
 			$default = $default === '' ? $field->{$col} : $default;
 
-			$new_value = isset( $values['field_options'][ $col . '_' . $field->id ] ) ? $values['field_options'][ $col . '_' . $field->id ] : $default;
+			$new_value         = isset( $values['field_options'][ $col . '_' . $field->id ] ) ? $values['field_options'][ $col . '_' . $field->id ] : $default;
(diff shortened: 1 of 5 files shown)


lauramekaj1 (approved):
@AbdiTolesa I tested this and verified that the issue has been fixed. Thank you!

classes/models/FrmForm.php
```diff
@@ -516,7 +516,8 @@ private static function prepare_field_update_values( $field, $values, &$new_fiel
 		foreach ( $field_cols as $col => $default ) {
 			$default = $default === '' ? $field->{$col} : $default;
 
-			$new_field[ $col ] = isset( $values['field_options'][ $col . '_' . $field->id ] ) ? $values['field_options'][ $col . '_' . $field->id ] : $default;
+			$new_value = isset( $values['field_options'][ $col . '_' . $field->id ] ) ? $values['field_options'][ $col . '_' . $field->id ] : $default;
+			$new_field[ $col ] = ( '9999' === $new_value && $field->type === 'submit' && 'field_order' === $col ) ? $field->field_order : $new_value;
```
truongwp, line 520: I think we should use `intval( $new_value )` when comparing to `9999`.
We should create a constant in `FrmSubmitHelper` for the `9999` value, maybe `DEFAULT_ORDER`, then we can use it here and in `FrmFormsController::create_submit_button_field()`.

Crabcyborg, line 520: @AbdiTolesa There are errors here coming from phpcs as well. These parentheses really don't add anything to make this more readable, and there's an error for that in the workflow.

classes/models/FrmForm.php
```diff
@@ -516,7 +516,8 @@ private static function prepare_field_update_values( $field, $values, &$new_fiel
 		foreach ( $field_cols as $col => $default ) {
 			$default = $default === '' ? $field->{$col} : $default;
 
-			$new_field[ $col ] = isset( $values['field_options'][ $col . '_' . $field->id ] ) ? $values['field_options'][ $col . '_' . $field->id ] : $default;
+			$new_value         = isset( $values['field_options'][ $col . '_' . $field->id ] ) ? $values['field_options'][ $col . '_' . $field->id ] : $default;
+			$new_field[ $col ] = FrmSubmitHelper::DEFAULT_ORDER === intval( $new_value ) && $field->type === 'submit' && 'field_order' === $col ? $field->field_order : $new_value;
```
Crabcyborg, line 520: Instead of checking this in a loop, could we just check after the loop?

Something like..
```
if ( $field->type === 'submit' && isset( $new_field['field_order'] ) && (int) $new_field['field_order'] === FrmSubmitHelper::DEFAULT_ORDER ) {
    $new_field['field_order'] = $field->field_order;
}
```

Crabcyborg: I made this change. I want this in tomorrow's release, so I made a quick update.

Crabcyborg (approved):

🚀

PR #2086, opened 2024-11-07
Stop showing onboarding inbox item after visiting onboarding page or …

Crabcyborg:
…clicking dismiss

**Related ticket** https://secure.helpscout.net/conversation/2756274044/214547/

**The issue**
Clicking to dismiss the onboarding message does not dismiss it permanently. It re-appears after every page load. This is because it continues to get added when `has_onboarding_been_skipped()` returns true.

There are also issues where the inbox message continues to appear on the onboarding wizard page after you click it.

**The update**
1. Deletes the skipped option when the inbox message is dismissed.
2. Removes the inbox item when the onboarding page is loading.

**To replicate**
You just need to add the `frm_onboarding_skipped` option to the database with a value like "yes".

image

### DIFF
--- a/classes/controllers/FrmInboxController.php
+++ b/classes/controllers/FrmInboxController.php
@@ -60,9 +60,15 @@ class FrmInboxController {
 		if ( ! empty( $key ) ) {
 			$message = new FrmInbox();
 			$message->dismiss( $key );
+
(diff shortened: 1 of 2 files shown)


Crabcyborg:
Thanks Sherv!

PR #2108, opened 2024-11-13
fix textarea default value in multipage forms

engahmeds3ed:
Fixes https://github.com/Strategy11/formidable-pro/issues/5470

We validated the fix in the customer's site and it's working properly.

### DIFF
--- a/classes/models/fields/FrmFieldTextarea.php
+++ b/classes/models/fields/FrmFieldTextarea.php
@@ -94,7 +94,7 @@ class FrmFieldTextarea extends FrmFieldType {
 
 		return '<textarea name="' . esc_attr( $args['field_name'] ) . '" id="' . esc_attr( $args['html_id'] ) . '" ' .
 			$rows . $input_html . '>' .
-			FrmAppHelper::esc_textarea( $this->field['value'] ) .
+			FrmAppHelper::esc_textarea( ! empty( $this->field['value'] ) ? $this->field['value'] : $this->field['default_value'] ) .
 			'</textarea>';
 	}
 }


classes/models/fields/FrmFieldTextarea.php
```diff
@@ -94,7 +94,7 @@ public function front_field_input( $args, $shortcode_atts ) {
 
 		return '<textarea name="' . esc_attr( $args['field_name'] ) . '" id="' . esc_attr( $args['html_id'] ) . '" ' .
 			$rows . $input_html . '>' .
-			FrmAppHelper::esc_textarea( $this->field['value'] ) .
+			FrmAppHelper::esc_textarea( ! empty( $this->field['value'] ) ? $this->field['value'] : $this->field['default_value'] ) .
```
Crabcyborg, line 97: Thanks @engahmeds3ed.

Unfortunately these sorts of issues are a bit more complex.

This code is also used if I'm editing an entry for example with no value. It will show the default even though I had previously put nothing. There's other code that sets `$this->field['value']` to the default, and other code that clears these values when conditionally hidden.

We have a few similar issues that we need to figure out still.
- https://github.com/Strategy11/formidable-pro/issues/4081
- https://github.com/Strategy11/formidable-pro/issues/4110

We generally fix these sorts of issues a bit differently.

The `! empty` also wouldn't be totally safe since it would break `0` values.

I plan to try looking further into this.

Crabcyborg:
Closing this for now as I don't think we'll be going in this direction. I think the fix will likely be in Pro code since this issue isn't related to Lite either.

PR #2115, opened 2024-11-18
Remove buttons settings tab

truongwp:
Related PR: https://github.com/Strategy11/formidable-pro/pull/5479

In the related PR, we won't use `frm_add_form_style_tab_options` and `frm_add_form_button_options` hooks to print hidden inputs anymore. If these hooks are still used by custom code, we will keep the Buttons settings tab, and show the deprecation notices. If these hooks aren't used, we remove the Buttons settings tab completely.

### DIFF
--- a/classes/views/frm-forms/settings-buttons.php
+++ b/classes/views/frm-forms/settings-buttons.php
@@ -2,11 +2,20 @@
 if ( ! defined( 'ABSPATH' ) ) {
 	die( 'You are not allowed to call this page directly.' );
 }
+
+// Do not deprecate if the Pro version still use these hooks.
+$should_deprecate_hook = class_exists( 'FrmProDb' ) && version_compare( FrmProDb::$plug_version, '6.16.2' ) >= 0;
 ?>
 <input type="hidden" name="options[custom_style]" value="<?php echo esc_attr( $values['custom_style'] ); ?>" />
 
 <table class="form-table">
-	<?php do_action( 'frm_add_form_style_tab_options', $values ); ?>
+	<?php
+	if ( $should_deprecate_hook ) {
+		do_action_deprecated( 'frm_add_form_style_tab_options', compact( 'values' ), '6.16.2' );
+	} else {
+		do_action( 'frm_add_form_style_tab_options', $values );
+	}
+	?>
 	<tr>
 		<td colspan="2">
 			<h3><?php esc_html_e( 'Buttons', 'formidable' ); ?></h3>
@@ -19,5 +28,11 @@ if ( ! defined( 'ABSPATH' ) ) {
 			</div>
 		</td>
 	</tr>
-	<?php do_action( 'frm_add_form_button_options', $values ); ?>
+	<?php
+	if ( $should_deprecate_hook ) {
+		do_action_deprecated( 'frm_add_form_button_options', compact( 'values' ), '6.16.2' );
+	} else {
+		do_action( 'frm_add_form_button_options', $values );
+	}
+	?>
 </table>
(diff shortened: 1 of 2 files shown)


classes/views/frm-forms/settings-buttons.php
```diff
@@ -2,11 +2,20 @@
 if ( ! defined( 'ABSPATH' ) ) {
 	die( 'You are not allowed to call this page directly.' );
 }
+
+// Do not deprecate if the Pro version still use these hooks.
+$should_deprecate_hook = class_exists( 'FrmProDb' ) && version_compare( FrmProDb::$plug_version, '6.16.2' ) >= 0;
```
Crabcyborg, line 7: I think we would also want to deprecate this right away if Pro isn't available.

I'd change the check to `! class_exists( 'FrmProDb' ) ||  version_compare...`.

truongwp: I updated it. Thank you.

Crabcyborg (approved):
Thanks @truongwp!

🚀

Crabcyborg:
It seems something about this update causes the e2e tests to fail.

Crabcyborg:
@lauramekaj1 Maybe you can help with this failing unit test issue we're having here.

This update really just removes the "Buttons" section in form settings. Is there a test that is checking for this somewhere that needs to be updated?

Thank you!

lauramekaj1:
> @lauramekaj1 Maybe you can help with this failing unit test issue we're having here.
> 
> 
> 
> This update really just removes the "Buttons" section in form settings. Is there a test that is checking for this somewhere that needs to be updated?
> 
> 
> 
> Thank you!

@Crabcyborg yes sure I will look into this!

lauramekaj1:
@Crabcyborg @truongwp When I switch to the `remove-buttons-settings` branch, I'm unable to click on the sections in the form settings. In comparison, everything works fine in the master branch. I've attached a screen recording for reference: https://www.loom.com/share/dc1286c9152c482d839d33fdb8ebacec?sid=41c58beb-a01b-4000-ab2a-acd0f52d85f7 

Note: If you have Formidable Forms Pro activated, then deactivate and you should be able to reproduce this issue. Let me know if you need further information! 
Thank you!

Crabcyborg:
Thank you @lauramekaj1!

So there's a real bug here.

@truongwp I'm seeing this fatal error in my logs
```
PHP Fatal error:  Uncaught Error: Unknown named parameter $values
```

classes/views/frm-forms/settings-buttons.php
```diff
@@ -2,11 +2,20 @@
...
+// Do not deprecate if the Pro version still use these hooks.
+$should_deprecate_hook = ! class_exists( 'FrmProDb' ) || version_compare( FrmProDb::$plug_version, '6.16.2' ) >= 0;
 ?>
 <input type="hidden" name="options[custom_style]" value="<?php echo esc_attr( $values['custom_style'] ); ?>" />
 
 <table class="form-table">
-	<?php do_action( 'frm_add_form_style_tab_options', $values ); ?>
+	<?php
+	if ( $should_deprecate_hook ) {
+		do_action_deprecated( 'frm_add_form_style_tab_options', compact( 'values' ), '6.16.2' );
```
Crabcyborg, line 14: It looks like the issue, as far as I can tell, is because of this `compact` code. It looks like we want to just pass `array( $values )`, with `0` as the key instead.

Crabcyborg:
@truongwp I don't believe that passing a `'values'` string works. I think we need to use the `$values` variable directly.

truongwp:
@Crabcyborg Oh sorry, I fixed it.

PR #2116, opened 2024-11-19
Update FrmAppHelper::is_form_builder_page to return true only on the form builder

AbdiTolesa:
This updates makes sure `FrmAppHelper::is_form_builder_page` returns true only on the form builder page.

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -314,7 +314,7 @@ class FrmAppHelper {
 	 * @return bool
 	 */
 	public static function is_form_builder_page( $check_for_settings = true ) {
-		$action = self::simple_get( 'frm_action', 'sanitize_title' );
+		$action        = self::simple_get( 'frm_action', 'sanitize_title' );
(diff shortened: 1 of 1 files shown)


classes/helpers/FrmAppHelper.php
```diff
@@ -307,12 +307,19 @@ public static function pro_is_connected() {
...
+	 *
+	 * @param bool $check_for_settings
 	 *
 	 * @return bool
 	 */
-	public static function is_form_builder_page() {
+	public static function is_form_builder_page( $check_for_settings = true ) {
 		$action = self::simple_get( 'frm_action', 'sanitize_title' );
-		return self::is_admin_page( 'formidable' ) && ( $action === 'edit' || $action === 'settings' || $action === 'duplicate' );
+		$check_actions = array( 'edit', 'duplicate' );
```
Crabcyborg, line 318: @AbdiTolesa PHPCS is failing here. The equals alignment for the `$action` line should match the `$check_actions` line.

Crabcyborg: I updated this. I'm planning on a release today.

PR #2117, opened 2024-11-19
Add redirect delay settings

truongwp:
Handoff: https://linear.app/strategy11/document/feature-handoff-479d8bd5186c


lauramekaj1:
Hi @truongwp, I am testing this feature, and this is the issue I encountered:
- Negative values are accepted in the input field for seconds. However, after a page update, the input field value is automatically set to 8 (https://imgur.com/GwWoa0K).

truongwp:
@lauramekaj1 That's intentional. If the Delay time isn't correct, it will be set to `8`.

truongwp:
I added some code to not allow negative value for that input.

lauramekaj1 (approved):
@truongwp I tested it and it's working as expected. Thank you!

Crabcyborg (approved):
I haven't tested this but it looks good to me!

I'll wait to merge this as I'm not sure if we want to delay it for marketing.

Thanks Truong and Laura for working on this!

PR #2118, opened 2024-11-19
Pro issue 5418 / Use WordPress date and time settings for entry timestamp in edit/view entry sidebar

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/5418

Now I'm using our more commonly used `get_formatted_time` function that adds the time data, and some custom logic for the date format (I change long month names to the short month name so if you have it set to show "November", it will show "Nov" instead to save space).

This generally uses the WordPress setting, and not our global setting in Pro, as the Pro settings are all formats like `Y/m/d` and `Y-m-d`, none of which look as good as the more verbose format options that you can use in the WordPress date setting.

The `@` is now `at` by default, but I think that's okay. It makes it more consistent with our other dates with times which is more important than the possible space saved.

**Using the following settings** _These are set at /wp-admin/options-general.php_
Screen Shot 2024-11-19 at 12 36 36 PM

**Before**
Screen Shot 2024-11-19 at 12 35 45 PM

**After**
Screen Shot 2024-11-19 at 12 35 20 PM

### DIFF
--- a/classes/helpers/FrmAppHelper.php
+++ b/classes/helpers/FrmAppHelper.php
@@ -2695,7 +2695,14 @@ class FrmAppHelper {
 		return $formatted;
 	}
 
-	private static function add_time_to_date( $time_format, $date ) {
+	/**
+	 * @since x.x This function was made public.
+	 *
+	 * @param string $time_format
+	 * @param string $date
+	 * @return string
+	 */
+	public static function add_time_to_date( $time_format, $date ) {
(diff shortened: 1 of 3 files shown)


lauramekaj1 (approved):
@Crabcyborg I tested it and it's working as expected. Thank you!

Crabcyborg:
Thank you Laura!

🚀

PR #2133, opened 2024-11-21
New duplicate check

Crabcyborg:
**Related ticket** https://secure.helpscout.net/conversation/2765614867/215349/

**The issue**
Reopening a page on an iPhone re-submits the POST request. Our duplicate check defaults to 60 seconds, but it's likely someone will open a page again hours later, maybe even days later.

**The solution**
I'm adding a unique string that gets added right before the form is submitted.

If this string is detected, it is checked in the duplicate check, and saved to entry meta under field 0, serialized as an array like `a:1:{s:9:"unique_id";s:9:"a34d2ypgn";}`.

When this exact item meta row is detected, during the duplicate check, it will return true immediately if there is a match.

The refresh events don't re-trigger the JavaScript, so all of these requests will send the same unique string.

**Pre-release**
[formidable-6.16.3b.zip](https://github.com/user-attachments/files/17909053/formidable-6.16.3b.zip)

**To opt out** _In case the database query causes performance issues_
```
add_filter( 'frm_extend_duplicate_entry_time_on_unique_id_match', '__return_false' );
```

_**This has been confirmed by the customer**_

### DIFF
--- a/classes/models/FrmEntry.php
+++ b/classes/models/FrmEntry.php
@@ -51,8 +51,7 @@ class FrmEntry {
 			return false;
 		}
 
-		$is_duplicate = self::maybe_check_for_unique_id_match( $new_values['created_at'] );
-		if ( $is_duplicate ) {
+		if ( self::maybe_check_for_unique_id_match( $new_values['created_at'] ) ) {
 			return true;
 		}
 
--- a/classes/models/FrmEntry.php
+++ b/classes/models/FrmEntry.php
@@ -47,6 +47,13 @@ class FrmEntry {
 	public static function is_duplicate( $new_values, $values ) {
 		$duplicate_entry_time = apply_filters( 'frm_time_to_check_duplicates', 60, $new_values );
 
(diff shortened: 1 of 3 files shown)


Achilles28:
@Crabcyborg The customer has replied and confirmed that the resubmission issue in Safari has been fixed with the latest patch.

Customer tested under the following cuircumstances:

- I submitted a form with a confirmation screen via Safari on an iPhone.
- I immediately tried to refresh the confirmation page and got the resubmit warning. When I clicked OK, I got the "already submitted" message (properly displayed).
- I closed Safari and opened it and got the "already submitted" message (partially displayed).
- I closed Safari and waited 5 minutes, then opened it and got the "already submitted" message (partially displayed).
- I closed Safari and waited overnight, then opened it and got the "already submitted" message (partially displayed).

The **partially displayed** in the tests refer to how the message was displayed which is illustrated here: https://cln.sh/vHdqjNZ6

Crabcyborg:
Thank you @Achilles28!

PR #2142, opened 2024-11-27
Add more global settings to usage tracking

Crabcyborg:
While working on setting up some metabase examples, I noticed a few things we weren't tracking.

- Captcha Type
- Whether Summary Emails is enabled
- Whether the Customer Header IPs setting is enabled

@truongwp This probably requires some changes in the other repo as well. I could use your help getting that updated.


truongwp (approved):
This looks good. We don't need to update the usage repo.
Thanks @Crabcyborg!

Crabcyborg:
Awesome. Thanks Truong!

PR #2143, opened 2024-11-28
Unschedule send usage event if it is disabled

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/4291

To test, you can use the WP Crontrol to see the list of scheduled events, turn on and off the Send usage feature in the Miscellaneous tab in the Global settings and check if the `formidable_send_usage` event is added or removed.


Crabcyborg (approved):
Thank you @truongwp!

🚀

PR #2146, opened 2024-11-29
Stop using font icon in section field icons in form builder

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5504
CleanShot 2024-11-29 at 14 37 02

### DIFF
--- a/css/frm_admin.css
+++ b/css/frm_admin.css
@@ -6577,12 +6577,12 @@ li.selected .divider_section_only:before {
 }
 
 .edit_field_type_divider .frm-divider-icon.frmsvg {
-    padding: 2px;
-    position: absolute;
-    left: -12px;
-    color: #fff;
-    background-color: var(--primary-700);
-    border-radius: 50%;
+	padding: 2px;
+	position: absolute;
+	left: -12px;
+	color: #fff;
+	background-color: var(--primary-700);
+	border-radius: 50%;
 }
 
 .edit_field_type_divider .start_divider .frm-divider-icon.frmsvg {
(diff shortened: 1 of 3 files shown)


Crabcyborg (changes requested):
@AbdiTolesa It looks like this update causes a lot of issues with other icons inside of sections. The move field icon, the more options dropdown, the option in that dropdown, etc.

Screen Shot 2024-11-29 at 9 07 20 AM
Screen Shot 2024-11-29 at 9 07 29 AM

css/frm_admin.css
```diff
@@ -6573,30 +6571,39 @@ li.selected .divider_section_only:before {
...
 	font-family: 's11-fp';
 	content: '\f363';
 }
 
-.edit_field_type_divider:after {
-	width: 5px;
-	height: 5px;
-	left: -4px;
+.edit_field_type_divider .frm-divider-icon.frmsvg {
+    padding: 2px;
```
Crabcyborg, line 6580: @AbdiTolesa It looks like you're using spaces here. The phpcs workflow has some errors.

Can you get it passing?

Screen Shot 2024-12-02 at 10 04 48 AM

Crabcyborg (approved):
Thanks @AbdiTolesa!

🚀

Crabcyborg:
@AbdiTolesa It looks like I merged a bit early.

I forgot that this update would also impact repeaters. Can you get it looking like it did before?

**Before this update**
Screen Shot 2024-12-03 at 9 26 52 AM

**After this update**
Screen Shot 2024-12-03 at 9 27 54 AM

Thank you!

PR #2167, opened 2024-12-05
Lite update for applications dropdown

AbdiTolesa:
Related issue: https://github.com/Strategy11/formidable-pro/issues/5463


lauramekaj1:
Hi @AbdiTolesa, 
I don't see the list of application templates after switching to this branch. Could you please check as well?
image

AbdiTolesa:
> I don't see the list of application templates after switching to this branch. Could you please check as well?

Thank you for catching that @lauramekaj1! I just fixed that issue but I still see the Cypress workflow fail, is that something related to this update?

lauramekaj1:
> Thank you for catching that @lauramekaj1! I just fixed that issue but I still see the Cypress workflow fail, is that something related to this update?

@AbdiTolesa The failing Cypress tests are actually not related to this update. The tests are now fixed in this [branch](https://github.com/Strategy11/formidable-forms/pull/2185)!

lauramekaj1 (approved):
@AbdiTolesa I tested this and it's working now as expected. Thank you!

classes/views/frm-forms/settings-advanced.php
```diff
@@ -44,6 +44,13 @@
 	</p>
 
 <?php
+/**
+ * Fires after the form description field.
+ *
+ * @since x.x
+ */
+do_action( 'frm_settings_after_form_description' );
```
truongwp, line 52: Where is this hook used?

AbdiTolesa: In the other related Pro update: https://github.com/Strategy11/formidable-pro/pull/5525/files#diff-cb53198069014818d2202e56c5be8ba6bdd1ed222f31ecceadae456c60f2dec0R516

truongwp: Thank you. I didn't notice it when I sent this comment.

classes/controllers/FrmApplicationsController.php
```diff
@@ -69,11 +69,17 @@ private static function get_view_path() {
 	/**
 	 * Get information about applications via AJAX action.
 	 *
-	 * @return void
+	 * @return array|void
 	 */
-	public static function get_applications_data() {
+	public static function get_applications_data( $is_ajax_request = true ) {
```
truongwp, line 74: Please add `@since` tag to this method.

classes/views/frm-forms/settings-advanced.php
truongwp: @AbdiTolesa Should we add `$values` to the hook?

Crabcyborg (approved):
Thanks @AbdiTolesa!

This looks good to me.

I removed the version change in this code, and it would have needed to change in Pro anyway since it was checking for 6.16.3 and we're already on v6.17.1.

🚀

PR #2168, opened 2024-12-05
Include license type in usage data

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/5519

### DIFF
--- a/classes/models/FrmUsage.php
+++ b/classes/models/FrmUsage.php
@@ -101,6 +101,10 @@ class FrmUsage {
 			'subscriptions'     => $this->payments( 'frm_subscriptions' ),
 		);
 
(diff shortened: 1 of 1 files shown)


Crabcyborg:
@truongwp Would it make sense to just put this in Pro instead?

We can use the `frm_usage_snapshot` filter.

truongwp:
> @truongwp Would it make sense to just put this in Pro instead?
> 
> We can use the `frm_usage_snapshot` filter.

That makes sense. I changed here because I don't want to create an `add_filter()` call.

Crabcyborg (approved):
Thanks @truongwp!

If it's here to avoid a hook, I think that's fine.

I'll merge this as-is.

PR #2187, opened 2024-12-19
Support custom options for autocomplete dropdown

truongwp:


Crabcyborg:
@truongwp It might be nice to get this merged if it's required for another update, so a core release isn't blocking the other update.

It looks like there are some workflow issues at the moment.

Do you think you could get this PR ready to get merged soon?

Thank you!

truongwp:
@Crabcyborg I fixed all PHPCS errors. The PHP CS Fixer doesn't give me the line of error so I can't fix it. The Psalm errors should be fixed in another PR.

classes/views/shared/views-info.php
```diff
@@ -5,6 +5,28 @@
 ?>
 <div class="frm_wrap frm-views-info-page">
 	<?php
+	if ( method_exists( 'FrmAppHelper', 'maybe_autocomplete_options' ) ) {
```
Crabcyborg, line 8: @truongwp Is this here for testing?

It's showing a funny dropdown on the views page.

Screen Shot 2025-03-07 at 5 45 39 PM

classes/helpers/FrmAppHelper.php
```diff
@@ -1707,6 +1707,101 @@ class="frm_autocomplete_value_input"
...
+			<?php
+		}//end if
+	}
+
+	/**
+	 * Gets dropdown value and label from autodropdown option.
+	 *
+	 * @since x.x
+	 *
+	 * @param string|array $option Autocomplete option.
```
Crabcyborg, line 1788: The php cs fixer workflow doesn't line this line. There's a rule that enforces that types are in alphabetical order (with the exception of always requiring `null` as the last item).

So this needs to be `array|string` instead.

truongwp: Thank you. I updated it.

classes/helpers/FrmAppHelper.php
```diff
@@ -1682,7 +1682,7 @@ public static function maybe_autocomplete_pages_options( $args ) {
 
 		$pages_count = wp_count_posts( $args['post_type'] );
```
Crabcyborg, line 1683: Do we want to skip this call for the Google Sheets dropdown? Is that possible?

truongwp: Now GoogleSpreadsheet add-on doesn't use this method anymore.

truongwp: It uses `maybe_autocomplete_options()` method. Should I rename it to avoid confusion?

Crabcyborg: I guess I'm wondering why this change was made then. I keep looking at this function because of the change here.

Crabcyborg: I feel like we might want to call `wp_pages_dropdown` if it isn't set instead of the autocomplete.

What do you think about changing this to `! isset( $pages_count->publish ) || $pages_count->publish <= 50`?

truongwp: You're right. We should call the normal dropdown in that case. I updated the code.

Crabcyborg (approved):
Thank you @truongwp!

This works great!

🚀

PR #2195, opened 2025-01-02
Fix centered submit button bottom margin value with modern dark theme

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5557
Quoting Mike's workaround the issue implemented in this PR:

> We should check for this, and use the bottom value.
> 
> So if the submit_margin has 2 units (vertical, horizontal), we want to use the first value (10px in this case). If there are 4 (top, right, bottom, left), we'll want to use the bottom value (the 3rd one).

### DIFF
--- a/classes/helpers/FrmStylesHelper.php
+++ b/classes/helpers/FrmStylesHelper.php
@@ -944,4 +944,23 @@ class FrmStylesHelper {
 			? FrmProAppHelper::use_chosen_js()
 			: true;
 	}
+
+	/**
+	 * Returns the bottom part from a margin/padding value.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $value The margin/padding value.
+	 * @return string
+	 */
+	public static function get_bottom_value( $value ) {
+		if ( ! $value ) {
+			return $value;
+		}
+		$parts = explode( ' ', $value );
+		if ( count( $parts ) < 3 ) {
+			return $parts[0];
+		}
+		return $parts[2];
+	}
 }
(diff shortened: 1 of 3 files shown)


css/_single_theme.css.php
```diff
@@ -254,7 +254,19 @@
 }
 
 .<?php echo esc_html( $style_class ); ?>.frm_center_submit .frm_submit .frm_ajax_loading{
-	margin-bottom:<?php echo esc_html( $submit_margin ); ?>;
+		<?php
+		if ( $submit_margin ) {
+			$_submit_margin = explode( ' ', $submit_margin );
```
Crabcyborg, line 259: Thanks @AbdiTolesa!

Can we use another variable name? I try to avoid this style of variable formatting. I would opt for something like `$submit_margin_parts` instead. It might also be good to treat this file like a view and avoid adding more logic here.

I think we could add something to `FrmStylesHelper` to call so we can avoid adding extra variables to this scope that we only need for one thing.

Crabcyborg: If it's a new function too, it would allow us to use more simple variable names since we aren't likely to run into naming conflicts.

In this case, we could go with a more simple variable name like `$parts`.

classes/helpers/FrmStylesHelper.php
```diff
@@ -944,4 +944,23 @@ public static function use_chosen_js() {
...
+
+	/**
+	 * Returns the bottom part from a margin/padding value.
+	 *
+	 * @since x.x
+	 *
+	 * @param string $value The margin/padding value.
+	 * @return string
+	 */
+	public static function get_bottom_value( $value ) {
```
Crabcyborg, line 956: @AbdiTolesa Could you add a unit test for this new function? It should be pretty quick and it would be nice to have covered.

Crabcyborg (approved):
Thanks @AbdiTolesa!

🚀

PR #2206, opened 2025-01-08
Improve email validation

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5527

This update just adds an extra check to valid email values so that if there is a `.` at the end of the email local part (just before `@`), it would be treated as an invalid value.

### Test steps
1. Create a form and add an email field to it.
2. Preview the form and fill the email field with a value that has `.` just before the `@` symbol.
3. Try submitting the form and confirm that an error is thrown indicating an invalid email value.

### DIFF
--- a/classes/models/fields/FrmFieldEmail.php
+++ b/classes/models/fields/FrmFieldEmail.php
@@ -48,7 +48,7 @@ class FrmFieldEmail extends FrmFieldType {
 			return array();
 		}
 		$errors = array();
-		if ( ! is_email( $args['value'] ) || strpos( $args['value'], '.@' ) !== false ) {
+		if ( strpos( $args['value'], '.@' ) !== false || ! is_email( $args['value'] ) ) {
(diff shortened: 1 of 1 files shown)


classes/models/fields/FrmFieldEmail.php
```diff
@@ -44,11 +44,20 @@ protected function field_settings_for_type() {
...
 	public function validate( $args ) {
-		$errors = array();
-		if ( $args['value'] != '' && ! is_email( $args['value'] ) ) {
+		if ( ! $args['value'] ) {
+			return array();
+		}
+		$errors   = array();
+		$is_email = false;
+		if ( is_email( $args['value'] ) ) {
+			$parts = explode( '@', $args['value'] );
```
truongwp, line 53: I think we just need to check if the value contains `.@`.

AbdiTolesa: Thanks Truong! That sounds simpler!

truongwp (approved):

Crabcyborg (approved):

I think this looks good.

🚀

PR #2208, opened 2025-01-08
Fix auto width issue with different label positions

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5561

### DIFF
--- a/css/_single_theme.css.php
+++ b/css/_single_theme.css.php
@@ -100,6 +100,16 @@ if ( '' === $field_height || 'auto' === $field_height ) {
 }
 <?php } ?>
 
+.<?php echo esc_html( $style_class ); ?> .frm_left_container select.auto_width,
+.<?php echo esc_html( $style_class ); ?> .frm_right_container select.auto_width {
+	width: max-content<?php echo esc_html( $important ); ?>;
+}
+
+.<?php echo esc_html( $style_class ); ?> .frm_right_container .frm_primary_label,
+.<?php echo esc_html( $style_class ); ?> .frm_right_container select.auto_width {
+	margin-left: auto<?php echo esc_html( $important ); ?>;
+}
+
 .<?php echo esc_html( $style_class ); ?> .frm_form_field.frm_left_container{
 	grid-template-columns: <?php echo esc_html( $width ); ?> auto;
 }


truongwp (approved):

css/_single_theme.css.php
```diff
@@ -100,6 +100,16 @@
 }
 <?php } ?>
 
+.<?php echo esc_html( $style_class ); ?> .frm_left_container select.auto_width,
```
Crabcyborg, line 103: @AbdiTolesa Does this work if we put it in `custom_theme.css.php` instead? Do we really need this printed for every single style?

If we move it to `custom_theme.css.php` it will take up less file size if there are multiple styles installed.

AbdiTolesa: Yeah it works and I believe we better put it that file.

css/custom_theme.css.php
```diff
@@ -1445,6 +1445,16 @@
 	padding-top:4px;
 }
 
+.with_frm_style .frm_left_container select.auto_width,
```
Crabcyborg, line 1448: I'm not sure if this would have any nesting issues, but are we able to narrow this scope and use `.with_frm_style .frm_left_container > select.auto_width,`, etc, adding the `>` to the selector so it only impacts direct. children?

css/custom_theme.css.php
```diff
@@ -1445,6 +1445,16 @@
 	padding-top:4px;
 }
 
+.with_frm_style .frm_left_container > select.auto_width,
+.with_frm_style .frm_right_container > select.auto_width {
+	width: max-content<?php echo esc_html( $important ); ?>;
+}
+
+.with_frm_style .frm_right_container .frm_primary_label,
+.with_frm_style .frm_right_container select.auto_width {
```
Crabcyborg, line 1454: Can we use `>` on these as well?

Crabcyborg (approved):
This looks good now. I haven't tested it though.

I'll merge this once Laura has tested and approves it.

🚀

lauramekaj1 (approved):
@AbdiTolesa I tested this and verified as fixed. Thank you!

Crabcyborg:
Thank you Laura!

🙌

PR #2209, opened 2025-01-09
Prevent duplicating field from breaking field group limit

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5223

### Screen record
CleanShot 2025-01-09 at 14 43 37

### Test steps
1. Create a form and create a field group in the form which is limited to a maximum of 6 fields.
2. Try duplicating one of the fields in the field group.
3. Confirm that the duplication fails and a modal with a message about the error shows instead.


lauramekaj1 (approved):
@AbdiTolesa I tested this and it's working as expected. Thank you!

js/formidable_admin.js
```diff
@@ -2101,13 +2101,34 @@ function frmAdminBuildJS() {
...
+	function canDuplicateField( field ) {
+		if ( field.classList.contains( 'frm-page-collapsed' ) ) {
+			return false;
+		}
+		const fieldGroup = field.closest( 'li.frm_field_box:not(.form-field)' );
+		if ( ! fieldGroup ) {
+			return true;
+		}
+		const fieldsInGroup = fieldGroup.querySelectorAll( 'li.form-field' ).length;
+		return fieldsInGroup < 6;
```
truongwp, line 2121: We should create a constant instead of using `6` here.

js/formidable_admin.js
```diff
@@ -2101,13 +2101,34 @@ function frmAdminBuildJS() {
...
+
 	function duplicateField() {
 		let $field, fieldId, children, newRowId, fieldOrder;
 
 		$field = jQuery( this ).closest( 'li.form-field' );
 
-		if ( $field.hasClass( 'frm-page-collapsed' ) ) {
-			return false;
+		if ( ! canDuplicateField( $field[0] ) ) {
+			infoModal( __( 'You can only have a maximum of 6 fields in a field group. Delete or move out a field from the group and try again.', 'formidable' ) );
```
truongwp, line 2130: We should use placeholder for the number `6`, so if we change this number in the future, this string won't need to be updated.

js/formidable_admin.js
```diff
@@ -2101,13 +2101,38 @@ function frmAdminBuildJS() {
...
+
 	function duplicateField() {
 		let $field, fieldId, children, newRowId, fieldOrder;
+		const maxFieldsInGroup = 6;
 
 		$field = jQuery( this ).closest( 'li.form-field' );
 
-		if ( $field.hasClass( 'frm-page-collapsed' ) ) {
-			return false;
+		if ( ! canDuplicateField( $field[0], maxFieldsInGroup ) ) {
```
Crabcyborg, line 2132: I typically try to use `.get( 0 )`. I feel like this makes it more clear that it's a jQuery object, since this way it's hard to tell apart from an array.

Crabcyborg:
@AbdiTolesa It looks like there might be some sort of bug with the duplicating. Right now it's trying to show the group on multiple lines if I create a new group, then duplicate a field.

https://github.com/user-attachments/assets/1739ee31-5a05-48d4-a00a-3826796b4fdd

AbdiTolesa:
> @AbdiTolesa It looks like there might be some sort of bug with the duplicating. Right now it's trying to show the group on multiple lines if I create a new group, then duplicate a field.

@Crabcyborg I looked into it and found out that the layout classes for the field that is duplicated were not saved in the database and PHP is not aware of that when it copies the field. That's why the copy field's layout is not in sync but that is fixed now.

Crabcyborg:
Thank you @AbdiTolesa!

This is looking great.

I just found one more issue though. It looks like I cannot duplicate a group of fields with 6 fields in it. I made a short recording. You can see how I can duplicate a group with 5 fields just fine, and it creates a new group with 5 fields. This 6 field check is preventing me from duplicating a group that has 6 fields, which should work fine. I should just have 2 groups with 6 fields in each.

https://github.com/user-attachments/assets/22c69d50-ddb9-4b69-9ac7-9d15a673a69c

Crabcyborg (approved):

I'm seeing some other issues with sections, but it's nothing new. I'll log another issue shortly for that.

For now, I'm going to merge this.

🚀

PR #2211, opened 2025-01-09
Avoid selecting text when multiselecting fields

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5224

This is just an implementation improvement than a feature to test. We have been unselecting selected text after fields are multiselected but we are now preventing text selection at the first place.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -4411,9 +4411,6 @@ function frmAdminBuildJS() {
 						}
 					}
 				);
-
-				// when holding shift and clicking, text gets selected. unselect it.
-				document.getSelection().removeAllRanges();
 			}
 		} else {
 			// not multi-selecting
@@ -10706,6 +10703,9 @@ function frmAdminBuildJS() {
 			handleShowPasswordLiveUpdate();
 			document.addEventListener( 'scroll', updateShortcodesPopupPosition, true );
 			document.addEventListener( 'change', handleBuilderChangeEvent );
+			document.onselectstart = function( e ) {
+				return e.target.closest( '.frm_form_builder' );
+			};
 		},
 
 		settingsInit: function() {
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -10706,6 +10703,9 @@ function frmAdminBuildJS() {
 			handleShowPasswordLiveUpdate();
 			document.addEventListener( 'scroll', updateShortcodesPopupPosition, true );
 			document.addEventListener( 'change', handleBuilderChangeEvent );
+			document.onselectstart = function( e ) {
```
Crabcyborg, line 10706: @AbdiTolesa Instead of setting this globally, can we use `addEventListener` so we're not at risk of overwriting this function at all (if it were used in WP or in the theme for example).

https://developer.mozilla.org/en-US/docs/Web/API/Node/selectstart_event

Crabcyborg (commented):
Thanks @AbdiTolesa.

It looks like this update is working in Chrome but now the text is getting selected in Firefox and not getting unselected (since the other code was also removed).

Could you test it in FIrefox as well?

Crabcyborg (approved):
Thank you @AbdiTolesa and @truongwp!

🚀

Crabcyborg:
Actually, I think this was working, but I'm still seeing the same issue sometimes.

🫤

AbdiTolesa:
> Actually, I thought this was working, but I'm still seeing the same issue sometimes.
> 
> 🫤

I tried hooking into `mousedown` and the result seems more reliable now.

Crabcyborg (approved):

I've tagged Laura for QA as well. Once she's approved it, I'll merge this.

🚀

lauramekaj1:
@AbdiTolesa I tested this and it's working as expected. Thank you!

lauramekaj1 (approved):

PR #2213, opened 2025-01-10
fix failling unit test in php 7.4

Liviu-p:
Fixes a failing unit test in php 7.4: https://github.com/Strategy11/formidable-pro/actions/runs/12711625531/job/35435346578?pr=5544. I noticed it while running unit tests in this PR: https://github.com/Strategy11/formidable-pro/pull/5544

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -2797,7 +2797,7 @@ class FrmFormsController {
 
 				ob_start();
 				self::show_lone_success_message( $args );
-				$response_data['content'] = ob_get_clean();
+				$response_data['content'] .= ob_get_clean();
 			}//end if
 
 			$response_data['fallbackMsg'] = self::get_redirect_fallback_message( $args['success_url'], $args );
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -2794,7 +2794,7 @@ class FrmFormsController {
 
 				ob_start();
 				self::show_lone_success_message( $args );
-				$response_data['content'] .= ob_get_clean();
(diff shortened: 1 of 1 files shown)


Crabcyborg:
@truongwp I think this is related to thew new redirect delay updates.

Can you look into this? Is this fix correct?

Liviu-p:
> Thanks @Liviu-p. This fixes the problem. But I believe [this line](https://github.com/Strategy11/formidable-forms/pull/2213/files#diff-1b880a2d11e84622570383ea0503da990c04c1ccbffe4f5f81ab21d68fa00280R2777) will cause the PHP notice too. And if redirect delay and open in new tab are both enabled, this will break the message.
> 
> We can fix this be adding `'content' => ''` to this line https://github.com/Strategy11/formidable-forms/pull/2213/files#diff-1b880a2d11e84622570383ea0503da990c04c1ccbffe4f5f81ab21d68fa00280R2773

Hey @truongwp, I've just updated it

classes/controllers/FrmFormsController.php
```diff
@@ -2794,7 +2797,7 @@ private static function get_ajax_redirect_response_data( $args ) {
 
 				ob_start();
 				self::show_lone_success_message( $args );
-				$response_data['content'] .= ob_get_clean();
+				$response_data['content'] = ob_get_clean();
```
truongwp, line 2800: We should keep the old code here because this will remove the content when redirect delay is enabled.

Crabcyborg: I just updated this.

Crabcyborg (approved):
I haven't confirmed this yet, but this looks good to me.

I applied the fix for the comment Truong left as well.

🚀

PR #2215, opened 2025-01-13
QoL Release - Lite Version

shervElmi:
Provides a series of small fixes to improve QoL when making forms.

### Pro PR:
https://github.com/Strategy11/formidable-pro/pull/5575

### QA URL:
https://qa.formidableforms.com/sherv11/wp-admin/admin.php?page=formidable&frm_action=edit&id=16

### How to test
See the **[Feature Brief document](https://linear.app/strategy11/document/feature-brief-ae6efa195695)** for the task in Linear. You can find the items we improved there with full descriptions.


classes/controllers/FrmHooksController.php
```diff
@@ -127,6 +127,7 @@ public static function load_admin_hooks() {
 		add_filter( 'admin_footer_text', 'FrmAppController::set_footer_text' );
 		add_action( 'admin_footer', 'FrmAppController::add_admin_footer_links' );
 		add_action( 'current_screen', 'FrmAppController::filter_admin_notices' );
+		add_action( 'current_screen', 'FrmAppController::remember_custom_sort', 0 );
```
Crabcyborg, line 130: @shervElmi To avoid too many hooks, is it possible to just call one function on the `current_screen` hook and have that function call both `FrmAppController::filter_admin_notices` and `FrmAppController::remember_custom_sort`?

The added benefit, is that we should be able to make `FrmAppController::remember_custom_sort` private as well.

js/formidable_admin.js
```diff
@@ -10375,7 +10376,38 @@ function frmAdminBuildJS() {
...
+	 */
+	function focusNextSingleOptionInput( currentInput ) {
+		const optionsList = currentInput.closest( '.frm_single_option' ).parentElement;
+		const inputs = optionsList.querySelectorAll( '.frm_single_option input[name^="field_options[" ], .frm_single_option input[name^="rows_"]' );
+		const inputsArray = Array.from( inputs );
+
+		// Find the index of the currently focused input
+		const currentIndex = inputsArray.indexOf( currentInput );
+
+		if ( currentIndex >= 0 ) {
```
Crabcyborg, line 10394: It looks like we should flip this condition and `return` early so we can reduce the indentation here.

js/formidable_admin.js
```diff
@@ -10375,7 +10376,38 @@ function frmAdminBuildJS() {
...
+			// Find the next visible input field
+			const nextInput = inputsArray.slice( currentIndex + 1 ).find( input =>
+				input.offsetParent !== null &&
+				input.offsetWidth > 0 &&
```
Crabcyborg, line 10398: @shervElmi Do we really need these extra checks? I think just the `offsetParent` check is enough?

classes/controllers/FrmAppController.php
```diff
@@ -1353,6 +1353,51 @@ public static function filter_admin_notices() {
...
+
+		if ( is_array( $preferred_list_sort ) && ! empty( $preferred_list_sort['orderby'] ) ) {
+			$orderby = $preferred_list_sort['orderby'];
+			$order   = ! empty( $preferred_list_sort['order'] ) ? $preferred_list_sort['order'] : 'asc';
+		} else {
+			$orderby = FrmAppHelper::get_param( 'orderby', 'name' );
+			$order   = FrmAppHelper::get_param( 'order', 'desc' );
+		}
+
+		$_GET['orderby']     = $orderby;
```
Crabcyborg, line 1395: @shervElmi Can we do this in a cleaner way? Generally, setting `$_GET` variables is consider a very bad practice.

This can make code really hard to debug as well.

Crabcyborg:
@shervElmi QoL `#8` is incorrect.

> A link to Views for the current form right next to embed and styles.

**This isn't a link to the preview page.**

We already link to that in Quick Actions. We don't want a redundant URL here.
Screen Shot 2025-02-03 at 11 12 49 AM

The page we're referring to is the link to the Views tab (See https://qa.formidableforms.com/sherv11/wp-admin/admin.php?page=formidable-views&form=16 / https://qa.formidableforms.com/sherv11/wp-admin/edit.php?post_type=frm_display&show_nav=0&form=16).

classes/views/frm-fields/back-end/settings.php
```diff
@@ -202,6 +202,8 @@
 			include FrmAppHelper::plugin_path() . '/classes/views/frm-fields/back-end/field-description.php';
 		}
 
+		do_action( 'frm_field_options_after_description', compact( 'field', 'display', 'values' ) );
```
Crabcyborg, line 205: @shervElmi Can we add some comments here?

Typically a new filter has a `@since` comment, and `@param` comments to describe the parameters. Since this an array, we also typically try to use a `@type` comment for each array key.

This way the filter is well documented and easier to use in the future.

classes/helpers/FrmListHelper.php
```diff
@@ -391,7 +391,7 @@ protected function bulk_actions( $which = '' ) {
 			$verify = $this->confirm_bulk_delete();
 
 			if ( $verify ) {
-				echo "<a id='confirm-bulk-delete-" . esc_attr( $which ) . "' class='frm-hidden' href='confirm-bulk-delete' data-loaded-from='" . esc_attr( $this->loaded_from() ) . "' data-frmverify='" . esc_attr( $verify ) . "' data-frmverify-btn='frm-button-red'></a>";
+				echo "<a id='confirm-bulk-delete-" . esc_attr( $which ) . "' class='frm-hidden' tabindex='-1' aria-hidden='true' href='confirm-bulk-delete' data-loaded-from='" . esc_attr( $this->loaded_from() ) . "' data-frmverify='" . esc_attr( $verify ) . "' data-frmverify-btn='frm-button-red'></a>";
```
Crabcyborg, line 394: It might be nice to update this to use `FrmAppHelper::array_to_html_params`. It's pretty ugly right now with the single quotes.

classes/controllers/FrmHooksController.php
shervElmi: Addressed with the [Consolidate hooks into handle_current_screen](https://github.com/Strategy11/formidable-forms/pull/2215/commits/8126f76e05f15e29c7155a54de9a3d0b7a0a2558) commit.

js/formidable_admin.js
shervElmi: Addressed with the [Refactor to early return for reduced indentation](https://github.com/Strategy11/formidable-forms/pull/2215/commits/9d0bad2170801d2880075f3e7bbe0fc58ffe2cd9) commit.

js/formidable_admin.js
shervElmi: In this context, they aren’t necessary. 🚀 

Addressed with the [Remove unnecessary visibility checks](https://github.com/Strategy11/formidable-forms/pull/2215/commits/e914ed0acf42a23d2da475a2b7461018a193ee14) commit.

classes/controllers/FrmAppController.php
shervElmi: WP Table sorting and ordering work only with the URL string parameters orderby and order. I haven’t found any other method, as the source code relies on them.

Could you provide more feedback on this?

Crabcyborg: @shervElmi I believe we aren't really using the WP code much?

We have our own custom `FrmListHelper` class, that doesn't use a lot of WP code.

We should be able to fix this in the child classes. `FrmFormsListHelper`, etc.

shervElmi: Addressed with the following commits:

- [Refactor remember_custom_sort](https://github.com/Strategy11/formidable-forms/pull/2215/commits/e85e95431e51fd1ea61f04d531a8802405e9bd7c)
- [Load saved sorting preferences for the forms list](https://github.com/Strategy11/formidable-forms/pull/2215/commits/7e9de8b02b97de15fadebad39d6ddfabffaef611)
- [Add apply_saved_sort_preference method](https://github.com/Strategy11/formidable-forms/pull/2215/commits/5ab72b045ec1dc610ca71a0291d40d17ae622f87)
- [Apply saved sort preferences to Forms and History lists](https://github.com/Strategy11/formidable-forms/pull/2215/commits/50c6a0f3e8658167bf9ba18abef9464ce712f659)

shervElmi:
> @shervElmi QoL `#8` is incorrect.
> 
> > A link to Views for the current form right next to embed and styles.
> 
> **This isn't a link to the preview page.**
> 
> We already link to that in Quick Actions. We don't want a redundant URL here.
> ...

Addressed with the [Add column_views](https://github.com/Strategy11/formidable-forms/pull/2215/commits/406f4af58a89062dbf5ef74b050d954249b26346) commit.

classes/controllers/FrmAppController.php
```diff
@@ -1325,15 +1325,27 @@ private static function in_our_pages() {
...
+		self::filter_admin_notices();
+		self::remember_custom_sort();
+	}
+
+	/**
+	 * Hide all third-parties admin notices only in our admin pages.
+	 *
+	 * @return void
+	 */
+	private static function filter_admin_notices() {
```
Crabcyborg, line 1348: @shervElmi I don't think we can suddenly make `filter_admin_notices` a private function. We typically never make a function's scope less than what it was, since it would introduce errors if it was getting called somewhere else.

classes/controllers/FrmAppController.php
```diff
@@ -1358,6 +1370,60 @@ public static function filter_admin_notices() {
...
+	 */
+	private static function remember_custom_sort() {
+		$screen  = get_current_screen();
+		$orderby = FrmAppHelper::get_param( 'orderby' );
+
+		if ( ! $screen || ! $orderby ) {
+			return;
+		}
+
+		update_user_meta(
```
Crabcyborg, line 1386: @shervElmi Is there some way to avoid this getting called on every load?

Also, are we able to figure out a way to get the arrows to highlight based on the preference on load?

Screen Shot 2025-02-18 at 4 49 33 PM

classes/views/frm-fields/back-end/default-value-setting.php
```diff
@@ -12,7 +12,11 @@ class="frm-has-modal frm-default-value-wrapper default-value-section-<?php echo
 	</label>
 	<span class="frm-with-right-icon">
 		<?php
-		$field_obj->display_smart_values_modal_trigger_icon( $field );
+		$display_smart_values_icon = isset( $display_smart_values_icon ) ? $display_smart_values_icon : true;
```
Crabcyborg, line 15: A comment here might be nice since this is never happening in this plugin. We could mention that it's false for toggle fields in Pro.

classes/controllers/FrmAppController.php
Crabcyborg: I still think this is getting called too often. It would if `update_user_meta` wasn't called after refreshing, where my sort order never changed.

classes/views/frm-fields/back-end/default-value-setting.php
shervElmi: I’ve removed the changes based on feedback from the Pro version, so I’ve marked this feedback as resolved.

classes/controllers/FrmAppController.php
shervElmi: Addressed in the [Prevent redundant DB updates on sort refresh](https://github.com/Strategy11/formidable-forms/pull/2215/commits/1a5ca0355f41f3d1ed3c0615a15fe8cf92977762) commit.

classes/helpers/FrmFormsListHelper.php
```diff
@@ -370,6 +374,26 @@ protected function column_style( $form ) {
...
+	 * Generate the HTML for the form Views page.
+	 *
+	 * @since x.x
+	 *
+	 * @param stdClass $form Form object.
+	 * @return string
+	 */
+	protected function column_views( $form ) {
+		$attributes = array(
+			'href'   => admin_url( 'admin.php?page=formidable-views&form=' . absint( $form->id ) ),
```
Crabcyborg, line 387: @shervElmi Can we also pass `show_nav=1`? Without the nav at the top this page can look a bit funny.

classes/helpers/FrmListHelper.php
```diff
@@ -390,8 +390,20 @@ protected function bulk_actions( $which = '' ) {
 		if ( isset( $this->_actions['bulk_delete'] ) ) {
 			$verify = $this->confirm_bulk_delete();
 
+			$confirm_delete_attributes = array(
```
Crabcyborg, line 393: @shervElmi I think we could declare this value inside of the `if ( $verify )` check to avoid it when we don't need it.

classes/helpers/FrmFormsListHelper.php
shervElmi: Addressed with the [Add 'show_nav=1' to 'View Form" url](https://github.com/Strategy11/formidable-forms/pull/2215/commits/004c3deb23be6b28db2ab8e61e4f4f6af2fcd5df) commit.

classes/helpers/FrmListHelper.php
shervElmi: Addressed with the [Move $confirm_delete_attributes inside the $verify condition check](https://github.com/Strategy11/formidable-forms/pull/2215/commits/4c0f9f67049bfdced53244156a61d976a1861743) commit.

Crabcyborg (approved):
Thanks @shervElmi!

I think we're good to merge the QoL update.

🚀

PR #2224, opened 2025-01-15
GDPR Update

Liviu-p:
FF PRO - https://github.com/Strategy11/formidable-pro/pull/5585#issuecomment-2678379736


classes/views/frm-fields/front-end/gdpr/gdpr-field.php
```diff
@@ -0,0 +1,31 @@
...
+		<input type="checkbox" aria-required="true" name="item_meta[<?php echo esc_attr( $field_id ); ?>]" id="<?php echo esc_attr( $label_id ); ?>" value="1" 
+		<?php echo $checked . ' '; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
+		<?php do_action( 'frm_field_input_html', $field ); ?>
+		/>
+		<?php echo esc_html( $agreement_text ); ?>
+	</label>
+</div>
+
+<div class="frm_gdpr_field_description">
+	<?php echo esc_html( $field_desc ); ?>
```
Crabcyborg, line 30: Maybe we want to support some basic HTML in here?

We could use `FrmAppHelper::kses_echo`?

Crabcyborg:
@Liviu-p

I see a funny issue here.

If I check out this branch in Lite, but not in Pro, I see "Other" twice, and the first one is empty.

Is it possible to do something about that?

Screen Shot 2025-02-24 at 3 57 48 PM

Crabcyborg:
Hm, the Pro PR doesn't have a lot. I see weird settings even with Pro up to date.

Screen Shot 2025-02-24 at 4 00 50 PM

Crabcyborg:
Removed the top Other section that would always be empty with https://github.com/Strategy11/formidable-forms/pull/2224/commits/2f7c17a00289dd0442a0371da09eaba9a7309956

classes/views/frm-settings/general.php
```diff
@@ -49,26 +49,45 @@
...
+	$gdpr_options_wrapper_params['class'] .= ' frm_hidden';
+}
 $custom_header_ip_wrapper_params = array( 'class' => 'frm_custom_header_ip_cont' );
 if ( $frm_settings->no_ips ) {
 	$custom_header_ip_wrapper_params['class'] .= ' frm_hidden';
 }
 ?>
+<p>
+	<label>
+		<input type="checkbox" name="frm_enable_gdpr" id="frm_enable_gdpr" value="1" <?php checked( $frm_settings->enable_gdpr, 1 ); ?> data-frmshow=".frm_gdpr_options" />
```
Crabcyborg, line 70: @Liviu-p This would need to be on automatically if `Do not store user IPs with form submissions.` was on before. Otherwise, the old setting, that was always on, is hidden.

classes/views/frm-settings/general.php
```diff
@@ -49,26 +49,45 @@
...
+		<input type="checkbox" name="frm_no_ips" id="frm_no_ips" value="1" <?php checked( $frm_settings->no_ips, 1 ); ?> data-frmhide=".frm_custom_header_ip_cont" />
+		<?php esc_html_e( 'Do not store user IPs with form submissions.', 'formidable' ); ?>
+	</label>
+</p>
+
+<h3><?php esc_html_e( 'Other', 'formidable' ); ?></h3>
+
 <p <?php FrmAppHelper::array_to_html_params( $custom_header_ip_wrapper_params, true ); ?>>
 	<label>
 		<input type="checkbox" name="frm_custom_header_ip" id="frm_custom_header_ip" value="1" <?php checked( $frm_settings->custom_header_ip, 1 ); ?> />
```
Crabcyborg, line 93: @Liviu-p I think the custom header IP setting needs to go under GDPR as well.

You can see it goes away when the other IP setting is removed. It seems weird to me that it's toggling in another category. It should really still be under the other setting.

https://github.com/user-attachments/assets/89c262f0-a229-43c1-b246-2aa9cc3cb02c

Crabcyborg (changes requested):
@Liviu-p I'm seeing some issues with the GDPR field.

1. We don't want the format setting.

Screen Shot 2025-02-24 at 4 17 03 PM

2. Can we come up with some decent default text? I feel like Agreement Text isn't very helpful.
 
Screen Shot 2025-02-24 at 4 18 03 PM

3. It looks like I can uncheck the "Required" setting. This shouldn't be possible. Maybe we can gray it out or hide it?

4. Let's get rid of Default value. This should always be off by default.

5. Placeholder text isn't applicable. Let's remove that setting as well.

6. I also see `Max Characters` is a setting. That shouldn't be there. It seems like we're getting settings for text fields?

7. There's also an Invalid Format validation message setting, but that might go away when we get rid of the format setting?

classes/views/frm-settings/general.php
```diff
@@ -49,26 +49,45 @@
...
+	<label>
+		<input type="checkbox" name="frm_enable_gdpr" id="frm_enable_gdpr" value="1" <?php checked( $frm_settings->enable_gdpr, 1 ); ?> data-frmshow=".frm_gdpr_options" />
+		<?php esc_html_e( 'Enable GDPR related features and enhancements.', 'formidable' ); ?>
+	</label>
+</p>
+
+<p <?php FrmAppHelper::array_to_html_params( $gdpr_options_wrapper_params, true ); ?>>
+	<label>
+		<input type="checkbox" name="frm_no_gdpr_cookies" id="frm_no_gdpr_cookies" value="1" <?php checked( $frm_settings->no_gdpr_cookies, 1 ); ?> />
+		<?php esc_html_e( 'Disable user tracking cookies. This will disable the user tracking feature as well as the ability to limit form entries to one per user by cookie.', 'formidable' ); ?>
```
Crabcyborg, line 78: @Liviu-p Am I missing something? I see this mention `user tracking feature`, but I don't see any changes anywhere related to the user tracking feature.

classes/models/fields/FrmFieldGdpr.php
```diff
@@ -0,0 +1,111 @@
...
+	 * Get the field settings for the field type.
+	 *
+	 * @since x.x
+	 * @return bool[]
+	 */
+	protected function field_settings_for_type() {
+		return array(
+			'size'           => true,
+			'clear_on_focus' => true,
+			'invalid'        => true,
```
Crabcyborg, line 61: Let's set invalid to false here. That will get rid of the validation message.

classes/controllers/FrmHooksController.php
```diff
@@ -112,6 +112,7 @@ public static function load_hooks() {
 
 		FrmTransLiteHooksController::load_hooks();
 		FrmStrpLiteHooksController::load_hooks();
+		FrmFieldGdprHelper::load_hooks();
```
Crabcyborg, line 115: @Liviu-p I think it might be better for performance if we just add the hooks normally, and check the settings in the functions themselves. This way we only check our settings when we need to know, instead of every page load.

classes/helpers/FrmAppHelper.php
```diff
@@ -4410,4 +4410,16 @@ public static function print_setting_error( $args ) {
...
+	/**
+	 * Check if GDPR cookies are disabled.
+	 *
+	 * @since x.x
+	 *
+	 * @return bool
+	 */
+	public static function no_gdpr_cookies() {
+		$frm_settings = self::get_settings();
+		return (bool) $frm_settings->no_gdpr_cookies;
```
Crabcyborg, line 4423: @Liviu-p It looks like, if I disable the gdpr setting, this is still enabled?

Do we need to check `$frm_settings->enable_gdpr && $frm_settings->no_gdpr_cookies`?

classes/helpers/FrmFieldGdprHelper.php
```diff
@@ -0,0 +1,77 @@
...
+	/**
+	 * Load hooks
+	 *
+	 * @since x.x
+	 * @return void
+	 */
+	public static function load_hooks() {
+		$settings = FrmAppHelper::get_settings();
+		if ( $settings->enable_gdpr ) {
+			add_filter( 'frm_available_fields', array( __CLASS__, 'add_gdpr_field' ), 10, 1 );
```
Crabcyborg, line 43: I don't love how GDPR fields just break when you disable the GDPR setting.

It looks like it turns into a text field.

Maybe we can keep the field, but maybe hide the option in the sidebar (possibly just with CSS).

Then we should show a note (like with `frm_note_style`), that this field is not configured, and that it needs to be enabled in global settings. We could link to global settings as well.

On the front end, we should just try to render nothing possibly.

Screen Shot 2025-02-24 at 4 54 08 PM

classes/models/FrmField.php
```diff
@@ -378,7 +383,9 @@ public static function create( $values, $return = true ) {
 		$values = FrmAppHelper::maybe_filter_array( $values, array( 'name', 'description' ) );
 
 		foreach ( array( 'name', 'description', 'type', 'default_value' ) as $col ) {
-			$new_values[ $col ] = $values[ $col ];
+			if ( isset( $values[ $col ] ) ) {
```
Crabcyborg, line 386: I'm curious why this change is needed @Liviu-p?

Liviu-p: As the GDPR field doesn't have a `default_value`, I removed it from field options. The unit tests were failing because of that line. $values[ 'default_value' ] doesn't exists for GDPR field

classes/views/frm-settings/general.php
Liviu-p: good catch @Crabcyborg, I believe it should show up now for this particular case

classes/views/frm-settings/general.php
Liviu-p: it should be fixed

classes/models/fields/FrmFieldGdpr.php
Liviu-p: I updated it

classes/views/frm-settings/general.php
Liviu-p: This one it's related to "Save Cookie" option from "Limit Entries" from form settings.  I haven't manage to find something else related to tracking user cookies.
Screenshot 2025-03-06 at 22 17 30

classes/controllers/FrmHooksController.php
Liviu-p: Yep, I agree. I moved away from using hooks and loaded the field directly

classes/helpers/FrmAppHelper.php
Liviu-p: I updated it

classes/helpers/FrmFieldGdprHelper.php
Liviu-p: I fixed this one @Crabcyborg as suggested

classes/views/frm-fields/front-end/gdpr/gdpr-field.php
Liviu-p: I removed the description part as it will use the default available one

classes/views/frm-settings/general.php
```diff
@@ -67,7 +67,7 @@
 ?>
 <p>
 	<label>
-		<input type="checkbox" name="frm_enable_gdpr" id="frm_enable_gdpr" value="1" <?php checked( $frm_settings->enable_gdpr, 1 ); ?> data-frmshow=".frm_gdpr_options" data-frmuncheck="#frm_no_gdpr_cookies, #frm_no_ips, #frm_custom_header_ip" />
+		<input type="checkbox" name="frm_enable_gdpr" id="frm_enable_gdpr" value="1" <?php checked( $frm_settings->enable_gdpr || $frm_settings->no_ips, 1 ); ?> data-frmshow=".frm_gdpr_options" data-frmuncheck="#frm_no_gdpr_cookies, #frm_no_ips, #frm_custom_header_ip" />
```
Crabcyborg, line 70: @Liviu-p I think we might want to also check `custom_header_ip`.

But for consistency, I think we should just check everywhere the same.

Like not just here, but everywhere. We should see the GDPR field if it looks like it's checked here. If it didn't, it wouldn't feel right to the user.

classes/models/FrmField.php
Liviu-p: The `default_value` is provided by this `get_new_field_defaults` in `FrmFieldType` The FrmFieldGDPR is rewriting that one and doesn't include the 'default_value' option in order to remove the default field setting.

classes/views/frm-settings/general.php
Crabcyborg: @Liviu-p The user tracking is this add-on here https://github.com/Strategy11/formidable-user-tracking/

All that plugin does is track users, but since it is an add-on, it's already opt-in only.

I don't want to confuse people and say it disables user tracking when it isn't.

The setting here is just the "as well as the ability to limit form entries to one per user by cookie" part of the string.

We can reduce this string to just "This will disable the option to limit form entries to one per user by cookie.".

classes/views/frm-settings/general.php
Crabcyborg: Oh, it also looks like the other stuff that relies on the setting is all off.

I see this. The settings are invisible even though it appears checked.

Screen Shot 2025-03-06 at 4 34 45 PM

Crabcyborg (commented):
Screen Shot 2025-03-06 at 4 39 06 PM

@Liviu-p Is it possible to make the required checkbox appear greyed out? It looks like I can't disable it, but it doesn't look like a disabled field.

classes/views/frm-settings/general.php
```diff
@@ -49,33 +49,56 @@
...
 	<label>
 		<input type="checkbox" name="frm_custom_header_ip" id="frm_custom_header_ip" value="1" <?php checked( $frm_settings->custom_header_ip, 1 ); ?> />
 		<?php esc_html_e( 'Use custom headers when retrieving IPs with form submissions.', 'formidable' ); ?>
 		<?php FrmAppHelper::tooltip_icon( __( 'Only turn this on if IP addresses are incorrect in entries. Some server setups may require spoofable headers to determine an accurate IP address.', 'formidable' ) ); ?>
 	</label>
 </p>
+<p class="frm-text-xs frm-mb-0">
+	<?php
+	// translators: %s: Knowledge base URL
+	printf( 'Learn more about our GDPR settings <a href="%s" target="_blank">%s</a>', 'https://formidableforms.com/knowledgebase/gdpr-settings/', esc_html__( 'here', 'formidable' ) );
```
Crabcyborg, line 98: @Liviu-p It looks like the `here` is translated here but not the `'Learn more about our GDPR settings` text. Can we make this all translatable?

classes/models/FrmField.php
Crabcyborg: Sounds good. Thanks!

classes/views/frm-settings/general.php
Liviu-p: hey @Crabcyborg, I believe it will be more accurate from now

classes/views/frm-settings/general.php
Liviu-p: It should be fixed

Crabcyborg (approved):
Thank you @Liviu-p!

I think this looks good to merge now.

🚀

PR #2234, opened 2025-01-28
Fix phpstan name field constructor issue

Crabcyborg:
I'm going through our PHPStan exceptions and trying to clean code up a bit.

**The issue**
`$field = ''` is not a good default value since `FrmFieldType` expects `@param array|int|object $field`

**The fix**
I'm updating the constructor parameter default value to match the value used for `FrmFieldType`.


truongwp (approved):

Crabcyborg:
Thanks Truong!

PR #2237, opened 2025-01-28
Fix JS error when searching fields

Crabcyborg:
I noticed when searching fields, the section titles were disappearing.

It was because of this JS error introduced in v6.16.3.

Screen Shot 2025-01-28 at 1 13 19 PM


AbdiTolesa (approved):
Sorry about missing that check, thanks @Crabcyborg!

PR #2243, opened 2025-02-11
Fix html entities display issue

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5613

### DIFF
--- a/classes/models/fields/FrmFieldText.php
+++ b/classes/models/fields/FrmFieldText.php
@@ -47,4 +47,25 @@ class FrmFieldText extends FrmFieldType {
 
 		return $errors;
 	}
+
+	/**
+	 * If the value includes intentional entities, don't lose them.
+	 *
+	 * @since x.x
+	 *
+	 * @return string
+	 */
+	protected function prepare_esc_value() {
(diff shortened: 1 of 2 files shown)


classes/models/fields/FrmFieldText.php
```diff
@@ -47,4 +47,22 @@ public function validate( $args ) {
...
+	 *
+	 * @return string
+	 */
+	protected function prepare_esc_value() {
+		$value = $this->field['value'];
+		if ( is_null( $value ) ) {
+			return '';
+		}
+		if ( strpos( $value, '&lt;' ) !== false ) {
+			$value = str_replace( '&amp;', '&', htmlentities( $value ) );
```
truongwp, line 64: Should we move this to a separate method to reuse it in other field types?

AbdiTolesa: It looks other field types might not likely have similar issue and we may not want to refactor it for now.

truongwp: But I see this is duplicated in the two fields in this PR. So it's worth refactoring a bit.

PR #2245, opened 2025-02-11
Lite Version: Fix number field step validation

shervElmi:
**Pro PR:**
https://github.com/Strategy11/formidable-pro/pull/5618

### DIFF
--- a/classes/models/fields/FrmFieldNumber.php
+++ b/classes/models/fields/FrmFieldNumber.php
@@ -114,7 +114,7 @@ class FrmFieldNumber extends FrmFieldType {
 			// Translators: %1$s: the first nearest value; %2$s: the second nearest value.
 			__( 'Please enter a valid value. Two nearest valid values are %1$s and %2$s', 'formidable' ),
 			is_numeric( $result[0] ) ? floatval( $result[0] ) : $result[0],
-			is_numeric( $result[1] ) ? floatval( $result[1] ) : $result[1],
+			is_numeric( $result[1] ) ? floatval( $result[1] ) : $result[1]
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


classes/models/fields/FrmFieldNumber.php
```diff
@@ -113,8 +113,8 @@ protected function validate_step( &$errors, $args ) {
 		$errors[ 'field' . $args['id'] ] = sprintf(
 			// Translators: %1$s: the first nearest value; %2$s: the second nearest value.
 			__( 'Please enter a valid value. Two nearest valid values are %1$s and %2$s', 'formidable' ),
-			floatval( $result[0] ),
-			floatval( $result[1] )
+			is_numeric( $result[0] ) ? floatval( $result[0] ) : $result[0],
```
Crabcyborg, line 116: @shervElmi What is the case here where `$result[0]`/`$result[1]` is not a number? We should probably document that better.

shervElmi: For example, if a user sets a decimal separator like `*` or any other special character, we should adjust the number to use the custom decimal separator accordingly in the error message.

Crabcyborg (approved):
Thanks @shervElmi!

classes/models/fields/FrmFieldNumber.php
```diff
@@ -113,8 +113,8 @@ protected function validate_step( &$errors, $args ) {
...
+			is_numeric( $result[1] ) ? floatval( $result[1] ) : $result[1],
```
Crabcyborg, line 117: Oh, this comma here is breaking a PHP syntax check. I think it's still checking for old versions of PHP. But the extra comma here isn't necessary and we can just remove it.

PR #2248, opened 2025-02-17
Lite version: Fix number field validation for custom decimal separators

shervElmi:
### Main PR:
https://github.com/Strategy11/formidable-pro/pull/5629#issuecomment-2663420187

### DIFF
--- a/classes/models/fields/FrmFieldType.php
+++ b/classes/models/fields/FrmFieldType.php
@@ -1116,14 +1116,12 @@ DEFAULT_HTML;
 	 * @return void
 	 */
 	protected function add_min_max( $args, &$input_html ) {
-		$decimal_separator = FrmField::get_option( $this->field, 'custom_decimal_separator' );
-
-		$min = FrmCurrencyHelper::replace_custom_decimal_char( FrmField::get_option( $this->field, 'minnum' ), $decimal_separator );
+		$min = FrmField::get_option( $this->field, 'minnum' );
 		if ( ! is_numeric( $min ) ) {
 			$min = 0;
 		}
 
-		$max = FrmCurrencyHelper::replace_custom_decimal_char( FrmField::get_option( $this->field, 'maxnum' ), $decimal_separator );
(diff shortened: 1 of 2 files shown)


classes/models/fields/FrmFieldType.php
```diff
@@ -1116,12 +1116,14 @@ protected function add_extra_html_atts( $args, &$input_html ) {
...
-		$min = FrmField::get_option( $this->field, 'minnum' );
+		$decimal_separator = FrmField::get_option( $this->field, 'custom_decimal_separator' );
+
+		$min = FrmCurrencyHelper::replace_custom_decimal_char( FrmField::get_option( $this->field, 'minnum' ), $decimal_separator );
 		if ( ! is_numeric( $min ) ) {
 			$min = 0;
 		}
 
-		$max = FrmField::get_option( $this->field, 'maxnum' );
+		$max = FrmCurrencyHelper::replace_custom_decimal_char( FrmField::get_option( $this->field, 'maxnum' ), $decimal_separator );
```
Crabcyborg, line 1126: It would be nice if we could do this without requiring changes in Lite.

This is also rather dangerous, since we don't really want to use `custom_decimal_separator` in Lite even if it does exist, since the setting could no longer be available to someone who is no longer using Pro.

shervElmi: I’ve reverted these changes and will handle them in Pro. However, let’s keep `replace_custom_decimal_char` in the Lite version with this PR. Does that sound good to you?

Crabcyborg: If it isn't being used in Lite I don't see why it would make sense in Lite.

shervElmi: I thought it might be useful in the future, but you’re right. 👍🏻

shervElmi: @Crabcyborg, could you close this PR? I’ve made the necessary changes in the Pro version.

PR #2251, opened 2025-02-18
Fix views issue 28

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-views/issues/28

The issue, is that this call to `intval` changes `''` to `0`, which we don't want. Keeping it blank is better.

I tested with this basic CSV. My form has a "Text" field, and a User ID field. The User ID field isn't in the CSV, since we're trying to import empty.

[250218182012_user-id-test_formidable_entries.csv](https://github.com/user-attachments/files/18851714/250218182012_user-id-test_formidable_entries.csv)


AbdiTolesa (approved):
Thanks @Crabcyborg!

PR #2257, opened 2025-02-21
Fix Currency Task Issues - Lite

shervElmi:
### Main PR:
https://github.com/Strategy11/formidable-pro/pull/5645


Crabcyborg:
@shervElmi Ideally, we won't require changes in all of our payments code to fix this.

If we were able to break it changing Pro, we should be able to fix it changing Pro.

shervElmi:
@Crabcyborg, could you please close this PR? We no longer need it at this time.

Crabcyborg:
Thank you Sherv!

PR #2258, opened 2025-02-24
Enhance hasClass function

AbdiTolesa:
Just adding a check for HTML element before its classList is attempted to be read.


Crabcyborg:
@AbdiTolesa

My concern with this check, is that it would likely be masking issues. In most cases, I'm sure, if the element does not exist, we don't really want to be checking its class.

I had fixed this last week with https://github.com/Strategy11/formidable-forms/pull/2255/files

Crabcyborg:
I'm going to close this for now.

PR #2259, opened 2025-02-24
Redefine white labelling

AbdiTolesa:
Related PR: https://github.com/Strategy11/formidable-pro/pull/5668


classes/helpers/FrmAppHelper.php
```diff
@@ -200,7 +200,7 @@ public static function get_menu_name() {
 	 */
 	public static function is_formidable_branding() {
 		if ( ! self::pro_is_installed() ) {
-			return true;
+			return false;
```
Crabcyborg, line 203: @AbdiTolesa What is the reason for this change? I think we do want the Formidable branding whenever pro is not installed?

AbdiTolesa: @Crabcyborg I thought Steph's request from the [issue checklist](https://github.com/Strategy11/formidable-pro/issues/4687) says so:

> 3. Update the FrmAppHelper::is_formidable_branding() to always return false if pro is not active.

Crabcyborg: @AbdiTolesa I think that item in the checklist is a mistake.

We want to always show the floating links to Lite users. This update does the opposite and hides it for all of them.

js/src/common/components/icon.js
```diff
@@ -8,7 +8,7 @@ export default class FormidableIcon extends Component {
 
 	loadCustomSvgIcon( ) {
 		const icon = formidable_form_selector.icon;
-		if ( icon.match( /frm_white_label_icon/ ) ) {
+		if ( icon.match( /frm_white_label_icon/ ) || 1 == parseInt( icon ) ) {
```
Crabcyborg, line 11: @AbdiTolesa Would it maybe make sense to just call the new setting `frm_white_label_icon` so we don't need to make changes in Lite?

AbdiTolesa: @Crabcyborg We are changing the setting's type to boolean in Pro to make it easy for backward compatibility, that means we are just tracking whether it is on/off as that is changed to a toggle now.
https://github.com/Strategy11/formidable-pro/pull/5668/files#diff-bc6968583edc1dad89ab73f71753369fd8c39732daeb8e937c68727989180f17R118

Crabcyborg: Yeah, I don't love this idea of changing `menu_icon` from a string to a boolean though. It sounds like we could run into a lot of type conflicts.

I think saving an empty string vs. `frm_white_label_icon` would be more backward compatible.

AbdiTolesa: Ok, I have updated my Pro update accordingly so that this update is not needed.

PR #2260, opened 2025-02-24
Only localize scripts once

truongwp:
Suggestion by Mike in https://github.com/Strategy11/formidable-views/pull/625#discussion_r1963812162


Crabcyborg (approved):
Thanks @truongwp!

I think this should be good.

PR #2281, opened 2025-03-24
Fix issue with backslashes being removed from entry values

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/1246

### DIFF
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -617,7 +617,7 @@ class FrmEntriesController {
 		}
 
 		if ( $errors == '' && ! $ajax ) {
-			$errors = FrmEntryValidate::validate( wp_unslash( $_POST ) ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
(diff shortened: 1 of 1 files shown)


AbdiTolesa:
@Crabcyborg This PR only covers entry create process, we can apply similar fix to Pro if it is accepted.

What I found to be causing the issue of backslashes getting removed from the entry values is that we are calling `wp_unslash` twice on each entry value when creating an entry and try to unslash the value in `FrmEntry::prepare_entry`.

Here is what I exactly mean,

When creating an entry `$_POST` is slashed somewhere and `FrmEntriesController::process_entry` is called, which unslashes the global `$_POST` when passing it to `FrmEntryValidate::validate`. Later on `FrmEntriesHelper::get_posted_meta` is called and that unslashes the entry meta values in the global `$_POST` variable for the second time. Now if you have `\I'm cool` in your value, only `I'm cool` is stored in database. That is the issue.

The values sent to the database should actually be slashed once to be retrieved properly later but I don't have that much of confidence on that. I have tried to think to come up with alternative solutions but I couldn't.

I would appreciate your thoughts on this.

Crabcyborg:
Closing this since I'm not confident this update is safe.

PR #2282, opened 2025-03-24
Make gdpr field required even if the setting is disabled

Crabcyborg:
Fixes the customer's issue in https://github.com/Strategy11/formidable-pro/issues/5708

It is possible for the GDPR field to not be required. I can at least replicate this by adding a field, disabling the GDPR setting, save the field, and then enabling it again. Since it is read only you end up with a checkbox that is uncheckable and unchecked.

Screen Shot 2025-03-24 at 10 04 47 AM

This update always sets the checkbox to appear as checked if the `readonly_required` setting is on. When this is on, we expect it to always be checked. I don't see us requiring a readonly disabled checkbox since we would likely just hide the setting in those cases.

I also added a check so it always does the required validation even when unchecked, and the required asterisk will still show even when unchecked.

**Pre-release**
[formidable-6.19.1b.zip](https://github.com/user-attachments/files/19430232/formidable-6.19.1b.zip)

### DIFF
--- a/classes/controllers/FrmHooksController.php
+++ b/classes/controllers/FrmHooksController.php
@@ -112,6 +112,9 @@ class FrmHooksController {
 
 		FrmTransLiteHooksController::load_hooks();
 		FrmStrpLiteHooksController::load_hooks();
+
+		// GDPR
+		add_filter( 'frm_is_field_required', 'FrmFieldGdpr::force_required_field', 10, 2 );
 	}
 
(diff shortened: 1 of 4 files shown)


Liviu-p (approved):
@Crabcyborg, it looks good to me. Thank you for fixing it!

Crabcyborg:
Thank you Liviu!

PR #2285, opened 2025-03-26
Antispam improvements

truongwp:
- [x] Honeypot improvements
- [x] Add stopforumspam API
- [ ] Add WordPress spam comments check
- [ ] Block list check
- [ ] Usage tracking

### DIFF
--- a/classes/models/FrmHoneypot.php
+++ b/classes/models/FrmHoneypot.php
@@ -5,11 +5,27 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 class FrmHoneypot extends FrmValidate {
 
+	protected $fields;
+
+	public function __construct( $form_id, $fields = null ) {
+		parent::__construct( $form_id );
+		if ( is_null( $fields ) ) {
+			$fields = FrmField::get_all_for_form( $form_id, '', 'include' );
+		}
+		$this->fields = $fields;
+	}
+
 	/**
 	 * @return string
 	 */
 	protected function get_option_key() {
 		return 'honeypot';
+	}
+
+	protected function is_option_on() {
+		$frm_settings = FrmAppHelper::get_settings();
+		$key          = $this->get_option_key();
+		return $frm_settings->$key;
 	}
 
 	/**
@@ -29,14 +45,24 @@ class FrmHoneypot extends FrmValidate {
 	private function is_honeypot_spam() {
 		$is_honeypot_spam = $this->is_legacy_honeypot_spam();
 		if ( ! $is_honeypot_spam ) {
-			// Check the newer honeypot input name which is randomly generated so it's more difficult to detect.
-			$class_name       = self::get_honeypot_class_name();
-			$honeypot_value   = FrmAppHelper::get_param( $class_name, '', 'get', 'sanitize_text_field' );
-			$is_honeypot_spam = '' !== $honeypot_value;
+			$field_id = $this->get_honeypot_field_id();
+			$value    = $this->get_honeypot_field_value( $field_id );
+			$is_honeypot_spam = '' !== $value;
 		}
 
-		$form = $this->get_form();
-		$atts = compact( 'form' );
+		$atts = array(
+			'form'   => $this->get_form(),
+			'fields' => $this->fields,
+		);
+
+		/**
+		 * Filters the honeypot spam check.
+		 *
+		 * @since x.x The `$atts` now contains `fields`.
+		 *
+		 * @param bool  $is_honeypot_spam Set to `true` if is spam.
+		 * @param array $atts             Contains `form` and `fields`.
+		 */
 		return apply_filters( 'frm_process_honeypot', $is_honeypot_spam, $atts );
 	}
 
@@ -68,12 +94,13 @@ class FrmHoneypot extends FrmValidate {
 	}
 
 	/**
-	 * @param int $form_id
+	 * @param int        $form_id Form ID.
+	 * @param array|null $fields  Array of fields.
 	 *
 	 * @return void
 	 */
-	public static function maybe_render_field( $form_id ) {
-		$honeypot = new self( $form_id );
+	public static function maybe_render_field( $form_id, $fields = null ) {
+		$honeypot = new self( $form_id, $fields );
 		if ( $honeypot->should_render_field() ) {
 			$honeypot->render_field();
 		}
@@ -90,28 +117,66 @@ class FrmHoneypot extends FrmValidate {
 	 * @return void
 	 */
 	public function render_field() {
-		$honeypot    = $this->check_honeypot_setting();
-		$form        = $this->get_form();
-		$class_name  = self::get_honeypot_class_name();
+		$field_id    = $this->get_honeypot_field_id();
+		$field_key   = $this->get_honeypot_field_key();
 		$input_attrs = array(
-			'id'    => 'frm_email_' . absint( $form->id ),
-			'type'  => 'strict' === $honeypot ? 'email' : 'text',
-			'class' => 'frm_verify',
-			'name'  => $class_name,
-			'value' => FrmAppHelper::get_param( $class_name, '', 'get', 'wp_kses_post' ),
+			'id'    => 'field_' . $field_key,
+			'type'  => 'text',
+			'class' => 'frm_form_field form-field frm_verify',
+			'name'  => 'item_meta[' . $field_id . ']',
+			'value' => $this->get_honeypot_field_value( $field_id ),
 		);
-
-		if ( 'strict' !== $honeypot ) {
-			$input_attrs['autocomplete'] = 'off';
-		}
 		?>
-			<div class="<?php echo esc_attr( $class_name ); ?>">
-				<label for="frm_email_<?php echo esc_attr( $form->id ); ?>" <?php FrmFormsHelper::maybe_hide_inline(); ?>>
+			<div class="frm_field_<?php echo intval( $field_id ); ?>_container">
+				<label for="<?php echo esc_attr( $input_attrs['id'] ); ?>" <?php FrmFormsHelper::maybe_hide_inline(); ?>>
 					<?php esc_html_e( 'If you are human, leave this field blank.', 'formidable' ); ?>
 				</label>
 				<input <?php FrmAppHelper::array_to_html_params( $input_attrs, true ); ?> <?php FrmFormsHelper::maybe_hide_inline(); ?> />
 			</div>
 		<?php
+	}
+
+	/**
+	 * Gets honeypot field ID. This should not exist in the form.
+	 *
+	 * @return int
+	 */
+	private function get_honeypot_field_id() {
+		$max = 0;
+		foreach ( $this->fields as $field ) {
+			if ( $field->id > $max ) {
+				$max = $field->id;
+			}
+			unset( $field );
+		}
+		return $max + 1;
+	}
+
+	private function get_honeypot_field_key() {
+		return FrmAppHelper::generate_new_key( 5 );
+	}
+
+	/**
+	 * Gets honeypot field value.
+	 *
+	 * @param string $field_id Field ID.
+	 *
+	 * @return string
+	 */
+	private function get_honeypot_field_value( $field_id ) {
+		$item_meta = FrmAppHelper::get_simple_request(
+			array(
+				'param'   => 'item_meta',
+				'default' => array(),
+				'type'    => 'request',
+			)
+		);
+
+		if ( ! $item_meta || ! is_array( $item_meta ) ) {
+			return '';
+		}
+
+		return isset( $item_meta[ $field_id ] ) ? $item_meta[ $field_id ] : '';
 	}
 
 	/**
--- a/classes/models/FrmHoneypot.php
+++ b/classes/models/FrmHoneypot.php
@@ -7,6 +7,8 @@ class FrmHoneypot extends FrmValidate {
 
 	protected $fields;
 
+	protected $option_type = 'global';
+
 	public function __construct( $form_id, $fields = null ) {
 		parent::__construct( $form_id );
 		if ( is_null( $fields ) ) {
@@ -20,10 +22,6 @@ class FrmHoneypot extends FrmValidate {
 	 */
 	protected function get_option_key() {
 		return 'honeypot';
-	}
-
-	protected function is_option_on() {
-		return self::is_enabled();
 	}
 
 	private static function is_enabled() {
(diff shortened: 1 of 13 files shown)


blacklist/domain-partial.txt
Crabcyborg, line 1: @truongwp Can we go with another term than blacklist? WordPress doesn't like this word since it's considered a derogatory term.

We could go with `denylist` to keep it short.

classes/views/frm-settings/captcha/captcha.php
```diff
@@ -103,3 +103,36 @@
...
 	?>
 </div>
+
+<h2 class="frm-h2"><?php esc_html_e( 'Spam', 'formidable' ); ?></h2>
+
+<p>
+	<label for="frm-honeypot">
+		<input type="checkbox" id="frm-honeypot" name="frm_honeypot" value="1" <?php checked( $frm_settings->honeypot, 1 ); ?> />
+		<?php esc_html_e( 'Use Honeypot to check entries for spam', 'formidable' ); ?>
+		<?php FrmAppHelper::tooltip_icon( __( 'Include an invisible field in your form to trick bots. Setting to strict will catch more spam but issues with autocomplete may prevent real people from submitting on some browsers.', 'formidable' ), array( 'data-container' => 'body' ) ); ?>
```
Crabcyborg, line 113: @truongwp It looks like the tooltip hasn't been updated here. The strict option is gone - we should stop referencing it.

classes/controllers/FrmAntiSpamController.php
```diff
@@ -0,0 +1,78 @@
...
+		return $matches[0];
+	}
+
+	/**
+	 * Gets whitelist IP addresses.
+	 *
+	 * @return string[]
+	 */
+	public static function get_whitelist_ip() {
+		return apply_filters( 'frm_whitelist_ip', array( '', '127.0.0.1' ) );
```
Crabcyborg, line 76: Can we call this `frm_allowed_ips`?

classes/controllers/FrmAntiSpamController.php
```diff
@@ -0,0 +1,78 @@
...
+		preg_match_all( '/[\._a-zA-Z0-9-]+@[\._a-zA-Z0-9-]+/i', $values, $matches );
+		return $matches[0];
+	}
+
+	/**
+	 * Gets whitelist IP addresses.
+	 *
+	 * @return string[]
+	 */
+	public static function get_whitelist_ip() {
```
Crabcyborg, line 75: And `get_allowed_ips`?

classes/controllers/FrmAntiSpamController.php
```diff
@@ -0,0 +1,78 @@
...
+		$spam_check = new FrmSpamCheckUseWPComments( $values );
+		return $spam_check->is_spam();
+	}
+
+	public static function contains_wp_disallowed_words( $values ) {
+		$spam_check = new FrmSpamCheckWPDisallowedWords( $values );
+		return $spam_check->is_spam();
+	}
+
+	public static function is_blacklist_spam( $values ) {
```
Crabcyborg, line 44: Can we call this `is_denylist_spam`?

classes/models/FrmSpamCheckStopforumspam.php
```diff
@@ -0,0 +1,53 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+
+class FrmSpamCheckStopforumspam extends FrmSpamCheck {
```
Crabcyborg, line 6: Are we able to change the case here?

I feel like `FrmSpamCheckStopForumSpam` looks better.

classes/models/FrmSpamCheckBlacklist.php
```diff
@@ -0,0 +1,296 @@
...
+		);
+
+		$custom_blacklist = $this->get_words_from_setting( 'blacklist' );
+		if ( $custom_blacklist ) {
+			$blacklist_data['custom'] = array(
+				'words' => $custom_blacklist,
+			);
+		}
+
+		return apply_filters( 'frm_blacklist_data', $blacklist_data );
```
Crabcyborg, line 61: Let's rename this to `frm_denylist_data`.

classes/models/FrmSpamCheckStopforumspam.php
```diff
@@ -0,0 +1,53 @@
...
+		return $this->response_is_spam( $response );
+	}
+
+	protected function is_enabled() {
+		$form = FrmForm::getOne( $this->values['form_id'] );
+		return $form && ! empty( $form->options['stopforumspam'] );
+	}
+
+	private function send_request( $request_data ) {
+		$url = add_query_arg( $request_data, 'https://api.stopforumspam.org/api' );
```
Crabcyborg, line 43: @truongwp Maybe we can offer a way for someone to add an API key to this URL with a filter?

It isn't required, but I think it might come into play with rate limiting. If it's used too much, I think you can get more requests when you pass an API key.

Crabcyborg (commented):
This looks great @truongwp!

A lot of my comments are about renaming things. The functionality all looks solid.

🙌

blacklist/domain-partial.txt
truongwp: Do we need to change the setting name?

Crabcyborg:
@truongwp It looks like one of the unit tests is failing.

Screen Shot 2025-05-05 at 9 57 16 AM

truongwp:
> @truongwp It looks like one of the unit tests is failing.
> 
> Screen Shot 2025-05-05 at 9 57 16 AM

Yes. I haven't completed all your feedback yet. Something hasn't been updated in the tests.

classes/models/FrmSpamCheckWPDisallowedWords.php
```diff
@@ -0,0 +1,63 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+
+class FrmSpamCheckWPDisallowedWords extends FrmSpamCheck {
+
+	public function check() {
+		$mod_keys = trim( $this->get_disallowed_words() );
```
Crabcyborg, line 9: @truongwp It looks like we aren't checking against this list of disallowed words yet https://github.com/splorp/wordpress-comment-blacklist/blob/master/blacklist.txt

Can we also include this file and disallow these words?

classes/models/FrmSpamCheckStopforumspam.php
truongwp: I added 2 filters, one for the request data and one for the URL. Users might want to change the region of the API.

truongwp:
@Crabcyborg I resolved all feedback. I added the Splord WP comment denylist. I'm fixing PHPCS errors and adding more doc comments now.

Crabcyborg (commented):
@truongwp I think it might be nice to allow trusted users to bypass a lot of this. We don't really want to block all of the comment blacklist data for someone who we know isn't a spammer.

I think we could add a check for `current_user_can( 'frm_create_entries' )` early?

truongwp:
> @truongwp I think it might be nice to allow trusted users to bypass a lot of this. We don't really want to block all of the comment blacklist data for someone who we know isn't a spammer.
> 
> I think we could add a check for `current_user_can( 'frm_create_entries' )` early?

Which spam check do you want to add that check?

Crabcyborg:
> Which spam check do you want to add that check? Or all of them?

I think we can keep a lot of the checks. But maybe the splorp file we can ignore by default?

Crabcyborg:
@truongwp It looks like the e2e test is picking up an accessibility issue.

I tried the global settings in axe and confirmed this error.

Screen Shot 2025-05-07 at 1 36 03 PM

truongwp:
@Crabcyborg I have one question. Previously, we use a different message for blacklist spam: https://github.com/Strategy11/formidable-forms/blob/1de0ae72f8fb1fc9bd59f526381daff4410170a6/classes/models/FrmEntryValidate.php#L436

Do we need to keep it in this version? Currently, I'm using `Your entry appears to be spam!` for all spam checks.

Crabcyborg:
@truongwp We also have a `isHoneypotSpam` function in Pro.

Right now it checks.

```
var honeypotField = document.getElementById( 'frm_email_' + formID );
if ( honeypotField === null ) {
	honeypotField = document.getElementById( 'frm_verify_' + formID );
}
return honeypotField !== null && honeypotField.value !== '';
```

I think we'll need to update that as well.

classes/models/FrmSpamCheckDenylist.php
```diff
@@ -0,0 +1,464 @@
...
+	 *
+	 * @return array[]
+	 */
+	protected function get_denylist_array() {
+		$denylist_data = array(
+			array(
+				'file' => FrmAppHelper::plugin_path() . '/denylist/domain-partial.txt',
+			),
+			array(
+				'file' => FrmAppHelper::plugin_path() . '/denylist/splorp-wp-comment.txt',
```
Crabcyborg, line 47: @truongwp Can we also avoid this for file fields? I feel like some of the substrings in this might match valid file names.

truongwp:
> @truongwp
> 
> We have a function in JS, `makeHoneypotFieldsUntabbable`.
> 
> It includes a `input.id.indexOf( 'frm_email_' )` check. With this update, the ID has changed, so now the honeypot field is tabblable.
> 
> We'll need to update this function as well. Just removing the condition should be fine. I don't think `.frm_verify` is really used for anything else.
> 
> We'll probably want to also drop `.frm_verify` in the future, but there are a couple places where things would break. I think we should leave it for this update.

For the `frm_verify`, I think we can remove it from PHP side and add it using JS, like we do with the `<style>` tag for honeypot field.

For making it untabbable: I think we can just make honeypot field `display:none` instead of `overflow:hidden;...`. The hidden style is loaded only when JS runs, so when bots retrieve the HTML, it's visible.

What do you think?

Crabcyborg:
> I think we can just make honeypot field display:none instead of overflow:hidden;.... The hidden style is loaded only when JS runs, so when bots retrieve the HTML, it's visible.

Thanks @truongwp. I think it was something like that before. The concern is that it's too easy to check for `display: none`, and less likely they'd check for the overflow.

truongwp:
> > I think we can just make honeypot field display:none instead of overflow:hidden;.... The hidden style is loaded only when JS runs, so when bots retrieve the HTML, it's visible.
> 
> Thanks @truongwp. I think it was something like that before. The concern is that it's too easy to check for `display: none`, and less likely they'd check for the overflow.

But without JS, there is no CSS to hide it, it's visible.

Crabcyborg:
@truongwp I think some bots can still run JS. And it sounds like we might not use JS because of the API form issue? I think we should keep the frm_verify for this version, and we can revisit after we release. I don't want to change too much, and I'm concerned requiring JS could cause issues.

What about adding `visibility: hidden;`? It looks like it prevents tabbing.

classes/models/FrmSpamCheckDenylist.php
```diff
@@ -0,0 +1,464 @@
...
+	const COMPARE_EQUALS = 'equals';
+
+	protected $posted_fields;
+
+	protected $denylist;
+
+	public function __construct( $values ) {
+		parent::__construct( $values );
+
+		$this->posted_fields = FrmField::get_all_for_form( $values['form_id'] );
```
Crabcyborg, line 19: @truongwp Sometimes I see `PHP Warning:  Undefined array key "form_id"`.

I think it might be the black list check when files are uploaded? It doesn't pass anything but the `item_item` for the file field ID.

truongwp: I added `form_id` in the linked PR. I also added a fix if user uses the old Pro version.

truongwp:
@Crabcyborg I completed everything. Please test again. 

Now denylist array supports `skip` and `skip_field_types`. I changed `field_type` to `field_types`.

classes/models/FrmHoneypot.php
```diff
@@ -59,26 +107,77 @@ private function check_honeypot_filter() {
...
+	 */
+	public static function maybe_print_honeypot_css() {
+		// Print the CSS if form is loaded by API.
+		if ( ! FrmFormsHelper::form_is_loaded_by_api() ) {
+			return;
+		}
+
+		$css = self::get_honeypot_field_css();
+		if ( $css ) {
+			echo '<style>' . esc_attr( $css ) . '</style>';
```
Crabcyborg, line 160: PHPCS is throwing an error here. This should be `esc_html`.

classes/models/FrmHoneypot.php
```diff
@@ -59,26 +107,77 @@ private function check_honeypot_filter() {
...
-		return $form->options[ $key ];
-	}
-
-	/**
-	 * @param int $form_id
+	 * @param int $form_id Form ID.
 	 *
 	 * @return void
 	 */
 	public static function maybe_render_field( $form_id ) {
```
Crabcyborg, line 114: @truongwp It feels a little funny that this field is after the submit button.

Do you think it's possible that we could inject it somewhere else on the page, where it's closer to another fields?

classes/views/frm-forms/spam-settings/stopforumspam.php
```diff
@@ -0,0 +1,12 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+?>
+<p class="frm6 frm_form_field frm_first">
+	<label for="stopforumspam">
```
Crabcyborg, line 7: Our AXE tests are picking up errors on these.

I see `no-redundant-for`.

Basically, we don't need a `for` attribute when the `input` is already a child of the `label`.

So we should remove this `for` here. And for the Honeypot and Spam Comments settings in global settings too.

Screen Shot 2025-05-08 at 2 11 59 PM

Crabcyborg:
@truongwp A unit test is also breaking.

Screen Shot 2025-05-08 at 4 41 15 PM

Crabcyborg:
Also @truongwp I got the e2e tests passing on master. So all new tests failing are because of changes in this branch.

classes/models/FrmHoneypot.php
truongwp: Moving the honeypot field to the top causes the e2e error so I reverted it.

Crabcyborg (approved):
Thank you @truongwp!

I'll go ahead and merge this now.

I think it's good to go. If anything else comes up, feel free to make a new PR.

🚀

classes/models/FrmHoneypot.php
Crabcyborg: That's fine. We can consider this for a future update.

PR #2286, opened 2025-03-26
Only editors should be allowed to preview

matt-pierce-montclair:
Given that:
* on the creation of a new site, Formidable creates a form called 'contact-form' that is active
* the form has an Email Notification action set to send to [site_admin]
* the form does not have any strong spam mitigation enabled
* the preview page is not restricted in any way

This results in a situation where any unauthenticated user can go to `/wp-admin/admin-ajax.php?action=frm_forms_preview&form=contact-form` and spam the site admin's email without any authentication, whether or not the default example form has been published on any page.

The solution proposed here is to limit the Preview function to people who have `frm_edit_forms` access since it's clear from the interface that the only way you should normally be able to access the Preview is to do so from the Editor.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -389,6 +389,7 @@ class FrmFormsController {
 	 * @return string|null
 	 */
 	public static function page_preview() {
+		FrmAppHelper::permission_check( 'frm_edit_forms' );
 		$params = FrmForm::list_page_params();
 		if ( ! $params['form'] ) {
 			return null;
@@ -415,6 +416,7 @@ class FrmFormsController {
 	 * @return void
 	 */
 	public static function preview() {
+		FrmAppHelper::permission_check( 'frm_edit_forms' );
 		do_action( 'frm_wp' );
 		FrmAppHelper::set_current_screen_and_hook_suffix();
 


matt-pierce-montclair:
Hey Mike,

Thanks for getting back to me so quickly. It's a bit of a bummer that there
are plugins relying on this behavior because it's really just a very
strange way of handling previews - but I do understand.

At least for me, and the onslaught of MySQL injection attempts that flooded
my inbox yesterday (which is what brought about this question) the proposed
solution of (a) randomizing the ID number and key for the sample Contact
form and (b) not displaying a preview when there's no form specified - this
definitely would have completely solved the problem. Just so long as
there's no API endpoint that exposes form keys to unauthenticated users.

Crabcyborg:
Thank you @matt-pierce-montclair.

We're planning on adding new spam protection features soon. We have a few really good ideas in the works, including new blocklists and new options for API checks.

Your suggestions are great! I'll look into the possibility of also randomizing the form key of the contact form. I was just trying to protect `contact-form`, as I'm also concerned about existing sites (and the new contact form form keys would only be possible for new sites).

It's helpful to know whatever we can about the new spam. I don't usually get enough to work with. Knowing it's the default contact form you're struggling with is really useful.

Crabcyborg:
Closing this.

Currently working on some new improvements https://github.com/Strategy11/formidable-forms/pull/2287

Still a work in progress.

PR #2241, opened 2025-02-05
Make admin nav responsive

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/4076
Design: https://www.figma.com/design/DQbZMgeNmJWkxOWYfHjDZX/Formidable-System?node-id=316-1642&t=w5JzVxzSrEUHXsVY-0

image

image

image

PR #2291, opened 2025-04-01
Hide posts filter when showing mobile go back option

Crabcyborg:
This fixes the issues I see in https://github.com/Strategy11/formidable-forms/pull/2241#issuecomment-2757943304

There's a `posts-filter` element on the forms list page that was increasing the width of the page. Hiding it fixes the issue.


AbdiTolesa (approved):
Thanks @Crabcyborg, I noticed the issue on forms list page and this fixes it for me.

PR #2302, opened 2025-04-07
Limit pages small device message appears in

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5728

### Test procedure
Go to different WP admin pages that are not Formidable pages and confirm the 'More on bigger devices' message does not appear anywhere on the page.

### DIFF
--- a/classes/controllers/FrmFormsController.php
+++ b/classes/controllers/FrmFormsController.php
@@ -1928,6 +1928,13 @@ class FrmFormsController {
 	 * @return void
 	 */
 	public static function include_device_too_small_message() {
+		if ( ! FrmAppHelper::is_formidable_admin() ) {
+			return;
+		}
+		if ( 'frm_logs' !== FrmAppHelper::simple_get( 'post_type', 'sanitize_title' ) ) {
+			return;
+		}
+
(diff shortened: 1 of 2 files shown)


classes/controllers/FrmFormsController.php
```diff
@@ -1928,6 +1928,13 @@ public static function process_bulk_form_actions( $errors ) {
 	 * @return void
 	 */
 	public static function include_device_too_small_message() {
+		if ( ! FrmAppHelper::is_formidable_admin() ) {
+			return;
+		}
+		if ( 'frm_logs' !== FrmAppHelper::simple_get( 'post_type', 'sanitize_title' ) ) {
```
Crabcyborg, line 1934: @AbdiTolesa This is never true. `is_formidable_admin` is never true on the logs page.

That being said - it looks like the logs page is fine on a really small screen. Can we possibly allow that to continue to work normally instead?

Crabcyborg (approved):
Thanks @AbdiTolesa!

I made a small commit to fix an issue I had introduced by hiding `#posts-filter`. Nothing on the logs page was really appearing on the mobile width because of that. Now that should only get hidden on the "formidable" top level page.

PR #2304, opened 2025-04-07
Fix sorting issues

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/5722

This function was all messed up. It was copied from the usage controller, but that has a different context.

1. It was checking for the trash page and returning false, so sorting on the trash page was not possible.
2. It was missing the delete action.

I also fixed the usage controller logic, so it doesn't miss the delete action either.


Crabcyborg:
Thanks Sherv!

PR #2321, opened 2025-04-22
Fix issue 5751

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/5751

This broke with the QoL update. This code that adds the `+0` is important, but get overwritten when `FrmAppController::apply_saved_sort_preference` is called.

I moved this after so it still happens.


shervElmi (approved):
LGTM. Thanks @Crabcyborg! 🚀

Crabcyborg:
Thanks Sherv!

PR #2324, opened 2025-04-25
Pull bulk delete entries into Lite

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5704

Required PR:https://github.com/Strategy11/formidable-pro/pull/6082

This update makes it possible to do bulk entry delete without needing the Pro version of Formidable Forms.
CleanShot 2025-05-15 at 17 31 36


classes/controllers/FrmEntriesController.php
```diff
@@ -104,6 +113,44 @@ private static function load_manage_entries_hooks() {
...
+	 *
+	 * @param string $action
+	 * @return void
+	 */
+	public static function bulk_actions( $action = 'list-form' ) {
+		$errors = array();
+
+		$items = FrmAppHelper::get_param( 'item-action', '', 'get', 'sanitize_text_field' );
+		if ( empty( $items ) ) {
+			$errors[] = __( 'No entries were specified', 'formidable' );
```
truongwp, line 127: I think we can display errors here and return earlier. We also don't need the `$errors = array()` line because there is only one error can happen in this method.

classes/controllers/FrmEntriesController.php
```diff
@@ -104,6 +113,44 @@ private static function load_manage_entries_hooks() {
...
+		} else {
+			if ( $action === 'list-form' ) {
+				$request_bulkaction  = FrmAppHelper::get_param( 'bulkaction', '-1', 'request', 'sanitize_text_field' );
+				$request_bulkaction2 = FrmAppHelper::get_param( 'bulkaction2', '-1', 'request', 'sanitize_text_field' );
+				$bulkaction          = $request_bulkaction !== '-1' ? $request_bulkaction : $request_bulkaction2;
+			} else {
+				$bulkaction = str_replace( 'bulk_', '', $action );
+			}
+			if ( $bulkaction === 'delete' ) {
+				$frm_settings = FrmAppHelper::get_settings();
```
truongwp, line 137: This should be can when user doesn't have permission only.

classes/controllers/FrmEntriesController.php
```diff
@@ -104,6 +113,44 @@ private static function load_manage_entries_hooks() {
...
+
+				if ( ! is_array( $items ) ) {
+					$items = explode( ',', $items );
+				}
+				if ( ! current_user_can( 'frm_delete_entries' ) ) {
```
truongwp, line 142: This should be checked before exploding `$items`.

classes/controllers/FrmEntriesController.php
```diff
@@ -117,13 +164,127 @@ public static function route() {
...
+
+		return $wpdb->query( $entry_query ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+	}
+
+	/**
+	 * @since 4.02.04
+	 *
+	 * @param int         $form_id
+	 * @param bool|string $implode
+	 */
```
truongwp, line 261: This is missing `@return` tag. And all `@since` tags of new methods should be updated.

AbdiTolesa:
> Hi @AbdiTolesa! Could you please check this issue?
> 
> * When I try to bulk delete entries, the popup that requires typing “Delete All” appears at first. Then if I click "Cancel" or close the dialog box using the "X" icon, the popup no longer appears the next time I attempt to bulk delete entries, unless I reload the page. This happens in both the Pro and Lite versions.
>   Please check screen recording for reference: https://www.loom.com/share/9ad916e77e2442c5a11964f54737872f

Hey @lauramekaj1 That is fixed now.

classes/controllers/FrmEntriesController.php
```diff
@@ -104,6 +113,51 @@ private static function load_manage_entries_hooks() {
...
+				return;
+			}
+			if ( ! is_array( $items ) ) {
+				$items = explode( ',', $items );
+			}
+			if ( is_array( $items ) ) {
+				foreach ( $items as $item_id ) {
+					FrmEntry::destroy( $item_id );
+				}
+			}
```
truongwp, line 142: We can move these into a separate method to make it cleaner and reduce indentation.

truongwp (approved):

lauramekaj1:
> Hey @lauramekaj1 That is fixed now.

Hey @AbdiTolesa, I tested and it's working as expected. Thank you! 🚀

lauramekaj1 (approved):

classes/controllers/FrmEntriesController.php
```diff
@@ -117,13 +181,128 @@ public static function route() {
...
+
+	/**
+	 * @since x.x
+	 *
+	 * @param int $form_id
+	 */
+	private static function delete_form_entries( $form_id ) {
+		global $wpdb;
+
+		$form_ids = self::get_child_form_ids( $form_id );
```
Crabcyborg, line 256: Can we keep the child form logic in Pro? This isn't really relevant to Lite.

classes/helpers/FrmEntriesListHelper.php
```diff
@@ -257,6 +266,60 @@ protected function extra_tablenav( $which ) {
...
+		if ( ! apply_filters( 'frm_show_delete_all', current_user_can( 'frm_delete_entries' ), $form_id ) ) {
+			return;
+		}
+
+		$entries_count = ( isset( $args['entries_count'] ) ? $args['entries_count'] : 0 );
+		$verify        = ( isset( $args['bulk_delete_confirmation_message'] ) ? $args['bulk_delete_confirmation_message'] : '' );
+
+		?>
+		<span class="frm_uninstall">
+			<a href="<?php echo esc_url( wp_nonce_url( '?page=formidable-entries&frm_action=destroy_all' . ( $form_id ? '&form=' . absint( $form_id ) : '' ) ) ); ?>" class="button frm-button-secondary" data-loaded-from="entries-list" data-total-entries="<?php echo esc_attr( $entries_count ); ?>" data-frmverify="<?php echo esc_attr( $verify ); ?>" data-frmverify-btn="frm-button-red">
```
Crabcyborg, line 300: This form ID check is a bit concerning.

We never show the destroy all button on the All Entries page, right? I don't really think we should.

We likely only want to show this button where there is a valid `$form_id`.

classes/controllers/FrmEntriesController.php
```diff
@@ -216,16 +216,6 @@ public static function destroy_all() {
 		$form_id = (int) $params['form'];
 
 		if ( $form_id ) {
-			if ( self::should_trigger_on_delete_entry_actions( $form_id ) ) {
```
Crabcyborg: @AbdiTolesa You can't just delete this. We need hooks, and an update in Pro.

It also looks like you didn't update the JS at all. It should be in Pro as well.

This update isn't ready for review at all.

Crabcyborg: @AbdiTolesa We need any logic related to triggering delete actions in Pro.

There is no such thing as triggering actions on entry delete in Lite, so we should only have hooks here in Lite, and another update in Pro that uses those filters to keep the Pro-only functionality in-place.

classes/controllers/FrmEntriesController.php
```diff
@@ -127,15 +195,114 @@ public static function route() {
...
-				self::display_list();
-				return null;
+				return self::display_list();
+		}//end switch
+	}
+
+	/**
+	 * Delete all entries in a form when the 'delete all' button is clicked.
+	 *
+	 * @since x.x
```
truongwp: Please add `@return` tag.

classes/controllers/FrmEntriesController.php
```diff
@@ -127,15 +195,114 @@ public static function route() {
...
+
+	/**
+	 * @since x.x
+	 *
+	 * @param int         $form_id
+	 * @param bool|string $implode
+	 *
+	 * @return array|string
+	 */
+	private static function get_child_form_ids( $form_id, $implode = ',' ) {
```
truongwp: I don't think passing the second parameter and returning both `array` and `string` is a good idea. It can make us confused.

Should we return `array` here and use `implode()` where we use it?

Crabcyborg:
I don't believe this was ever implemented properly with support for the Pro functionality.

PR #2326, opened 2025-04-25
Add a text that clarifies whether a field column is for field value or label

AbdiTolesa:
CleanShot 2025-04-25 at 22 33 55@2x

### DIFF
--- a/classes/controllers/FrmEntriesController.php
+++ b/classes/controllers/FrmEntriesController.php
@@ -238,10 +238,29 @@ class FrmEntriesController {
 		$has_separate_value = ! FrmField::is_option_empty( $field, 'separate_value' );
 		$is_post_status     = FrmField::is_option_true( $field, 'post_field' ) && $field->field_options['post_field'] === 'post_status';
 		if ( $has_separate_value && ! $is_post_status ) {
-			$columns[ $form_id . '_frmsep_' . $col_id ] = FrmAppHelper::truncate( $field->name, 35 );
+			$columns[ $form_id . '_frmsep_' . $col_id ] = self::maybe_format_field_name_for_title( $field );
 		}
 
-		$columns[ $form_id . '_' . $col_id ] = FrmAppHelper::truncate( $field->name, 35 );
+		$columns[ $form_id . '_' . $col_id ] = self::maybe_format_field_name_for_title( $field, false );
+	}
+
+	/**
+	 * Appends "(Value)" or "(Label)" to the field name if it's an option field.
+	 *
+	 * @since x.x
+	 *
+	 * @param object $field
+	 * @param bool   $is_value
+	 *
+	 * @return string
+	 */
+	private static function maybe_format_field_name_for_title( $field, $is_value = true ) {
+		if ( ! in_array( $field->type, array( 'select', 'radio', 'checkbox' ), true ) ) {
+			return $field->name;
+		}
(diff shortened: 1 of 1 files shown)


classes/controllers/FrmEntriesController.php
```diff
@@ -238,10 +238,31 @@ private static function add_field_cols( $field, $form_id, &$columns ) {
...
+	 *
+	 * @since x.x
+	 *
+	 * @param object $field
+	 * @param bool   $has_separate_value
+	 * @param bool   $is_value
+	 *
+	 * @return string
+	 */
+	private static function maybe_format_field_name_for_title( $field, $has_separate_value, $is_value = true ) {
```
Crabcyborg, line 258: @AbdiTolesa It looks like this is only ever called when `$has_separate_value` is true. This could likely be simplified.

classes/controllers/FrmEntriesController.php
```diff
@@ -238,10 +238,31 @@ private static function add_field_cols( $field, $form_id, &$columns ) {
...
-			$columns[ $form_id . '_frmsep_' . $col_id ] = FrmAppHelper::truncate( $field->name, 35 );
+			$columns[ $form_id . '_frmsep_' . $col_id ] = self::maybe_format_field_name_for_title( $field, $has_separate_value );
 		}
 
-		$columns[ $form_id . '_' . $col_id ] = FrmAppHelper::truncate( $field->name, 35 );
+		$columns[ $form_id . '_' . $col_id ] = self::maybe_format_field_name_for_title( $field, $has_separate_value, false );
+	}
+
+	/**
+	 * Appends "(Value)" or "(Label)" to the field name if it's an option field it it has a separate value/label
```
truongwp, line 248: There is a typo here: `it it`.

truongwp (approved):
Thanks @AbdiTolesa. This looks good.

@Crabcyborg Do we need to shorten the field label because we append some string?

classes/controllers/FrmEntriesController.php
```diff
@@ -238,10 +238,31 @@ private static function add_field_cols( $field, $form_id, &$columns ) {
 		$has_separate_value = ! FrmField::is_option_empty( $field, 'separate_value' );
 		$is_post_status     = FrmField::is_option_true( $field, 'post_field' ) && $field->field_options['post_field'] === 'post_status';
 		if ( $has_separate_value && ! $is_post_status ) {
-			$columns[ $form_id . '_frmsep_' . $col_id ] = FrmAppHelper::truncate( $field->name, 35 );
+			$columns[ $form_id . '_frmsep_' . $col_id ] = self::maybe_format_field_name_for_title( $field, $has_separate_value );
 		}
 
-		$columns[ $form_id . '_' . $col_id ] = FrmAppHelper::truncate( $field->name, 35 );
+		$columns[ $form_id . '_' . $col_id ] = self::maybe_format_field_name_for_title( $field, $has_separate_value, false );
```
Crabcyborg, line 244: @AbdiTolesa It seems to me that the condition isn't really `$has_separate_value`, but `$has_separate_value && ! $is_post_status` (above).

When it is post status, it doesn't add a second column, so there isn't really a need in adding `(Value)` in that case.

Maybe another variable would be more clear. Something like `$include_column_for_sep_val = $has_separate_value && ! $is_post_status;`?

PR #2337, opened 2025-04-29
Add option validation

Crabcyborg:
More details in Slack https://strategy11.slack.com/archives/CJFQ599V5/p1745951863090849

**This update will throw errors if an option is used that isn't defined in the form builder data**

**To see the error** - Edit the form HTML and change the value to something like "INVALID". Select that option, then submit. It should throw an error.

**What to look for when testing**
- That no unexpected errors occur.
- That no invalid options can be used.
- Try validation for many field types.

I added a `frm_option_is_valid` filter to allow for customization. If I detect that the option values were modified using the `frm_setup_new_fields_vars` filter, it doesn't validate, to avoid issues where the field options could be based on a condition that we can't replicate when submitting form data.

To revert, you can use `add_filter( 'frm_option_is_valid', '__return_true' );`.

Documentation examples for changing options dynamically are available here https://formidableforms.com/knowledgebase/frm_setup_new_fields_vars/

**Pre-release**
[formidable-6.20.1b.zip](https://github.com/user-attachments/files/19965805/formidable-6.20.1b.zip)


lauramekaj1 (approved):
@Crabcyborg I tested this and it's working as expected. Thank you! 🚀

Crabcyborg:
Thank you Laura!

🙌

PR #2346, opened 2025-05-06
Range Slider project -- Required Lite update

AbdiTolesa:
This PR is required for: https://github.com/Strategy11/formidable-pro/pull/5783


js/formidable_admin.js
```diff
@@ -1604,15 +1646,7 @@ function frmAdminBuildJS() {
 
 		jQuery.ajax({
 			type: 'POST', url: ajaxurl,
-			data: {
-				action: 'frm_insert_field',
-				form_id: formId,
-				field_type: fieldType,
-				section_id: sectionId,
```
truongwp, line 1611: The `section_id` is set to `0` in this PR. Is this intentional?

js/formidable_admin.js
```diff
@@ -1571,12 +1572,53 @@ function frmAdminBuildJS() {
...
+		const args = {
+			action: 'frm_insert_field',
+			form_id: formId,
+			field_type: fieldType,
+			section_id: sectionId,
+			nonce: frmGlobal.nonce,
+			has_break: hasBreak,
+			last_row_field_ids: getFieldIdsInSubmitRow()
+		};
+		return wp.hooks.applyFilters( 'frm_insert_field_args', args, fieldType );
```
truongwp, line 1596: We should add `sectionId`, `formId`, `hasBreak` to the filter params too. We should create an object for them.

AbdiTolesa: @truongwp Those params are already accessible through `args` right? Did I probably miss something?

truongwp: Another function hooked into that filter can change those data in `args`. We want to pass the original ones.

js/formidable_admin.js
```diff
@@ -1571,12 +1572,54 @@ function frmAdminBuildJS() {
...
+	 * Returns true if it's a range field type and slider type is not selected.
+	 *
+	 * @since x.x
+	 *
+	 * @param {string} fieldType
+	 * @param {Event}  event
+	 * @returns {boolean}
+	 */
+	function shouldStopInsertingField( fieldType, event ) {
+		return 'range' === fieldType && event.originalEvent?.detail?.showModal !== 0 && builderPage.dataset.supportsRangeSlider === '1';
```
Crabcyborg, line 1610: @AbdiTolesa I think this should would likely not always work.

Checking `event.originalEvent?.detail?.showModal !== 0` seems problematic since it isn't accounting for the `null` values.

Something like `( event.originalEvent?.detail?.showModal ?? 0 ) !== 0` should catch that nulls.

Crabcyborg (approved):
Thanks @AbdiTolesa!

I think these changes are ready to merge.

🚀

PR #2347, opened 2025-05-07
Improve honeypot compatibility with API loaded forms

Crabcyborg:
I'm seeing issues when using `[frm-api]` and the `<script>` tag to include an API loaded form, where the honeypot is appearing because it looks like the footer function is never called.

This update effectively disables the honeypot if it detects that the form is loaded from an API.

What do you think @truongwp? There's probably a better fix. I want to keep this contained to this plugin as well for now if possible.

### DIFF
--- a/classes/models/FrmHoneypot.php
+++ b/classes/models/FrmHoneypot.php
@@ -150,7 +150,40 @@ class FrmHoneypot extends FrmValidate {
 	 * @return bool
 	 */
 	public function should_render_field() {
+		if ( $this->form_is_loaded_by_api() ) {
+			return false;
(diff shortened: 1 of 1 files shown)


truongwp:
What do you think if we print the `<style>` in the form HTML?

Crabcyborg:
@truongwp Sure! I think that sounds good for now.

truongwp:
Ok. I will do that. I noted this in the Team QA issues.

Crabcyborg:
Thank you @truongwp!

I'll close this since this shouldn't be needed.

PR #2352, opened 2025-05-09
Also print honeypot css on gutenberg editor

Crabcyborg:
I'm also changing the `0` `strpos` check because I'm testing with WPML and I'm seeing `/fr/` in the front of some URLs which breaks that check.


truongwp (approved):

Crabcyborg:
Thanks Truong!

PR #2356, opened 2025-05-10
Try alternative way of getting honeypot field

Crabcyborg:


classes/models/FrmHoneypot.php
```diff
@@ -67,14 +44,18 @@ public function validate() {
 	private function is_honeypot_spam() {
 		$is_honeypot_spam = $this->is_legacy_honeypot_spam();
 		if ( ! $is_honeypot_spam ) {
-			$field_id         = $this->get_honeypot_field_id();
+
+			$field_id = $this->get_honeypot_field_id();
+			if ( ! $field_id ) {
+				return false;
```
truongwp, line 50: This will skip the filter hook `frm_process_honeypot`. Should we run it in this case?

classes/views/frm-entries/form.php
```diff
@@ -61,6 +61,13 @@
 <?php } ?>
 <?php
 if ( ! $only_contain_submit ) {
+	$max_field_id = 0;
+	foreach ( $values['fields'] as $field ) {
+		if ( $field['id'] > $max_field_id ) {
```
truongwp, line 66: I think we can use `$max_field_id = max( $field['id'], $max_field_id )`.

classes/models/FrmHoneypot.php
Crabcyborg: I think it's okay. That filter is mostly for turning the honeypot off.

classes/models/FrmHoneypot.php
```diff
@@ -111,10 +92,17 @@ private function check_honeypot_filter() {
 	 *
 	 * @return void
 	 */
-	public static function maybe_render_field( $form_id ) {
+	public static function maybe_render_field( $form_id, $honeypot_field_id ) {
```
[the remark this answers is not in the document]
Crabcyborg: Thanks! Updated.

classes/views/frm-entries/form.php
Crabcyborg: Thanks! I've updated this now.

truongwp (approved):

truongwp:
One small thing, please remove this `@since`: https://github.com/Strategy11/formidable-forms/pull/2356/files#diff-9ffbec27738838ff9ec46a0e3d28960f34eddf58858a083aae1c52f7c1496f04L83

PR #2357, opened 2025-05-10
Denylist spam performance improvement

truongwp:
If denylist spam is disabled, the get fields call should not be run.


Crabcyborg (approved):
Thank you @truongwp!

This looks good!

PR #2361, opened 2025-05-13
Denylist spam check improvements

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/5801

Changes:
- Added a toggle to disable denylist spam check.
- All spam checks now don't run when importing.
- Denylist spam check doesn't check password, captcha, signature, radio, checkbox, select field types. But it still check the Other value of radio, checkbox, select field types. To skip for Other value, add `other` to the `skip_field_types` of denylist.


Crabcyborg (approved):
Thanks @truongwp!

This looks good to me.

🚀

PR #2363, opened 2025-05-14
Show spam keywords detected on spam settings page

Crabcyborg:


classes/models/FrmSpamCheckDenylist.php
```diff
@@ -260,6 +260,7 @@ protected function single_line_check_values( $line, $args ) {
 			foreach ( $values_to_check as $value ) {
 				$value = $this->convert_to_lowercase( $value );
 				if ( $line === $value ) {
+					self::add_spam_keyword_to_option( $line );
```
truongwp, line 263: I think this shouldn't be here. It should be on line 173: https://github.com/Strategy11/formidable-forms/pull/2363/files#diff-4c7ffc5a1ccacdb8d25cc36d707ca5b65b51f5ef8575e833050954470b25df1dL173

classes/models/FrmSpamCheckDenylist.php
```diff
@@ -455,6 +456,8 @@ protected function read_lines_and_check( $file_path, $callback, $callback_args =
 
 			$is_spam = $callback( $line, $callback_args );
 			if ( $is_spam ) {
+				self::add_spam_keyword_to_option( $line );
```
truongwp, line 459: This will add the spam IP as well. Do we want to do this?

Crabcyborg: What is the best way to tell if it an IP?

truongwp: We can check the $callback, or $callback_args. They are different between IP check and word check.

Crabcyborg: Thanks @truongwp! I added a check for `$callback[1]`. What do you think?

https://github.com/Strategy11/formidable-forms/pull/2363/commits/1abe402472d8ed1619ef31de5e1587b5e5a67815

Crabcyborg:
@truongwp Currently that's where I have it. I've asked Razvan about his opinion in Slack about the position but haven't heard.

My thoughts are that we want it close to the allowed words, so they know what to add.

classes/views/frm-settings/captcha/captcha.php
```diff
@@ -135,6 +135,17 @@
 </p>
 
 <p>
+	<?php
+	$transient = get_transient( 'frm_recent_spam_detected' );
+	if ( ! empty( $transient ) ) {
+		?>
+		<div class="frm_note_style">
+			<strong><?php esc_html_e( 'Spam keywords detected recently:', 'formidable' ); ?></strong>
+			<i><?php echo esc_html( '&nbsp;' . implode( ', ', $transient ) ); ?></i>
```
truongwp, line 144: Do we need `$nbsp;`? There has already been a space between `<strong>` and `<i>`.

Crabcyborg: Thanks for catching that. I removed it.

Also, I've moved the note below the setting now, after talking to Razvan.

PR #2364, opened 2025-05-14
Add filter for honeypot field ID

Crabcyborg:
Related ticket https://github.com/Strategy11/formidable-pro/issues/5802

This allows someone to pick a custom honeypot ID, in case the field ID is also loaded on the form.

### DIFF
--- a/classes/models/FrmHoneypot.php
+++ b/classes/models/FrmHoneypot.php
@@ -96,6 +96,15 @@ class FrmHoneypot extends FrmValidate {
 			return;
 		}
 
+		/**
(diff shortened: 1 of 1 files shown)


truongwp:
@Crabcyborg What do you think if we get the latest field ID (in all form), then plus 1 to get the honeypot field ID? So that field doesn't exist.

In case multiple forms on a page, we can store the previous honeypot field ID in a global variable, then plus 1 for latter forms?

Crabcyborg:
Thanks @truongwp. I made a new PR https://github.com/Strategy11/formidable-forms/pull/2366/files

PR #2373, opened 2025-05-23
Fix inline modal open error after field delete

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5810

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -4778,6 +4778,14 @@ function frmAdminBuildJS() {
 				if ( settings.is( ':visible' ) ) {
 					document.getElementById( 'frm_insert_fields_tab' ).click();
 				}
+				const openModal = settings[0].querySelectorAll( '.frm-inline-modal' );
+				if ( openModal.length ) {
+					openModal.forEach( modal => {
+						modal.classList.add( 'frm_hidden' );
+						modal.removeAttribute( 'data-fills' );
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -4761,6 +4761,26 @@ function frmAdminBuildJS() {
...
+	/**
+	 * Appends open modals to the field options form.
+	 *
+	 * @since x.x
+	 *
+	 * @param {Object} settings
+	 * @returns {void}
+	 */
+	function appendOpenModalsToFieldOptionsForm( settings ) {
+		const openModals = settings[0].querySelectorAll( '.frm-inline-modal' );
```
truongwp, line 4773: We can update the query to only get modal with `data-fills`.

The function isn't clear to me. I prefer `moveOpenModalsOutOfFieldOptions()`.

AbdiTolesa: I have updated the function name, the other looks good to me as it is.

truongwp: The open modal will have the `data-fills` attribute, so we can update this query to only get open modal.

Crabcyborg (approved):
This looks good!

I haven't tested it yet though. I'll merge once Laura approves.

🚀

Crabcyborg:
Thank you Laura!

🚀

PR #2385, opened 2025-06-06
Still load honeypot js when loading formidable js

Crabcyborg:
I was seeing too much flickering of the honeypot when it loads so late.

This new update tracks what was already hidden with CSS, so it doesn't keep adding the same CSS twice (which happens if you have multiple forms).

This update calls the honeypot render twice, so it happens early and late, so it catches everything without the flickering. If there were no more forms the second time, it doesn't print more CSS.


truongwp (approved):
This looks good. Thanks @Crabcyborg.

PR #2396, opened 2025-06-19
fix the timeline view option on import

Liviu-p:
Fixes https://github.com/Strategy11/formidable-views/issues/650


Crabcyborg:
Thank you Laura!

PR #2398, opened 2025-06-19
Fix JS error when interacting with field options in builder

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5851

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -5011,7 +5011,7 @@ function frmAdminBuildJS() {
 	 * @returns {Object}
 	 */
 	function getChoiceOldValueAndLabel( choiceElement ) {
-		const usingSeparateValues   = choiceElement.closest( '.frm-single-settings' ).querySelector( '.frm_toggle_sep_values' ).checked;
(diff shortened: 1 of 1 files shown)


js/formidable_admin.js
```diff
@@ -5011,7 +5011,7 @@ function frmAdminBuildJS() {
 	 * @returns {Object}
 	 */
 	function getChoiceOldValueAndLabel( choiceElement ) {
-		const usingSeparateValues   = choiceElement.closest( '.frm-single-settings' ).querySelector( '.frm_toggle_sep_values' ).checked;
+		const usingSeparateValues   = choiceElement.closest( '.frm-single-settings' ).querySelector( '.frm_toggle_sep_values' )?.checked;
```
Crabcyborg, line 5014: @AbdiTolesa Can we make sure this is still a boolean? If it doesn't exist, then `usingSeparateValues` would be `undefined`.

Something like `...?.checked ?? false` should be fine.

Crabcyborg (approved):
Thanks @AbdiTolesa!

PR #2401, opened 2025-06-21
Use "found" parameter in wp_cache_get

andrewklimek:
I've seen views that make heavy use of frm-stats where this returns hundreds of empty arrays that don't get cached, because the current check which won't cache empty arrays.


Crabcyborg (approved):
Thank you @andrewklimek.

I think this change looks good.

I'll merge this now.

🚀

PR #2403, opened 2025-06-25
Update no entries help message

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5777


Crabcyborg:
@AbdiTolesa When I look at this again, I think this is really intentional. The documentation it directs you to is on how to take your form and put it somewhere where it would then get entries. When someone first makes their form, and sees that they have no entries, they might not know how to put their form on a page to draw attention to it to get new entries.

PR #2406, opened 2025-07-01
Add an action that triggers before field label option in the builder

AbdiTolesa:


Crabcyborg (approved):
Thanks @AbdiTolesa! I think this looks good to merge.

🚀

PR #2408, opened 2025-07-07
fix deleteEntries.cy.js

lauramekaj1:


lauramekaj1:
@Crabcyborg The test was failing due to a change in a selector. I've updated the test accordingly, and it's now passing.

PR #2420, opened 2025-07-10
Fix can't deactivate plugin when popup has limited height

truongwp:
User shared a screenshot:
image


Crabcyborg (approved):
Thanks @truongwp!

PR #2421, opened 2025-07-10
Improve tables accessibility

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-quizzes/issues/229

### DIFF
--- a/classes/models/FrmTableHTMLGenerator.php
+++ b/classes/models/FrmTableHTMLGenerator.php
@@ -258,7 +258,7 @@ class FrmTableHTMLGenerator {
 	 * @return string
 	 */
 	public function generate_table_header() {
-		return '<table cellspacing="0" role="presentation" ' . $this->table_style . '><tbody>' . "\r\n";
+		return '<table ' . $this->table_style . '><tbody>' . "\r\n";
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


classes/models/FrmTableHTMLGenerator.php
```diff
@@ -287,8 +287,8 @@ public function generate_two_cell_table_row( $label, $value ) {
 		$row .= $this->add_row_class( $value === '' );
 		$row .= '>';
 
-		$label = '<th' . $this->td_style . '>' . wp_kses_post( $label ) . '</th>';
-		$value = '<td' . $this->td_style . '>' . wp_kses_post( $value ) . '</td>';
+		$label = '<th scope="row" role="presentation" ' . $this->td_style . '>' . wp_kses_post( $label ) . '</th>';
```
Crabcyborg, line 290: @AbdiTolesa

I've been doing some research on table accessibility, and I don't think we want to use `role="presentation"` in most cases. When it is added, it tells the screen reader that the table is not a table. It's not meant for actual tables with data relationships defined in the `th`/`td` tags.

I think we do want this specifically for the customer's table in quizzes, which really is just a single score and not really a table of data.

Screen Shot 2025-07-24 at 5 40 11 PM

AbdiTolesa: Ok, I'll add `role="presentation"` in Quiz but I'm not sure what you mean by `customer's table`? Is that `Show user's score` option in the Quiz action? That looks a bit different to your screenshot for me.

Crabcyborg:
@AbdiTolesa It looks like they're showing the correct answers. There's just a lot of CSS to customize the appearance of it.

If I strip the custom CSS I get it down to this.
Screen Shot 2025-07-25 at 2 29 00 PM

Screen Shot 2025-07-25 at 2 24 25 PM

They're really manipulating the table so it isn't a table anymore. If I remove that CSS, you can see our tables are really still there.
Screen Shot 2025-07-25 at 2 35 18 PM

It looks like we use valid tables in the Quizzes add-on. It seems to me that their accessibility issue is because of their custom CSS.

So I think we don't need to add `role="presentation"` anywhere.

PR #2428, opened 2025-07-16
add filter that will extended the form field container html attribues…

Liviu-p:
…, add js filter that will get fired after a field is successfully added in the form build, export 2 js functions that are related to inserting a new field and deleting it from form builder in order to reuse these in other plugins, toggle a jQuery notice if jQuery is selected as library from general settings

### DIFF
--- a/classes/controllers/FrmFieldsController.php
+++ b/classes/controllers/FrmFieldsController.php
@@ -64,12 +64,13 @@ class FrmFieldsController {
 		FrmAppHelper::permission_check( 'frm_edit_forms' );
 		check_ajax_referer( 'frm_ajax', 'nonce' );
 
-		$field_type = FrmAppHelper::get_post_param( 'field_type', '', 'sanitize_text_field' );
-		$form_id    = FrmAppHelper::get_post_param( 'form_id', 0, 'absint' );
+		$field_type    = FrmAppHelper::get_post_param( 'field_type', '', 'sanitize_text_field' );
+		$form_id       = FrmAppHelper::get_post_param( 'form_id', 0, 'absint' );
+		$field_options = FrmAppHelper::get_post_param( 'field_options', array(), 'wp_kses_post' );
 
 		do_action( 'frm_before_create_field', $field_type, $form_id );
 
-		$field = self::include_new_field( $field_type, $form_id );
+		$field = self::include_new_field( $field_type, $form_id, $field_options );
 
 		// this hook will allow for multiple fields to be added at once
 		do_action( 'frm_after_field_created', $field, $form_id );
@@ -82,11 +83,16 @@ class FrmFieldsController {
 	 *
 	 * @param string $field_type
 	 * @param int    $form_id
+	 * @param array  $field_options
 	 *
 	 * @return array|false
 	 */
-	public static function include_new_field( $field_type, $form_id ) {
+	public static function include_new_field( $field_type, $form_id, $field_options = array() ) {
 		$field_values = FrmFieldsHelper::setup_new_vars( $field_type, $form_id );
+
+		if ( ! empty( $field_options ) ) {
+			$field_values['field_options'] = array_merge( $field_values['field_options'], $field_options );
+		}
 
 		// When a new field is added to the form, flag it as draft and hide it from the front-end.
 		$field_values['field_options']['draft'] = 1;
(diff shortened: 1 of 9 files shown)


classes/controllers/FrmFieldsController.php
```diff
@@ -315,6 +332,10 @@ public static function load_single_field_settings( $atts ) {
 			$field['read_only'] = false;
 		}
 
+		if ( ! isset( $field['range_field'] ) ) {
+			$field['range_field'] = false;
```
Crabcyborg, line 336: @Liviu-p Is there a way to do this without adding this into Lite? Ideally Lite shouldn't include any references to "range fields".

classes/views/frm-fields/back-end/settings.php
```diff
@@ -426,7 +425,7 @@
 
 				if ( $display['unique'] ) {
 					?>
-					<p class="frm_unique_details<?php echo esc_attr( $field['id'] . ( $field['unique'] ? '' : ' frm_hidden' ) ); ?>">
+					<p class="frm_unique_details<?php echo esc_attr( $field['id'] . ( $field['unique'] ? '' : ' frm_hidden' ) ); ?> <?php echo 'time' === $field['type'] ? 'frm-has-linked-date-field' : ''; ?>">
```
Crabcyborg, line 428: @Liviu-p Can we do this with a hook or something instead? Ideally we wouldn't have any references to time fields like this in Lite.

classes/views/frm-forms/add_field.php
```diff
@@ -3,7 +3,7 @@
 	die( 'You are not allowed to call this page directly.' );
 }
 ?>
-<li id="frm_field_id_<?php echo esc_attr( $field['id'] ); ?>" class="<?php echo esc_attr( $li_classes ); ?>" data-fid="<?php echo esc_attr( $field['id'] ); ?>" data-formid="<?php echo esc_attr( 'divider' === $field['type'] && isset( $field['form_select'] ) ? $field['form_select'] : $field['form_id'] ); ?>" data-ftype="<?php echo esc_attr( $display['type'] ); ?>" data-type="<?php echo esc_attr( $field['type'] ); ?>">
+<li id="frm_field_id_<?php echo esc_attr( $field['id'] ); ?>" class="<?php echo esc_attr( $li_classes ); ?>" data-fid="<?php echo esc_attr( $field['id'] ); ?>" data-formid="<?php echo esc_attr( 'divider' === $field['type'] && isset( $field['form_select'] ) ? $field['form_select'] : $field['form_id'] ); ?>" data-ftype="<?php echo esc_attr( $display['type'] ); ?>" data-type="<?php echo esc_attr( $field['type'] ); ?>" <?php echo esc_attr( $extra_field_attributes ); ?>>
```
Crabcyborg, line 6: I don't believe `echo esc_attr( $extra_field_attributes );` really works.

You end up with `&quot;` instead of `"` characters.

We could possibly just call `do_action` here instead and echo from Pro.

js/admin/settings.js
```diff
@@ -13,6 +13,11 @@
 		if ( 'frm_currency' === e.target.id) {
 			syncCurrencyOptions( e.target );
 		}
+
+		if ( 'frm_datepicker_library' === e.target.id ) {
+			toggleDatepickerJqueryRangeSupportNoteOnChange( e );
```
Crabcyborg, line 18: Is it possible to listen for this in Pro JS instead? We don't have a datepicker library in Lite.

js/formidable_admin.js
```diff
@@ -2082,12 +2104,91 @@ function frmAdminBuildJS() {
...
+
+			if ( 'summary' === fieldType ) {
+				hasBreak = $newFields.children( 'li[data-type="break"]' ).length > 0 ? 1 : 0;
+			}
+
+			jQuery.ajax({
+				type: 'POST',
+				url: ajaxurl,
+				data: {
+					action: 'frm_insert_field',
```
Crabcyborg, line 2146: @Liviu-p I just merged Abdi's new range slider requirements into master, and I've merged those updates into this branch.

It includes a new `getInsertNewFieldArgs` function we can call here. We could be able to do something like:

```
const data = getInsertNewFieldArgs( fieldType, sectionId, formId, hasBreak );
data.field_options: fieldOptions;
```

classes/controllers/FrmFieldsController.php
Liviu-p: yep, that's not used anymore, I removed it.

classes/views/frm-forms/add_field.php
Liviu-p: sure @Crabcyborg, I switched to FrmAppHelper::array_to_html_params

js/admin/settings.js
Crabcyborg: I've dropped this for now. I'll try to add it back in the Pro PR shortly.

PR #2431, opened 2025-07-21
Fix undefined array key when using default value with Slider fields

AbdiTolesa:
### Test steps

1. Add a Slider field to your form
2. Set a default value in the field options
3. Save the builder
4. Confirm that `PHP Warning:  Undefined array key "placeholder" ...` error is not triggered

### DIFF
--- a/classes/controllers/FrmFieldsController.php
+++ b/classes/controllers/FrmFieldsController.php
@@ -730,7 +730,7 @@ class FrmFieldsController {
 	 * @return void
 	 */
 	private static function add_frmval_to_input( $field, &$add_html ) {
-		if ( $field['placeholder'] != '' ) {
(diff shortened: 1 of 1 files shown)


classes/controllers/FrmFieldsController.php
```diff
@@ -730,15 +730,15 @@ private static function add_placeholder_to_input( $field, &$add_html ) {
 	 * @return void
 	 */
 	private static function add_frmval_to_input( $field, &$add_html ) {
-		if ( $field['placeholder'] != '' ) {
+		if ( isset( $field['placeholder'] ) && $field['placeholder'] != '' ) {
```
Crabcyborg, line 733: @AbdiTolesa Are there more steps required to replicate this? I tried adding a range field, and it never throws a warning here.

PR #2434, opened 2025-07-22
Fix currency format stays selected

Crabcyborg:
This fixes an issue I was seeing, where "Currency"/"Number" would stay selected after the format options were selected and the upgrade modal was shown.

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -5437,12 +5437,17 @@ function frmAdminBuildJS() {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


shervElmi:
Thanks @Crabcyborg for catching and working on this issue.

I found that the changes prevent the Upgrade modal from showing when the Format setting is changed to currency/format.

CleanShot 2025-07-22 at 22 21 58

Crabcyborg:
Thanks Sherv!

PR #2444, opened 2025-07-30
Fix js error when deleting a field group

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5917

I wasn't able to replicate this issue but I believe adding extra checks would fix the issue.
Image

### DIFF
--- a/js/formidable_admin.js
+++ b/js/formidable_admin.js
@@ -4783,7 +4783,7 @@ function frmAdminBuildJS() {
 	 * @returns {void}
 	 */
 	function moveOpenModalsOutOfFieldOptions( settings ) {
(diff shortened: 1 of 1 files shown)


Crabcyborg:
Thanks for working on this @AbdiTolesa.

I'm able to replicate this now after some help from Sherv. The issue is that he deleted another field first, and it sets a `data-deletefield` attribute. This doesn't get removed, so when you delete a field group, it would retry deleting the field that was already deleted (or one that you had previously cancelled).

https://github.com/Strategy11/formidable-forms/pull/2455 fixes the root of the issue.

I'm going to close this as I don't think we want to fix it this way. If a field is getting deleted, and the settings don't exist, the root of the issue is that we're trying to delete a non-existent field, which is something we should investigate.

PR #2445, opened 2025-07-30
Fix spam check issue with backslashes

Crabcyborg:
I noticed that some strings in our `splorp-wp-comment.txt` file were not throwing spam errors.

This is because of the `maybe_json_encode` call, which changes the array into JSON, where the slashes are escaped.

As a basic PHP example:
```
var_dump( json_encode( [ 'key' => '/joomla/' ] ) );
```

Outputs
```
string(20) "{"key":"\/joomla\/"}" 
```

**Before**
`/joomla/` would not trigger a denylist error.

**After**
Screen Shot 2025-07-30 at 3 46 56 PM


truongwp (approved):
@Crabcyborg This looks good. Thanks for working on this.

PR #2453, opened 2025-08-08
Rock: Email Gathering Modal - Form Templates

shervElmi:
## Linear

- **Handoff:** https://linear.app/strategy11/document/handoff-667bc26a70e5
- Design: https://www.figma.com/design/x8Epd7CJLLY6JIyPoCMPvc/Form-Templates?node-id=3617-395&t=hWNJus6WNJqC1WB6-1
- Technical Pitch: https://linear.app/strategy11/document/technical-pitch-a2549ade6d5e

## Screen recording demo

https://share.cleanshot.com/VqNGfzmQ


shervElmi:
@Crabcyborg, after I merged this branch with the new ESLint and fixed the ESLint errors, I didn’t do QA because I’ve been sick. Please don’t merge it until I’m back.

js/src/core/ui/counter.js
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	element._counterLastTimestamp = timestamp;
+
+	// Continue animation or finish
+	if ( progress < 1 ) {
+		element._counterAnimation = requestAnimationFrame( timestamp =>
+			animateCounter( timestamp, element, startValue, targetValue, duration, change, easing )
+		);
+	} else {
```
Crabcyborg, line 124: @shervElmi Could we return in the `if` here so we don't need to indent the last block of code here?

classes/models/FrmFormTemplateApi.php
```diff
@@ -44,7 +44,7 @@ protected function set_cache_key() {
 	protected function api_url() {
 		$url = self::$base_api_url . 'list';
 
-		if ( empty( $this->license ) ) {
+		if ( empty( $this->license ) && self::get_free_license_code() ) {
```
Crabcyborg, line 47: @shervElmi We might need to do this differently.

I think we want to still always include this license in the URL, then possibly filter out some results based on if they've provided an email.

In our welcome tour we're going to suggest that people try a template right away, without blocking them with the extra step of requiring an email.

The plan is to not block "Contact Us" and "Stripe Payment" but to still block the rest.

js/src/core/ui/counter.js
shervElmi: [Use early return in animateCounter to reduce indentation](https://github.com/Strategy11/formidable-forms/pull/2453/commits/d7f2a009752ed6580541e686a6d15c466f23965a)

resources/scss/admin/frm_admin.scss
```diff
@@ -269,9 +269,10 @@
...
 @import 'components/square';
-@import 'components/sales/banner';
+@import 'components/banner/sale-banner';
+@import 'components/banner/get-free-templates-banner';
 
 /**
  * Backwards Compatibility
  */
-@import 'backwards-compatibility';
+ @import 'backwards-compatibility';
```
Crabcyborg, line 278: It looks like this space was added accidentally.

shervElmi:
@Crabcyborg and @lauramekaj1 it looks like the E2E test needs to be updated since we changed our strategy for the initial free templates list.

PR #2455, opened 2025-08-08
Fix field delete getting called for already deleted field bug

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/5917

The issue is that the `data-deletefield` was never getting removed, so it would also try to remove that field when deleting a field group because it uses the same confirmation button.


Crabcyborg:
Thanks Sherv!

PR #2464, opened 2025-08-18
Refactor admin JavaScript build system and structure

shervElmi:
This PR modernizes the admin JavaScript codebase by:

### Key Benefits

- **Minification & Optimization**: The large `formidable_admin.js` file is now automatically minified and optimized through webpack's production build process
- **Modularization Foundation**: Sets up the groundwork for breaking down the monolithic admin JavaScript into smaller, manageable modules

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -1599,6 +1599,9 @@ class FrmFieldsHelper {
 			$replace_with[] = 'field_id="' . $new . '"';
 			$replace[]      = 'field_id=\"' . $old . '\"';
 			$replace_with[] = 'field_id=\"' . $new . '\"';
+			// This covers conditional logic.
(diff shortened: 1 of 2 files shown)


Crabcyborg (commented):
@shervElmi We need to make sure the new file is being linted as well. I don't want more files that aren't getting validated.

It looks like I can add `./node_modules/.bin/eslint ./js/src/admin/admin.js` to our `jscs.yml`, but if I try to add another file, like `./node_modules/.bin/eslint ./js/src/styles.js`, it seems to have issues. I think because of the `import` lines.

I see this error instead:
```
Cannot find module '.../wp-content/plugins/formidable/node_modules/@wordpress/i18n/webpack.config.js
```

Could you help with fixing that?

Crabcyborg (approved):
Thanks @shervElmi!

I think we can merge this now.

🚀

PR #2466, opened 2025-08-19
Rock: Welcome Tour

shervElmi:
## Linear

- **Handoff:** https://linear.app/strategy11/document/handoff-bdb6bb0e37bf
- **Design**: https://www.figma.com/design/jqOxr9w1AyoJmmmCauCH0t/Welcome-Tour?node-id=3979-2699&t=M08Z5FwTjw9O2lZd-4
- Technical Pitch: https://linear.app/strategy11/document/technical-pitch-9e3b124ac810

## Screen recording demo

https://share.cleanshot.com/WKt5rXZk

### DIFF
--- a/js/src/welcome-tour/elements/beginTourModalElement.js
+++ b/js/src/welcome-tour/elements/beginTourModalElement.js
@@ -53,7 +53,7 @@ function createBeginTourModalElement() {
 }
 
 /**
- * Inject begin tour modal elements into the DOM and the elements object.
+ * Add the begin tour modal to the elements registry.
  *
  * @private
  * @param {HTMLElement} beginTourModal The begin tour modal element.


shervElmi:
@Crabcyborg, I’m wondering if I should add the welcome tour data to the tracking system or not?

Crabcyborg:
> I’m wondering if I should add the welcome tour data to the tracking system or not?

@shervElmi I think we want to know which steps are completed, and if the tour was dismissed. It was be nice to know which step it was dismissed at.

We probably could use `FrmUsageController::update_flows_data` for this.

shervElmi:
@Crabcyborg,

I’ve added usage tracking based on your comment. I also have a suggestion:

Since the plugin shows the checklist until it’s dismissed, there are only two states: dismissed or completed. When a user completes it, it means all steps have been finished.

For tracking all completed steps, I’ve added this:

```php
foreach ( self::$checklist['completed_steps'] as $step_key => $completed ) {
    FrmUsageController::update_flows_data( 'welcome_tour_completed_steps', $step_key );
}
```

However, it could alternatively be:

```php
FrmUsageController::update_flows_data( 'welcome_tour', 'completed' );
```

And for dismissing, instead of the current implementation:

```php
FrmUsageController::update_flows_data( 'welcome_tour_dismissed_step', self::$checklist['active_step_key'] );
```

We could use:

```php
FrmUsageController::update_flows_data( 'welcome_tour', 'dismissed: ' . self::$checklist['active_step_key'] );
```

truongwp:
@shervElmi I have some questions:

1. Can user complete the items in the checklist NOT following the order? For example, can user complete "Embed in a page" without "Style your form"?
2. We do we save the completed steps, or do we save them?

shervElmi:
Thanks for your comment, @truongwp.

1. It’s possible under certain conditions.
2. We save the completed steps and some other data in the DB wp_options table.

Based on your comment, I got your point, and here’s what I’m thinking now:

There are only two states in the Welcome Tour, **the user must either complete all steps or dismiss it**.

1. **Dismissed state**: I can use the following code to save user data:

   ```php
   FrmUsageController::update_flows_data( 'welcome_tour', 'dismissed: ' . self::$checklist['active_step_key'] );

   foreach ( self::$checklist['completed_steps'] as $step_key => $completed ) {
       FrmUsageController::update_flows_data( 'welcome_tour', $step_key );
   }
   ```

2. **Completed state**: I can just save the completed steps using the code below:

   ```php
   foreach ( self::$checklist['completed_steps'] as $step_key => $completed ) {
       FrmUsageController::update_flows_data( 'welcome_tour', $step_key );
   }
   ```

Crabcyborg:
> The welcome tour will also appear for existing users.

Thanks @shervElmi.

I'm running this by Dalton now. I'm not sure.

truongwp:
@shervElmi The `add_filter` to the snapshot data should run globally, not just in those methods. The snapshot is grabbed daily via cron job.

For the complete steps, I like adding it to the `$snapshot['settings']`, and one key per step, it will be easier for us to do statistics.
```
['settings'][0]['welcome_tour_step_1'] = 1; // completed
['settings'][0]['welcome_tour_step_2'] = 0; // incompleted
```
We should also add a flag to know whether this site showed the welcome tour or not.

We should save which step users click on the "Dismiss" in the DB.

We might want to track which link users click on after completing the welcome tour.

I know this is complex. I can take this usage tracking part. Do you want that?

truongwp:
@shervElmi I have another suggestion:

We should add `@since` tag to the PHPDOC of new files to mark when those files are added.
For the new `FrmWelcomeTourController`, since you added `@since` to the class, its methods don't need `@since` anymore, unless you add a new method in future versions.

classes/views/welcome-tour/checklist.php
```diff
@@ -0,0 +1,33 @@
+<?php
+/**
+ * Welcome Tour's Checklist component.
+ *
```
truongwp: @shervElmi We should add `@since` tag to the new view files.

classes/models/FrmMigrate.php
```diff
@@ -80,6 +80,7 @@ private function update_settings_for_new_install() {
 		$settings = FrmAppHelper::get_settings();
 
 		$settings->denylist_check = 1;
+		$settings->is_new_install = 1;
```
Crabcyborg: @shervElmi I don't love the way this is named, since it looks like it would be `1` forever, even long after a site is a "new install".

I think we could go with a more descriptive name. Something like `installed_after_welcome_tour_update`.

This way we can continue to add others if this comes up in another context.

css/formidableforms11.css
Crabcyborg: @shervElmi It looks like this file was committed incorrectly.

You should really only have a `formidableforms.css` file. I'm not sure why you're running into so many other ones.

classes/controllers/FrmWelcomeTourController.php
```diff
@@ -0,0 +1,604 @@
...
+					case 'create-form':
+						$completed_step = self::more_than_the_default_form_exists();
+						break;
+					case 'embed-form':
+						$completed_step = self::check_for_form_embeds();
+						break;
+				}
+			}
+
+			if ( $completed_step && ! isset( self::$checklist['completed_steps'][ $step_key ] ) ) {
```
Crabcyborg: This `isset` check doesn't seem necessary. If we set `self::$checklist['completed_steps'][ $step_key ] = true;` without checking, it shouldn't make a significant performance difference, and it makes the code a bit smaller.

js/src/welcome-tour/elements/beginTourModalElement.js
```diff
@@ -0,0 +1,70 @@
...
+ * Inject begin tour modal elements into the DOM and the elements object.
+ *
+ * @private
+ * @param {HTMLElement} beginTourModal The begin tour modal element.
+ * @return {void}
+ */
+function addBeginTourModalToElements( beginTourModal ) {
+	const elements = getElements();
+
+	if ( elements.beginTourModal || undefined === beginTourModal ) {
```
Crabcyborg: Do we really need to check if `beginTourModal` is `undefined`? It doesn't seem possible based on the code.

js/src/welcome-tour/elements/beginTourModalElement.js
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Create and return the begin tour modal HTML element.
+ *
+ * @private
+ * @return {HTMLElement} The begin tour modal element.
+ */
+function createBeginTourModalElement() {
+	const beginTourModal = maybeCreateModal(
+		'frm-get-started-modal',
```
Crabcyborg: @shervElmi Generally, I try to make all of the IDs with underscores instead of hyphens.

You can see most code calling `maybeCreateModal` uses underscores.

For consistency, can we use underscores here?

classes/controllers/FrmWelcomeTourController.php
```diff
@@ -0,0 +1,604 @@
...
+	 *
+	 * @return void
+	 */
+	public static function enqueue_assets() {
+		$plugin_url = FrmAppHelper::plugin_url();
+		$version    = FrmAppHelper::plugin_version();
+
+		wp_enqueue_style( self::SCRIPT_HANDLE, $plugin_url . '/css/admin/welcome-tour.css', array(), $version );
+
+		wp_register_script( self::SCRIPT_HANDLE, $plugin_url . '/js/welcome-tour.js', array( 'wp-i18n' ), $version, true );
```
Crabcyborg: @shervElmi If we're adding translations in Lite, we also need to call `wp_set_script_translations` or the translations will not work.

classes/controllers/FrmWelcomeTourController.php
```diff
@@ -0,0 +1,604 @@
...
+	 * @return bool True if welcome tour should be shown, false otherwise.
+	 */
+	private static function should_show_welcome_tour() {
+		// Only show welcome tour for new installs on Formidable admin pages.
+		if ( ! FrmAppHelper::is_formidable_admin() || empty( FrmAppHelper::get_settings()->installed_after_welcome_tour_update ) ) {
+			return false;
+		}
+
+		self::$checklist = self::get_checklist();
+		if ( self::is_tour_completed() || ! empty( self::$checklist['dismissed'] ) ) {
```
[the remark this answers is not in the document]
shervElmi: [feat: show welcome tour completion step only once](https://github.com/Strategy11/formidable-forms/pull/2466/commits/8bfb1156d9df0fbda891a371be82783a54f8592e)

classes/controllers/FrmWelcomeTourController.php
shervElmi: [Add translation support for welcome tour script](https://github.com/Strategy11/formidable-forms/pull/2466/commits/a9859a10c266da2e75aa848948d267f63d6b14f5)

js/src/welcome-tour/elements/beginTourModalElement.js
```diff
@@ -0,0 +1,69 @@
...
+ * @param {HTMLElement} beginTourModal The begin tour modal element.
+ * @return {void}
+ */
+function addBeginTourModalToElements( beginTourModal ) {
+	const elements = getElements();
+	if ( elements.beginTourModal ) {
+		return;
+	}
+
+	addElements( { beginTourModal } );
```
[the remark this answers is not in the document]
shervElmi: [refactor: simplify beginTourModal element addition logic](https://github.com/Strategy11/formidable-forms/pull/2466/commits/2f368e2d4357031333780fa2e96d1ec96022c270)

Crabcyborg (approved):
Thank you @shervElmi!

🚀

PR #2469, opened 2025-08-21
Add class required for single column layout in builder when One column layout is selected

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-surveys/issues/136

### Test steps
1. Add Radio field to your form
2. Select 'Button' format and check `Use images with buttons` option in the field settings
3. Under Advanced field settings, select One Column for the `Option Layout` setting
4. Save and reload the builder
5. Go to the radio field Option Layou settings and try switching between the `One Column` and any other layout and confirm that the field preview layout reflects the option selected.

CleanShot 2025-08-21 at 22 05 23


lauramekaj1:
@AbdiTolesa When the display format of buttons is set to "Images", the Option Layout is not visible under the Advanced field settings. But it is actually shown when the display format is set to "Simple".
image

AbdiTolesa:
> @AbdiTolesa When the display format of buttons is set to "Images", the Option Layout is not visible under the Advanced field settings. But it is actually shown when the display format is set to "Simple".

@lauramekaj1 I checked this and it behaves like that on the main branch too and is unrelated to this update. I believe that is intentional and not an issue that we are hiding the layout options for image buttons. Does this PR fix the issue mentioned for you?

truongwp:
@AbdiTolesa Can you fix the conflict with the master branch?

PR #2480, opened 2025-08-27
Update email gathering modal tests

lauramekaj1:


lauramekaj1:
@Crabcyborg I updated e2e tests to match the new updates and also added coverage for the email gathering modal. It seems that `admin-a11y.cy.js` keeps failing because of a heading order. When you have time can you check please? 
Thanks!

Crabcyborg (approved):
Thanks @lauramekaj1!

It looks like there are some new accessibility issues, but we can fix those separately.

We've started enforcing our syntax checks in all of our JS files. I updated the test to get it passing.

You can mostly auto-format it by running the CLI commands. For example, `./node_modules/.bin/eslint --fix  './tests/cypress/e2e/Form Templates/FormTemplates.cy.js'` should work.

Our linting workflow starts with `npm ci --legacy-peer-deps --include=dev`, which I also had to run to get it working properly.

I use a find/replace as well to change 4 spaces into tabs. Moving forward, indentation should use tabs and not spaces.

I'll merge this now.

PR #2482, opened 2025-08-29
issue 5929 alternative fix

Crabcyborg:
Related PR https://github.com/Strategy11/formidable-forms/pull/2456

Fixes https://github.com/Strategy11/formidable-pro/issues/5929

I don't love the list of hard coded keys. I think we really just want to check the prepared values instead of the other value.

I tried this out on Emma's QA site and it looks like it works well.


Liviu-p (approved):
Looks good to me @Crabcyborg, I can confirm that it fixes the issue reported by Emma as well.

PR #2486, opened 2025-09-09
Lite: Replace special characters in CSS class names with standard naming

shervElmi:
This change improves CSS class naming by removing special characters that can cause issues with escaping and parsing. Updated class names to use standard hyphen-separated naming convention for better compatibility and cleaner code.

The change ensures consistent CSS class naming across the codebase without requiring escape characters, making the styles more maintainable and less error-prone.


Crabcyborg (approved):
Thank you Sherv!

PR #2495, opened 2025-09-15
Fix keyboard accessibility and tab order for inputs with "more icon"

shervElmi:
This PR fixes keyboard accessibility issues with form inputs that include a "more icon" functionality. Previously, users navigating with keyboard could not properly tab through these inputs due to incorrect DOM order and missing keyboard event handlers.

The fix repositions the more icon after the input field (instead of before) and adds proper keyboard navigation support, ensuring users can access all functionality using only the keyboard.

### Related Issue
Fixes https://github.com/Strategy11/formidable-pro/issues/5971

### Add-ons Changes

- Pro: https://github.com/Strategy11/formidable-pro/pull/5990
- Dates: https://github.com/Strategy11/formidable-dates/pull/180


js/src/admin/admin.js
```diff
@@ -6688,7 +6688,10 @@ window.frmAdminBuildJS = function() {
 			}
 			if ( input !== null ) {
 				if ( ! inputTrigger ) {
-					input.focus();
+					const { key } = event ?? {};
+					if ( key !== 'Enter' && key !== ' ' ) {
```
Crabcyborg, line 6692: @shervElmi It looks like if there is no event, that `key` would be `undefined`. Do we want to focus when it's undefined? I'd imagine we don't.

lauramekaj1 (approved):
@shervElmi I tested it and it's working as expected. Thank you!

PR #2496, opened 2025-09-18
Fixed Text Domain to one string in captcha.php

DAnn2012:
Fixed Text Domain to one string in captcha.php


Crabcyborg (approved):
Thanks for catching that @DAnn2012!

PR #2498, opened 2025-09-18
Enhance token input field initialization for AJAX-loaded fields

shervElmi:
This PR enhances the token input field initialization system to properly support fields that are dynamically loaded via AJAX. Previously, token input fields would not initialize correctly when added through AJAX operations.

## Before

CleanShot 2025-09-18 at 17 23 22

## After

CleanShot 2025-09-18 at 17 55 14


Crabcyborg (approved):
Thanks @shervElmi!

I haven't tested this but this looks good to me.

🚀

PR #2519, opened 2025-09-30
Fix Firefox display issue with input icon containers

shervElmi:
This PR fixes a Firefox-specific display issue where input icon containers in the custom format field did not render correctly. It affected the appearance and positioning of icons within date input fields when viewed in Firefox.

### Before

CleanShot 2025-09-30 at 22 30 47@2x

### After

CleanShot 2025-09-30 at 22 31 11@2x


Crabcyborg (approved):
Thanks Sherv!

PR #2524, opened 2025-10-01
add a sql db function that will retrieve a entry and its fields as ro…

Liviu-p:


Crabcyborg:
I feel like this query is maybe a bit too complex.

It's tough to know how this would perform for every customer.

I feel like there's another approach we can consider.

PR #2525, opened 2025-10-01
fix base font size inconsistency

Liviu-p:


Crabcyborg (approved):
Thanks @Liviu-p!

This appears to work well.

🚀

PR #2526, opened 2025-10-03
Update tests email gathering modal

lauramekaj1:


Crabcyborg (approved):
Thanks @lauramekaj1!

The CSS validation is just because of a `Status: 429 - Too Many Requests` error.

The API is being rate limited.

You can ignore that. It will start passing again after a short period.

PR #2527, opened 2025-10-03
Fix range selection that includes fields in section

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/6017

### Test procedure
1. Create a form and add fields to it.
2. Add a Section field and add a few field rows to it.
3. Try multi selecting range of fields that include fields both in and out of the Section with Shift + click.
4. Confirm that you can select the range of fields as you wish. The whole row is selected by default when you click on a field inside a section for multi selection.

https://github.com/user-attachments/assets/617448cb-be11-4f47-9279-4336cc5e6d6b


AbdiTolesa:
> I’ve retested the fixes for Issues 2 and 3, and I was still able to reproduce the same unexpected behavior. Could you please check on your side as well?

Hi @lauramekaj1 Yes I think I had an illusion it was working as I haven't had a good number of fields inside my section. Those two issues are fixed for me. Could you verify that please?

lauramekaj1 (approved):

Crabcyborg:
Closing this for now. There are a lot of conflicts. We can start over from the main branch later.

PR #2534, opened 2025-10-08
Prevent new choice options aside from the choice label from displaying as separate elements in the builder

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/6011
Before:

https://github.com/user-attachments/assets/0fd018cc-258b-473e-92b5-c69cb7cc4b6e

After:

https://github.com/user-attachments/assets/7c93bb69-29ce-4284-a74a-27420ef2a27c


js/src/admin/admin.js
```diff
@@ -6204,7 +6204,7 @@ window.frmAdminBuildJS = function() {
 			opts = [],
 			imageUrl = '';
 
-		const optVals = jQuery( 'input[name^="field_options[options_' + fieldId + ']"]' );
+		const optVals = jQuery( 'input[name^="field_options[options_' + fieldId + ']"][name*="label"], input[name^="field_options[options_' + fieldId + ']"][name*="value"]' );
```
Crabcyborg: @AbdiTolesa I'm pretty sure we want to support price here as well.

Crabcyborg (approved):
@AbdiTolesa I made some changes.

This change removes the need for some other funny logic underneath.

I got mixed up when I said we wanted to include `[price]`. In reality, it calls `continue` right away for anything that is `[value]`, `[price]`, or `[image]` (all of the other options we support here).

This logic below clearly is only intended for labels.

So I've updated this, so we only query for labels, with the `[`/`]` wrapper characters to make sure there aren't any funny matches, and I removed all of the checks for `[price]`, `[image]` and `[value]`, so this is all more simple.

I think this is good now.

🚀

PR #2535, opened 2025-10-08
Speed Up CI/CD Workflows by Caching Dependency Installations

shervElmi:


Crabcyborg:

It's gotten a bit old, and it looks a fair bit more difficult to maintain for what doesn't appear to be a significant performance boost.

PR #2543, opened 2025-10-15
Fix the order of submit field in some cases

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/6027

Besides the first case in the issue, I found the second case that causes this issue (screencast below):
- Click to add a field.
- Click on the submit field to show its settings.
- Update the form.

https://github.com/user-attachments/assets/7c69185d-08c1-4626-b472-94d4a400947b

For the first case, when adding a field to a repeater or section, it has a different field order, we should not update the submit row fields order.

For the second case, when showing the submit settings, it has the old field_order input value, which is used when updating the form.


Crabcyborg (approved):
Thank you @truongwp!

I think this looks good.

I'll merge this once Laura approves it.

lauramekaj1 (approved):
@truongwp @Crabcyborg I tested it and it's working as expected. Thank you!

PR #2549, opened 2025-10-20
Fix missing add option button for product fields in mixed version setups

shervElmi:
Fixes missing "Add Product" button in product fields when Lite is updated but Pro is still on an older version (before the field settings refresh).

Product fields now display the add option button correctly regardless of Pro version, matching the behavior of regular radio fields.

Fixes https://github.com/Strategy11/formidable-pro/issues/6054

### Before

image

### After

CleanShot 2025-10-20 at 18 34 59@2x


Crabcyborg (approved):
Thanks @shervElmi!

This looks like it works well now.

🚀

PR #2559, opened 2025-10-27
Update tests bootstrap 5

lauramekaj1:


lauramekaj1:
@Crabcyborg After the Bootstrap update, some locators were affected, causing the E2E tests to fail (for example, the selector data-original-title was renamed to data-bs-original-title).
I also made some adjustments to improve test stability.

PR #2570, opened 2025-10-31
Add new upsells in Lite for missing Pro features

AbdiTolesa:
Closes https://github.com/Strategy11/formidable-pro/issues/5873

- [x] Unique
- [x] Read only
- [x] Before/After Input
- [x] Autocomplete
- [x] Field visibility
- [x] AI generated forms button

## Summary by CodeRabbit

* **New Features**
  * Pro gating for advanced field options (Unique, Read‑Only, Autocomplete, Visibility, Before/After content) with inline upgrade prompts
  * AI Create/Generate button supports customizable text, styling, and optional BETA pill

* **Style**
  * Upsell/disabled controls visually dimmed for clarity in the admin UI

* **UX**
  * Upgrade prompts and related controls moved earlier in field settings for better discoverability

### DIFF
--- a/classes/helpers/FrmSettingsUpsellHelper.php
+++ b/classes/helpers/FrmSettingsUpsellHelper.php
@@ -25,9 +25,7 @@ class FrmSettingsUpsellHelper {
 			'class' => 'frm_mark_unique',
 		);
 
-		$pro_is_installed = FrmAppHelper::pro_is_installed();
-
-		if ( $pro_is_installed ) {
+		if ( FrmAppHelper::pro_is_installed() ) {
 			if ( ! empty( $field['unique'] ) ) {
 				$unique_element_atts['checked'] = 'checked';
 			}


classes/controllers/FrmFieldsController.php
```diff
@@ -362,6 +362,15 @@ public static function load_single_field_settings( $atts ) {
 			$field['placeholder'] = implode( ', ', $field['placeholder'] );
 		}
 
+		$pro_is_installed = FrmAppHelper::pro_is_installed();
+		$no_allow         = ! $pro_is_installed ? 'frm_noallow' : '';
+
+		$field_has_unique_option                = in_array( $field['type'], array( 'address', 'checkbox', 'data', 'date', 'email', 'file', 'lookup', 'name', 'number', 'password', 'phone', 'radio', 'rte', 'text', 'textarea', 'time', 'url' ), true );
```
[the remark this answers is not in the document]
AbdiTolesa: @Crabcyborg Definitely, thanks!

classes/controllers/FrmFieldsController.php
```diff
@@ -367,6 +367,22 @@ public static function load_single_field_settings( $atts ) {
 			$field['placeholder'] = implode( ', ', $field['placeholder'] );
 		}
 
+		$pro_is_installed = FrmAppHelper::pro_is_installed();
+		$no_allow         = ! $pro_is_installed ? 'frm_noallow' : '';
+
+		if ( ! $pro_is_installed ) {
+			$field_has_unique_option                = in_array(
```
Crabcyborg: Maybe we could make these new functions?

To make it more readable still, with the `field_has_unique_option` name, while pulling logic out of this function.

I'd try `$show_upsell_for_unique_value = self::field_type_has_unique_option( $field['type'] );`.

classes/views/form-templates/index.php
```diff
@@ -55,7 +55,14 @@
 					 * @since 6.10
 					 */
 					do_action( 'frm_after_create_blank_form_button' );
-					?>
+					if ( ! FrmAppHelper::pro_is_installed() ) {
+						?>
+					<button id="frm-form-templates-create-ai-form" class="frm-flex-box frm-items-center frm-form-templates-create-button frm_show_upgrade" data-upgrade="<?php esc_attr_e( 'Create with AI options', 'formidable' ); ?>">
```
Crabcyborg: The alignment here feels funny.

I think we should consider adding a new view for button and just using `include` here so we don't have to jump in and out of PHP and HTML.

classes/views/frm-fields/back-end/settings.php
```diff
@@ -330,8 +334,72 @@
...
+						'class'          => 'frm-flex',
+					)
+				);
+				?>
+			</label>
+			<select name="field_options[autocomplete_<?php echo absint( $field['id'] ); ?>]" id="field_options_autocomplete_<?php echo absint( $field['id'] ); ?>" data-upgrade="<?php esc_attr_e( 'Autocomplete options', 'formidable' ); ?>">
+				<option value=""><?php esc_html_e( '&mdash; Select &mdash;' ); ?></option>
+			</select>
+		</p>
+		<p class="frm6 frm_form_field frm_show_upgrade">
```
Crabcyborg: I think we also need a check for field type here for Visibility.

Some field types, at least not hidden fields, don't have this setting.

classes/views/frm-fields/back-end/settings.php
```diff
@@ -102,22 +102,22 @@
...
 				?>
 				<div class="frm_form_field">
-					<label for="frm_uniq_field_<?php echo esc_attr( $field['id'] ); ?>" class="frm_help frm-mb-0" title="<?php esc_attr_e( 'Unique: Do not allow the same response multiple times. For example, if one user enters \'Joe\', then no one else will be allowed to enter the same name.', 'formidable' ); ?>" data-trigger="hover">
-						<input type="checkbox" name="field_options[unique_<?php echo esc_attr( $field['id'] ); ?>]" id="frm_uniq_field_<?php echo esc_attr( $field['id'] ); ?>" value="1" <?php checked( $field['unique'], 1 ); ?> class="frm_mark_unique" />
+					<label for="frm_uniq_field_<?php echo esc_attr( $field['id'] ); ?>" class="frm_help frm-mb-0 <?php echo esc_attr( $pro_is_installed ? '' : 'frm_show_upgrade' ); ?>" title="<?php esc_attr_e( 'Unique: Do not allow the same response multiple times. For example, if one user enters \'Joe\', then no one else will be allowed to enter the same name.', 'formidable' ); ?>" data-trigger="hover">
+						<input <?php FrmAppHelper::array_to_html_params(self::get_unique_element_atts( $field, $pro_is_installed, $no_allow ), true ); ?> />
```
Crabcyborg: @AbdiTolesa

This line has a few issues.

1. The static analysis doesn't like that this references `self`. It works better to just reference the class name with the function instead.
2. It looks like there's a missing space between the `(` and `self::`.

classes/controllers/FrmFieldsController.php
```diff
@@ -367,9 +367,78 @@ public static function load_single_field_settings( $atts ) {
...
+	private static function get_unique_element_atts( $field, $pro_is_installed, $no_allow ) {
+		$unique_element_atts = array(
+			'type'  => 'checkbox',
+			'name'  => 'field_options[unique_' . esc_attr( $field['id'] ) . ']',
+			'id'    => 'frm_uniq_field_' . esc_attr( $field['id'] ),
+			'value' => '1',
+			'class' => 'frm_mark_unique ' . esc_attr( $no_allow ),
+		);
+		if ( ! empty( $field['unique'] ) ) {
+			$unique_element_atts['checked'] = true;
```
Crabcyborg: @AbdiTolesa I think we generally use `checked="checked"`.

It looks like WP does as well (https://developer.wordpress.org/reference/functions/__checked_selected_helper/).

Can we use `= 'checked'` here instead of `= true`?

classes/controllers/FrmFieldsController.php
```diff
@@ -367,9 +367,78 @@ public static function load_single_field_settings( $atts ) {
...
+	private static function get_read_only_element_atts( $field, $pro_is_installed, $no_allow ) {
+		$read_only_element_atts = array(
+			'type'  => 'checkbox',
+			'name'  => 'field_options[read_only_' . esc_attr( $field['id'] ) . ']',
+			'id'    => 'frm_read_only_field_' . esc_attr( $field['id'] ),
+			'value' => '1',
+			'class' => esc_attr( $no_allow ),
+		);
+		if ( ! empty( $field['read_only'] ) ) {
+			$read_only_element_atts['checked'] = true;
```
Crabcyborg: Can we use `= 'checked`' here as well?

classes/views/form-templates/ai-upsell-button.php
```diff
@@ -0,0 +1,34 @@
...
+$oneclick_data = FrmAddonsController::install_link( 'ai' );
+if ( isset( $oneclick_data['url'] ) ) {
+	$ai_install_span_attrs['data-oneclick'] = json_encode( $oneclick_data );
+}
+?>
+<div class="frm-activate-addon frm-flex-box frm-form-templates-create-button frm-items-center frm_show_upgrade">
+	<?php FrmAppHelper::icon_by_class( 'frmfont frm-ai-form-icon', array( 'aria-label' => _x( 'Create', 'form templates: create an AI generated form', 'formidable' ) ) ); ?>
+	<span <?php FrmAppHelper::array_to_html_params( $ai_install_span_attrs, true ); ?>>
+	<?php
+	FrmAddonsController::conditional_action_button(
```
Crabcyborg: @AbdiTolesa Looking at this PR, we probably can't just use `conditional_action_button` since it does just render a blue button.

A better example would be to look at `FrmFieldsHelper::render_ai_generate_options_button`.

There we have a similar button already, for generating options with AI.

Screen Shot 2025-12-10 at 11 05 27 AM

classes/controllers/FrmFormTemplatesController.php
```diff
@@ -701,6 +701,11 @@ public static function enqueue_assets() {
 		wp_register_style( self::SCRIPT_HANDLE, $plugin_url . '/css/admin/form-templates.css', array(), $version );
 		wp_enqueue_style( self::SCRIPT_HANDLE );
 
+		$page = FrmAppHelper::simple_get( 'page' );
+		if ( FrmFormTemplatesController::PAGE_SLUG === $page ) {
```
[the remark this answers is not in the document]
Crabcyborg: All of the JS you should need is available already in the admin.js code. You shouldn't need new scripts.

classes/helpers/FrmFieldsHelper.php
```diff
@@ -2777,9 +2777,15 @@ public static function render_ai_generate_options_button( $args, $should_hide_bu
 	 *
 	 * @since 6.24
 	 *
+	 * @param false|string $classes
+	 *
 	 * @return string Button class.
 	 */
-	private static function get_ai_generate_options_button_class() {
+	private static function get_ai_generate_options_button_class( $classes = false ) {
+		if ( is_string( $classes ) ) {
```
Crabcyborg: @AbdiTolesa It seems to me that we don't really need to add a parameter here or really call this function at all if we're passing `$classes`. If `$args['class']` is not empty, can we just not call `get_ai_generate_options_button_class`?

classes/views/form-templates/ai-upsell-button.php
```diff
@@ -0,0 +1,14 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'You are not allowed to call this page directly.' );
+}
+/**
+ * @since x.x
+ */
+$atts = array(
```
Crabcyborg: I don't think we really need to set this to a variable first. Since the variable name is so generic, it's probably better to just pass this array into `render_ai_generate_options_button` directly.

classes/controllers/FrmFieldsController.php
```diff
@@ -374,9 +374,78 @@ public static function load_single_field_settings( $atts ) {
...
 
+	/**
+	 * @since x.x
+	 *
+	 * @param array $field
+	 * @param bool  $pro_is_installed
+	 *
+	 * @return array
+	 */
+	public static function get_unique_element_atts( $field, $pro_is_installed ) {
```
Crabcyborg: @AbdiTolesa It feels kind of messy to pass `$pro_is_installed` around in as a parameter.

I don't think an extra call to `FrmAppHelper::pro_is_installed()` should be too big of a deal for performance, and keeping that logic internally in this function makes it more simple.

Crabcyborg:
@AbdiTolesa I also see this Learn More button but it tries to link me to `https://dev-site.local/wp-admin/undefined`. I think we need to do something about this.

Screen Shot 2026-01-06 at 12 00 47 PM

PR #2572, opened 2025-11-03
Ignore generated CSS files for all multisite installations

shervElmi:
Updates the gitignore pattern to ignore CSS files generated for any multisite blog ID, not just blog 1.

Previously only `formidableforms1.css` was ignored. Now all numbered variants (formidableforms2.css, formidableforms17.css, etc.) are ignored.

### Related Code

image


Crabcyborg (approved):
Thanks @shervElmi!

Yeah, this seems to work great!

🚀

PR #2573, opened 2025-11-03
handle welcome tour checklist to prevent Cypress test failures

lauramekaj1:


lauramekaj1:
Hey @Crabcyborg, 
The welcome tour popup and the checklist modal were blocking some elements, so I added code to admin-a11.cy.js to dismiss them during test run. After dismissing the popups, the tests now pass.

I’m still seeing an accessibility issue on the Entries page for a heading element, but I’m not sure what’s causing it. Could you please take a look?

PR #2575, opened 2025-11-04
Limit fields in group to 12

Crabcyborg:
This update allows users to add up to 12 fields in a group. Up from the previous limit of 6.

I have this set up on my QA site. You can see it here https://qa.formidableforms.com/mike2/wp-admin/admin.php?page=formidable&frm_action=edit&id=937

When there are more than 6 fields in a group, the row layout options are more limited.
Screen Shot 2025-11-19 at 9 29 33 AM

And we hide the field ID on hover.
Screen Shot 2025-11-19 at 9 30 37 AM

**Pre-release**
[formidable-6.25.3b.zip](https://github.com/user-attachments/files/23628958/formidable-6.25.3b.zip)


lauramekaj1 (approved):
@Crabcyborg I tested it and it's working as expected. Thank you! 🚀

PR #2576, opened 2025-11-05
Rock: Project Elevate - Part 1

shervElmi:
## Linear

- **Handoff:** https://linear.app/strategy11/document/handoff-693db9ae1792
- **Design**: https://www.figma.com/design/DQbZMgeNmJWkxOWYfHjDZX/Formidable-System?node-id=1-20&t=KdWMt6UNRG8kIbcG-1

## Screen recording demo

https://share.cleanshot.com/GBMXWNjn

### DIFF
--- a/tests/cypress/e2e/Entries/EntriesPageDataValidations.cy.js
+++ b/tests/cypress/e2e/Entries/EntriesPageDataValidations.cy.js
@@ -54,6 +54,7 @@ describe( 'Entries submitted from a form', () => {
 				'Upgrade to Pro.',
 				'Get 60% Off Pro!',
 				'Black Friday Deals',
+				'Cyber Monday Deals',
 			] );
 		} );
 
--- a/tests/cypress/e2e/Forms/formPageDataValidation.cy.js
+++ b/tests/cypress/e2e/Forms/formPageDataValidation.cy.js
@@ -24,6 +24,7 @@ describe( 'Forms page', () => {
 					'Get the Deal',
 					'upgrading for 60% off during our No Brainer Sale!',
 					'Black Friday Deals',
+					'Cyber Monday Deals',
 				];
 
 				if ( href && substrings.some( substring => text.includes( substring ) ) ) {
@@ -42,6 +43,7 @@ describe( 'Forms page', () => {
 								'build powerful forms, web apps, dashboards, and more',
 								'black friday sales are happening now!',
 								'save 65% on the best wordpress form builder',
+								'don\'t miss the best deals for cyber monday!',
 							] ).to.include( headingText );
 						} );
 					} );


classes/models/FrmField.php
```diff
@@ -97,98 +97,156 @@ private static function get_captcha_field_name() {
 	}
 
 	public static function pro_field_selection() {
-		$images_url = FrmAppHelper::plugin_url() . '/images/';
-		$fields     = array(
+		$make_tracked_url = static function ( $url ) {
```
Crabcyborg: @shervElmi I think we could just add this as a new function? I don't see why it's inside of this function.

classes/models/FrmField.php
```diff
@@ -97,98 +97,156 @@ private static function get_captcha_field_name() {
...
+				'learn-more'   => $make_tracked_url( 'https://formidableforms.com/knowledgebase/password-2/' ),
 			),
 			'tag'             => array(
-				'name' => __( 'Tags', 'formidable' ),
-				'icon' => 'frm_icon_font frm_price_tags2_icon',
+				'name'         => __( 'Tags', 'formidable' ),
+				'icon'         => 'frm_icon_font frm_price_tags2_icon',
+				'message'      => esc_html__( 'Facilitate easy categorization and analysis by letting users select or create multiple relevant keywords or labels for their entry.', 'formidable' ),
+				'upsell_image' => $upsell_images_url . 'tags-field-preview.webp',
+				'learn-more'   => $make_tracked_url( 'https://formidableforms.com/knowledgebase/tags/' ),
```
Crabcyborg: @shervElmi

Do we call `$make_tracked_url` for every item with a `learn-more` key?

I think this could be cleaner if we just checked for `learn-more` and prepended the URL with `https://formidableforms.com/knowledgebase/`, so we don't need to repeat this everywhere.

Something like `'learn-more' => 'tags/'` would go here, so it's easy to read.

classes/models/FrmField.php
```diff
@@ -97,98 +97,156 @@ private static function get_captcha_field_name() {
...
+				'message'      => esc_html__( 'Let users quickly select values within a range using a dynamic, interactive slider, creating a modern and enjoyable data input experience.', 'formidable' ),
+				'upsell_image' => $upsell_images_url . 'slider-field-preview.webp',
+				'learn-more'   => $make_tracked_url( 'https://formidableforms.com/knowledgebase/slider/' ),
 			),
 			'toggle'          => array(
-				'name' => __( 'Toggle', 'formidable' ),
-				'icon' => 'frm_icon_font frm_toggle_on_icon',
+				'name'         => __( 'Toggle', 'formidable' ),
+				'icon'         => 'frm_icon_font frm_toggle_on_icon',
+				'message'      => esc_html__( 'Link entries together, dynamic display information, and even make selections from data entered on another form.', 'formidable' ),
```
Crabcyborg: @shervElmi It looks like Toggles have the message for Dynamic/Lookup fields. This needs to be updated.

js/src/admin/upgrade-popup.js
```diff
@@ -38,6 +37,22 @@ export function addOneClick( link, context, upgradeLabel ) {
...
 
+	let modalIcon;
+	let modalIconSvg;
+	const modalIconWrapper = container.querySelector( '.frm-circled-icon' );
+	if ( modalIconWrapper ) {
+		modalIconSvg = modalIconWrapper.querySelector( 'svg' );
+		modalIcon = modalIconSvg.querySelector( 'use' );
+		modalIconWrapper.classList.remove( 'frm-circled-icon-green' );
+		modalIconSvg.classList.remove( 'frm_svg32' );
+		modalIcon.setAttribute( 'xlink:href', '#frm_filled_lock_icon' );
```
Crabcyborg: Here as well. Ideally we won't use the deprecated `xlink:href` anywhere.

classes/helpers/FrmFieldsHelper.php
```diff
@@ -2189,7 +2189,17 @@ public static function show_add_field_buttons( $args ) {
...
+			$li_params['data-learn-more'] = $field_type['learn-more'];
+		}
+
+		if ( isset( $upgrading['url'] ) ) {
+			$li_params['data-message'] = sprintf(
+				// translators: %s: Field name
+				esc_html__( 'You already have access to %s fields, you\'ll just need to activate to start using them.', 'formidable' ),
+				$field_name
+			);
+		} else if ( $upgrade_message ) {
```
Crabcyborg: Can we use `elseif`? We don't use `else if` in PHP.

Crabcyborg (approved):
Thanks @shervElmi!

Once we figure out the failing e2e test, I think we can merge this.

resources/scss/admin/components/upgrade/_upgrade-modal.scss
```diff
@@ -6,7 +6,73 @@
...
+#frm_upgrade_modal {
+
+	.cta-inside {
+		display: flex;
+		flex-direction: column;
+		gap: var(--gap-sm);
+	}
+
+	.frm-upgrade-message img {
+		width: 100%;
```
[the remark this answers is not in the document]
Crabcyborg: @shervElmi Most people will see the cached image unless they've never loaded it before.

Can you add something to the URL so it's updated for this?

Something like `?v=2`?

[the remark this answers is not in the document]
Crabcyborg: Hm, it's a webp. I think I must've just had the wrong branch checked out.

lauramekaj1:
> @lauramekaj1 It looks like an e2e test in `deleteForms` is failing here.
> 
> Can you look into this?
> 
> Thank you!

@Crabcyborg I checked and it wasn’t related to functionality, so I updated the failing test and improved the selectors to make the flow stable.
Here is the PR with the update: https://github.com/Strategy11/formidable-forms/pull/2617

Thank you!

PR #2579, opened 2025-11-10
Add aria-describedby attribute where missing for fields with error

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/6078

- [x] Add missing `aria-describedby` attributes when field errors happen.
- [x] Make sure `aria-describedby` values are removed from elements when errors are cleared, during Ajax submission

1. Create a form and add Checkbox, Radio and Dropdown fields to it.
2. Set constraints like `Required` in the fields options.
3. Preview the forms and try submitting the form without filling the fields so that errors are triggered.
4. Confirm that `aria-describedby` attribute is added to the elements to link them to the error message element programatically.
5. Repeat the test under different form settings, like JS validation, Ajax submission etc.

## Summary by CodeRabbit

* **Enhancements**
  * Improved form field accessibility with enhanced ARIA support for screen readers on input fields and error messages.
  * Better error message handling with refined alert role management for users relying on assistive technologies.


lauramekaj1:
Hi @AbdiTolesa, there are some error warnings showing in this branch. Please see the attached screenshot:
image

lauramekaj1 (approved):
@AbdiTolesa I tested it and it's working as expected. Thank you!

classes/models/FrmFieldFormHtml.php
```diff
@@ -232,12 +232,16 @@ private function replace_error_shortcode() {
 		$this->maybe_add_error_id();
 		$error = $this->pass_args['errors'][ 'field' . $this->field_id ] ?? false;
 
-		if ( ! empty( $error ) && false === strpos( $this->html, 'role="alert"' ) && FrmAppHelper::should_include_alert_role_on_field_errors() ) {
+		$include_alert_role_on_field_errors = FrmAppHelper::should_include_alert_role_on_field_errors();
```
truongwp: In the old code, this won't be reached if `$error` is empty. The new code should also follow that. Unless it's too complicated in the new code.

classes/controllers/FrmFieldsController.php
```diff
@@ -524,13 +524,18 @@ public static function input_html( $field, $echo = true ) {
 		self::add_pattern_attribute( $field, $add_html );
 
 		$add_html = apply_filters( 'frm_field_extra_html', $add_html, $field );
-		$add_html = ' ' . implode( ' ', $add_html ) . '  ';
+		$html     = ' ' . implode( ' ', $add_html ) . '  ';
```
truongwp: Do we really need two separate variables `$html` and `$add_html`? Using just one will save the memory usage.

truongwp (approved):

js/formidable.js
```diff
@@ -1198,8 +1198,23 @@ function frmFrontFormJS() {
 
 	function removeAllErrors() {
 		jQuery( '.form-field' ).removeClass( 'frm_blank_field has-error' );
-		jQuery( '.form-field .frm_error' ).replaceWith( '' );
```
Crabcyborg: @AbdiTolesa Why exactly is this line being removed?

classes/models/FrmFieldFormHtml.php
```diff
@@ -281,15 +281,23 @@ private function replace_error_shortcode() {
 		$this->maybe_add_error_id();
 		$error = $this->pass_args['errors'][ 'field' . $this->field_id ] ?? false;
 
-		if ( ! empty( $error ) && false === strpos( $this->html, 'role="alert"' ) && FrmAppHelper::should_include_alert_role_on_field_errors() ) {
+		if ( empty( $error ) ) {
+			FrmShortcodeHelper::remove_inline_conditions( false, 'error', $error, $this->html );
+		}
```
Crabcyborg: @AbdiTolesa Should there be a `return` here?

AbdiTolesa:
@Crabcyborg There is a PHP CS Fixer I couldn't reproduce when running the check in my local setup:

image

Is that a false positive?

js/formidable.js
```diff
@@ -1209,9 +1207,9 @@ function frmFrontFormJS() {
 	function getErrorElementId( key, input ) {
 		if ( isNaN( key ) || ! input || ! input.id ) {
 			// If key isn't a number, assume it's already in the right format.
-			return 'frm_error_field_' + key;
+			return 'frm_error_field_' + key.split( '-' )[ 0 ];
 		}
-		return 'frm_error_' + input.id;
+		return 'frm_error_' + input.id.split( '-' )[ 0 ];
```
Crabcyborg: @AbdiTolesa I think we want to still include part of this array.

If there is a repeater, the error IDs are something like `abepx-0`. So we need to only remove part of the ID and not everything right of the first dash.

garretlaxton (approved):
This works great!

Crabcyborg:
Thank you Garret!

🚀

PR #2595, opened 2025-11-21
Add `autocomplete` attribute to the first name field of the form

truongwp:
Except the name fields inside Repeater.

Fixes https://github.com/Strategy11/formidable-pro/issues/6099

### DIFF
--- a/classes/models/fields/FrmFieldName.php
+++ b/classes/models/fields/FrmFieldName.php
@@ -136,6 +136,7 @@ class FrmFieldName extends FrmFieldCombo {
 	 *
 	 * @param mixed $value Field value before processing.
 	 * @param array $atts  Shortcode attributes.
+	 *
 	 * @return string      Most of cases, this will return string.
 	 */
 	protected function prepare_display_value( $value, $atts ) {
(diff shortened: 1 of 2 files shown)


classes/models/fields/FrmFieldName.php
```diff
@@ -285,4 +293,36 @@ public function show_after_default( $args ) {
...
+	 *
+	 * @param array $sub_field Subfield data.
+	 * @param array $args      Field output args. See {@see FrmFieldCombo::load_field_output()}.
+	 * @return array
+	 */
+	protected function get_sub_field_input_attrs( $sub_field, $args ) {
+		$attrs = parent::get_sub_field_input_attrs( $sub_field, $args );
+
+		$parent_form_id = (int) FrmField::get_option( $args['field'], 'parent_form_id' );
+		$form_id        = (int) is_array( $args['field'] ) ? $args['field']['form_id'] : $args['field']->form_id;
```
Crabcyborg: @truongwp I'm seeing issues with this line. It looks like it's casting `is_array( $args['field'] )` to an int, and then `$form_id` is still a string.

If I add braces, like `(int) ( is_array( $args['field'] ) ? $args['field']['form_id'] : $args['field']->form_id );`, it works as expected.

PR #2600, opened 2025-11-26
style custom css option

Liviu-p:


js/src/admin/styles.js
```diff
@@ -100,12 +160,29 @@ class frmStyleOptions {
...
 	 */
 	initStyleClassCopyToClipboard( successMessage ) {
-		const copyLabel = document.querySelector( '.frm-copy-text' );
-		copyLabel.addEventListener( 'click', event => {
-			const className = event.currentTarget.innerText;
-			navigator.clipboard.writeText( className ).then( () => {
-				this.success( successMessage );
-			} );
+		if ( ! navigator.clipboard || ! navigator.clipboard.writeText ) {
+			return;
```
Crabcyborg: @Liviu-p I just has Jonathan share that the copy to clipboard doesn't work for him, because his Chrome browser doesn't have `navigator.clipboard` for some reason.

We should probably add a fallback solution. I have this one in the Coupons add-on at the moment.

```
/**
 * Copy to clipboard if the Clipboard API is not available.
 *
 * @param {string} couponCode      The string being copied to the clipboard.
 * @param {HTMLElement} copyButton Used to position the temporary input element.
 * @return {bool}
 */
function fallbackCopy( couponCode, copyButton ) {
	if ( 'function' !== typeof document.execCommand ) {
		return false;
	}

	let copySuccess;

	const temp = document.createElement( 'input' );
	temp.setAttribute( 'type', 'text' );
	temp.value = couponCode;

	copyButton.parentElement.appendChild( temp );

	temp.focus();
	temp.select();
	temp.setSelectionRange( 0, 99999 );

	// Hide the input so it doesn't show up in the UI.
	temp.style.position = 'absolute';
	temp.style.left = '-9999px';
	temp.style.top = '-9999px';

	try {
		copySuccess = document.execCommand( 'copy' );
	} catch ( error ) {
		copySuccess = false;
	}

	temp.remove();

	return copySuccess;
}
```

classes/helpers/FrmCssScopeHelper.php
```diff
@@ -0,0 +1,253 @@
...
+			
+			foreach ( $lines as $line ) {
+				$trimmed = trim( $line );
+				if ( '' !== $trimmed ) {
+					$formatted_lines[] = "\n\t" . $trimmed;
+				}
+			}
+			
+			return implode( '', $formatted_lines ) . "\n";
+		} else {
```
Crabcyborg: @Liviu-p Since there is a `return` above this, we don't really need the `else`.

PR #2603, opened 2025-11-26
Support iframed Gutenberg

truongwp:
This increases the `apiVersion` of blocks to support the iframed Gutenberg.

This also fixes some warnings:
Screen Shot 2025-11-27 at 01 33 11

**Before:** the editor is not in any `<iframe>`

https://github.com/user-attachments/assets/d28d49d5-c259-446e-9bd7-db0e232634b6

**After:** the editor is in an `<iframe>` tag

https://github.com/user-attachments/assets/bb73d814-c0d6-4960-992c-9aeb73a00eb1

This also fixes JS error with the Formidable Views block after running `npm run build` (when Pro isn't installed):
Screen Shot 2025-11-27 at 22 39 21


Crabcyborg (approved):
I haven't tested this yet but these changes look good.

As for the failing tests, those should be fixed in master now. I've merged those changes into this branch and I think it should all pass.

Thanks Truong!

truongwp:
I'm not sure how to fix those ESLint errors.

PR #2630, opened 2025-12-04
Pass email object to frm_notification hook

AbdiTolesa:
Required for: https://github.com/Strategy11/formidable-logs/pull/54


Crabcyborg:
I don't love that this passes the whole `$this` object.

It looks like it only even needs `package_atts`.

PR #2631, opened 2025-12-04
Revert black text is used for errors

Crabcyborg:
This changed during the Quick Settings update https://github.com/Strategy11/formidable-forms/commit/30c73f1b33ec772f649a804c9194e0cb950e81e8

But it's wrong. I figured this was intentional or something, but talking to Razvan, it sounds like this is just a bug.

- `$text_color_error` is intended for the input text color (inside of the input). By default this is a shade of gray.
- We want our errors to be red by default, and we want it to match the red error colour of the borders. It should still be using `$border_color_error`, which was intentionally set that way.

Screen Shot 2025-12-04 at 4 10 11 PM


Liviu-p (approved):
Looks good to me @Crabcyborg. Thank you for taking care of this!

PR #2648, opened 2025-12-12
Fix issues with duplicating Section field

AbdiTolesa:
Fixes https://github.com/Strategy11/formidable-pro/issues/5630

This PR fixes a few issues related to duplicating section fields in/outside of a field group. Screen casts of those issues are available in the issue description for reference to the test this update.


lauramekaj1:
> @lauramekaj1 Thanks for mentioning that I should test before saving the fields! I was able to reproduce those issues and have worked on fixing them. Could you give it another around of test from your end?

@AbdiTolesa thank you for addressing the issues. I tested the fixes and now it's working fine. However I noticed this error shown on forms, entries, styles and some other pages: 
image

lauramekaj1:
> Thanks @lauramekaj1! That error likely isn't related to this update since this update only includes JS files update but that error is coming from PHP. I need your confirmation on that too by switching to the master branch since I don't see that error in the main branch or this one. Could you create a new issue to track that with the steps needed, if you can still reproduce that error?

@AbdiTolesa I wasn't able to reproduce the PHP error again. But I tested the other fixes and it's working fine now. Thank you!

lauramekaj1 (approved):

truongwp (approved):
This looks good to me! Just one thing is that the CSS files in `master` are minified, but the ones in this PR are not. Can you run `npm run build` again to minify them? @AbdiTolesa 

@Crabcyborg Look like there are some problems with Cypress in the `master` branch.

Crabcyborg:
Closing this PR. It's a bit old, and it'll be easier to start from master when we fix this.

PR #2653, opened 2025-12-12
Clean usage data before sending

truongwp:
Related issue: https://github.com/Strategy11/formidable-pro/issues/6112

This PR does:
- Change field options and form action settings from JSON to array.
- Data contains `@` or key ends with `_email` will be replaced by `{{contain_email}}`.
- Some long text data (`description` and key ends with `_msg`) is replaced by `{{long_text}}`. We clean more of them at the usage site.
- Keys that end with `_url` are removed.
- More keys are removed before sending. I copy them from the usage site.

This requires a change in the usage site to support the array instead of the old JSON data: https://github.com/Strategy11/ff-usage-2/pull/6

## Bug Fixes

* Enhanced data sanitization with improved filtering of sensitive information before transmission
* Email addresses and sensitive fields are now masked with placeholders to protect user privacy
* Long text values replaced with placeholders to optimize data transmission size
* Admin email data removed from usage snapshots for improved privacy compliance

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/classes/controllers/FrmAddonsController.php
+++ b/classes/controllers/FrmAddonsController.php
@@ -1362,7 +1362,7 @@ class FrmAddonsController {
 		}
 
 		// If empty license, save it now.
-		if ( empty( self::get_pro_license() ) && function_exists( 'load_formidable_pro' ) ) {
+		if ( ! self::get_pro_license() && function_exists( 'load_formidable_pro' ) ) {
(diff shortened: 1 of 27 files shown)


Crabcyborg (approved):
Awesome.

Thanks @truongwp!

🚀

PR #2661, opened 2025-12-15
Try breaking field group when frm_first class is added to a field in a group

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/5228

https://github.com/user-attachments/assets/2b66e31b-caea-4bf6-b539-f25f5354f016


Crabcyborg:

The solution here doesn't really use any of the proper helper functions and doesn't appear to handle field groups properly still.

PR #2664, opened 2025-12-17
Lite: Fix number field validation appearing when min/max values are empty

shervElmi:
### Pro PR

https://github.com/Strategy11/formidable-pro/pull/6137

## Summary by CodeRabbit

* **New Features**
  * Added validation for range and numeric field settings in the form builder: invalid configs (e.g., min ≥ max, non-positive or oversized step) now show clear error messages and highlight affected fields.
  * Improved field-type/ID detection to make validations more reliable.
  * Validation utilities are now exposed so other admin tools can trigger or reuse these checks.


Crabcyborg (approved):
I made several changes.

This no longer requires a JS var to determine if Pro is connected.

I also made some changes in Pro. When Lite wasn't up to date, the errors were not appearing.

And the step validation was weird. I updated it to check for the difference between max and min.

PR #2681, opened 2025-12-23
Lite: Fix repeater button styling in form builder

shervElmi:
Fixes https://github.com/Strategy11/formidable-pro/issues/6142

Updated the repeater Add/Remove buttons in the form builder to match the form preview.

### Add-ons Changes

- https://github.com/Strategy11/formidable-pro/pull/6170


resources/scss/admin/components/builder/_form-rows.scss
```diff
@@ -23,15 +29,15 @@
...
-.frm-page-break::before,
-.frm-page-break::after {
-	content: '';
-	position: absolute;
-	background-color: #fff;
-	border-radius: 50%;
-	border: 1px solid var(--primary-700);
-	text-rendering: auto;
-	-webkit-font-smoothing: antialiased;
+	&::before,
```
Crabcyborg: @shervElmi Shouldn't we still be referencing `.frm-page-break` here?

shervElmi: Yeah, we should, and it is still referencing that.

Image

It seems you need to uncollapse the editor where it is not showing any changed lines.

Crabcyborg: Thanks. The way GitHub is showing the diff is bizarre.

resources/scss/admin/components/builder/_form-rows.scss
```diff
@@ -2,17 +2,23 @@
...
-#frm_builder_page .frm_remove_form_row i::before,
-#frm_builder_page .frm_add_form_row .frmsvg,
-#frm_builder_page .frm_add_form_row i::before {
-	color: var(--primary-700);
+#frm_builder_page {
+	.frm_remove_form_row.frm_button,
+	.frm_add_form_row.frm_button {
+		display: inline-flex;
+		align-items: center;
+		gap: 8px;
```
Crabcyborg: @shervElmi What do you think about pulling any new rules into Pro's `builder.css` file instead?

I don't love having this in Lite.

Crabcyborg:
Also @shervElmi,

It looks like the icon buttons are using other styles still.

Screen Shot 2026-01-02 at 12 44 32 PM

shervElmi:
> Also @shervElmi,
> 
> It looks like the icon buttons are using other styles still.

@Crabcyborg the PR also includes Pro changes, and you can find the Pro PR in the description.

Crabcyborg:
@shervElmi That's only a size change? I'm seeing issues with colour.

Crabcyborg:
@shervElmi If I have that branch checked out, it still doesn't reflect the icon I have selected in my style.

I'm wondering if we want to bother with parts of this update though. The builder isn't really 100% the same as the style. Since the builder does not have a background colour, we don't really want to show a white button here in the builder.

I think we should at least show the right icon though, since that shouldn't cause any issues.

Screen Shot 2026-01-02 at 12 50 16 PM

Crabcyborg:
I'm just going to close this for now since it still changes colours.

Since the background is always white, a lot of buttons could just disappear.

I think we're better off the way it currently is.

PR #2682, opened 2025-12-23
Reposition frm_extra_form_instructions_hook

AbdiTolesa:
Related ticket: https://secure.helpscout.net/conversation/3174611369/244321


Crabcyborg:
@AbdiTolesa I think this is too risky. I left a comment in Helpscout about how this is a breaking change https://secure.helpscout.net/conversation/3174611369/244321#thread-9700394418

I think we could add a new hook instead, and consider deprecating the old hook.

PR #2708, opened 2026-01-05
Fix denylist spam check warnings for nested field values (Ranking)

shervElmi:
Fixes https://github.com/Strategy11/formidable-surveys/issues/342

This PR fixes repeated PHP warnings triggered when editing an entry with a Ranking field. The spam denylist checker attempted to `implode()` field values that can include nested arrays, causing “Array to string conversion” warnings.  
It now uses the existing `FrmAppHelper::safe_implode()` helper to safely flatten arrays before imploding, preserving existing behavior for normal field values while preventing warnings for nested data.


Crabcyborg:
The issue looks like a duplicate of https://github.com/Strategy11/formidable-pro/issues/5806

Thanks for catching that it's related to ranking fields! This has been tricky to actually replicate.

PR #2728, opened 2026-01-07
Add ranking fields to denylist skip list

Crabcyborg:
It looks like ranking fields were causing form validation issues.

I think we should probably ignore ranking fields as well since they're similar to the other skipped field types (radio, checkbox, select).


truongwp (approved):

PR #2748, opened 2026-01-09
fix custom css prefixing class on style duplicate

Liviu-p:


Crabcyborg:
Replaced with a safer and more robust fix https://github.com/Strategy11/formidable-forms/pull/3116

PR #2777, opened 2026-01-12
Fix hidden column reset when inbox count changes

shervElmi:
Fixes https://github.com/Strategy11/formidable-pro/issues/5794

This PR fixes an issue where hidden column preferences reset when the inbox unread count changes.

### Pro PR

- https://github.com/Strategy11/formidable-pro/pull/6189

## Summary by CodeRabbit

* **Refactor**
  * Improved menu badge handling and display logic in admin interface
  * Simplified dashboard and entries screen options filter registration
  * Standardized filter hook naming patterns across controllers

* **Deprecation**
  * Deprecated unread inbox count retrieval helper method


Crabcyborg (approved):
I think this should be good.

We'll likely lose some settings for people who were at a certain inbox number, but it's probably not worth any extra work trying to pull the last setting they had.

🚀

Crabcyborg:
Actually, I'm going to hold it here.

It looks like `get_visible_unread_inbox_count` is also called in the coupons add-on.

PR #2815, opened 2026-01-15
New sniff to add missing return void comments

Crabcyborg:


Crabcyborg:
Giving up on this one.

PR #2843, opened 2026-01-20
Fix HTML field ignoring layout classes

AbdiTolesa:
Related issue: https://secure.helpscout.net/conversation/3197229748/245409

**Field setup**:
image

**Before**:
image

**After**:
image

## Release Notes

* **Style**
  * Improved text wrapping behavior in form field containers to prevent text overflow and enhance readability of long text in HTML container fields.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


lauramekaj1 (approved):
@AbdiTolesa I tested this and verified that the issue is fixed. Thank you!

truongwp (approved):

Crabcyborg (approved):
I moved the new CSS outside of the if condition since it isn't related.

I think this change makes sense otherwise.

🚀

PR #2920, opened 2026-02-03
Fix backslash accumulation in Stripe and Square settings

shervElmi:
## Related
Fixes https://github.com/Strategy11/formidable-pro/issues/6294

## Root Cause
WordPress adds slashes to `$_POST` data via `wp_magic_quotes()`. Both `FrmStrpLiteSettingsController::process_form()` and `FrmSquareLiteSettingsController::process_form()` passed raw `$_POST` to the settings update method without calling `wp_unslash()` first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed improper handling of special characters when saving Square and Stripe integration settings, ensuring data is processed correctly before storage.
  * Improved data sanitization for Stripe processing messages to prevent character encoding issues.

### DIFF
--- a/stripe/models/FrmStrpLiteSettings.php
+++ b/stripe/models/FrmStrpLiteSettings.php
@@ -91,11 +91,13 @@ class FrmStrpLiteSettings {
 		foreach ( $settings as $setting => $default ) {
 			if ( isset( $params[ 'frm_' . $this->param() . '_' . $setting ] ) ) {
 				$value = $params[ 'frm_' . $this->param() . '_' . $setting ];
+
 				// Clean up any accumulated backslashes from previous bug.
 				// @see https://github.com/Strategy11/formidable-pro/issues/6294
 				if ( 'processing_message' === $setting ) {
 					$value = str_replace( '\\', '', $value );
 				}
+
 				$this->settings->{$setting} = sanitize_text_field( $value );
 			}
 		}


shervElmi:
@Crabcyborg,

> I think, to fix this issue on existing sites, it would be nice if we trimmed all of the excess backslashes if possible the next time they save.

I looked into auto-cleaning existing corrupted data on save, but it's tricky. The simple approach in the model's [update()](https://github.com/Strategy11/formidable-forms/blob/fix/payment-settings-backslash-accumulation/stripe/models/FrmStrpLiteSettings.php#L88) method would be:

```php
$value = $params[ 'frm_' . $this->param() . '_' . $setting ];
// Clean up any accumulated backslashes from previous bug.
if ( 'processing_message' === $setting ) {
    $value = wp_unslash( $value );
}
$this->settings->{$setting} = trim( sanitize_text_field( $value ) );
```

However, this creates two issues:
1. **Removes intentional backslashes** - If a user types `\This payment...`, it would strip the `\` on every save
2. **Partial cleanup** - For heavily corrupted data like `\\\\\\\\Hello`, it only removes half the backslashes per save

The alternative is pattern detection:

```php
// Only clean if value has 2+ consecutive backslashes (bug pattern)
if ( preg_match( '/\\\\{2,}/', $value ) ) {
    $value = preg_replace( '/\\\\+/', '', $value ); // Remove all consecutive backslashes
}
```

But this would also remove intentional escaped backslashes (e.g., `\\` that users want to display as `\`).

However, it doesn’t follow the “do no harm” principle. Could I have your input on my concern? Thanks!

PR #2932, opened 2026-02-03
Rock: Redesigned forms list

truongwp:
Handoff: https://linear.app/strategy11/document/feature-handoff-aa4fc7646f11

This adds a new gear icon to change the screen options more quickly.

## Summary by CodeRabbit

* **New Features**
  * Settings column and dropdown added to forms list for quick access
  * Option to show/hide form descriptions in the list, persisted per user
  * New screen options and templates to customize visible columns and per-page pagination

* **Style**
  * New admin styles for forms list UI, settings panel, collapsible sections, and controls

* **Chores**
  * Client-side behavior for settings panel, apply/save actions, and collapsible interactions


lauramekaj1 (approved):
@truongwp I tested it and it's working as expected. Thank you!

Crabcyborg (commented):
@truongwp I believe with this update we're planning to also remove the screen options tab, since the options are now all covered with the new gear icon.

Image

Crabcyborg:
@truongwp It's okay if there are some DeepSource errors.

I'd prioritize the other workflows and ignore DeepSource if necessary.

classes/controllers/FrmFormsController.php
```diff
@@ -1300,6 +1316,9 @@ public static function hidden_columns( $hidden_columns ) {
 	 * @return mixed
 	 */
 	public static function save_per_page( $save, $option, $value ) {
+		// phpcs:ignore WordPress.Security.NonceVerification.Missing
+		update_user_option( get_current_user_id(), 'frm_forms_show_desc', ! empty( $_POST['frm_forms_show_desc'] ) );
```
Crabcyborg: @truongwp Do we really need a new option?

Can't we use the excerpt view setting?

truongwp: The excerpt view setting doesn't show the full long description. So I created this option.

Crabcyborg: Do we even want the full long description? I feel like the design wasn't intending on a new setting - just a misunderstood redesign of the existing setting.

Maybe we can just name it "Show Excerpt" instead. I can ask Mike Dalton.

Crabcyborg: @truongwp 

So, I've sent a question to Dalton and Razvan.

Razvan has confirmed that this is meant to replace the excerpt view toggle. It isn't intended to be a new feature, and we're not trying to just ignore the old excerpt view setting for any one who was already using it.

For now, we can go with naming the setting "Description Excerpt" instead.

Everything should work based on the previous setting.

Image

Crabcyborg: Also, Truong. Looking at my screenshot, the number for "Items per page" seems really close to the up/down arrows. This is on Firefox. I'm not sure if it's different in other browsers, but if we can add some )(just a tiny bit) extra padding between the number and the arrows I think it would look nicer.

truongwp: @Crabcyborg I removed the new user option for the form description and updated the padding of that number input.

js/admin/forms-list.js
```diff
@@ -0,0 +1,180 @@
...
+	function closeSettingsPanel( panel ) {
+		panel.classList.remove( 'frm-forms-list-settings--visible' );
+		panel.addEventListener( 'transitionend', () => {
+			panel.classList.add( 'frm_hidden' );
+		}, { once: true } );
+	}
+
+	function handleClickFormsListSettings( event ) {
+		event.preventDefault();
+		const btn = 'A' === event.target.tagName ? event.target : event.target.closest( 'a' );
```
Crabcyborg: @truongwp Is this check necessary?

If `event.target.closest( 'a' )` is an anchor tag, then `event.target.closest( 'a' )` returns `event.target` already.

So `const btn = 'A' === event.target.tagName ? event.target : event.target.closest( 'a' );` and `const btn = event.target.closest( 'a' );` should always have the same results but the second check is more simple.

truongwp: Thank you. I changed this.

Crabcyborg (approved):
Thanks @truongwp!

I made some changes. I left a summary in this comment in the other PR https://github.com/Strategy11/formidable-pro/pull/6304#issuecomment-4681963737

🚀

PR #2941, opened 2026-02-05
Usage tracking for test mode in Lite

truongwp:
This tracks the number of times user views the test mode without the add-on installed and with the chat form.

This is tracked in the flows data with the format:
```
[view_test_mode] => Array
    (
        [no_addon] => 1
        [chat_form] => 2
    )
```

## Summary by CodeRabbit

* **Enhancements**
  * Test mode now records views of chat-enabled forms.
  * Test mode now records when the required addon is unavailable.
* **New Features**
  * The Test Mode interface now supports additional form-action controls, allowing extended actions to appear alongside the enabled form actions.
* **Documentation**
  * Clarified usage tracking behavior for test mode activity.


classes/views/test-mode/container.php
```diff
@@ -88,6 +88,17 @@
 				}
 				?>
 				</select>
+
+				<?php
```
Crabcyborg, line 92: @truongwp This is for the add-on tracking?

Can we add a new hook here instead so we can move this logic into the add-on?

truongwp: Can we use `$enabled` to check, like we did with other test mode add-on features?

Crabcyborg: @truongwp Like the hook is only called if `$enabled` is `true`?

That would be fine.

Otherwise I'm not sure if I follow.

truongwp: @Crabcyborg No. I meant we won't need to add a hook there, just check `if ( $enabled ) { foreach ... }`. Because that file contains many test mode add-ons code.

Crabcyborg: @truongwp

I just think there's a difference.

This file is intended to use the whole UI, as it's shown in Lite as disabled placeholders.

And these inputs are only used when the add-on is actually active, and have no visual benefit for being in Lite.

Since they only impact the add-on, it would be nice if we only needed to modify the add-on when we needed to make changes.

truongwp: Ok. I added the hook and moved the code to the add-on.

Crabcyborg (approved):
Thank you @truongwp!

🚀

PR #2945, opened 2026-02-06
Fix Conditional Logic not copied over when duplicating a field

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/6308

## Steps to reproduce

1. Add a field to your form and setup conditional logic on it and save your changes.
2. Duplicate the field and save the form.
3. Confirm that the Conditional logic is copied over to the new field after page is reload.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parent-form detection when loading individual fields, reducing errors and making field editing and display more reliable—especially for fields lacking an explicit parent reference. This update ensures the correct form relationship is determined from available field or form data, decreasing incorrect form assignments and improving consistency and stability when opening, editing, and displaying field-to-form associations.

### DIFF
--- a/classes/controllers/FrmFieldsController.php
+++ b/classes/controllers/FrmFieldsController.php
@@ -216,13 +216,13 @@ class FrmFieldsController {
 	 * @return int
 	 */
 	private static function get_form_id_from_field_or_form( $field_object, $values ) {
-		$form_id = $field_object->form_id;
+		$form_id = absint( $field_object->form_id );
 
 		if ( isset( $values['form_key'] ) ) {
 			return $values['id'] ?? $form_id;
 		}
 
-		$parent_form_id = FrmDb::get_var( 'frm_forms', array( 'id' => $form_id ), 'parent_form_id' );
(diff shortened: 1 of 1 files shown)


lauramekaj1 (approved):
@AbdiTolesa I tested it and it's working as expected. Thank you!

classes/controllers/FrmFieldsController.php
```diff
@@ -205,6 +205,28 @@ public static function load_single_field( $field_object, $values, $form_id = 0 )
...
+	 * Get the parent form id for the field.
+	 *
+	 * @since x.x
+	 *
+	 * @param object $field_object The field object.
+	 * @param array  $values       Either form or field values.
+	 *
+	 * @return int
+	 */
+	private static function get_parent_form_id( $field_object, $values ) {
```
truongwp: I think we should rename this method, like `get_parent_form_id_from_...`. We have already had the `FrmProFieldHelper:get_parent_form_id()` method.

truongwp (approved):

Crabcyborg:
Looks like this was recently fixed another way.

Closing this.

PR #2965, opened 2026-02-23
Fix min version notice overlapping header tabs on form builder

shervElmi:
Fixes https://github.com/Strategy11/formidable-pro/issues/6330

Adds CSS rule to adjust the form builder container position when the min version notice is displayed, preventing the notice from overlapping the header tabs.

## Testing

1. Use an old version of Pro (v6.19 or lower) with a newer Lite version
2. Navigate to the form builder page
3. Verify the min version notice appears above the header tabs without overlapping

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed page layout spacing when a minimum-version notice banner appears to prevent content overlap.
  * Recalculated container offsets on window resize to keep layout consistent.
  * Improved full-screen/admin toolbar styling to ensure correct top padding when the banner is present.


Crabcyborg:
I'm just going to close this for now. The ideal solution wouldn't require JS. I don't love that this issue is still an issue on load and only gets fixed when the JS runs.

PR #2979, opened 2026-02-25
[AI] Fix inline Submit button position

AbdiTolesa:


Crabcyborg:
I'm just going to close this.

I don't love this solution, and I'm not seeing any requests to fix this.

PR #3021, opened 2026-03-13
Fix "After Payment" setting not imported for Payment actions

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/6377

### Test steps
1. Create a Payment form
2. Go to the Payment Action and add "After Payment" updates mapping fields as needed for test
3. Export the form and import it back.
4. Go to the Payment Action and confirm that the "After Payment" field mappings are imported successfully.
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Form duplication now correctly updates nested and array-valued field entries so duplicated forms keep proper field references.
  * Duplicate-ID mappings are applied earlier in the duplication process to ensure accurate field ID replacement for complex form configurations.

### DIFF
--- a/classes/helpers/FrmFieldsHelper.php
+++ b/classes/helpers/FrmFieldsHelper.php
@@ -1847,7 +1847,7 @@ class FrmFieldsHelper {
 
 			$val[ $k ] = str_replace( $replace, $replace_with, $v );
 			unset( $k, $v );
-		}
+		}//end foreach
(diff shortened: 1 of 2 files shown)


Crabcyborg:

The fix seems to touch a lot of code that wouldn't usually be relevant. I'm concerned this is adding too much complexity given the issue.

PR #3049, opened 2026-03-27
Wrap custom html field labels in error class when showing in field validation messages

AbdiTolesa:
Fix https://github.com/Strategy11/formidable-pro/issues/6392

### Test procedure
1. Enable AJAX submission in your form
2. Add a field and set it to be required
3. For the field label, use html that has a `<div>` tag in it. Eg. "Enter your answer<div>< i >here< /i ></div>"
4. Preview the form and try submitting empty values
5. Observe that there are validation errors printed
6. Submit again (before reloading the page and clearing the previous errors)
7. Confirm that the errors are not duplicated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error message handling with consistent HTML formatting and sanitization for all field error displays.


garretlaxton (approved):
This fix works great! 

I do want to confirm that script tags should be allowed in the field labels though?

truongwp:
@garretlaxton No, it shouldn't be allowed in the label.

garretlaxton:
@truongwp I am able to add a script tag to the label that runs when viewing the form, unless I have `define( 'DISALLOW_UNFILTERED_HTML', true);` set. This also happens in the main branch, so it's not something new with this PR.
image

truongwp:
@Crabcyborg Should we pass the label through `wp_kses_post()`?

@garretlaxton If that's an issue, we will fix it in a separate PR.

Crabcyborg:
@truongwp I think it should be fine. I don't expect that people are actually using scripts in field labels. I expect that people use descriptions or custom HTML if they do need to add scripts.

Just in case, I think a new hook would be nice. Like `frm_filter_field_labels`. We can default it to true, but in case someone is using scripts in labels, they could just add a `__return_false` to revert the behaviour.

truongwp:
@Crabcyborg I think we can leave it as is.

@garretlaxton Currently, we allow <script> tag in field label.

garretlaxton:
@truongwp Thanks Truong! This looks good then!

js/formidable.js
```diff
@@ -1685,6 +1709,19 @@ function frmFrontFormJS() {
 		return `${ uniqueKey }-${ timestamp }`;
 	}
 
+	/**
+	 * @since x.x
+	 *
+	 * @param {string} html
+	 * @return {string} Sanitized HTML string.
+	 */
+	function purifyHtml( html ) {
```
Crabcyborg: @truongwp I don't think this really purifies in its current state. It looks like it would still allow every tag and every attribute.

The `frmDom.cleanNode` functionality is pretty important.

truongwp: Ok. I don't remember why I found this code. Is it ok if we load the dom.js file in frontend?

Crabcyborg: I think we probably shouldn't.

Maybe there's a safer and more simple way to parse the message?

truongwp: Is it ok if we don't parse the message?

Crabcyborg: It would be nice if it were filtered somewhere (can we do so in PHP?).

It looks like the previous code is already doing the same sort of thing, so it doesn't appear to be making things worse than they currently are.

truongwp: Ok. I removed the usage if purifyHtml() function.

Crabcyborg:
Closing this. I think that adding the class isn't really the ideal solution. I'd like to fix this without introducing styling.

PR #3075, opened 2026-04-14
Fix price validation

Crabcyborg:
I'm seeing this validation error coming from Lite when I change my price.

The issue is that `target` is really an `HTMLEvent`, not a `HTMLElement`.

> Uncaught TypeError: can't access property "trim", e.value is undefined

Screen Shot 2026-04-14 at 10 18 24 AM


truongwp:
@Crabcyborg I've already seen this error too.

In "formidable-pro/js/src/builder/builder.js", in `handleChangeEvent()` function, we call the validate function with target: `validateProductPriceValue( target );`.

Crabcyborg:
@truongwp Yeah. The difference is that in Pro it's called inside of a `handleChange` function, not directly in the jQuery `.on` code.

I don't think a `handleChange` function is as useful in the `admin.js` file currently. It would need to handle too much. We need to break it all down into smaller files still. And a new function just for this change event that just calls `validateProductPriceValue` would likely be redundant as well.

We can always change this later if we want to call the function without an event.

Crabcyborg:
@truongwp I just disabled Pro.

The way the preview syncs looks really weird. It's showing inputs and every option.

https://github.com/user-attachments/assets/7784965a-9387-450f-868d-fd7e49c83fce

Crabcyborg:
I just created a new issue for my previous comment (https://github.com/Strategy11/formidable-pro/issues/6416).

It looks like this is an issue on older branches as well. We can work on that later.

truongwp:
> @truongwp Yeah. The difference is that in Pro it's called inside of a `handleChange` function, not directly in the jQuery `.on` code.
> 
> I don't think a `handleChange` function is as useful in the `admin.js` file currently. It would need to handle too much. We need to break it all down into smaller files still. And a new function just for this change event that just calls `validateProductPriceValue` would likely be redundant as well.
> 
> We can always change this later if we want to call the function without an event.

Ok. If you decided to go with this fix, let me update the Pro code to use `event` instead of `target`. This error happens with old version so I think we don't need to handle the backward compatibility.

Crabcyborg:
@truongwp I think we can leave Pro alone.

The functions don't need to be identical.

This error is only in Lite with the new pricing fields updates.

truongwp (approved):
Sorry for messing this up. This is good. Thank you.

Crabcyborg:
Thanks Truong!

PR #3080, opened 2026-04-17
Fix DeepSource Warnings 

shervElmi:
DeepSource was flagging view templates for undefined variables because those variables are actually passed in from the including controller. This adds PHPDoc headers to document the inherited variables so that DeepSource (and IDEs) can understand the template context.

### DIFF
--- a/.github/workflows/mago.yml
+++ b/.github/workflows/mago.yml
@@ -14,11 +14,8 @@ jobs:
       - name: Install dependencies
         run: composer install --dev --prefer-dist --no-progress
 
-      - name: "⚡ Install Mago"
-        uses: nhedger/setup-mago@v1
-
       - name: "✅ Mago Lint"
-        run: mago lint
+        run: ./vendor/bin/mago lint
 
       - name: "🔎 Mago Analyze"
-        run: mago analyze
+        run: ./vendor/bin/mago analyze
--- a/.github/workflows/mago.yml
+++ b/.github/workflows/mago.yml
@@ -15,12 +15,10 @@ jobs:
         run: composer install --dev --prefer-dist --no-progress
 
       - name: "⚡ Install Mago"
-        run: |
-          mkdir -p bin
-          curl --proto '=https' --tlsv1.2 -sSf https://carthage.software/mago.sh | bash -s -- --install-dir=bin
+        uses: nhedger/setup-mago@v1
 
       - name: "✅ Mago Lint"
-        run: ./bin/mago lint
+        run: mago lint
 
       - name: "🔎 Mago Analyze"
-        run: ./bin/mago analyze
+        run: mago analyze
--- a/.github/workflows/mago.yml
+++ b/.github/workflows/mago.yml
@@ -14,8 +14,13 @@ jobs:
       - name: Install dependencies
         run: composer install --dev --prefer-dist --no-progress
 
+      - name: "⚡ Install Mago"
+        run: |
+          mkdir -p bin
+          curl --proto '=https' --tlsv1.2 -sSf https://carthage.software/mago.sh | bash -s -- --install-dir=bin
+
       - name: "✅ Mago Lint"
-        run: ./vendor/bin/mago lint
+        run: ./bin/mago lint
 
       - name: "🔎 Mago Analyze"
-        run: ./vendor/bin/mago analyze
+        run: ./bin/mago analyze
--- a/classes/views/frm-entries/sidebar-shared.php
+++ b/classes/views/frm-entries/sidebar-shared.php
@@ -4,7 +4,7 @@
  *
  * @package Formidable
  *
- * @var int          $id      Entry ID.
+ * @var string       $id      Entry ID.
  * @var stdClass     $entry   Entry object.
  * @var string|null  $browser Browser info string if available.
  * @var array|null   $data    Additional entry meta data (referrer, user_journey, etc.).
--- a/classes/views/frm-entries/sidebar-shared.php
+++ b/classes/views/frm-entries/sidebar-shared.php
@@ -1,4 +1,15 @@
 <?php
+/**
+ * Shared entry sidebar.
+ *
+ * @package Formidable
+ *
+ * @var int          $id      Entry ID.
+ * @var stdClass     $entry   Entry object.
+ * @var string|null  $browser Browser info string if available.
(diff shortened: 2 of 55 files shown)


.github/workflows/mago.yml
```diff
@@ -15,12 +15,10 @@ jobs:
         run: composer install --dev --prefer-dist --no-progress
 
       - name: "⚡ Install Mago"
-        run: |
-          mkdir -p bin
-          curl --proto '=https' --tlsv1.2 -sSf https://carthage.software/mago.sh | bash -s -- --install-dir=bin
+        uses: nhedger/setup-mago@v1
```
Crabcyborg: @shervElmi I'm not sure about this change. It looks like this workflow you're referencing has 6 stars.

I don't really trust it.

classes/views/frm-fields/back-end/settings.php
```diff
@@ -1,4 +1,26 @@
...
+ * @var array        $field_types         Available field types this field can switch to.
+ * @var array        $disabled_fields     Field types that are disabled in the type dropdown.
+ * @var array        $default_value_types Default value types available for this field.
+ * @var array        $unique_values_label_atts     Label attributes for the Unique checkbox.
+ * @var array        $read_only_label_atts         Label attributes for the Read Only checkbox.
+ * @var bool         $show_upsell_for_unique_value Whether to show upsell for Unique.
+ * @var bool         $show_upsell_for_read_only    Whether to show upsell for Read Only.
+ * @var bool         $show_upsell_for_before_after_contents Whether to show upsell for before/after contents.
+ * @var bool         $show_upsell_for_autocomplete Whether to show upsell for autocomplete.
+ * @var bool         $show_upsell_for_visibility   Whether to show upsell for visibility.
```
Crabcyborg: @shervElmi These `$show_upsell_*` variables are only set conditionally. They use `! empty` intentionally as they're all really `bool|null` types.

js/src/admin/admin.js
```diff
@@ -818,18 +818,18 @@ window.frmAdminBuildJS = function() {
 
 	function clickNewTab() {
 		/*jshint validthis:true */
-		const t = this.getAttribute( 'href' );
-		if ( t === undefined ) {
+		const href = this.getAttribute( 'href' );
+		if ( href === undefined ) {
```
Crabcyborg: @shervElmi I don't actually think this `href === undefined` here is possible.

If there is no `href` at all, it's really `null`.

It looks like this is somewhere where `jQuery().att()` differs. The jQuery function returns `undefined`.

phpstan.neon
```diff
@@ -325,6 +326,7 @@ parameters:
 				- classes/helpers/FrmFormsHelper.php
 				- classes/helpers/FrmXMLHelper.php
 				- classes/models/FrmEntryValidate.php
+				- classes/views/frm-entries/_sidebar-shared-pub.php
```
Crabcyborg: @shervElmi It seems like you could just add `|null` to a type definition instead of ignoring this error.

classes/views/frm-entries/sidebar-shared.php
```diff
@@ -1,4 +1,15 @@
 <?php
+/**
+ * Shared entry sidebar.
+ *
+ * @package Formidable
+ *
+ * @var int          $id      Entry ID.
```
Crabcyborg: @shervElmi I believe this is really a `string`.

classes/views/frm-fields/back-end/field-credit-card.php
```diff
@@ -1,4 +1,12 @@
 <?php
+/**
+ * Credit card field form builder view.
+ *
+ * @package Formidable
+ *
+ * @var array|object $field Field data including 'form_id'.
```
Crabcyborg: @shervElmi From the code it looks like `$field` is always an `array`, never an object.

classes/views/frm-fields/back-end/default-value-field.php
```diff
@@ -1,4 +1,15 @@
 <?php
+/**
+ * Text input for the default value setting.
+ *
+ * @package Formidable
+ *
+ * @var FrmFieldType $this          Field type handler that included this template.
+ * @var array        $field         Field data including 'id' and 'field_key'.
+ * @var string       $default_name  HTML name attribute for the input.
+ * @var mixed        $default_value Current default value.
```
Crabcyborg: @shervElmi Are we sure `$default_value` is `mixed`?

It seems to go through `echo esc_attr();` implying it may be a `int|float|string` at most. An `array` or `object` or any other type would result in an error.

.github/workflows/mago.yml
shervElmi: Thanks for the feedback! I've reverted the Mago changes back to master.

classes/views/frm-fields/back-end/settings.php
shervElmi: https://github.com/Strategy11/formidable-forms/pull/3080/commits/a0d5bf0bc

js/src/admin/admin.js
shervElmi: https://github.com/Strategy11/formidable-forms/pull/3080/commits/89569fdf4

phpstan.neon
shervElmi: https://github.com/Strategy11/formidable-forms/pull/3080/commits/7f9c93f0e

classes/views/frm-entries/sidebar-shared.php
shervElmi: https://github.com/Strategy11/formidable-forms/pull/3080/commits/18ad5b318

classes/views/frm-fields/back-end/field-credit-card.php
shervElmi: https://github.com/Strategy11/formidable-forms/pull/3080/commits/548a5b86c

classes/views/frm-fields/back-end/default-value-field.php
shervElmi: - https://github.com/Strategy11/formidable-forms/pull/3080/commits/9c5811e38
- https://github.com/Strategy11/formidable-forms/pull/3080/commits/40a1cfcf9

Crabcyborg (approved):
Thank you @shervElmi!

I think these changes look good!

🚀

PR #3085, opened 2026-04-21
Form actions refresh paypal compat

shervElmi:
This adds a small, scoped style patch in Lite so the PayPal action settings keep a clean look until the PayPal add-on catches up.

### Before

CleanShot 2026-04-21 at 21 10 59

### After

CleanShot 2026-04-21 at 21 09 02


resources/scss/admin/frm-legacy-paypal.scss
```diff
@@ -0,0 +1,46 @@
+// Start @since x.x
+// Temporary PayPal styles for legacy table-based markup, scoped to .frm_single_paypal_settings and removable after view refresh.
+.frm_single_paypal_settings {
```
Crabcyborg, line 2: @shervElmi Is it possible to only include this CSS conditionally when the PayPal add-on is active?

Since most users don't require this, it seems unnecessary to always having it loading.

Maybe something similar to `enqueue_legacy_views_assets`.

A "enqueue_legacy_paypal_assets" would work here.

And we can change "removable after view refresh." to something like "removable after the PayPal add-on is officially no longer supported.

shervElmi: - https://github.com/Strategy11/formidable-forms/pull/3085/commits/4d536c1dafce4572d15cc783339fb1b8e9c83f61
- https://github.com/Strategy11/formidable-forms/pull/3085/commits/a956f61a44a35f4f7ac3ffcd8181c207fb0dca7a
- https://github.com/Strategy11/formidable-forms/pull/3085/commits/4ff852692f7136073c34653baa781910bab47a79

Crabcyborg (approved):
Thanks @shervElmi!

🚀

PR #3095, opened 2026-05-08
Update gateway setting toggle logic

Crabcyborg:
This update should prevent this issue where "Repeat" appears twice.

**The problem**
- Stripe uses "Repeat Interval" / "Repeater Period".
- Square uses "Repeat Cadence".
- The JS to sync which setting is visible misses this case and ends up displaying both the Stripe and Square settings.

**To replicate**
1. Create a payment action. Set it to One Time, Square.
2. Change to Stripe.
3. Change to Recurring.

**When testing**
- We'll want to test with the Stripe add-on active, and inactive, is these may work differently.
- It's also good to test Authorize.Net active and inactive, as the gateways use checkboxes when Authorize.Net is available. While you shouldn't have both Stripe and Square selected (it won't really work), both settings should appear if both gateways are selected.

**Before**

https://github.com/user-attachments/assets/0dbdd0a2-45c5-48eb-9a69-426ff62db173

**After**

https://github.com/user-attachments/assets/41e62151-add3-42f2-9868-e015ffa3a3dc

## Summary by CodeRabbit

* **Refactor**
  * Improved payment gateway settings visibility so sub-options consistently show or hide based on transaction type (recurring vs. one-time) and selected gateways.
  * Ensures layout for gateway sub-option groups remains intact when displayed.
  * Fixes cases where repeat-cadence controls could remain hidden after being added, improving UI reliability.

### DIFF
--- a/square/js/action.js
+++ b/square/js/action.js
@@ -15,12 +15,18 @@
 		}
 
 		if ( 'recurring' === typeDropdown.value ) {
-			const activeGateway = settings.querySelector( '[name*="[post_content][gateway]"]:checked' )?.value;
+			const activeGateways = Array.from( settings.querySelectorAll( '[name*="[post_content][gateway]"]:checked' ) ).map( function( el ) {
+				return el.value;
+			} );
 
 			settings.querySelectorAll( '.frm_trans_sub_opts' ).forEach(
 				function( subOpts ) {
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
This fix works great!

Crabcyborg:
Thank you @garretlaxton!

🚀

PR #3098, opened 2026-05-22
Update colorpicker logic

Crabcyborg:
Requires this Pro PR. Without the Pro changes, it will change from hex to RGB when picking a color https://github.com/Strategy11/formidable-pro/pull/6467

This was a request from Razvan.

**What this update changes**
The format would change back and forth between hex and rgb, and wasn't really in control of the user. With this update, the format should generally remain. If you set an alpha value (some transparency), it should change to rgba.

**Before**

https://github.com/user-attachments/assets/6f76be05-ee37-4be6-8167-e73d4ea13e13

**After**

https://github.com/user-attachments/assets/26fb79ff-c133-4a5c-a5a1-5657851f8d38

## Summary by CodeRabbit

* **Bug Fixes**
  * Color picker now preserves and respects the input's original color format (hex, rgb/rgba, hsl/hsla), forces rgba for transparency, and no longer overwrites non-hex inputs.
  * Color output logic updated so rgb(...) values are not misclassified as hex or auto-converted to rgba, improving saved color fidelity.

* **Chores**
  * Improved legacy text-insertion handling to better maintain cursor position when inserting content.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Strategy11/formidable-forms/pull/3098?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


garretlaxton:
This works great, but when I am focused on a field, there is a visual glitch.
image

Crabcyborg:
Thanks @garretlaxton!

Do you see the same issue on the master branch?

I'm having issues replicating it. I didn't touch styles though, so I'm wondering if that's a separate issue.

garretlaxton:
@Crabcyborg, yes I am seeing this on the master as well.
image

This only happens when I move the slider:
image

I am also seeing it on Chrome & Safari. Maybe it's my smaller screen size? There is some inline CSS that is forcing the width to 130px.
Screenshot 2026-06-03 at 12 04 25 PM

Crabcyborg:
Thank you @garretlaxton!

I'm using a small screen as well.

I think it could possibly be your theme?

If you can look more closely at which CSS rule is causing it, and which file it is coming from, that would really help.

For now I'm just going to merge this. I think your issue can be handled separately.

garretlaxton:
I just tried a few different themes. The CSS is inline so it's not coming from a file.

Screenshot 2026-06-03 at 1 54 07 PM

PR #3103, opened 2026-05-26
Fix PHP notice related to script registering in Styles settings

truongwp:
Fixes https://github.com/Strategy11/formidable-pro/issues/6338

## Summary by CodeRabbit

* **Refactor**
  * Improved internal script and style registration logic in the form controller system, optimizing how dependencies are initialized for the form builder and Visual Styler components.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Strategy11/formidable-forms/pull/3103?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/classes/controllers/FrmStylesController.php
+++ b/classes/controllers/FrmStylesController.php
@@ -127,6 +127,8 @@ class FrmStylesController {
 		}
 
 		FrmStyleComponent::register_assets();
+		FrmFormsController::register_formidable_script();
+		self::register_style_script();
 		self::load_pro_hooks();
 
 		$version = FrmAppHelper::plugin_version();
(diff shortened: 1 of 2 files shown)


classes/controllers/FrmStylesController.php
```diff
@@ -127,6 +127,8 @@ public static function admin_init() {
 		}
 
 		FrmStyleComponent::register_assets();
+		FrmFormsController::register_formidable_script();
```
Crabcyborg, line 130: @truongwp I think we might want to fix this in a different way.

We don't want the front end scripts loading in the styler. Those are off intentionally.

truongwp: In `FrmProFormsController::admin_footer()`, we have a call to `enqueue_footer_js()`, which enqueues the Pro JS.

Crabcyborg: Hm, I think we'll want to fix this in Pro then. It sounds like Pro shouldn't be doing that either.

truongwp: Do we need to load the frontend JS on the Styles page?

Crabcyborg: @truongwp No, the front end JS is only meant for the front end, and the entry editing page (which uses front end forms).

Pages like the styler, builder, and form settings, should never load the front end JS.

truongwp: Ok. I will dequeue the frontend JS on the Styles page.

truongwp: I created PR https://github.com/Strategy11/formidable-pro/pull/6487 in Pro to fix this issue. We can close this for now.

Crabcyborg:
Thank you @truongwp!

PR #3116, opened 2026-06-01
Style settings improvements (fix scoping issues on duplicate and when…

Crabcyborg:
… renaming, use a more accurate style class name based on the name in the input

Fixes https://github.com/Strategy11/formidable-pro/issues/6157

The custom CSS is saved with the style name prefixed. The automatic scoping is there to prevent conflicts and unnested in the settings so it is more user friendly.

On duplicate, it would use the old style name.

In this update, the unnested/nesting happens when the style name changes appropriately.

**In addition**
- The "Style Class" would previously show a random string, and then change on save based on the style name setting. Now, since the style name is initially typed in a modal, the style class on load is based on that string so it's closer to what it will be on submit. If the style name is already in use, it will end up with `-2`, `-3`, appended to it.

https://github.com/user-attachments/assets/ab7a632f-4266-435a-af39-b3d15752fb6e

### DIFF
--- a/classes/controllers/FrmStylesController.php
+++ b/classes/controllers/FrmStylesController.php
@@ -631,8 +631,8 @@ class FrmStylesController {
 				// The single style custom CSS is nested under the old style's scope. Re-scope it to the
 				// new style's scope so it unnests correctly for display and re-nests correctly on save.
 				if ( ! empty( $style->post_content['single_style_custom_css'] ) ) {
-					$css_scope_helper = new FrmCssScopeHelper();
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This fix works good for me! 

I do want to confirm that when you rename a style the `Style Class` shouldn't be renamed, correct?

Image

Crabcyborg:
@garretlaxton Yes. We don't want to frequently change the style class since it may be in use. There is another open issue I believe to make the style name editable, but it hasn't been implemented yet.

We only try to update it for a new style.

Testing this, I am seeing another funny behaviour though - renaming a new style, then clicking save, reverts the name change.

I'll merge this and consider making further improvements.

Thank you!

PR #3149, opened 2026-06-15
Extend sales api cache timeout

Crabcyborg:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API resilience and application stability by implementing intelligent cache timeout adjustments. When the external API service experiences rate limiting, the system automatically extends the cache duration to gracefully handle the situation, reducing failed requests and maintaining consistent user experience.

### DIFF
--- a/classes/models/FrmSalesApi.php
+++ b/classes/models/FrmSalesApi.php
@@ -70,6 +70,22 @@ class FrmSalesApi extends FrmFormApi {
 	}
 
 	/**
+	 * If the last check was a rate limit, we'll need to check again sooner.
+	 *
+	 * @since 6.25.1
+	 *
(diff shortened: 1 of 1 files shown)


srwells:
the doc comment is copy-pasted from the parent and still says "we'll need to check again *sooner*," but the code now backs off *longer*. And `@since x.x` is a placeholder that needs a real version before merge.

Crabcyborg:
@srwells We always use `@since x.x` and update this in the release process. That way the versions are never incorrect and I don't waste any time fixing versions.

Also, it still does back off for less time, but I agree we can improve the comment. I merged this before seeing your comment. I'll tweak it in my release changes I'm making today.

Thank you!

PR #3160, opened 2026-06-18
Remove payment cron when all gateways are disconnected

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/3358713219/253763

_**Summary**_
- The `frm_payment_cron` cron is scheduled when the payments tables are created.
- It would only get unscheduled when deactivating Lite, or when disconnecting both Stripe gateways.

**The solution**
- This update adds new hooks for when a payment gateway is disconnected. It checks in this hook if all gateways are disconnected. If they are, the payment cron should automatically get unscheduled (and no longer appear in the WP Control settings).

You can use the WP Crontrol plugin to view your Cron events.

Screenshot 2026-06-18 at 4 58 30 PM

### DIFF
--- a/classes/controllers/FrmAppController.php
+++ b/classes/controllers/FrmAppController.php
@@ -1330,14 +1330,18 @@ class FrmAppController {
 
 	/**
 	 * The payment cron is unscheduled when Formidable is deactivated.
-	 * We need to add it back again on activation if Stripe is configured.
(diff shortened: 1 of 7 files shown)


garretlaxton (approved):
This works good!

Crabcyborg:
Thank you @garretlaxton!

🚀

PR #3189, opened 2026-07-16
Fix archive page show 404 because of the gated content

truongwp:
This fixes https://secure.helpscout.net/conversation/3388018170/255310

## Summary by CodeRabbit

* **Bug Fixes**
  * Gated content unlocking now runs only on individual post or page views.
  * Prevented access-code checks, redirects, and restricted-content handling from affecting archive, listing, and other non-singular pages.


Crabcyborg (approved):
Thanks @truongwp!

🚀

PR #3193, opened 2026-07-16
More accessible captcha labels

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/3386866379/255259

Fixes this WAVE issue.

The CAPTCHA inputs are always in an iframe, so a label can't actually direct to any inputs.

Screenshot 2026-07-16 at 1 58 21 PM

In this update,
1. If the label position is set to "None" (the default value), the label is removed altogether.
2. If the label position is not set to "None", the label element is converted to a span tag.


garretlaxton (approved):
This works great!

Crabcyborg:
Thank you Garret!

PR #3197, opened 2026-07-22
Get all form actions for usage tracking

Crabcyborg:
It looks like our usage tracking would only send data for 100 form actions. This update uses batches to pull all form actions, 100 at a time.

This is memory sensitive, so I added some `unset` code and a call to `gc_collect_cycles`. I also removed our caching layer here, as I think it would just add to memory in most cases since this query isn't really happening in any other code. Without the extra logic, the query is significantly faster as well.

## Summary by CodeRabbit

* **Bug Fixes**
  * Usage snapshots now capture all available actions, including installations with more than 100 actions.
  * Improved reliability when loading large sets of usage data.

### DIFF
--- a/classes/models/FrmUsage.php
+++ b/classes/models/FrmUsage.php
@@ -627,6 +627,8 @@ class FrmUsage {
 
 			$offset += $limit;
 			unset( $saved_actions );
+
+			gc_collect_cycles();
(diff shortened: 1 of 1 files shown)


truongwp (approved):

Crabcyborg:
Thanks Truong!

PR #3198, opened 2026-07-22
Add unsaved changes popup to global settings

Crabcyborg:
This adds the confirmation pop-up if you update a global setting and then try to leave without saving.

When testing, we want to make sure there aren't any non-settings inputs that trigger this (like a search).

And that it actually triggers for all unsaved global settings changes.

The pop-up contents vary from browser to browser. In Firefox it looks like this.

Screenshot 2026-07-22 at 1 47 15 PM

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented search, add-on license, and payment-related fields from incorrectly marking global settings as having unsaved changes.
  * Improved unsaved-change tracking for other global settings fields.


garretlaxton:
This works pretty good. The only spot it doesn't popup is the Custom CSS tab. Should it trigger there?

Crabcyborg:
Thanks @garretlaxton!

The custom CSS uses code mirror, so it uses special listeners.

This should work now as well.

garretlaxton (approved):
This works great! 

The only minor suggestion I have is disabling this for the `Uninstall Formidable Forms and permanently delete all data.` option on the miscellaneous tab. That way, if someone misclicks, they don't accidentally save and delete all their data. But I'll leave that up to you to decide.

PR #3211, opened 2026-07-27
Fix gated content and custom post type capabilities

### LINKED ISSUE #3203: Gated Content: private custom post types with their own capability_type are forced to 404 for all non-editor users
**Plugin version:** Formidable Forms 6.33.1 (also present in 6.33)
**WordPress:** 7.0.2 · **PHP:** 8.4 (live) / 8.1 (local)
**Severity:** high — every member of a real production site lost access to all restricted content

### Summary

`FrmGatedContentController::maybe_unlock_post()` decides whether a private post is restricted
using a **hard-coded** capability:

```php
// classes/controllers/FrmGatedContentController.php:163
$is_restricted_private = 'private' === $post->post_status
    && ! current_user_can( 'read_private_posts', $post_id );
```

`read_private_posts` is only the correct capability for post types that use the default
`capability_type`. A custom post type registered with its own `capability_type` is governed by
`read_private_<something>s` instead, and `read_private_posts` is meaningless for it.

For any such post type the check is therefore always "restricted" for every user who does not
happen to also hold `read_private_posts` (i.e. everyone below Editor), and `force_404()` is
called on content the user is fully entitled to read.

Two smaller points in the same line:

- `current_user_can( 'read_private_posts', $post_id )` passes an object ID to a **primitive**
  capability. `map_meta_cap` does not map `read_private_posts`, so the argument is silently
  ignored — the call is equivalent to `current_user_can( 'read_private_posts' )`.
- The post type's capability is available right there via `get_post_type_object()`.

### Steps to reproduce

No third-party plugin needed:

1. Register a public CPT with its own capability type:

   ```php
   register_post_type( 'demo_doc', array(
       'public'          => true,
       'capability_type' => 'demo_doc',   // => read_private_demo_docs
       'map_meta_cap'    => true,
   ) );
   ```

2. Create a `demo_doc` post and set its status to **private**.
3. Give a Subscriber-level role the capability `read_private_demo_docs`
   (and **not** `read_private_posts`).
4. Log in as that user and open the post.

**Expected:** the post is rendered — `current_user_can( 'read_post', $id )` is `true` and
WordPress itself includes the post in the main query.

**Actual:** `maybe_unlock_post()` calls `force_404()` on the `wp` hook (priority 10) and the
visitor gets a 404. The main query had already resolved the post correctly
(`is_single = true`, `post_count = 1`) before the hook ran.

### How this looked in production

Site with the Echo Knowledge Base suite: each knowledge base is its own CPT
(`epkb_post_type_N`, `capability_type = amgr_N_cpt`), all articles are `private`, and read
access is granted per KB group by the Access Manager. Measured for one affected member:

| Capability | Value |
|---|---|
| `read_private_amgr_8_cpts` (governs the post type) | `true` |
| `read_private_posts` (checked by Gated Content) | `false` |

Result after updating 6.30 → 6.33.1: **628 articles across 10 knowledge bases became
unreachable for all 306 member accounts.** Administrators, Editors and custom editorial roles
were unaffected because they hold `read_private_posts`, so the regression was invisible to the
site owner and only surfaced through member complaints.

Confirmed by removing only this one callback for a single request — the article renders with
HTTP 200 and all other access control stays in effect.

Gated Content is not configured at all on this site (0 gated content form actions, 0 tokens),
so the hook was rejecting content it had no reason to touch.

### Suggested fix

Resolve the capability from the post type instead of hard-coding it:

```php
$post_type_object = get_post_type_object( $post->post_type );
$read_private_cap = $post_type_object
    ? $post_type_object->cap->read_private_posts
    : 'read_private_posts';

$is_restricted_private = 'private' === $post->post_status
    && ! current_user_can( $read_private_cap );
```

Alternatively, and arguably more robust, use the meta capability and let `map_meta_cap` resolve
everything (private status, password, own posts) in one call:

```php
$is_restricted_private = 'private' === $post->post_status
    && ! current_user_can( 'read_post', $post_id );
```

### Related

6.33.1 already fixed a sibling case in the same feature — *"Public taxonomy pages could result
in 404 errors, caused by a missing check in the new Gated Content actions."* This report is the
same class of problem on the single-post path.

### Workaround currently in place

A small site plugin removes `FrmGatedContentController::maybe_unlock_post` on the `wp` hook at
priority 9, but **only** when the post is private, the post type has its own `read_private`
capability, and that capability passes for the current user. Gated Content keeps working
everywhere else. This will be removed once the fix ships.

truongwp:
Fix #3203 and https://secure.helpscout.net/conversation/3398946279/255863

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved private gated-content unlocking to respect content-type-specific “read private” permissions instead of a single hardcoded capability.
  * Prevented incorrect 404 behavior for taxonomy archive pages and for private items not referenced by gated-content actions.
  * Ensured gated private items are properly restricted (including 404 when no valid token exists).
  * Refined gated-content item selection to only surface private and password-protected content where applicable.
* **Tests**
  * Added/expanded PHPUnit coverage for gated-content access edge cases and for correct post selection/grouping behavior.

### DIFF
--- a/classes/controllers/FrmGatedContentController.php
+++ b/classes/controllers/FrmGatedContentController.php
@@ -160,10 +160,10 @@ class FrmGatedContentController {
 			return;
 		}
 
-		$is_password_protected  = '' !== $post->post_password;
-		$post_type_obj          = get_post_type_object( $post->post_type );
-		$read_private_cap       = $post_type_obj ? $post_type_obj->cap->read_private_posts : 'read_private_posts';
-		$is_restricted_private  = 'private' === $post->post_status && ! current_user_can( $read_private_cap, $post_id );
+		$is_password_protected = '' !== $post->post_password;
+		$post_type_obj         = get_post_type_object( $post->post_type );
+		$read_private_cap      = $post_type_obj ? $post_type_obj->cap->read_private_posts : 'read_private_posts';
+		$is_restricted_private = 'private' === $post->post_status && ! current_user_can( $read_private_cap, $post_id );
 		$access_code_from_url  = FrmAppHelper::simple_get( 'access_code' );
 
 		// Nothing to unlock — post is publicly accessible.
--- a/classes/controllers/FrmGatedContentController.php
+++ b/classes/controllers/FrmGatedContentController.php
@@ -160,8 +160,10 @@ class FrmGatedContentController {
 			return;
 		}
 
-		$is_password_protected = '' !== $post->post_password;
-		$is_restricted_private = 'private' === $post->post_status && ! current_user_can( 'read_private_posts', $post_id );
+		$is_password_protected  = '' !== $post->post_password;
+		$post_type_obj          = get_post_type_object( $post->post_type );
+		$read_private_cap       = $post_type_obj ? $post_type_obj->cap->read_private_posts : 'read_private_posts';
+		$is_restricted_private  = 'private' === $post->post_status && ! current_user_can( $read_private_cap, $post_id );
 		$access_code_from_url  = FrmAppHelper::simple_get( 'access_code' );
 
 		// Nothing to unlock — post is publicly accessible.
--- a/classes/controllers/FrmGatedContentController.php
+++ b/classes/controllers/FrmGatedContentController.php
@@ -228,14 +228,16 @@ class FrmGatedContentController {
 	private static function has_gated_action_for_item( FrmGatedItem $item ): bool {
 		global $wpdb;
 
-		$action_ids = FrmDb::get_col(
-			$wpdb->posts,
-			array(
-				'post_type'    => FrmFormActionsController::$action_post_type,
-				'post_excerpt' => FrmGatedContentAction::$slug,
-				'post_status'  => 'publish',
-			),
-			'ID'
+		// Direct query — FrmDb caches results per-request which would hide newly
+		// created actions until the cache expires. action_contains_item() already
+		// handles per-action caching via transients.
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.NoCaching,WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		$action_ids = $wpdb->get_col(
+			$wpdb->prepare(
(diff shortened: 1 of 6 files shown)


garretlaxton (approved):
This looks good!

classes/controllers/FrmGatedContentController.php
```diff
@@ -205,6 +214,41 @@ public static function maybe_unlock_post() {
...
+	private static function has_gated_action_for_item( FrmGatedItem $item ): bool {
+		global $wpdb;
+
+		// Direct query — FrmDb caches results per-request which would hide newly
+		// created actions until the cache expires. action_contains_item() already
+		// handles per-action caching via transients.
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.NoCaching,WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		$action_ids = $wpdb->get_col(
+			$wpdb->prepare(
+				"SELECT ID FROM {$wpdb->posts} WHERE post_type = %s AND post_excerpt = %s AND post_status = 'publish'",
```
Crabcyborg: @truongwp Can we use `%i` here for the table name?

classes/controllers/FrmGatedContentController.php
```diff
@@ -205,6 +214,41 @@ public static function maybe_unlock_post() {
...
+	 *
+	 * @return bool True if any published gated content action lists this item.
+	 */
+	private static function has_gated_action_for_item( FrmGatedItem $item ): bool {
+		global $wpdb;
+
+		// Direct query — FrmDb caches results per-request which would hide newly
+		// created actions until the cache expires. action_contains_item() already
+		// handles per-action caching via transients.
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.NoCaching,WordPress.DB.PreparedSQL.InterpolatedNotPrepared
```
Crabcyborg: I think we can remove `WordPress.DB.PreparedSQL.InterpolatedNotPrepared` from here if we use `%i`.

truongwp: I added `%i` and remove this rule.

Crabcyborg (approved):
Thank you @truongwp and @garretlaxton for working on this!

I think this is good now.

🚀

PR #3212, opened 2026-07-30
Stripe Connect: scheduled backstop pull for Connect events

NathanaelJonesIreland:
## Summary

Adds a site-side, hourly (filterable) WP-Cron **backstop** that pulls unprocessed Stripe **Connect** events, so event ingestion no longer depends solely on the Strategy11 relay pinging the site. The relay push stays the primary, near-real-time path; this cron only bounds the worst case.

## Problem

Formidable Stripe **Connect** does not receive Stripe webhooks directly. The flow is:

1. Stripe → Strategy11 relay (the Connect platform account).
2. The relay pings the site: `admin-ajax.php?action=frm_strp_process_events` (`FrmStrpLiteEventsController::process_connect_events`).
3. The site then **pulls** unprocessed event IDs from the relay and handles them.

`process_connect_events` is wired **only** to that ajax hook (`FrmStrpLiteHooksController::load_ajax_hooks`) — there is **no site-side scheduled pull**. So if the relay ping is delayed, dropped, or slow for a given site, the events (`invoice.payment_failed`, `charge.refunded`, `customer.subscription.deleted`, …) are not ingested until the *next* ping. The only existing backstop, `frm_payment_cron` (daily, `FrmTransLiteAppController::run_payment_cron`), is subscription-only and coarse, and does not cover non-subscription events such as refunds.

### Why it matters

Everything that reacts to `frm_payment_status_*` inherits this unbounded latency — Payments itself, and especially the payment-reactive add-ons:

- **Memberships** — renewal failure → suspend/downgrade, cancellation, refund → revoke access.
- **Workflows** — the `payment_failed` / `refunded` / `subscription_canceled` triggers (dunning / at-risk-subscriber automations).

These add-ons are — and should remain — purely *reactive*; they must not each poll the relay themselves (the event queue is core-owned; parallel pollers would race core's relay-triggered pull and each other, risking double-processed events). Fixing ingestion **once here in core** gives every consumer bounded, consistent delivery.

*(Observed in QA: a subscription renewal declined in Stripe (`past_due`), but the site showed the subscription still active and fired no `frm_payment_status_failed` until the endpoint was pulled manually — the relay had not yet pinged, and neither cron had run.)*

## What this changes

- **`pull_connect_events()`** — the get-unprocessed + `process_event_ids()` body is extracted from `process_connect_events()` into this shared method (no output, no `wp_die`), so it is safe to call from WP-Cron. The ajax handler keeps identical behaviour: `flush_response(); pull_connect_events(); wp_send_json_success();`. A short `frm_strp_connect_pull_lock` transient (with `try/finally` release) serializes the ajax and cron paths, which also closes a **pre-existing** relay-vs-relay double-process race.
- **`run_scheduled_pull()`** — the cron callback. Skips when the paid Stripe add-on is active (`class_exists( 'FrmStrpHooksController', false )`).
- **`maybe_schedule_connect_pull()`** (on `admin_init`) — schedules `frm_strp_pull_connect_events` when a Stripe mode is set up, reschedules if the interval changed, and **self-heals** (clears) when Stripe is no longer connected. **`maybe_remove_connect_pull()`** clears it on `frm_disconnected_gateway`.
- **`frm_strp_connect_pull_interval`** filter — cadence, default `hourly`, validated against registered WP-Cron schedules (falls back to `hourly`).
- Hooks registered in `load_hooks()` (runs on all requests, including WP-Cron — unlike `load_ajax_hooks`, which is `DOING_AJAX`-gated and would not bind during a real `wp-cron.php` request).

## Why it's safe

- **No double-processing.** `process_event_ids()` already dedupes per event (60s `frm_last_process_<id>` transient + the `events_to_skip` option + relay-side `process_event()`); the new shared lock adds top-level serialization. Net: this is *safer* than before (the pre-existing relay-vs-relay TOCTOU is now closed).
- **No new attack surface.** No new endpoint or external input. The cron runs server-side and authenticates to the relay via the existing stored Connect credentials — strictly less exposed than the existing `nopriv` ajax endpoint. No nonce/cap surface.
- **Gated + cheap.** Scheduled only when `FrmStrpLiteConnectHelper::at_least_one_mode_is_setup()`; on a connected but idle site it is one relay call/hour returning an empty set. Non-payment sites never schedule it. Skipped entirely when the paid Stripe add-on owns events.
- **Backward compatible.** Purely additive; the ajax path is behaviourally unchanged; no DB/schema change; rollback = unschedule.

## Performance / cadence

Default **hourly** (built-in schedule). The relay push remains the near-real-time primary, so this is a reconciliation backstop — hourly bounds worst-case latency without meaningful relay-platform load. Sites needing tighter dunning can opt into a shorter interval via `frm_strp_connect_pull_interval` (returning a registered schedule name) without changing the core default.

## Testing

Adds `tests/phpunit/stripe/test_FrmStrpLiteEventsController.php` (`@group stripe`): schedules hourly when connected; clears when not (self-heal); reschedules on interval change; interval-filter fallback + honouring a registered custom schedule; and the locked-pull no-op. These run in the standard WP-integration suite in CI. The end-to-end path (a real failed renewal → pull → `frm_payment_status_failed` → a Workflows `payment_failed` run) was also verified manually.

## Follow-up (separate PR)

A **just-in-time per-subscription** pull — schedule a single targeted pull at each subscription's expected renewal time (+grace), reusing `pull_connect_events()` — would give precise, low-load coverage for the renewal case. It belongs in core (which holds the period-end) and complements, but does not replace, this periodic backstop (Stripe retries span days; refunds have no predictable time). Kept out of this PR to keep a payment-critical change minimal.

## Notes for reviewers

- `@since 6.34` is a placeholder — please set to the intended release.
- Pre-existing, out of scope: if the paid Stripe add-on becomes active *after* Lite scheduled this cron, the Lite-scheduled event lingers as an inert no-op (both `run_scheduled_pull` and the schedulers early-return on the `FrmStrpHooksController` check). Worth the paid add-on clearing `frm_strp_pull_connect_events` on activation.

## Summary by CodeRabbit

* **New Features**
  * Added a scheduled backstop to retrieve Stripe Connect events when relay delivery is unavailable.
  * Prevented overlapping event pulls to avoid duplicate processing.
  * Automatically manages event-pull scheduling as Connect settings change.

* **Bug Fixes**
  * Improved reliability of Stripe Connect event ingestion through scheduled retries and safer processing coordination.


NathanaelJonesIreland:
**CI status update**

- **DeepSource: PHP — now passing.** I'd initially included a PHPUnit test (`test_FrmStrpLiteEventsController.php`), but DeepSource's PHP analyzer flagged a *newly-introduced* WP-integration test: it reported the inherited assertions (`assertSame`/`assertNotFalse`/`assertFalse`) as "undefined method" (it doesn't resolve the `FrmUnitTest → WP_UnitTestCase` chain for a file new in the diff — existing tests use them fine as baseline), and it doesn't narrow `wp_get_scheduled_event()`'s `object|false` after `assertNotFalse`. These are analyzer-resolution artifacts, not real defects, and can't be cleared without repo-wide `.deepsource.toml` changes or `skipcq` noise. I removed the test so the check is green — happy to add coverage in whatever way you prefer new tests to land (the two controller files themselves have zero DeepSource findings).

- **Mago — failing on pre-existing issues, not this PR.** The Mago annotations point at `classes/models/FrmGatedItem.php:116/120` and `classes/controllers/FrmStylesController.php:1542` — files this PR doesn't touch (`git diff --name-only master...HEAD` = only `stripe/controllers/FrmStrpLiteEventsController.php` + `FrmStrpLiteHooksController.php`). Mago is also red on `master`, so these are pre-existing repo issues rather than something introduced here.

- All other checks green (PHPStan, Psalm, Scrutinizer, CodeRabbit, DeepScan, DeepSource: JavaScript, Typos).

NathanaelJonesIreland:
@crabcyborg heads-up on the one remaining red check: **Mago** is failing, but on **pre-existing** issues in files this PR doesn't touch — `classes/models/FrmGatedItem.php:116` (invalid `clone` on `object|false` + a `never`-type assignment) and `:120` (redundant `(string)` cast), and `classes/controllers/FrmStylesController.php:1542` (impossible `$post_type` condition). `master`'s own Mago run is also red on these, so they're not introduced here — this PR's diff is only the two `stripe/controllers/` files.

I've deliberately left them out of scope to keep this PR focused on the Connect-events backstop. Happy to open a separate PR to clear the Mago debt if that's useful — just let me know. Everything else on this PR is green (DeepSource, PHPStan, Psalm, Scrutinizer, CodeRabbit, DeepScan, Typos).

Crabcyborg:
@NathanaelJonesIreland

Thanks for working on this.

Some notes:

- Can you never use hard coded since versions? We always use `x.x` and replace this at release, or the versions are going to likely be wrong in most cases.
- Can you assign me as a reviewer, not an assignee moving forward? I check for things where I've been assigned as a reviewer, but will likely miss anything where I'm an assignee.
- I believe most of our Stripe users are still using the Stripe Pro add-on. I'd really put this in the add-on first.
- This could introduce massive performance issues with our API with so many extra calls being made. I would love that if a ping ever comes across successfully that this cron be turned off indefinitely, as this would only cause performance issues for reachable sites.

NathanaelJonesIreland:
Closing in favour of **Strategy11/formidable-stripe#446**.

Per @Crabcyborg's review: most Stripe users run the paid **Stripe add-on**, and this Lite code even early-returns when that add-on is active — so a Lite-only fix would help almost nobody. The replacement PR puts the fix where the users are, and also redesigns it per the performance feedback: instead of an always-on hourly poll, the fallback cron now **self-disables permanently the first time the relay successfully pings the site**, so reachable sites make zero extra relay calls.

(If a Lite-only counterpart is wanted later for free-Stripe sites, it can mirror #446 — but leading with the add-on.)

PR #3214, opened 2026-08-03
Fix some mago issues

Crabcyborg:
Looks like the workflow is mostly failing after some gated content changes.

The `get_permalink` call would potentially get called with `null` because `$post` was only validated in the if condition above.

I haven't tested this.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gated content links by safely handling missing or invalid posts.
  * Preserved permalink generation for private posts.
  * Updated password-protected content testing with a generated password for greater reliability.

### DIFF
--- a/tests/phpunit/forms/test_FrmGatedContentAction.php
+++ b/tests/phpunit/forms/test_FrmGatedContentAction.php
@@ -53,7 +53,7 @@ class test_FrmGatedContentAction extends FrmUnitTest {
 		$post    = $this->factory->post->create_and_get(
 			array(
 				'post_status'   => 'publish',
-				'post_password' => 'secret',
+				'post_password' => wp_generate_password( 12 ),
 			)
 		);
 		$grouped = FrmGatedContentAction::get_posts();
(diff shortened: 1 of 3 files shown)


tests/phpunit/forms/test_FrmGatedContentAction.php
```diff
@@ -53,7 +53,7 @@ public function test_get_posts_includes_password_protected_post() {
 		$post    = $this->factory->post->create_and_get(
 			array(
 				'post_status'   => 'publish',
-				'post_password' => 'secret',
+				'post_password' => wp_generate_password( 12 ),
```
truongwp, line 56: I think this is unnecessary.

truongwp (approved):
Except my comment for the password change in unit tests, other things looks good to me. Thanks @Crabcyborg!

tests/phpunit/forms/test_FrmGatedContentAction.php
Crabcyborg: Mago sees "secret" as a password committed in code and flags it as compromised.

So it's really just to avoid a hard coded password.

Crabcyborg:
Thank you @truongwp!

🚀

PR #3217, opened 2026-08-04
Prevent duplicate data token attributes when loading multiple forms

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/3408027516/256229

**The issue**
The hooks load for every form, so additional forms loaded on a page include duplicate `data-token` attributes.

**The solution**
The hooks now only load once, and the code that adds the token checks that the current form actually has the JS spam validation setting enabled.

**Things to test**
- That `data-token` only appears once per element. It does appear on both the `div` and the `form`, but that's okay both are required depending on context.
- That `data-token` is not included at all for a form where the setting is disabled.
- That JS spam validation still works as expected.

**Pre-release**
[formidable-6.33.2b.zip](https://github.com/user-attachments/files/30710457/formidable-6.33.2b.zip)

## Summary by CodeRabbit

* **Bug Fixes**
  * Anti-spam tokens are now added only to forms with anti-spam protection enabled.
  * Prevented duplicate anti-spam filter registration, improving form processing reliability.

* **Tests**
  * Added coverage to verify correct token behavior and one-time filter registration.

### DIFF
--- a/classes/models/FrmAntiSpam.php
+++ b/classes/models/FrmAntiSpam.php
@@ -13,6 +13,18 @@ if ( ! defined( 'ABSPATH' ) ) {
 class FrmAntiSpam extends FrmValidate {
 
 	/**
+	 * Track when the token filters have been added so they are only added once.
+	 * The callback checks the Anti-Spam setting of the form being rendered, so
+	 * a single callback covers every form on the page. Adding one callback for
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
This works good!

Crabcyborg:
Thanks Garret!

🚀

PR #3241, opened 2026-08-14
Support for conditional PayPal commerce

Crabcyborg:
This update aims to match recent updates we did for Square (https://github.com/Strategy11/formidable-pro/issues/6335 and https://github.com/Strategy11/formidable-forms/pull/3278).

To allow for more than 1 PayPal Commerce action, there's a pro PR as well https://github.com/Strategy11/formidable-pro/pull/6612

- PayPal Commerce actions should work well with submit button conditional logic.
- The submit button should not be disabled if a PayPal commerce payment field is conditionally disabled.
- If using 2 PayPal Commerce actions with conditional logic, the amount should be for the action that should be conditionally enabled.

**Pre-release**
[formidable-6.35.1b.zip](https://github.com/user-attachments/files/32071504/formidable-6.35.1b.zip)

PR #3236, opened 2026-08-13
Break field group when frm_first class is added

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/5228

**Before**

https://github.com/user-attachments/assets/49437d2c-a14b-42da-9bb1-531d42f15d52

**After**

https://github.com/user-attachments/assets/a8b0cea3-7f65-4eae-95b7-118b739f5f71

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved field grouping behavior when applying a class that starts a new field group.
  * Automatically moves the selected field and subsequent fields into a newly created group.
  * Refreshes group controls to keep the form editor state accurate.

PR #3242, opened 2026-08-14
Derive the generated stylesheet's cache-busting version from its content

NathanaelJonesIreland:
## What

`FrmStylesController::get_css_version()` reads the `frm_last_style_update` option and appends it as the `?ver=` on the enqueued generated stylesheet (`css/formidableforms.css`). `FrmStyle::save_settings()` set that option to `gmdate( 'njGi' )` — unpadded month, day and hour, with no year — so the value is not a reliable marker of the file's content:

- **Distinct dates collide.** `1 Jan 10:59`, `11 Jan 00:59` and `1 Nov 00:59` all produce `111059`. Sampling one year at three minutes per hour (24,192 timestamps), 1,980 version strings are produced by more than one date, covering 4,212 timestamps — **17.4%**.
- **No year**, so every value repeats annually.
- **Minute resolution**, so two saves within the same minute are indistinguishable.

A third-party CSS cache keyed on the enqueued URL can therefore keep serving a copy generated from superseded content, while the styler preview — served through `admin-ajax.php` and not cached — looks correct.

This replaces it with `substr( md5( $css ), 0, 12 )`, so the version changes if and only if the generated bytes change.

## Why `FrmCreateFile::create_file()` is in this PR

A content-derived version is **idempotent**, so it must only be published once the bytes are known to be on disk. `create_file()` returned nothing and no-ops silently when it lacks filesystem permission or cannot create its directories, and it discarded `put_contents()`'s return.

Publishing a hash after a failed write would be unrecoverable: the option would read `hash(B)` while disk still held `A`; the file remains readable so `get_url_to_custom_style()` never falls back to the AJAX endpoint; and every later save of the same content reproduces the same hash and the same URL, pinning a downstream cache to `A` permanently. The old timestamp self-healed here, so gating on a confirmed write is required rather than optional — without it this change would be a regression.

`create_file()` therefore returns `bool`, and the version is stored only on success and only after `frmpro_css` is populated. The two other callers (`append_file()`, `combine_files()`) already invoked it as a bare statement and are unaffected; there is a test asserting that.

## Migration

`migrate_to_107()` deletes the legacy value so `get_css_version()` falls back to the plugin version, guaranteeing the enqueued URL changes on upgrade even when the post-upgrade `$frm_style->update( 'default' )` is skipped by its `function_exists( 'get_filesystem_method' )` guard. `$db_version` 106 → 107. I checked the open PRs: none other adds a migration or touches `$db_version`, so 107 is free — worth re-checking at merge time.

## Tests

28 tests / 176 assertions across the touched files; full suite 416 passing aside from pre-existing network- and environment-dependent failures. Each test was verified to **fail against the previous implementation**, not merely to pass against the new one — including reconstructing the unconditional-write state to confirm the write-failure guard catches it. The write-failure test forces a real filesystem failure rather than mocking. The migration test drives `migrate_data()`'s dispatch, so a migration that never runs cannot pass.

## Provenance and one honest caveat

This came out of a support ticket where a customer's datepicker header rendered Pro's hardcoded default instead of their configured Head Color, intermittently after updates, with the served stylesheet missing the rule that consumes `--date-head-bg-color`. Their site self-resolved before the mechanism could be proven, so **the causal link to that ticket is inferred, not established** — the customer had WP Rocket, and toggling a WP Rocket CSS setting both changes behaviour and purges its asset caches, which I could not separate.

The defect fixed here is provable independently of that ticket: the version string collides and omits the year regardless of which cache is downstream. I'd rather state that plainly than over-claim the fix.

## Not included

A related hardening change in Pro would make the datepicker's hardcoded colour defaults fall back through the corresponding style variables (`--date-head-bg-color`, `--date-head-color`, `--date-band-color`), so a missing per-style rule degrades to the configured colour rather than the shipped default. It touches the vendored `ui-lightness/jquery-ui.css` and there is no demonstrable live case for it now, so it is deliberately left out. Happy to raise it separately if wanted.

## Summary by CodeRabbit

* **New Features**
  * Stylesheet versions now reflect CSS content, ensuring reliable cache updates when styles change.
  * Database migration 107 removes a legacy style-update setting.

* **Bug Fixes**
  * Stylesheet versions no longer advance when CSS files cannot be written.
  * File creation now reports whether the operation succeeded.

* **Tests**
  * Added coverage for migrations, file-writing failures, and content-sensitive stylesheet version updates.


NathanaelJonesIreland:
Two distinct pieces of feedback landed here — one actionable, 43 not.

**CodeRabbit's minute-boundary finding: fixed in `62defb1c`.** It's a real flake. `test_same_minute_saves_with_different_content_produce_different_versions` took `time()` before and after the two saves and asserted both fell in the same `gmdate( 'njGi' )` bucket. That was a *test setup assumption* asserted as a hard failure — two consecutive in-process calls usually share a minute, but nothing stops them straddling a boundary, and when they do the build goes red while the product is entirely correct.

I didn't take the suggested remedy verbatim (remove the assertion, rename, revise the docblock) because removing it alone would have left the test an exact duplicate of `test_update_css_version_is_sensitive_to_content_in_both_directions` two methods up. The premise was never needed — the version is a content hash, so it distinguishes the two saves whenever they happen — so instead the clock-independence claim is now asserted *directly*: each version must equal `substr( md5( $css ), 0, 12 )` for its own content. No minute-granularity value, or any other clock-derived one, can satisfy both assertions, so the reversion the old check was reaching for is now caught deterministically rather than only on a lucky run. Renamed and docblock rewritten to match, since "same minute" is no longer the premise.

**`DeepSource: PHP` (red) is not this PR's.** It's failing on `master` at the same time, so it's pre-existing rather than introduced here. The 43 inline comments break down as:

- **`md5()` "insecure function"** — the rule is about password hashing (`password_hash()`). This `md5()` is a cache-busting content digest for a stylesheet URL, which is exactly what a fast non-cryptographic digest is for. No change made, deliberately.
- **~40 × "Call to an undefined method `test_FrmMigrate::assertFalse()` / `assertSame()`"** — DeepSource isn't resolving PHPUnit's `TestCase` base class, so every assertion in a test file reads as undefined. Same false positive on every test file in the repo.

Nothing to fix on either, so nothing was pushed for them.

Added the `run tests` label so the PHPUnit job actually runs against the new commit — it was skipping, so the test suite hadn't been exercised on this branch at all. `php -l` is clean, and PHPCS finds the same 13 pre-existing errors in this file before and after the change (all in `test_save_settings_leaves_version_untouched_when_file_write_fails`, untouched here).

NathanaelJonesIreland:
CI confirmation on `62defb1c`: **PHP 7.4 / WP 6.9 and PHP 8 / WP 6.9 both pass** — 416 tests, 2189 assertions, 0 failures. Worth noting the suite had never run on this branch before the label went on, so the styles tests added here were unverified until now, not merely unreported.

`DeepSource: PHP` remains red for the pre-existing reasons above (same failure on `master`). Cypress and CodeRabbit were still in flight when I checked.

NathanaelJonesIreland:
Pushed a377edff2 in response to this round of feedback.

**CodeRabbit — `test_save_settings_leaves_version_untouched_when_file_write_fails` could pass without reaching `create_file()`** — fixed, and the finding was correct. `save_settings()` returns at `classes/models/FrmStyle.php:411` when `css/custom_theme.css.php` is missing, so the sentinel assertion would have held for the wrong reason. The test now asserts the source stylesheet exists before the call, and that `frmpro_css` was populated after it. That option is only written once `create_file()` has returned, so a populated `frmpro_css` sitting next to an untouched `frm_last_style_update` proves the write was genuinely attempted and genuinely reported failure — which is the whole point of the test.

**CodeRabbit — assert the ordering at the version-write boundary** — fixed, also correct. The claim is an ordering one and was only being checked against the end state, which a regression that advances the version first and stores the CSS afterwards would still satisfy. The test now hangs a `pre_update_option_frm_last_style_update` filter and records what `frmpro_css` (option and transient) held at the instant the version write was attempted, then asserts that content is exactly what the version hashes. The end-state assertions are kept, so both the ordering and the resulting consistency are covered. `pre_update_option_{$option}` fires ahead of both the no-change short-circuit and the `add_option()` branch, so it is reached once per write even on the fresh-option path this test sets up.

**DeepSource: PHP — "Call to an undefined method `test_FrmStyle::assertSame()`" (x2)** — not changed; these are false positives. `assertSame()` comes from `WP_UnitTestCase` via `FrmUnitTest`, which DeepSource cannot resolve because the WordPress test framework is not vendored here; the same call appears throughout `tests/phpunit/` and only these two are flagged because DeepSource comments on diff lines. Both PHPUnit matrix jobs pass, and `DeepSource: PHP` is already failing on `master`, so it is not a signal this PR introduced.

**Cypress** — not changed; unrelated to this PR. The single failure is `admin-html-validation.cy.js` → "Check the global settings page has valid HTML", from a duplicate `frm_connect_with_oauth` ID inside `#frm_strp_settings_container`. That is Stripe settings markup, nowhere near the style-version change, and the same job is red on every branch currently running it. Fixing it belongs in its own PR.

Also ran `php-cs-fixer` and PHPCS locally over the changed file (both are label-gated in CI and skipped on this PR). The fixer wanted a blank line before three `foreach` blocks added earlier in this PR — applied. PHPCS now reports the same 8 findings as before the change, all pre-existing filesystem-op and line-length ones in the test's setup; no new ones.

Leaving the PR open for a human review.

NathanaelJonesIreland:
Follow-up: my first push (a377edff2) turned the PHPUnit matrix red, and 506f5e59b fixes it. Both PHP 7.4 and PHP 8 jobs are green again.

The cause was in my own new assertion, not in the product. I had asserted `save_settings()` writes the version exactly once; it can write twice. `get_css_content()` renders `custom_theme.css.php`, which reads `FrmStyle::get_all()`, and when no style rows exist yet that creates a default style and calls `update( 'default' )` — so `save_settings()` again. Whether that re-entrant save happens depends on the state the rest of the suite leaves behind, so the count was never the right thing to pin.

The ordering invariant the test exists to protect is unaffected: each `save_settings()` stores `frmpro_css` before writing its own version. So it now asserts that invariant for *every* observed write, with the write index in the failure message, rather than for one expected write. That is stronger than what I first pushed — a regression that advanced the version ahead of the CSS in any of the writes now fails, where before an extra write could have gone unexamined.

Current check state on 506f5e59b: PHPUnit (7.4 and 8), PHPStan, Psalm, Mago, Typos, DeepScan, Scrutinizer and CodeRabbit all pass. The two reds are the ones described in my previous comment and unchanged by this PR — `DeepSource: PHP` (false-positive `assertSame()` resolution; also red on `master`) and `Cypress` (the duplicate `frm_connect_with_oauth` ID in the Stripe settings markup, still the identical single failure).

NathanaelJonesIreland:
Automated babysit pass. This round surfaced 13 new comments and two red checks. **No code change was warranted — nothing pushed.** Here's the account.

### The 11 DeepSource "undefined method" comments — false positives, no change

DeepSource flagged `assertTrue`, `assertNotEmpty`, `assertCount` and `assertSame` in `tests/phpunit/styles/test_FrmStyle.php` as *"Call to an undefined method"*, all at critical severity. They're analyser resolution failures, not defects:

- `test_FrmStyle extends FrmUnitTest` → `FrmUnitTest extends WP_UnitTestCase` (`tests/phpunit/base/FrmUnitTest.php:3`) → PHPUnit's `TestCase`. Every one of those assertions is inherited and defined. DeepSource isn't following the chain into the WP test library, which isn't in the analysed tree.
- **Both PHPUnit jobs pass on `506f5e59`** — `PHP 7.4 tests in WP 6.9` and `PHP 8 tests in WP 6.9` are green. A genuinely undefined method would fatal, not pass.
- The same calls appear in pre-existing code in this very file (e.g. `$this->assertSame` at line 36), untouched by this PR and never flagged before it.

### The two red checks — both pre-existing, neither caused by this PR

**`DeepSource: PHP`** is failing on `master` as well (`gh api repos/.../commits/master/status` → `DeepSource: PHP: failure`). It's a commit status rather than a check run, so it's red independently of this branch.

**`Cypress`** fails on one spec, `admin-html-validation.cy.js`:

```
no-dup-id: Duplicate ID "frm_connect_with_oauth"
selector: #frm_strp_settings_container > div:nth-child(2) > a
AssertionError: 1 error, 129 excluded: expected 1 to equal 0
```

The duplicate ID originates in `stripe/views/settings/connect.php:30` — a file this PR does not touch. This PR changes only `FrmAppHelper`, `FrmCreateFile`, `FrmMigrate`, `FrmStyle` and their tests. The identical failure reproduces on unrelated branches (e.g. `support_for_conditional_paypal_commerce`, run `31756505061`, same rule, same selector, same assertion), which makes it a repo-wide pre-existing failure rather than a regression here.

I deliberately did **not** fix the Stripe duplicate ID: it's outside this PR's scope, and folding an unrelated markup fix into a stylesheet-versioning change would muddy both. It looks worth its own issue.

### Status

Not merging — `reviewDecision` is `REVIEW_REQUIRED` and there's no approving review from a human yet. Branch is `MERGEABLE` with no conflicts; the PHPUnit matrix, PHPStan, Psalm, Mago, Typos, DeepScan and Scrutinizer are all green. Ready for human review.

NathanaelJonesIreland:
Automated babysit pass: both red checks here are pre-existing and unrelated to this PR — no change pushed.

**Cypress** — the single failure is `admin-html-validation.cy.js` → "Check the global settings page has valid HTML": `no-dup-id`, `Duplicate ID "frm_connect_with_oauth"`, selector `#frm_strp_settings_container > div:nth-child(2) > a`. The only source of that ID is [stripe/views/settings/connect.php:30](https://github.com/Strategy11/formidable-forms/blob/master/stripe/views/settings/connect.php#L30), so the global settings page is emitting that anchor more than once. This PR's diff is `FrmAppHelper`, `FrmCreateFile`, `FrmMigrate`, `FrmStyle` and their tests — it touches no `stripe/`, view or JS code, and a stylesheet cache-version change has no path to a duplicate ID. Confirmed repo-wide rather than assumed: Cypress is currently red on #3241, #3236, #3227 and #3223 as well.

**DeepSource: PHP** — already failing on `master` itself (`repos/Strategy11/formidable-forms/commits/master/status` reports `DeepSource: PHP` → `failure`), so it is not a signal about this branch.

Fixing the duplicate ID would mean editing the Stripe settings view, well outside this PR's scope, so I've left it alone. The PR is still `REVIEW_REQUIRED` and needs a human reviewer.

NathanaelJonesIreland:
CI triage for the two red checks — no code pushed, since neither failure is caused by this diff.

**Cypress (E2E Test)** — 1 of 18 specs fails, `admin-html-validation.cy.js`:

```
no-dup-id | Duplicate ID "frm_connect_with_oauth"
          | #frm_strp_settings_container > div:nth-child(2) > a
AssertionError: 1 error, 129 excluded: expected 1 to equal 0
```

That is the Stripe Connect settings view rendering one ID twice. This PR touches no Stripe or OAuth file (changed files are `FrmAppHelper`, `FrmCreateFile`, `FrmMigrate`, `FrmStyle` and their tests). There is no master baseline to diff against, because the E2E workflow is gated on the `run tests` label and is `skipped` on every recent master commit.

**DeepSource: PHP** — reports three issue types:

- `PHP-A1004` "insecure md5()" at `FrmStyle.php:456` — this one is genuinely on a line this PR adds. It is a false positive against a security rule: the hash is a cache-busting fingerprint of the generated stylesheet bytes, never a credential. Content-derived is the whole point of the change, so the value must be a pure function of the CSS. Seven other `md5()` calls already exist in `classes/` unannotated, so I did not want to invent a `skipcq` precedent unilaterally — happy to add `// skipcq: PHP-A1004`, or switch to a truncated sha256, if a reviewer prefers either.
- `PHP-E1002` (undefined `assertSame`/`assertInstanceOf`/…) and `PHP-W1067` (undefined `$factory`) in `tests/phpunit/database/test_FrmMigrate.php` at lines 20–88 — these are **pre-existing lines**, not added here; `git blame` puts line 20 on Mike Letellier, 2026-02-06. They surface as "introduced" only because this PR appends two test methods at line ~300, so the whole file is re-analysed. The analyser cannot resolve the `FrmUnitTest` base class; the untouched sibling `test_FrmDb.php` uses the same pattern. DeepSource: PHP is also red on 3 of the last 4 master commits.

PHPUnit (PHP 7.4 and PHP 8 on WP 6.9), PHPStan, Psalm, Mago, DeepScan, CodeRabbit and Scrutinizer are all green.

Leaving this for a human call rather than pushing a lint suppression into a security rule.

NathanaelJonesIreland:
Babysit pass. **Nothing pushed** — but one thing has changed since the last three triage rounds, and it changes the conclusion.

### DeepSource: PHP is *not* pre-existing on master. It is this PR's own line.

Earlier rounds here recorded it as red "for the pre-existing reasons above (same failure on `master`)". I checked, and that premise no longer holds:

```
gh api repos/Strategy11/formidable-forms/commits/master/status
→ DeepSource: PHP: success — Analysis passed
```

`master` is **green**. Of the six other open PRs I sampled, five are green on `DeepSource: PHP` too. The single finding on this PR is a critical-severity comment on `classes/models/FrmStyle.php:456` — the new `substr( md5( $css ), 0, 12 )` line — under **PHP-A1004**, *"Audit required: Use of an insecure hashing function"*. So this will not clear itself, and it is why `mergeStateStatus` is `BLOCKED`.

**On the merits it is still a false positive**, and clearly so: the rule is about password storage, and `md5( $css )` is a cache-busting fingerprint of a generated stylesheet. Its own suggested remedy, `password_hash()`, is salted and deliberately non-deterministic — it would destroy the exact property this PR is built on, that the version changes if and only if the bytes change. So the *finding* is wrong; what is new is that it is ours and it is permanent until someone acts.

Three ways to clear it, in the order I'd pick them:

1. **Audit the occurrence in the DeepSource dashboard.** PHP-A1004 is an *audit-required* rule, so this is the mechanism DeepSource intends: mark it intentional and the check goes green with no code change. Needs your dashboard access, which is why I have not done it.
2. **`// skipcq: PHP-A1004`** above the line, with a one-line WHY. This is already the house pattern — `skipcq` appears in five places under `classes/`, and the repo's own `CommentSpacingSniff` whitelists the prefix explicitly. Durable and visible in review, at the cost of a suppression comment in product code.
3. **Change the function** — e.g. `hash( 'crc32b', $css )`. Sidesteps the rule rather than suppressing it, and arguably states intent better, since a cache key wants a fast non-cryptographic fingerprint. But it is a change to your design on a finding that is wrong, so it is your call, not mine.

I did not push any of them: you have declined a change here three times, and this is product code on a false positive. Flagging the corrected facts so the fourth decision is made on accurate ones.

### Cypress — verified unrelated, unchanged

Independently confirmed from the run log rather than carried over:

```
no-dup-id | Duplicate ID "frm_connect_with_oauth"
          | #frm_strp_settings_container > div:nth-child(2) > a
AssertionError: 1 error, 129 excluded: expected 1 to equal 0
```

1 of 18 specs, `admin-html-validation.cy.js`. The ID comes from the Stripe Connect settings view; this diff touches `FrmAppHelper`, `FrmCreateFile`, `FrmMigrate`, `FrmStyle` and tests, and nothing under `stripe/`. Not this PR's.

Everything else is green: both PHPUnit matrices, Psalm, PHPStan, Mago, Typos, DeepScan, DeepSource: JavaScript.

### Note on the babysit loop

This PR has now been surfaced to four consecutive runs by the same two red checks, because the scan's watermark covers comments but not check state. Once you have picked a route for PHP-A1004 that stops, since Cypress alone would still flag it — a `babysit-hold` label would suppress the churn in the meantime. I have deliberately not added one, since it would also hide genuine new review feedback.

NathanaelJonesIreland:
Correction to my comment above: I said the DeepSource finding "is why `mergeStateStatus` is `BLOCKED`". That was wrong, and it matters for what you decide to do.

`reviewDecision` is `REVIEW_REQUIRED` — **the missing human review is what blocks this**, not the red status. A red `DeepSource: PHP` has not blocked merging on this repo in practice (#3243 went in with it failing). So clearing PHP-A1004 would tidy the check but would not by itself make this mergeable; it still needs a review either way.

Everything else in that comment stands: master is green on `DeepSource: PHP`, the finding is on this PR's own new `md5( $css )` line rather than pre-existing, it is a false positive on the merits, and the Cypress failure is the unrelated Stripe Connect duplicate ID.

NathanaelJonesIreland:
Babysit pass — **nothing pushed**, and no re-litigating the three routes above. One factual correction only.

### `master` is no longer green on `DeepSource: PHP`

My comment this morning stated `master` is green and used that to argue the finding is uniquely ours. That was true when measured; it is not true now:

```
master HEAD  452d5df12  2026-08-18T19:02:55Z  (Merge pull request #1379)
DeepSource: PHP  failure  updated_at 2026-08-18T19:05:31Z
```

`master` went red about seven hours after I measured it. I have not confirmed whether its failure is PHP-A1004 or a different rule — the per-occurrence detail is behind the DeepSource dashboard — so I am not claiming it is the same finding.

**What this does not change:** the occurrence on `classes/models/FrmStyle.php:456` is still this PR's own new line, still a false positive on the merits (PHP-A1004 is about password storage; `md5( $css )` is a cache-busting fingerprint, and the rule's suggested `password_hash()` is salted and non-deterministic, which would break the one property this PR exists to provide). The three routes — audit the occurrence, `// skipcq: PHP-A1004`, or `hash( 'crc32b', ... )` — stand as written, and the choice is still yours.

**What it does change:** the "this is ours and master is clean" framing is weaker than I put it this morning. A red `DeepSource: PHP` is now the state of `master` too, which is further evidence it is not functioning as a merge gate here.

### Otherwise unchanged

Head is still `506f5e59b` (2026-08-14); no new commits and no new review comments since. Cypress is the same unrelated Stripe Connect `no-dup-id` on `admin-html-validation.cy.js` — re-read from this run's log, not carried over. Both PHPUnit matrices, Psalm, PHPStan, Mago, Typos, DeepScan and DeepSource: JavaScript are green.

This still needs a human review (`REVIEW_REQUIRED`) before it can merge, independent of either red check.

NathanaelJonesIreland:
CI triage on the current head (`506f5e5`) — both red checks are pre-existing noise, not defects this PR introduced. Recording the evidence so nobody has to re-derive it. **No code pushed.**

**Cypress — repo-wide, unrelated to this diff.** One failure out of 51 tests: `admin-html-validation.cy.js`, `AssertionError: 1 error, 129 excluded: expected 1 to equal 0`. The error is `no-dup-id` → `Duplicate ID "frm_connect_with_oauth"` at `#frm_strp_settings_container > div:nth-child(2) > a` — the Stripe Connect settings view renders that ID twice. Nothing in this PR touches it (`FrmAppHelper`, `FrmCreateFile`, `FrmMigrate`, `FrmStyle` + their tests). Confirmed it isn't mine rather than assuming: Cypress is `fail` on **all 8 currently open PRs** in this repo (#3244, #3246, #3247, #3248, #3249, #3250, #3251 and this one). It needs fixing in the Stripe settings view, as its own change.

**DeepSource: PHP — diff-scoped analyzer noise.** `master`'s own `DeepSource: PHP` commit status is `success`, so the analyzer reports only new-in-diff findings; everything it raises here is a false positive on lines this PR happens to touch:

- `classes/models/FrmStyle.php:456` — "Use of insecure md5() function found". This is a **content hash for cache-busting**, not password or signature material, which is precisely the use the rule is aimed at. `md5()` is already the established idiom for this across the codebase — `FrmAddon:799`, `FrmApplicationApi:37`, `FrmFormApi:83`, `FrmStyleApi:37`, `FrmFormTemplateApi:42` all build cache keys with it, unsuppressed, on `master`.
- `tests/phpunit/database/test_FrmMigrate.php:320,335` and `tests/phpunit/misc/test_FrmCreateFile.php:23,49,50,75…` — "Call to an undefined method `assertFalse()`/`assertSame()`/`assertTrue()`", "Call to an undefined static method `FrmUnitTest::tearDown()`". DeepSource simply isn't resolving PHPUnit's `TestCase` inheritance through this repo's test base class. These are ordinary assertions; they only surface because this PR modified those two files.

I deliberately did **not** paper over these. A `// skipcq: PHP-A1004` on the `md5()` line would be defensible on its own (the repo already uses `skipcq` for `PHP-W1020`/`PHP-E1002`), but it would not turn the check green — the test-file findings dominate and cannot honestly be suppressed one assertion at a time. Making the check green here would mean either changing a cache-busting hash for a security rule that doesn't apply to it, or annotating ordinary assertions as exceptions. Both are worse than a red check with a written explanation, and neither is something to push to core without a human deciding it.

The real checks are green on this head: PHP 7.4 / PHP 8 tests on WP 6.9, PHPStan, Psalm, Mago, Typos, DeepScan, DeepSource: JavaScript.

NathanaelJonesIreland:
Babysit pass — **two commits pushed**, and unlike the previous eight passes this one is not a re-triage of the same two reds. Adding a missing label revealed that four gates had never actually run on this diff.

### What was actually wrong: `run analysis` was missing

This PR carried only the `run tests` label. The inspection workflows (`phpcs.yml`, `rector.yml`, `php-cs-fixer.yml`, `syntax.yml`, and the JS ones) are gated on:

```yaml
if: contains(github.event.pull_request.labels.*.name, 'run analysis')
```

so PHPCS, Rector, PHP-CS-Fixer, PHP Syntax, ESLint and Stylelint were all reporting **`skipping`** — which reads as harmless next to two red checks, but means they were *unmeasured*, not passing. Those workflows also trigger on `push: master`, but the `if:` reads `pull_request.labels`, which is empty for a push — so **they never run on master either**, and there is no baseline to compare against.

I added the label (the workflows trigger on `labeled`, so no push was needed) and two genuine failures in this diff appeared immediately:

**1. PHP-CS-Fixer — `phpdoc_separation`.** `update_css_version()`'s docblock needed a blank line between `@param` and `@return`. One line.

**2. PHPCS — 23 errors across the two new test files.** Both tests write to disk on purpose: they cannot exercise the *failed* CSS write without a real file blocking the path. That trips three VIP filesystem sniffs (`file_ops_unlink`, `file_ops_file_put_contents`, `directory_rmdir`), alongside alignment and line-length violations.

I excluded the three sniffs in `phpcs.xml` scoped to the two test filenames, following the precedent already in that file for `test_FrmCSVExportHelper.php`, rather than scattering inline `phpcs:ignore` comments. Worth noting the inline ignores already on those lines cite `WordPress.WP.AlternativeFunctions.*` — a *different, real* sniff — so they were never suppressing the VIP one. Everything else is formatting; **no test logic changed**, and the PHPUnit matrix confirms it.

### The second commit: Mago was version drift, not this diff

Pushing the fix turned `Mago` red. It was **not** caused by the change — every annotation pointed at test files the diff never touches (`test_FrmXMLHelper.php`, `FrmUnitTest.php`, `test_FrmFieldsAjax.php`).

Rather than guess, I re-ran the *old* Mago run against its original SHA `506f5e59b`: it **now fails on the commit it passed on six days ago**. That isolates drift from causation in one measurement.

Cause: this branch still had `"carthage-software/mago": "^1.40.1"`, a floating constraint that resolves to whatever 1.x is newest at run time. Master had already hit this and pinned it to `"1.46.0"` — a one-line `composer.json` change this branch never picked up. `mago.toml` and every flagged test file are byte-identical between this branch's base and master, so the inputs were unchanged and only the analyser had moved.

Merging master (clean, no conflicts) brought the pin and **Mago is green again**.

### Where CI stands on `f8d5e8d7`

`Mago`, `PHPCS`, `PHP-CS-Fixer`, `Rector`, `PHP Syntax` (7 and 8.4), `ESLint`, `Oxlint`, `Stylelint`, `PHPStan`, `Psalm`, `Typos`, `DeepScan`, `Scrutinizer` and `CodeRabbit` all pass — including the four gates that had never run on this diff before today. **both PHPUnit jobs (PHP 7.4 and PHP 8 on WP 6.9) pass** as well.

`Cypress` is now `skipped` (it needs `run e2e tests`, which this PR does not carry). Its earlier red was the repo-wide duplicate-ID failure in the Stripe settings view, tracked as #3258 and unrelated to this diff.

`DeepSource: PHP` will most likely stay red, and remains the known false positive: with `vendor/**` excluded it cannot resolve PHPUnit, so every `assert*()` on a changed or adjacent line reports "Call to an undefined method". `stubs.php` declares `WP_UnitTestCase extends WP_UnitTestCase_Base extends PHPUnit\Framework\TestCase`, and both `phpstan.neon` and `psalm.xml` load it explicitly — which is exactly why PHPStan and Psalm are green on the same files DeepSource calls broken. A suite cannot pass while calling undefined methods, and the PHPUnit matrix passes.

**This still needs a human review** — that, not the remaining red, is what `BLOCKED` reflects.

NathanaelJonesIreland:
Babysit pass — **nothing pushed**. One correction, and one line of accounting.

### Correction: the Cypress dup-ID issue number

My comment earlier today said the duplicate-ID failure is "tracked as #3258". That number is wrong, and in this repo it renders as a dead link. The tracking issue is **Strategy11/formidable-pro#6568** — *Duplicate element IDs in Stripe Connect settings view fail HTML validation* (open). Everything else in that comment stands.

### The two new DeepSource comments are the same false positive already accounted for

DeepSource flagged `test_FrmCreateFile::assertSame()` (line 158) and `test_FrmStyle::assertTrue()` (line 322) as calls to undefined methods. Both are the mechanism I described 11 minutes after they landed, so I am not re-litigating it — just recording that they were read and dismissed, not skipped:

- Both classes extend `FrmUnitTest`, which is `class FrmUnitTest extends WP_UnitTestCase` (`tests/phpunit/base/FrmUnitTest.php:3`).
- `stubs.php:537-540` declares `WP_UnitTestCase_Base extends PHPUnit\Framework\TestCase` and `WP_UnitTestCase extends WP_UnitTestCase_Base`. `phpstan.neon` and `psalm.xml` load that stub; DeepSource does not, so the chain to `TestCase` is unresolvable to it and every `assert*()` on a changed line reports as undefined.
- DeepSource reports diff-scoped, which is why only new lines are flagged while 81 test files on `master` use `$this->assert*()` untouched.
- `DeepSource: PHP` is **also red on `master` at `83b374e`**, so there is no green baseline this diff could have broken.
- Decisive: `PHP 7.4 tests in WP 6.9` and `PHP 8 tests in WP 6.9` both pass on this head. A suite cannot pass while fataling on undefined methods.

No `.deepsource.toml` change proposed — excluding `tests/**` from analysis is a repo-wide call, and out of scope for this diff.

**Still waiting on a human review**; that is what `BLOCKED` reflects, not the remaining red.

NathanaelJonesIreland:
Babysit pass. **Nothing pushed** — the diff has not changed, but two facts have since the last round, and one of them is a coverage loss rather than a red.

### Cypress is no longer failing here. It is no longer running.

The previous triage notes on this PR all recorded Cypress as red on the Stripe Connect duplicate-ID assertion (#3258). On the current head it reports `skipping`:

```
E2E Test   run 32382056907   head f8d5e8d72   conclusion: skipped
```

The cause is the master merge in `f8d5e8d72`, not anything in this diff. Master changed the workflow's gate in `3aaafd28a` ("Gate Cypress workflow on 'run e2e tests' label to match pro (#3254)"):

- before — `if: contains(github.event.pull_request.labels.*.name, 'run tests')`
- after  — `if: contains(github.event.pull_request.labels.*.name, 'run e2e tests')`

This PR carries `run tests` and `run analysis`, so PHPUnit — still gated on `run tests` — ran and passed on the same head while E2E skipped. That is not the dup-ID bug being fixed; it is no longer being measured here. Worth knowing it now applies to every open PR the moment it merges master: #3256 has not, which is why Cypress still runs there and still fails on byte-identical output.

Adding `run e2e tests` would restore the coverage at the cost of a red that nobody on this PR can clear, so I have left that call to a human rather than labelling it myself.

### DeepSource: PHP — same false-positive class, new positions, and master is red too

Two new inline comments landed on `f8d5e8d72`, both the unresolvable-PHPUnit class this repo already lives with:

- `tests/phpunit/misc/test_FrmCreateFile.php:158` — "Call to an undefined method test_FrmCreateFile::assertSame()"
- `tests/phpunit/styles/test_FrmStyle.php:322` — "Call to an undefined method test_FrmStyle::assertTrue()"

`stubs.php` declares `WP_UnitTestCase extends WP_UnitTestCase_Base extends PHPUnit\Framework\TestCase`, and both `phpstan.neon` and `psalm.xml` load it explicitly — which is exactly why PHPStan and Psalm are green on the same lines. DeepSource has no stub-path setting and excludes `vendor/**`, so every `assert*()` in every phpunit file is unresolvable to it. A suite cannot pass while calling undefined methods, and PHP 7.4/WP 6.9 and PHP 8/WP 6.9 both pass on this head.

Re-measured just now, because a stale reading of this has been quoted on this PR before: `DeepSource: PHP` is **failing on master itself** at `83b374e06` (2026-08-19T20:48:19Z). So it is not evidence about this branch in either direction.

Everything else on `f8d5e8d72` is green: PHPUnit ×2, PHPCS, PHP-CS-Fixer, PHPStan, Psalm, Mago, Rector, PHP syntax ×2, ESLint, Oxlint, Stylelint, Typos, DeepScan (0 new), Scrutinizer (no new issues), CodeRabbit.

`reviewDecision` is `REVIEW_REQUIRED` — a human review is the only thing actually outstanding here.

NathanaelJonesIreland:
Pushed `a8ed9aef1` for the red `DeepSource: PHP` check.

Its 5 findings split two ways:

- **1 in product code** — `PHP-A1004` "Use of insecure md5() function" on the new `update_css_version()` line. False positive: that value is a 12-character content-derived cache-busting token stored in `frm_last_style_update`, not a password. The check's suggested remedy is `password_hash()`, which is salted and non-deterministic — it would defeat the entire mechanism, since the point is that identical stylesheet bytes must reproduce an identical version string.

  Suppressed at the line with a stated reason rather than repo-wide, following the `skipcq:` convention already used in `classes/views/frm-entries/show.php`, `classes/views/frm-fields/back-end/field-options.php` and elsewhere — and whitelisted by this repo's own `CommentSpacingSniff`. Worth noting `classes/models/` already holds seven `md5()`-for-cache-key calls of exactly this kind (`FrmAddon`, `FrmAntiSpam`, `FrmApplicationApi`, `FrmFormApi`, `FrmFormTemplateApi`, `FrmStyleApi`, `FrmUsage`), none annotated. DeepSource only reports on lines a diff touches, which is why this one surfaced and those didn't. If the preference is to disable that check for PHP in `.deepsource.toml` instead, given how consistently this codebase uses md5 for cache keys, that's a repo-wide config call I've deliberately left alone.

- **4 in `tests/phpunit/`** — `test_FrmMigrate.php` (×2) and `test_FrmCreateFile.php` (×2). Not touched here: #3269 exists specifically to exclude `tests/phpunit` from DeepSource PHP analysis, and duplicating that change would just conflict. So this check stays red until #3269 lands, and `.deepsource.toml`'s existing `test_patterns` entry is not enough on its own — it classifies those files as tests without excluding them from analysis, which is the gap #3269 closes.

Everything else is green: PHPCS, PHPStan, Psalm, Rector, CS-Fixer, Mago, Scrutinizer, ESLint, Stylelint, Oxlint, DeepScan, DeepSource: JavaScript, and PHP 7.4 / PHP 8 tests on WP 6.9.

Verified locally before pushing: `php -l` clean; the custom `Formidable.Commenting.CommentSpacing` sniff passes on the file; `php-cs-fixer --dry-run --allow-risky=yes` exits 0 against the LF form CI analyses. (It reports exit 8 against the working-tree copy purely because that copy is CRLF under `autocrlf`; confirmed by running it on the same content with LF endings, and unrelated to this change.)

Still needs a human review — no non-bot review on it yet.

NathanaelJonesIreland:
Operator pass. **Nothing pushed** — this is a `core` repo, so the fix is a human's to make. Recording the triage of Franky's two findings, because they arrived *after* the last push (`a8ed9aef1`, 08-21 20:34) and nothing on this PR has answered them in the 106h since.

### 1. `FrmStyle.php` — the `$file_written` gate is a real regression. Accepted.

Franky is right, and the finding is in direct tension with the reason the gate exists. The gate was added so a *silent write failure* cannot leave a stale static file being served under an advanced version — otherwise the staleness is permanent. But on a host where `FrmCreateFile::has_permission` is permanently `false` (no FTP/SSH credentials, set on construction and never recovered for that request), `create_file()` never succeeds, so `update_css_version()` never runs, and `get_css_version()` falls back to the plugin version forever.

The important half is that those hosts are *not* serving a static file at all — `get_url_to_custom_style()` has already fallen through to `admin-ajax.php?action=frmpro_css`, and that endpoint serves the `frmpro_css` option, which `save_settings()` updates unconditionally a few lines above the gate. So the served CSS changes on every save while the `?ver=` never advances. That is a new break for that host class, not a pre-existing gap: before this PR the bump was unconditional.

Franky's suggested shape is the right one and I would take it as written: skip the bump only when `create_file()` fails **and** a readable file already exists at the target path — i.e. gate on "could a stale file be served" rather than "did this write succeed". No permission at all is not that case.

### 2. `test_FrmStyle.php` — `rmdir_recursive( $blocked_path )` on the real uploads dir. Accepted.

Also right. `wp-content/uploads/formidable/` is `FrmProFileField::default_formidable_uploads_dir()`'s default, so it holds genuine Pro file-upload submissions, not only generated cache data as the comment claims. Disposable CI is safe; a local or staging run sharing that path is not. The sibling pattern in `test_FrmCreateFile.php` (`'frm-test-…' . wp_generate_password(…)`) is the fix.

### Still red: `DeepSource: PHP`

Unchanged from the 08-21 note — the one product-code finding is `PHP-A1004` on `md5()`, a false positive: the value is a 12-character content-derived cache-busting token in `frm_last_style_update`, not a credential, and the suggested `password_hash()` remedy is salted and therefore cannot produce a stable token from stable content.

**Needs a human.** Finding 1 changes the guard's condition, which is the whole subject of this PR, so it is not a change to make on a core product without a person in the path.

NathanaelJonesIreland:
@truongwp handing this over to you — I'm not going to push further changes on it, so
it's yours to take in whatever direction you think is right. Summary of where it stands
so you don't have to reconstruct it.

**What the PR does:** derives the generated stylesheet's cache-busting version from the
CSS content instead of `gmdate( 'njGi' )`, which had minute resolution, no year
component, and so could collide (and could fail to advance across a real change).

**Blocking: two findings from the review of 2026-08-21, neither addressed.**

1. **`classes/models/FrmStyle.php:424` — functional regression, the real blocker.**
   The version bump is gated on `$file_written`. On hosts where
   `FrmCreateFile::has_permission` is permanently `false` (WP_Filesystem wanting
   FTP/SSH credentials that aren't configured), `create_file()` always returns `false`,
   so `update_css_version()` never runs and the version pins to the plugin version.
   Those hosts fall back to `admin-ajax.php?action=frmpro_css`, which serves the
   `frmpro_css` option that `save_settings()` updates *unconditionally* — so the served
   CSS changes on every save while the `?ver=` never advances. That's stale CSS
   indefinitely behind any CDN or browser cache. It's a regression, not a pre-existing
   gap: before this PR the version bump was unconditional.

   The gate is right for the case it was written for (a stale *existing* file served
   after an intermittent write failure). The suggested shape is to gate on "could a
   stale file be served whose content doesn't match `$css`" rather than "did this write
   succeed" — i.e. skip the bump only when `create_file()` fails **and** a readable file
   already exists at the target path, not when there's no write permission at all.

2. **`tests/phpunit/styles/test_FrmStyle.php:315` — test safety.**
   `rmdir_recursive( $blocked_path )` recursively deletes `wp-content/uploads/formidable/`
   before replacing it with a blocking file. That path is also
   `FrmProFileField::default_formidable_uploads_dir()`, i.e. where real Pro file-upload
   submissions live. Fine in this repo's disposable CI, destructive on a local or
   staging run that shares the uploads path. `test_FrmCreateFile.php` already does this
   safely with a randomised `'frm-test-' . wp_generate_password(...)` subfolder.

**On the red `DeepSource: PHP` check — I read it as noise, but please confirm.**
Nearly all of it is *"Call to an undefined method test_FrmCreateFile::assertTrue()"*
and siblings: DeepSource isn't resolving PHPUnit's assertion methods through
`FrmUnitTest` → `WP_UnitTestCase`. The one item that isn't purely analyser confusion is
*"Result of method `FrmCreateFile::append_file()` (void) is used"*, where the tests pass
a void call straight into `assertNull()`. Harmless at runtime, but a fair nit if you
want it cleaned up.

Nothing has been pushed since the review, and the branch is `MERGEABLE` against master.
Happy to answer questions on the diagnosis; I just won't be the one changing it.

PR #3247, opened 2026-08-17
Add hover states to admin inputs, checkboxes and field cards

tuguirazvan:
Most admin controls only changed on focus, so nothing confirmed they were interactive before being clicked. The ones that did have a hover used `--primary-25`, which made pointing at something read the same as selecting it.

In this update,
1. Text inputs, textareas and selects warm their stroke and lift their shadow on hover, staying neutral so it never reads as focus
2. Checkboxes and radios get the same treatment at a stronger mix, since a 16px box has roughly a tenth of an input's perimeter. Hovering the label lights the box
3. Their hit area grows from 16x16 to 16x40, vertically only so it cannot swallow clicks meant for the control beside it
4. Canvas field cards get a hover distinct from selected, which previously rendered almost identically
5. Icon buttons share one chip treatment with hover, press and focus: builder field card actions, both style options triggers, the form action header (duplicate, delete, collapse) and the forms list settings gear
6. The forms list settings panel matches the action dropdowns at a 10px radius with 6px padding and 6px rows, and its checkbox rows pick up the same 10px icon gap and `4px 8px` padding
7. The gateway and captcha selectors fill with `--surface-hover` on hover instead of turning the label blue, leaving the border alone so the blue stroke stays the sole marker of the selected card
8. Hover styling that was bound to `:focus` now uses `:focus-visible`, so clicking a control no longer leaves it painted while the pointer lights up another
9. Fixes a missing comma in `_inputs.scss` that turned two selectors into one descendant selector, leaving email and search inputs across the admin with none of the base input styling

Three things worth a reviewer's attention:

**The style options trigger was only half covered.** The rule was scoped to triggers inside a style card, so the one beside the style name on the edit page kept an 18x17 target with no hover at all. It is now keyed to the wrapper `getHamburgerMenu()` builds, which covers both and nothing else.

**Style cards paint in the style's own colours**, where an opaque grey chip reads as a white square. Those cards use a translucent wash instead, with the polarity flipped for `.frm-dark-style`, and the glyph on dark cards stays white rather than darkening.

**The forms list Apply button** was rendering at 12px in a 67x28 box, smaller than the labels above it, because a lower specificity rule was overriding the `.frm-button-primary` defaults. It now uses the component's own padding and font size.

Hover is neutral grey rather than the blue in the Figma: blue already means selected in the builder, so a blue hover left the two states differing only by intensity. Two other deviations from the file: the ID gap widened to 12px so the icons read as a pair, and the ID stays at 14px on the right. Worth a conversation with whoever owns the Figma.

Touches `images/icons.svg`: the move symbol hardcoded `stroke="#9EA9B8"` instead of `currentColor`, so it ignored every colour rule. CSS cannot reach inside a `<use>` shadow tree, so it had to be fixed in the sprite.

The collapse button in the form action header keeps its own transform transition rather than taking the shared press scale, because a scale would overwrite the `rotate(-180deg)` it uses for the open state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed search and email field styling for more consistent form controls.

* **Style**
  * Improved hover, selected, focus, and layout states for form-builder cards.
  * Added clearer feedback and larger hit areas for checkboxes and radio buttons.
  * Refined borders, shadows, transitions, and active states across form controls.
  * Standardized hover and active surface colors throughout the interface.
  * Improved dropdowns, action buttons, form lists, and settings controls with consistent sizing and feedback.
  * Added accessible focus states and reduced-motion support for interactive elements.


Crabcyborg:
@tuguirazvan This update seems to cause some issues.

Field inputs in the builder preview appear far too large now in some cases.

For example, if I have 12 fields in a row:

**Before**
Screenshot 2026-08-26 at 1 17 36 PM

**After**
Now they overflow off the page.
Screenshot 2026-08-26 at 1 18 33 PM

Before, if I had 4 fields in a row on my laptop screen, I could see the field IDs on hover.

**Before**
Screenshot 2026-08-26 at 1 23 16 PM

**After**
Now they're only partially visible and they fall to a new line. It also looks like the move icon matched the other icons better before.

Screenshot 2026-08-26 at 1 24 10 PM

tuguirazvan:
Thanks @Crabcyborg — one fixed, one I need your help with.

**Label/ID wrapping — fixed in 649a15e.** Worse than partial: measured 10 of 10 cards wrapping, full-width ones included. `.frm_primary_label` is a block element, so `flex: 1 1 auto` resolved its basis to the whole row and left no room for the icons. Basis `0` fixes it — 0 of 10 now. Happy to match the move icon back to its neighbours too if you'd prefer.

**Oversized inputs — couldn't reproduce.** Zero overflow at 4- and 6-across across 1440/1180/1000/860 widths, and my `box-sizing` theory measured false. `frm_sixth` is the narrowest preset so I can't build your 12-column row — could you share that form as an XML export?

One flag: the first hunk in `_inputs.scss` is a missing comma that had been denying base styling to `email`/`search` inputs. If that's behind the sizing, I'd rather not reshape that shared selector list here — it covers much of the admin. Own PR if it needs doing.

Crabcyborg:
@tuguirazvan You can add 12 fields in a row in a few ways. There just isn't a preset for our multi-select flow.

1. You can just drag/drop them all them. It goes up to 12.

https://github.com/user-attachments/assets/b9d5c3c3-88d1-459b-adfc-fb8b9489dbcb

2. They will automatically merge right now if you just had 12 fields in a row, all with `frm1` and none with `frm_first` to break up the rows. It won't auto-group them until you've saved and reloaded, but it will.

https://github.com/user-attachments/assets/68ad29cf-bc81-4826-8ed0-1ef39cecf380

PR #3248, opened 2026-08-17
Add icon button hover states and retire five deprecated colour aliases

tuguirazvan:
The shortcode box, the email chips and the full screen close button each had a focus state and nothing for hover.

Pairs with Strategy11/formidable-pro#6561 for the alias sweep on the Pro side.

In this update,
1. The ⋯ shortcode box and the CC / BCC / Reply To chips tint on hover and give slightly on press
2. The shortcode box swapped its glyph in a single frame; it now blooms in. A `MutationObserver` covers every call site — note `attributeFilter` cannot be used, as it misses the `xlink:href` writes `hideShortcodes()` makes
3. The close button is sized from the icon button component in the Formidable System file (40×40, `--small-radius`). Its anchor had no dimensions before, so the target was only as big as the 20px glyph
4. Adds `--primary-100`, derived to sit between 50 and 200 at the spacing the grey ramp uses for its own 100
5. Retires `--dark-grey`, `--grey`, `--grey-border`, `--primary-hover` and `--light-blue` in favour of the tokens they alias. `--primary-color` is deliberately left alone: Pro sets it inline as an override hook, so renaming its readers would silently break that and any customer CSS doing the same

---

**On the red DeepSource check:** it reports four issues in `admin.js` (lines 7235, 7636, 7640, 8049). All four are pre-existing and unmodified — this PR adds three lines to that file, shifting them down by one, so DeepSource counts them as newly introduced. Left as-is rather than touching unrelated code.

---

**Merge order:** none required. This stops core *using* the deprecated aliases but keeps all six defined in `_variables.scss`, so Formidable Pro and any add-on reading them are unaffected whichever way round these land. Pairs with Strategy11/formidable-pro#6561, which does the same sweep on the Pro side.

**Follow-up, deliberately not done here:** removing the definitions. That is the breaking change, and it can only happen once nothing references them — core is clean after this, Pro after #6561, but third-party add-ons and customer CSS are unknown. Worth its own PR in a release where a breaking-change note fits.

## Summary by CodeRabbit

- **New Features**
  - Added animated icon feedback for shortcode and inline-modal controls.
  - Added hover and pressed states to notification settings and shortcode controls.
  - Added a refined primary color option to the admin theme.

- **Style**
  - Improved the full-screen close control with centered sizing, hover feedback, and reduced-motion support.
  - Updated admin backgrounds, borders, labels, icons, and text colors for more consistent contrast.
  - Improved template, code-list, form-field, and notification styling.


Crabcyborg:
@tuguirazvan You never requested a review for this PR.

Is this complete, or still in progress?

tuguirazvan:
@Crabcyborg Yes, this is ready. Sorry, I thought I've requested a review from you. Thank you!

garretlaxton (approved):
This looks good!

Crabcyborg (approved):
Thank you Razvan and Garret!

🚀

PR #3251, opened 2026-08-18
Refine admin dropdown menus

tuguirazvan:
The Preview dropdown and the style options menu shared the blue hover used by the field and shortcode pickers, so hovering read the same as selecting. This makes those two action menus hover neutral and tidies their sizing.

Scope note: only the action menus change. Pickers, multiselect lists and the form switcher keep the existing blue hover.

In this update,
1. Items hover on `--grey-100` with `--grey-900` text instead of `--primary-25` / `--primary-500`.
2. Panel radius 4px -> 10px with 6px items inside it, and 6px panel padding.
3. Chip padding `4px 8px`, which reads as an even 8px on all four sides — the 24px line-height on 16px text adds ~4.4px of half-leading that equal padding values cannot account for.
4. Icons 18px -> 16px with a 10px gap, and `flex-shrink: 0` so they stop collapsing on a tight row.
5. Menus size to their contents. Three `min-width` overrides (230px `!important`, 205px, 200px) are removed, and `width: max-content` replaces `auto` so a menu near a container edge is not shrunk and then cropped by `overflow-x: hidden`.
6. The Preview chevron rotates 180deg while open, reusing the form action collapse timing.
7. Items get a keyboard focus ring.

One thing worth flagging for review: the shared rule in `_code-list.scss` carries `:link` to override jQuery UI, which lifts it to the same specificity as this block. It is imported later, so it was silently winning and the style options menu never received the new padding. Those pseudo-classes are repeated here to break the tie.

## Summary by CodeRabbit

* **Style**
  * Refined dropdown menus with content-sized widths, rounded corners, improved spacing, and clearer hover and focus states.
  * Improved menu icons, toggle rotation, transitions, and reduced-motion behavior.
  * Updated testing-mode and admin styling for consistent dropdown presentation.
  * Removed outdated fixed-width rules from preview and style-option menus.


garretlaxton (approved):
This looks good!

Crabcyborg (approved):
Thanks @tuguirazvan and @garretlaxton!

🚀

PR #3256, opened 2026-08-19
Accept internationalized domain names in Website/URL fields

NathanaelJonesIreland:
## What

The Website/URL field rejects valid internationalized domain names (IDNs) whose hostname contains
accented characters. Reported by an Elite customer in Help Scout ticket 257002 for a `.ch` domain
spelled with an umlaut. `.ch` explicitly permits accented vowels, so the domain is legitimately
registrable.

The cause is a single host-matching regex, duplicated in PHP and JS. Its host character class
`[\da-z\.-]` is ASCII-only:

- `classes/models/fields/FrmFieldUrl.php` in `FrmFieldUrl::validate()`
- `js/formidable.js` in `checkUrlField()`
- `js/formidable.min.js` (the literal survives verbatim, since `npm run minimize` compiles at
  `WHITESPACE` level)

Formidable Pro has zero copies of this pattern.

## Why this is safe rather than a loosening

Three things make this a consistency fix rather than a new capability:

1. **The punycode spelling of the same domain already passes.** `xn--`-encoded hosts have always
   validated, and they resolve to the identical host. Accepting the Unicode spelling adds nothing
   a submitter could not already do.
2. **Non-ASCII in the path, query and fragment already passes.** The regex only constrains the
   host, so the field already stored accented characters happily. Only the hostname was rejected.
3. **`esc_url_raw()` already preserves the bytes.** WordPress's URL sanitizer explicitly allows the
   high byte range, so sanitizing was never the thing rejecting these values.

Sanitizing, escaping, storage and the scheme allowlist are all untouched. Chrome's native
`type="url"` constraint already accepts every IDN form here, so the browser was not the blocker
either.

## Verification

- **Exhaustive ASCII equivalence.** The old and new patterns were compared across 1408 ASCII
  byte/position combinations and are byte-for-byte identical. No previously accepted or rejected
  ASCII url changes behaviour. This was brute-forced, not sampled.
- **Injection set unchanged.** `javascript:`, `data:`, `vbscript:`, `<script>` and CRLF-header
  inputs all still fail exactly as before. The only behavioural delta across the whole matrix is
  the IDN cases.
- **Pattern read back from the edited file** (rather than a copy) accepts 10 IDN forms, including
  Cyrillic and CJK, and rejects 8 malformed hosts.

## The trap worth knowing about in review

The two regexes deliberately do **not** use the same character range:

- PHP matches UTF-8 **bytes** (no `/u` modifier), so the raw byte range covers all non-ASCII.
- JS matches UTF-16 **code units**, so it needs `\u0080-\uFFFF`. The PHP byte range would cover an
  umlaut but **not** Cyrillic or CJK.

Copying one literal into both would accept those hosts server side while silently rejecting them in
the browser, which is the same class of split this ticket is about. `test_url_field_js_regex_parity()`
asserts the two JS files carry the code unit form and never the PHP one.

`/u` is also deliberately absent from the PHP pattern: `preg_match()` returns `false` on invalid
UTF-8, and because the result is negated that would report valid Latin-1 input as invalid.
`test_url_non_utf8_host_byte()` guards this, with a precondition so it cannot pass vacuously.

One more subtlety, confirmed empirically: appending the range *after* the trailing hyphen
(`[\da-z\.-\x80-\xff]`) is **not** a syntax error. PCRE reads it as the range `0x2E-0x80`, which
silently admits `/ ? : @ < > [` inside the hostname. The negative rows `https://a/b.com` and
`https://a?b.com` exist to fail if anyone ever reorders the class that way.

## Tests

- `test_url_idn_validation()` - 11 must-pass values (accented Latin, Cyrillic, CJK, uppercase
  non-ASCII, punycode, ASCII baseline, `localhost`, non-ASCII in path/query/fragment, and the
  no-scheme form) and 4 must-fail values.
- `test_url_non_utf8_host_byte()` - the `/u` guard.
- `test_url_field_js_regex_parity()` - asserts the rule rather than pinning the literal, and checks
  the minified artifact carries the same host class as its source. Mutation-proven: green on the
  real files, red when `formidable.min.js` is reverted alone, red when the PHP range is copied into
  the JS source.
- 3 rows appended to the shared `expected_format_errors()` table.
- One Cypress `it()` with "Validate this form with javascript" enabled, so it exercises the
  committed minified artifact through a real browser rather than passing via the server-side path.

`php -l` and `node --check` are clean. phpcs, phpstan, phpunit and eslint need `composer install`
plus a WordPress test database, which is not set up on this machine, so they run here in CI under
the labels below. Reviewer note: core CI is label-gated, so this PR carries `run tests`,
`run analysis` and `run e2e tests`. Without them only `typos` and `psalm` would run.

## Manual QA note

The front end serves the combined `js/frm.min.js`, which is rebuilt from `js/formidable.min.js`
only by `FrmAppHelper::save_combined_js()` (called from `FrmMigrate` and `FrmAddon`). When testing
this by hand, delete `js/frm.min.js` or set `SCRIPT_DEBUG`, otherwise a stale combined file will
serve the old regex and the fix will look like it did nothing.

## Deliberately not fixed here

These are pre-existing and adjacent, left out to keep the change narrow. Each is a separate
behaviour decision, and they are grouped into a follow-up issue rather than lost:

- Underscore in the host is rejected.
- IPv6 literals are rejected.
- `ftp:`, `mailto:`, `news:`, `feed:` and `telnet:` are rejected by the format check even though the
  scheme allowlist a few lines above explicitly admits them - a real internal contradiction.
- Garbage such as `https://-.-` and `https://....` passes, so this check was never a strong quality
  gate in the first place.

Help Scout ticket 257002.

## Summary by CodeRabbit

- **Bug Fixes**
  - URL fields now accept valid internationalized domain names, including accented or other non-ASCII characters.
  - Existing HTTP, HTTPS, and localhost validation rules remain supported.
  - Validation continues to reject malformed domains, dotless hostnames, invalid paths, and unsupported URL formats.
  - Browser and server-side validation now handle internationalized URLs consistently.

- **Tests**
  - Added coverage for internationalized domains, malformed URL inputs, non-UTF-8 host data, and validation consistency.


NathanaelJonesIreland:
### CI status: 18 green, 2 red, and neither red one comes from this PR

**Green, including the parts that matter here:** PHPUnit on PHP 7.4/WP 6.9 and PHP 8/WP 6.9, PHPCS,
PHPStan, Psalm, Mago, ESLint, Oxlint, PHP CS Fixer, Rector, Stylelint, Spell Check, DeepSource:
JavaScript, DeepScan, Scrutinizer, CodeRabbit.

The new Cypress case passed:

```
✓ should accept an internationalized domain name in a Website/URL field (8398ms)
Spec Ran: Forms/fieldsInFormBuilder.cy.js  (green)
```

That is the end-to-end proof that the committed `js/formidable.min.js` was actually rebuilt, since
wp-env activation regenerates `js/frm.min.js` from it.

---

**Red 1 — Cypress `admin-html-validation.cy.js`:** `no-dup-id`, `Duplicate ID
"frm_connect_with_oauth"` at `#frm_strp_settings_container > div:nth-child(2) > a`, on the global
settings page.

That ID lives in `stripe/views/settings/connect.php:30` and `stripe/js/connect_settings.js:5`, both
untouched here — this PR changes five files, none of them under `stripe/`.

Worth flagging separately: **Cypress has never actually run on master.** Every historical run shows
`skipped`, because the gate added in Strategy11/formidable-forms#3254 reads
`github.event.pull_request.labels`, which is `null` on `push` events, and the `run e2e tests` label
did not exist in this repo at all until I created it to label this PR. So there is no green baseline
to compare against, and this appears to be the first Cypress execution on any PR since the gate
landed. The duplicate ID looks like a genuine pre-existing HTML-validity and accessibility bug in
the Stripe Connect settings view (the anchor is rendered in two sibling containers) rather than
anything to do with URL fields. Happy to open a separate issue for it if useful — flagging rather
than silently folding it into this PR.

**Red 2 — DeepSource: PHP.** All of the reported findings are pre-existing lines that DeepSource
cannot resolve, not defects introduced here. Checked individually against `origin/master`:

- `Call to an undefined method assertEmpty()/assertNotEmpty()/assertArrayHasKey()` at lines 58, 76,
  78, 183, 198 — every one of those lines is unchanged master code. DeepSource is not resolving
  PHPUnit's `TestCase` base class.
- `Access to an undefined property $factory` at lines 16, 20, 145, 155, 170 — also unchanged master
  code; `$factory` comes from the WP test-case base class.
- `Cannot access property $required on array|int|object` at `FrmFieldUrl.php:89` — this is the
  pre-existing `} elseif ( $this->field->required == '1' ...` line, byte-identical to master's line
  88. It only moved down one line because the fix adds a comment above it.

DeepSource: JavaScript went green after this PR's second commit, which addressed the two findings
that genuinely were mine (`JS-0117` wanted the `u` flag; `JS-R1004` flagged four backtick strings
with no interpolation).

NathanaelJonesIreland:
@coderabbitai review

NathanaelJonesIreland:
Housekeeping so the review queue reflects reality.

**Resolved 5 DeepSource threads** — all raised against the first commit (`cf3f6c8`) and all genuinely
fixed in `df88b2d`. GitHub already marked each of them `outdated`, and `DeepSource: JavaScript` is
now **success** on the head commit. For the record, rather than just taking the check's word for it:

| Thread | Finding | Current state |
|---|---|---|
| `js/formidable.js:546` | JS-0117, use the `u` flag | Pattern is now `…/iu` with a code point range |
| `fieldsInFormBuilder.cy.js:258` | JS-R1004, needless template string | now `'li[id="text"] a[title="Text"]'` |
| `fieldsInFormBuilder.cy.js:259` | JS-R1004 | now `'li[id="url"] a[title="Website/URL"]'` |
| `fieldsInFormBuilder.cy.js:280` | JS-R1004 | now `'[id^="frm_error_field_"]'` |
| `fieldsInFormBuilder.cy.js:290` | JS-R1004 | now `'[id^="frm_error_field_"]'` |

Worth noting on JS-0117 specifically: adding `u` was not purely cosmetic. Under the `u` flag the
previous code *unit* range would have stopped matching astral-plane characters that the PHP side
accepts as bytes, which would have introduced a fresh server-accepts / client-rejects split — the
exact bug class this PR fixes. So the range was widened to a code *point* range at the same time,
and verified in node as identical to the previous behaviour across 640 generated ASCII cases with
astral hosts still matching.

**Left the 16 DeepSource PHP threads open deliberately.** They are false positives, but they are not
mine to dismiss and a reviewer should see them. The decisive evidence is that `PHP 7.4 tests in
WP 6.9` and `PHP 8 tests in WP 6.9` both pass: a suite cannot pass while calling undefined methods,
so `Call to an undefined method assertNotEmpty()` cannot be true. DeepSource is not resolving
PHPUnit's `TestCase`, which is also why it reports the same error on line 198 — untouched `master`
code. Similarly `FrmFieldUrl.php:89` is master's line 88 verbatim, moved down one line by the added
comment.

If the team wants those silenced repo-wide, excluding `tests/**` from the PHP analyzer in
`.deepsource.toml` would do it, but that is a policy call rather than something to fold into this PR.

**Filed Strategy11/formidable-pro#6568** for the pre-existing duplicate-ID failure in the Stripe Connect settings view that
this PR's Cypress run surfaced. It includes the root cause (the view is rendered once per mode with
hardcoded IDs) and a note that current behaviour is *not* broken, since the click handlers use
jQuery delegated events and derive mode from the `[data-test-mode]` ancestor.

NathanaelJonesIreland:
Babysit pass — **nothing pushed**, and no re-litigating the two triage comments above. One addition
only, because it answers the question a reviewer actually has to decide: *can this merge with
`DeepSource: PHP` red?*

**The repo has already answered that twice, this month.** Both of these merged on 2026-08-13 with
`DeepSource: PHP` failing and inline DeepSource threads open on the phpunit files they added:

| PR | DeepSource: PHP | Open phpunit threads at merge |
|---|---|---|
| [#3234](https://github.com/Strategy11/formidable-forms/pull/3234) — Import nulls without fatal errors | fail | 6, on `tests/phpunit/xml/test_FrmXMLHelper.php` |
| [#3235](https://github.com/Strategy11/formidable-forms/pull/3235) — Make field create handle invalid input better | fail | 9, on `tests/phpunit/fields/test_FrmField.php` |

Same finding class in both — `Call to an undefined method assertX()` and
`Access to an undefined property $factory`. So this is not a new judgement call about this PR; it is
what happens to every core PR that adds a phpunit test.

I also found the mechanical cause, which makes it unfixable from inside this diff rather than merely
a false positive. `.deepsource.toml` excludes `**/vendor/**`, so `PHPUnit\Framework\TestCase` is
invisible to the PHP analyzer. `stubs.php` does declare
`WP_UnitTestCase extends WP_UnitTestCase_Base extends PHPUnit\Framework\TestCase`, and `phpstan.neon`
and `psalm.xml` both load that file explicitly — which is exactly why PHPStan and Psalm are green
here while DeepSource is not. DeepSource has no equivalent stub setting, so every `assert*()` call in
all 78 phpunit test files is unresolvable to it. Nothing this PR can change; the fix is either a
stub path or a `tests/**` exclusion in `.deepsource.toml`, and that is a repo policy call.

Cypress is unchanged from the triage above: the sole failure is the pre-existing duplicate
`frm_connect_with_oauth` ID, now tracked as Strategy11/formidable-pro#6568. I re-read the log on the current head to confirm
it is byte-for-byte the same failure Strategy11/formidable-forms#3242 hits on an unrelated diff, which is the cleanest proof
available that it is master-side and not either PR's doing.

Still waiting on a human review — the only reviews on record are `deepsource-io[bot]` and
`coderabbitai[bot]`, so this is not merging itself.

NathanaelJonesIreland:
Thanks — pushed `9b6bfa4df` for the `@covers` nit, and I chased the red checks further than the review did, because one of the DeepSource findings isn't accounted for by the "PHPUnit inherited assertions" explanation.

**The `@covers` nit — fixed here rather than in a follow-up**, since the PR is still open and it's a docblock change. Removed rather than swapped for `@coversNothing`: this suite uses `@coversNothing` nowhere and sets neither `forceCoversAnnotation` nor `requireCoverageMetadata`, so an absent `@covers` is the existing convention for a test that isn't a unit test of one method. Left a note saying it deliberately carries none and pointing at `test_url_value()` as the test that does exercise `validate()` — verified that test exists and carries that annotation rather than assuming it.

**`DeepSource: PHP` — 20 findings, and they are not all the test-file false positive.** Breakdown by file:

| file | count | status |
|---|---|---|
| `tests/phpunit/fields/test_FrmFieldValidate.php` | 15 | the false positive you identified — unresolved inherited `assert*()` / `$factory` |
| `tests/cypress/e2e/Forms/fieldsInFormBuilder.cy.js` | 4 | test file; `DeepSource: JavaScript` passes, so not blocking |
| `classes/models/fields/FrmFieldUrl.php` | 1 | **product code, not covered by that explanation** |

That last one is `Cannot access property $required on array|int|object` at what is now line 89 — `$this->field->required == '1'`. I left it alone deliberately, and here's why rather than just asserting it:

- **It is not a line this PR changes.** The diff hunk is `@@ -85,2 +85,3 @@` — the comment and the `preg_match`. Line 89 sits outside the hunk and moved down one line because the new two-line comment pushed it. DeepSource analyses the whole changed file, so it surfaced pre-existing code.
- **The complaint is legitimate, not a false positive.** `FrmFieldType::$field` is declared `@var array|int|object` (since 3.0), so any `->property` access on it is genuinely unsound to a static analyser.
- **It's systemic, not local.** There are 15 `$this->field->…` accesses across `classes/models/fields/*.php`. Narrowing that type, or the accesses, is a core-wide change and has no business riding along in an IDN validation PR.

So `DeepSource: PHP` will stay red on this branch even after #3269 excludes `tests/phpunit`, for that one pre-existing finding. Filing the systemic `$field` typing gap separately rather than leaving it as a comment here — it will keep re-firing on every PR that touches a field class, so it's worth its own record.

**`Cypress`** is the pre-existing duplicate `frm_connect_with_oauth` id in `stripe/views/settings/connect.php`, tripping `admin-html-validation.cy.js` — nothing this branch touches, as your review noted. This PR's own spec passed 5/5.

Still needs a human review — the only approval on it is Franky's.

NathanaelJonesIreland:
Re-requesting a Franky pass: the last approving review landed at 20:04 UTC and the current head was pushed at 20:36, so that approval does not cover what is now on the branch — and the `franky-review` label had come off, so nothing was going to look at it. Re-added.

Nothing pushed here. Cypress and DeepSource: PHP are both red on this head; those are for a human to read rather than for me to touch on a core repo.

truongwp:
@Crabcyborg Should we add a checkbox in the field settings to enable this feature instead of doing that directly?

truongwp (approved):
Thanks @NathanaelJonesIreland and Franky. I think it's safe to make this feature without a checkbox to enable it.

I added a Vietnamese URL to the test and solved some DeepSource errors.

The code looks good for me now.

garretlaxton (approved):
This looks good!

PR #3260, opened 2026-08-19
Open dropdown menus under their trigger

tuguirazvan:
Style card menus opened to the left of the stacked dots, and sometimes flipped upward with room to spare below.

The cards sit inside several `overflow: hidden` ancestors. Popper treats those as clipping parents, reports the trigger as hidden, and positions from a broken rectangle. `data-bs-container="body"` was there to avoid exactly this, but Bootstrap 5 dropdowns have no `container` option, so it never did anything.

In this update,
1. Style card menus use static display and are placed by CSS, matching the More Options dropdown in `admin.js`.
2. Static display has no flip, so they open upwards when they would run past the bottom of the scroll panel.
3. The dead `data-bs-container` attribute is removed.
4. Right alignment is set with `--bs-position`, which is what Bootstrap 5 reads when choosing a placement. The `dropdown-menu-right` class this project uses is Bootstrap 4 naming and has had no effect since the upgrade.

Worth checking on review: this changes placement for the style card menus and right-aligns the Preview menu, so it is worth clicking through both plus the builder's More Options menu, which shares the same class.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved positioning of right-aligned dropdown menus.
  * Dropdowns in style panels now open upward when there isn’t enough space below.
  * Improved dropdown behavior within containers that restrict overflow.
  * Prevented incorrect upward-opening states when dropdowns are reopened.


Crabcyborg:
@tuguirazvan It looks like there is a bug with this update where it is always wrong on the 2nd click, 4th click, etc.

**Before**

https://github.com/user-attachments/assets/221b0591-8e0a-482e-9088-6389d6768f59

**After**

https://github.com/user-attachments/assets/1043dae6-3c1f-415d-acc0-890e9dd72061

tuguirazvan:
Good catch, thank you — reproduced and fixed in d47b276.

The cause: `frm-dropdown-menu-above` was only ever toggled in `shown.bs.dropdown`, and nothing removed it on close. So every second open measured a menu that was already sitting above its trigger, concluded it had room below, and dropped it back into the overflow it had just escaped. Hence right, wrong, right, wrong.

It now clears the class before reading the rectangle, so the check always runs against the default downward position.

Worth noting why it looked fine on my end: the class is only ever added when the menu actually overflows `#frm_style_sidebar`, so cards near the top never trigger it and there is nothing to alternate. It reproduces on a card far enough down the sidebar.

Verified both directions — a card near the bottom now opens upward on every open, and one near the top stays downward on every open.

garretlaxton (approved):
This looks good!

Crabcyborg (approved):
Thank you Razvan and Garret for working on this!

🚀

PR #3261, opened 2026-08-19
Build style card menus after Pro has registered its filter

tuguirazvan:
On the styles list page, a style card's menu sometimes showed a disabled "Duplicate" upsell, with Set as Default and Delete missing entirely, even with Pro active and authorized.

Core builds those menus 1ms after its own script runs, hoping Pro's separate script file has executed and registered its filter by then. When it has not, the menus are built without Pro's options and core adds the upsell to fill the gap. Reproduced on a cold load and not on five warm ones after it.

In this update,
1. The menus are built on DOMContentLoaded rather than after a 1ms timeout, which is the first point at which every enqueued script has run.

Pro's `formidable_pro_style_settings` already declares `formidable_style` as a dependency, so WordPress orders the two scripts correctly and no change is needed on the Pro side.

### DIFF
--- a/js/admin/style.js
+++ b/js/admin/style.js
@@ -184,9 +184,16 @@
 	 */
 	function initListPage() {
 		document.addEventListener( 'click', handleClickEventsForListPage );
-		// Add a timeout so Pro has a chance to add a filter first.
-		// 0 does not always work in Google Chrome, so use 1.
-		setTimeout( addHamburgerMenusToCards, 1 );
+		// Pro adds its options to these menus with a filter registered from a separate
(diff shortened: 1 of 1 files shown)


tuguirazvan:
Closing for now. This is a real intermittent bug — the same style card shows two different menus across page loads, and the short version carries a `frm_noallow` Duplicate that only core's upsell path produces — but it sits outside the dropdown placement and styling work this batch is scoped to. The diagnosis and the six-load evidence are preserved in the description above if anyone wants to pick it up.

PR #3275, opened 2026-08-25
Form builder optimizations

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/3427613497/257227. There's an XML they share in the ticket that I tested against. The form has ~1,100 fields.

Before it would take me ~2 minutes to properly load every field.
With this update, I'm seeing 40 seconds.

This update also aims to make the page less sluggish.

It's still a WIP though. I intend to continue working on this later after this part of the update is merged.

**This update includes several improvements.**
1. Fields AJAX loaded in long forms are now asynchronously pulled, so several batches can get requested at once rather than waiting for each bath before starting the next.
2. `setupSortable` was getting called for every field each time a field was loaded. This is used to make options for radio buttons, dropdown, etc sortable.
3. More icons are deferred until a field is actually selected and the field settings appear.
4. Updates the logic for determining the drag+drop placeholder. I wasn't able to drag a field in-between rows at the bottom of the 1,100 field form without removing `refreshPositions: true,` which appears to not be necessary.

## Performance Improvements

* Improved form-builder loading speed with larger batches and concurrent requests.
* Reduced unnecessary processing during field updates.
* Improved multiselect initialization for newly loaded fields.
* Improved drag-and-drop responsiveness and placement accuracy.
* Improved option handling for fields loaded after builder initialization.

## Bug Fixes

* Ensured page-break numbering and product-field options update after all fields finish loading.
* Corrected option-row controls when icons are disabled.
* Improved drag-and-drop behavior while scrolling or updating fields dynamically.


garretlaxton (approved):
This looks good!

Crabcyborg:
Thanks @garretlaxton!

🚀

PR #3278, opened 2026-08-26
Fix square amount issues when verifying buyer

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/3421447581/256847. They included an XML in the ticket.

This update fixes issues in a single location, where we make a `verifyBuyer` call to Square with JS before the form is submitted. This amount is used to verify the purchase and generally fails if it doesn't match the amount expected in PHP.

They likely only see this in production as the real purchase is likely showing a confirmation pop-up with the JS value. This endpoint now uses another amount preparing function we've had already that is used for Stripe.

Regression tests added in https://github.com/Strategy11/formidable-forms/pull/3279 to make sure that previous functionality is generally unchanged. Passed in master and passing in this branch.

**_Fix verified by the customer_**

**Pre-release**
[formidable-6.34.1b.zip](https://github.com/user-attachments/files/31697112/formidable-6.34.1b.zip)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Square Lite amount handling for verification and charge transactions.
  - Correctly normalizes literal amounts, decimal separators, zero-decimal currencies, and shortcode-resolved values.
  - Prevented issues when processing unresolved or empty shortcode amounts.
  - Improved handling of locale-specific number formatting.

- **Tests**
  - Added comprehensive automated coverage for currency and amount-formatting scenarios.

### DIFF
--- a/square/controllers/FrmSquareLiteAppController.php
+++ b/square/controllers/FrmSquareLiteAppController.php
@@ -103,6 +103,11 @@ class FrmSquareLiteAppController {
 	/**
 	 * Get the amount value for verification.
 	 *
+	 * Square's verifyBuyer expects the amount as a decimal string in the currency's
+	 * major units ("20.00" for twenty pounds), not the smallest denomination that the
+	 * Payments API uses. FrmSquareLiteActionsController::prepare_amount returns the
+	 * smallest denomination, so the parent is called here instead.
+	 *
 	 * @param WP_Post $action
 	 *
 	 * @return string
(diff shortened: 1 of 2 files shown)


garretlaxton:
I haven't found any issues with this PR, but I did run into something that I can also replicate on master. I can log it separately if you'd like.

I have conditional logic setup with 2 payment actions. If the radio button is yes, the slider should appear with the default value `53`, which is the total. If the radio button is no, the value should be `999`. When you select no, the value `53` is going through instead of `999`.

[square-pay-form.xml](https://github.com/user-attachments/files/31763576/square-pay-form.xml)

Crabcyborg:
Thanks @garretlaxton!

I made a new PR for that. In case it doesn't get merged between releases, I'm going to merge this separately.

To confirm, you have 2 Square actions, and this is happening just in the verify buyer endpoint?

There's a limit of 1 Square action per form, but this can obviously be manipulated by adding a Stripe action with the Stripe add-on active and then changing the gateway, or by using the Authorize.Net add-on and clicking the legacy Payment action type.

PR #3284, opened 2026-08-27
Fix denylist performance issue when there are no strings to check

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/3427613497/257227. There's an XML they share in the ticket that I tested against. The form has ~1,100 fields.

On my local site, changing between pages 1 and 2, I see page load times going from ~2.5 seconds to ~0.6 seconds with this update, saving about 2 seconds and most of the request time.

The issue is that it would iterate through each denylist line. This update exits early before the loop if there is nothing to actually check.

Regression tests added in https://github.com/Strategy11/formidable-forms/pull/3285

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spam denylist matching for regular expressions, exact values, contained text, and values containing slashes.
  * Empty submissions are skipped earlier for more reliable handling of blank content.
  * Option and password fields are excluded from denylist checks by default.
  * Denylist settings and filtering are now respected consistently.
  * Optimized repeated denylist checks for faster scanning of submitted content.

### DIFF
--- a/classes/models/FrmSpamCheckDenylist.php
+++ b/classes/models/FrmSpamCheckDenylist.php
@@ -180,6 +180,11 @@ class FrmSpamCheckDenylist extends FrmSpamCheck {
 			$this->fill_default_denylist_data( $denylist );
 			$denylist['allowed_words'] = $allowed_words;
 
+			if ( ! $this->add_values_to_check( $denylist ) ) {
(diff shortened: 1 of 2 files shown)


truongwp (approved):
Both the PRs looks good to me. Thanks @Crabcyborg!

Crabcyborg:
Thanks Truong!

🚀

PR #3286, opened 2026-08-27
Optimize denylist loop

Crabcyborg:
Carrying from https://github.com/Strategy11/formidable-forms/pull/3284

Attached are 50KB of text generated from https://www.lipsum.com/feed/html

[lorem-ipsum.txt](https://github.com/user-attachments/files/31528470/lorem-ipsum.txt)

This update optimizes the `str_contains` checks by creating a 4-gram index and checking against that first before the more expensive `str_contains` call.

The optimization is mostly seen when submitting large amount of text, like the 50KB example here.

This removes over 6 seconds from my form submission time.

The string is pretty repetitive so it might not be totally reflective of real cases, but I expect it to still be a huge improvement.

**Before**
Screenshot 2026-08-27 at 4 25 59 PM

**After**
Screenshot 2026-08-27 at 4 22 52 PM

### DIFF
--- a/classes/models/FrmSpamCheckDenylist.php
+++ b/classes/models/FrmSpamCheckDenylist.php
@@ -18,6 +18,31 @@ class FrmSpamCheckDenylist extends FrmSpamCheck {
 	const COMPARE_EQUALS = 'equals';
 
 	/**
+	 * How many leading characters of a denylist line are used as its index key.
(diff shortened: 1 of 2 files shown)


truongwp (approved):
This looks good.

Crabcyborg:
Thanks Truong!

🚀

PR #3288, opened 2026-09-01
Optimize embeds column and improve when syncing happens

Crabcyborg:
Related ticket https://secure.helpscout.net/conversation/3436854180/257585/

Related Pro PR https://github.com/Strategy11/formidable-pro/pull/6582

**This update:**
- Makes sure that the embed number updates on all post/page hooks (updated, deleted), rather than only ever tracking when a post is inserted.
- Handles all missing form data at once in a single query instead of once per missing row. Since this is based on a transient that lasts a day, and the transient is frequently deleted, it would be common for queries to happen for an entire page of form rows.
- Prevents attempts to delete the transient when it already was just deleted.
- Prevents the DB queries from happening at all if the embeds column is hidden.
- Fixes bugs where form key uses like `[formidable id=contact-form]` and `[formidable key=contact-form]` would not get counted.
- Delays when `prepare_posts` is called so it doesn't cache the extra empty properties in the DB, and only caches the necessary post object data to help reduce the size of the cache.
- Invalidate only specific form IDs from the cache instead of completely clearing it, so full large queries are less common.

**Pre-release**
[formidable-6.34.1b.zip](https://github.com/user-attachments/files/31697598/formidable-6.34.1b.zip)

## Summary by CodeRabbit

* **New Features**
  * Added embed-post information to the forms list, showing which posts and pages contain each form.
  * Improved embed detection across multiple forms for faster, more consistent results.
* **Bug Fixes**
  * Embed information now refreshes when posts are added, edited, trashed, restored, or deleted.
  * Prevented outdated embed results from appearing after content changes.
  * Improved handling of post titles, slugs, statuses, and embedded form references.
* **Performance**
  * Reduced redundant processing when displaying embed information for multiple forms.


classes/helpers/FrmFormsListHelper.php
```diff
@@ -576,12 +567,22 @@ protected function confirm_bulk_delete() {
 		return __( 'ALL selected forms and their entries will be permanently deleted. Want to proceed?', 'formidable' );
 	}
 
+	/**
+	 * @param stdClass $form
+	 *
+	 * @return string
+	 */
```
truongwp: This is duplicated and should be removed.

truongwp (approved):
Except the duplicated docblock, this PR looks good to me. Thanks @Crabcyborg for working on this.

classes/helpers/FrmFormsListHelper.php
Crabcyborg: Thanks Truong! Removed.

PR #3293, opened 2026-09-02
Reduce AJAX loaded fields overhead

Crabcyborg:
**_When loading a long form (1,100+ fields) with AJAX_**

**Before**
- JSON data for every AJAX loaded field would load in the page HTML.
- This JSON would get sent back to the server to load a field.
- Page source size on load: 21.4MB
- Example amount of data sent on each load field request: ~2.31MB

**After**
- The JSON is never included on page load.
- Just the field IDs are passed when loading a field, which then pulls the actual live field data instead of the JSON loaded in the page.
- Page source size on load: 9.2MB
- Example amount of data sent on each load field request: ~20KB

Cutting out 12.2MB of JSON in HTML that is never required on the front end, and >2MB of "upload" data when making a request.

This doesn't make significant changes in actual loading times, but it significantly cuts out a lot of overhead, should slightly improve performance, and likely helps more on slow connections. I also find that having less HTML in the page makes it generally less sluggish after it loads, and it should generally be less stressful on the server.

## Summary by CodeRabbit

- **Improvements**
  - Improved form builder field loading by retrieving field details server-side, supporting more reliable AJAX updates.
  - Updated field-loading responses to preserve each field’s ID and type during rendering.
- **Style**
  - Adjusted loading spinner behavior so it displays correctly for the full field list and individual field placeholders.
- **Bug Fixes**
  - Improved handling of field placeholders and asynchronous field rendering in the form builder.


garretlaxton (approved):
This looks good!

Crabcyborg:
Thank you Garret!

🚀

PR #3299, opened 2026-09-03
Check for action conditions when validating square payment

Crabcyborg:
Related comment Garret posted https://github.com/Strategy11/formidable-forms/pull/3278#issuecomment-5518995887

When verifying buyer, it would always just use the first Square action. This update adds checks for conditional logic so the right amount is used for the action that matches the fake entry object we make before the real entry is created.


garretlaxton (approved):
This looks good!

Crabcyborg:
Thank you Garret!

🚀

PR #3303, opened 2026-09-04
Fix style reset issues with quick settings

Crabcyborg:
There are 2 different "Reset to defaults" flows. One on the edit page and one on the list page. They differ because the edit page should not change the default in the DB until you save the settings page. The list page however should reset immediately.

With the quick settings update, some things stopped working.

This update passes along a `persist` param in `$_POST` now for the list page, and the edit page exits early with defaults rather than making any DB changes.

**Before**

https://github.com/user-attachments/assets/1d0d5d2b-dc66-49de-adbb-a61726ec4361

**After**

https://github.com/user-attachments/assets/5a5d7835-0253-4858-8477-4a17baa9c25a


garretlaxton (approved):
This looks good!

Crabcyborg:
Thank you Garret!

🚀

PR #3339, opened 2026-09-15
Fix pro issue 6392 (Properly replace errors without error classes when submitting with AJAX)

Crabcyborg:
Fixes https://github.com/Strategy11/formidable-pro/issues/6392

You can replicate this by going to the custom HTML for a required field, and replacing the error HTML with something like this:

```
[if error]<div>[error]</div>[/if error]
```

Where there is no identifying class or ID on the div anymore, which is where the JS was falling short.

Now every top level element is given `data-frm-error` so we have something that doesn't change styling at all but is still identifiable as error content.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form error handling for custom error markup.
  * Custom error messages are now consistently identified, focused, and removed during validation.
  * Revalidation now clears all matching error elements instead of only the first one.
  * Prevented validation focus handling from failing when no preceding error element is available.
  * Improved compatibility with custom templates that use error elements without the standard error class or identifier.


garretlaxton:
If I try to submit a blank field, I see my custom error I added in the custom HTML section.
image

However, when I correct the error, the original default error message goes away, but the custom error message remains.
image

Crabcyborg:
Thank you Garret!

That should now be fixed.

PR #3344, opened 2026-09-16
Remove Add-On request link to connect.formidableforms.com

stephywells:
## Summary

The Add-Ons page had a link called "Request Add-On." It sent users to `connect.formidableforms.com`. That site is going away, so this link had to go too.

This PR removes the link in two places:
- The small text at the bottom of the Add-Ons list: "Not finding what you need? Request Add-On."
- The button that shows up when a search finds no add-ons.

It also removes the code that built that link's web address, since nothing uses it anymore.

## Test plan

- [ ] Open the Add-Ons page in WP Admin. Check that the old "Request Add-On" text and link at the bottom are gone.
- [ ] Search for an add-on that does not exist. Check that the "No add-ons found" message shows, but no button appears under it.
- [ ] Run the Cypress test for the Add-Ons page and confirm it passes.

## Summary by CodeRabbit

* **Changes**
  * Removed the “Request Add-On” link and button from the add-ons page.
  * When no add-ons match the selected filters or search, the page now displays only the empty-state message.
  * Add-on category filtering and count displays remain unchanged.

* **Tests**
  * Updated add-ons page checks to verify that the request option is hidden in empty states.


Crabcyborg (approved):
I think this looks good.

There was no review request, but this looks like something we should get merged.

Thanks Steph!

🚀

PR #3377, opened 2026-09-18
Lift the admin footer text off the grey it could not be read against

tuguirazvan:
`--grey-400` measures 2.58:1 on the white admin body. The footer text is 12px, so it is normal text and answers to 4.5:1 — `--grey-500` is the lightest stop on the ramp that clears it, at 4.97:1. The social icons take the same step.

The social links also had no hover feedback: their resting rule and `.frm_wrap a:hover` are both (0,2,1) and this file loads later, so the grey won. The added pseudo-class settles it at 600, with `:focus-visible` alongside so the affordance is not mouse-only.

Token bumps only — no palette values changed. Built CSS diff is three footer rules and nothing else.

PROD-5662's missing-accessible-name half was already fixed on master in `8bb0bee28`; this covers the contrast half.


Crabcyborg (approved):
Thank you Razvan!

🚀

PR #3383, opened 2026-09-18
Fix low-contrast grey text across the admin

tuguirazvan:
Six grey-text contrast failures; three shared one cause.

`.frm_noallow` set `opacity: 0.5` globally, which fades text and background toward the same backdrop — so the ratio collapses whatever the colours are, and no palette change could fix it. The reported `#858992`, `#888c94` and `#89dbb5` are `--grey-900` and `--success-500` seen through it. The state is carried by colour now.

Worst pairs: BETA pill 1.63 → 7.68, inbox timestamp and Dismiss 2.58 → 4.97, sidebar tab 3.35 → 4.76, `--medium-grey` 4.46 → 5.21. Fixed at the shared rule or token in every case.


Crabcyborg (approved):
Thank you Razvan!

🚀

